PR #4, opened 2026-05-20
chore: add rustfmt.toml for uniform formatting

Pranav2612000:
### What does this PR do?
- Adds a basic rustfmt.toml file and runs `cargo fmt` to format codebase using this file

- Keep a consistent style guide that agents and human contributors can follow

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,2 @@
+tab_spaces = 4
+edition = "2024"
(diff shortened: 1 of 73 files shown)


rustfmt.toml
```diff
@@ -0,0 +1,2 @@
+tab_spaces = 2
```
vinaykakade: Can we make tab_spaces=4? I used to like 2 earlier, but visual nesting separation with 4 is pretty appealing. (And if we have so much nesting that 4 looks too much, that code is likely a candidate for refactor anyway.)

Pranav2612000: Done @vinaykakade

vinaykakade (approved):

PR #6, opened 2026-05-21
benchmarks: allow running superfile and supertable benchmarks seperately

Pranav2612000:
Before this change, running only superfile benchmarks also performed expensive setup for supertable


vinaykakade (approved):
We'll need a similar change for vector too - perhaps in a separate PR.

PR #7, opened 2026-05-21
Add peak-RSS stats to benchmarks

ekechinwokah:
## Summary
Adds a peak-RSS column to the vector and FTS bench tables and fixes a couple of bench-harness papercuts that were hiding numbers and breaking criterion's own delta reporting.

## What's in the PR
**Peak-RSS sampling**
- New `rss::PeakSampler` polls `/proc/self/status` on a background thread and records the peak `VmRSS` observed during each criterion group. Per-group peaks are written to `rss.json` next to criterion's `estimates.json`.
- Every vector + FTS bench (`superfile_*`, `supertable_*`, build + search groups) is now wrapped in a sampler. The markdown tables gain a `Peak RSS` column.
- Adds before/after comparisons like Criterion does for latency measurements

**`infino-bench-utils` crate**
- All bench helpers (`corpus`, `markdown`, `rss`) **and** every bench body (`fts_superfile`, `fts_supertable`, `vector_superfile`, `vector_supertable`) live in a new library crate at `benches/utils/`.


Cargo.toml
```diff
@@ -17,10 +17,9 @@ path = "src/lib.rs"
 # Decimal128 id columns, default tokenizers, default vector
 # configs. Used by integration tests, benches, and (via
 # `cfg(test)`) unit tests so all three contexts share one
-# source of truth for boilerplate-heavy fixtures. NOT part
-# of infino's stable API; signatures may change between any
-# two versions. The `dev-dependencies` self-reference below
-# auto-enables this for test + bench builds.
+# source of truth for boilerplate-heavy fixtures. The
+# `dev-dependencies` self-reference below auto-enables this for
```
Pranav2612000, line 21: Based on this diff, are we planning to make it a part of "infino's stable API"?

benches/utils/rss.rs
```diff
@@ -0,0 +1,299 @@
...
+/// work you want to bound and stop it after; the returned
+/// peak is the max VmRSS observed across the sampler's
+/// lifetime.
+///
+/// The thread reads `/proc/self/status` at `interval`
+/// cadence. Each read is a ~10 µs syscall — negligible next
+/// to the work the sampler watches.
+pub struct PeakSampler {
+    stop: Arc<AtomicBool>,
+    peak: Arc<AtomicU64>,
```
Pranav2612000, line 75: Lets also track median, p90 values ( maybe in a separate PR ).

benches/utils/rss.rs
```diff
@@ -0,0 +1,299 @@
...
+        let stop = Arc::new(AtomicBool::new(false));
+        let peak = Arc::new(AtomicU64::new(current_rss_bytes().unwrap_or(0)));
+
+        let stop_t = Arc::clone(&stop);
+        let peak_t = Arc::clone(&peak);
+        let handle = thread::Builder::new()
+            .name("rss-sampler".into())
+            .spawn(move || {
+                while !stop_t.load(Ordering::Acquire) {
+                    if let Some(rss) = current_rss_bytes() {
```
Pranav2612000, line 99: Is it better to sample absolute rss value or the difference between the value at the start and the current value?

Pranav2612000 (approved):

benches/utils/rss.rs
ekechinwokah: ok. will create separate PR.

benches/utils/rss.rs
ekechinwokah: good question. maybe we can answer it empirically.

Cargo.toml
ekechinwokah: no. it's a public repo so cleaning up comments that reference anything internal.

PR #8, opened 2026-05-22
bench: report RSS median and p90

ekechinwokah:
- Extend the bench RSS sampler to retain observed VmRSS samples and persist peak, median, and p90 values in rss.json.
- Keep peak_rss_bytes backward-compatible while adding median_rss_bytes and p90_rss_bytes for new runs.
- Add Median RSS and P90 RSS columns to FTS/vector superfile and supertable markdown output.


Pranav2612000 (approved):
LGTM! Not sure about the CPU/Memory impact now that we're capturing more samples and performing more work. But should be good if its tested and you're confident.

ekechinwokah:
Fair point - I didn't test specifically but we can adjust benchmarks if the numbers seem impacted. Feels like a low risk decision.

PR #11, opened 2026-05-22
chore ci: add basic fmt, build & test on every PR

Pranav2612000:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
...
+  check:
+    name: Format, Build & Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: rustfmt
+      - uses: Swatinem/rust-cache@v2
+      - run: cargo fmt -- --check
```
vinaykakade: Can you just use "make ci" from Makefile? That way, any setup needed in future locally or in CI is in one place.

vinaykakade (approved):
Good to go after using "make ci" instead of direct cargo commands

.github/workflows/ci.yml
Pranav2612000: Done

PR #12, opened 2026-05-22
Add streaming to vector index builds

ekechinwokah:
## Summary
This PR makes vector index building stream to disk instead of materializing the full build state in memory.
- Adds spill-backed vector build primitives and reservoir-sampled k-means so large vector builds do not keep all vectors resident.
- Adds streaming `VectorBuilder::finish_to<W: Write>` output, while keeping `finish()` available for callers that want an in-memory blob.
- Bounds concurrent vector shard builds in `Supertable` so large multi-segment builds do not fan out into excessive RSS.
- Updates benchmark harness compatibility with the current RSS reporting from `main`.


Pranav2612000 (approved):

Cargo.toml
Pranav2612000, line 183: Nit: We can remove this from dev-deps

src/superfile/vector/builder.rs
```diff
@@ -66,12 +123,58 @@ impl Default for VectorBuilder {
...
+    /// (under `$TMPDIR` via `tempfile::tempdir()`) and the
+    /// default 256 MiB spill threshold.
+    ///
+    /// Panics if creating the scratch tempdir fails — at construction
+    /// time there's no realistic recovery and the panic preserves
+    /// the existing public API. Operators running large builds
+    /// should prefer [`Self::with_scratch`] pointing at an
+    /// instance-store NVMe partition.
     pub fn new() -> Self {
+        let tmp = tempfile::tempdir().expect("create VectorBuilder scratch tempdir");
```
Pranav2612000, line 136: Should this tmp file creation occur only when we want to spill? Why should a user pay the costs of this file if their data is within the spill range?

src/superfile/vector/builder.rs
```diff
@@ -523,4 +1032,244 @@ mod tests {
...
+    }
+
+    /// `finish_to(Vec<u8>)` must produce byte-for-byte identical
+    /// output to `finish()` for the same logical builder state.
+    /// The build path is deterministic in everything that matters
+    /// (rot_seed, reservoir seed, bucket flush ordering), so any
+    /// drift here would indicate a regression in either the
+    /// streaming wrap or the underlying determinism contract.
+    #[test]
+    fn finish_to_matches_finish_byte_for_byte() {
```
Pranav2612000, line 1166: Do we need this test? `finish` calls `finish_to` so its always going to match byte for byte

Pranav2612000 (commented):
Implementation looks good. Can we have more tests around searching with the Spill writer ( I think most current tests dont run on data large enough to spill. Is that correct? )

Can you also update benchmark numbers with this approach ( and how smaller is it ). I'm also thinking there should be a slight effect on the accuracy of our results, but none of our tests fail - is it a concern? Do we have tests where we check our results with another db, to verify correctness? Where can I find them?

Cargo.toml
ekechinwokah: done: https://github.com/infino-ai/infino/pull/15

src/superfile/vector/builder.rs
ekechinwokah: changed to be lazy: https://github.com/infino-ai/infino/pull/15

src/superfile/vector/builder.rs
ekechinwokah: done: https://github.com/infino-ai/infino/pull/15

ekechinwokah:
> Implementation looks good. Can we have more tests around searching with the Spill writer ( I think most current tests dont run on data large enough to spill. Is that correct? )
> 
> Can you also update benchmark numbers with this approach ( and how smaller is it ). I'm also thinking there should be a slight effect on the accuracy of our results, but none of our tests fail - is it a concern? Do we have tests where we check our results with another db, to verify correctness? Where can I find them?

i believe we have some internal tools. will ping you offline.

on the benchmark numbers, done, but didn't move the needle


metric | parent | tip/fix set | delta
-- | -- | -- | --
correctness recall@10 | 1.000 | 1.000 | no change
Criterion mean build | 20.716s | 20.691s | -0.12%
throughput | 48.27 K/s | 48.33 K/s | no change
artifact size | 1515.94 MiB | 1515.94 MiB | no change
peak RSS | 10.47 GiB | 10.47 GiB | no change
median RSS | 4.90 GiB | 4.89 GiB | no change
p90 RSS | 7.57 GiB | 7.51 GiB | no change
cold open | 169.11ms | 159.68ms | small single-run improvement
first query | 0.61ms | 0.61ms | no change

PR #15, opened 2026-05-23
Add streaming to FTS index builds

ekechinwokah:
## Summary

This PR makes FTS index building spill to disk instead of holding every term's posting accumulator resident, then tunes the resulting hot path so the production build is materially faster than `main` was before this change.

- Adds spill-backed FTS build primitives (per-column threshold-based two-mode accumulator, fixed 12-byte spill triples, per-column hash-partitioned spill files) so large FTS builds no longer keep the full `term → postings` map in RAM.
- Adds streaming `FtsBuilder::finish_to<W: Write>` output with k-way merge across partition files and a `StreamingDictBuilder`-backed FST, while keeping `finish() -> Vec<u8>` available for callers that want an in-memory blob.
- Adds external-merge fallback for over-budget partitions plus configurable partition count / max-partition-bytes / scratch-directory knobs, so the bounded-memory guarantee holds even on pathologically skewed corpora.
- Optimises the spill and in-RAM `add_doc` hot paths (single-pass counting sort, SIMD `u8x16` tokenize scan, arena-backed term interner, `hashbrown::raw_entry_mut` on per-doc dedup, `Vec<Triple>` batch buffer above `BufWriter`, dense `Vec<u32>` per-doc tf accumulator, `TermScratch` reuse in encode).
- On-disk FTS blob format is unchanged; `FtsReader` is bit-identical. Regression-gated by a byte-for-byte equality test between the in-RAM and spilled+streaming-FST finish paths on the same corpus.
- Added larger instance with larger disk to CI yaml since the builds now spill over during tests.

## Benchmark Evidence

AWS `c7i.4xlarge`, 1M-doc Zipfian, 200 tokens/doc, 10K vocab. Criterion, 10 samples, `p < 0.05` on both build benches.

### Build timing

| bench | `main` | this branch | change |
|---|---:|---:|---:|
| `superfile_fts_build/infino_1thread_1000000docs` (single-threaded, spill path) | 17.39 s | **12.57 s** | **−27.7% faster** |
| `superfile_fts_build/infino_rayon_default_threads_1000000docs` (16 rayon shards, production path) | 1.75 s | **1.18 s** | **−32.6% faster** |

### Build peak RSS

| bench | `main` peak | this branch peak | change | `main` median | branch median | `main` p90 | branch p90 |
|---|---:|---:|---:|---:|---:|---:|---:|
| `infino_1thread_1000000docs` | 5.80 GiB | **5.37 GiB** | **−7.4%** | 4.67 GiB | 3.97 GiB | 5.27 GiB | 4.22 GiB |
| `infino_rayon_default_threads_1000000docs` | 5.81 GiB | **4.17 GiB** | **−28.2%** | 5.09 GiB | 4.17 GiB | 5.77 GiB | 4.17 GiB |

Both build benches are now faster than `main` *and* use less peak memory than `main`.

### Search timing (unchanged — PR is build-side only)

| query | `main` | this branch | change |
|---|---:|---:|---:|
| `single_rare_infino_top10` | 452 ns | 462 ns | +2.3% |
| `single_df1_infino_top10` | 201 ns | 206 ns | +2.3% |
| `single_common_infino_top10` | 10.6 µs | 11.5 µs | +9.0% |
| `two_term_or_infino_top10` | 169.1 µs | 163.9 µs | −3.1% |
| `three_wide_or_infino_top10` | 2.5 ms | 2.5 ms | +1.4% |
| `three_similar_or_infino_top10` | 10.5 ms | 10.9 ms | +3.3% |
| `five_term_or_infino_top10` | 17.9 ms | 18.1 ms | +1.6% |
| `two_term_and_infino_top10` | 67.0 ms | 64.1 ms | −4.3% |
| `three_wide_and_infino_top10` | 44.5 ms | 45.2 ms | +1.4% |
| `three_similar_and_infino_top10` | 26.3 ms | 24.6 ms | −6.4% |
| `five_term_and_infino_top10` | 34.2 ms | 33.6 ms | −1.7% |
| `similar_3_or_bmm_top10` | 10.5 ms | 10.9 ms | +3.4% |
| `similar_3_or_wand_top10` | 15.4 ms | 15.1 ms | −1.6% |
| `similar_5_or_bmm_top10` | 17.9 ms | 18.1 ms | +1.5% |
| `similar_5_or_wand_top10` | 45.1 ms | 45.6 ms | +1.1% |
| `wide_3_or_bmm_top10` | 2.5 ms | 2.5 ms | +1.1% |
| `wide_3_or_wand_top10` | 8.4 ms | 8.2 ms | −2.0% |

All 17 search benches stay within run-to-run variance of `main`. The one >5% delta (`single_common_infino_top10`, +9%) sits on a posting-block scan path this PR does not modify, and shows both +9% and −3% on the same commit pair across prior runs.

### Memory headroom

The larger 10M-doc supertable validation no longer OOMs on the 30 GiB bench box; observed RSS stayed bounded during the run instead of growing linearly in `n_docs × distinct_terms` per column.


benches/utils/corpus.rs
```diff
@@ -1,585 +0,0 @@
-//! Shared bench fixtures: deterministic corpora, query batteries,
```
Pranav2612000, line 1: Is the delete of this file intentional? @ekechinwokah 
`cargo bench` doesn't seem to work anymore

src/superfile/vector/builder.rs
```diff
@@ -96,23 +95,61 @@ struct ColumnState {
...
+                Some(parent) => tempfile::TempDir::new_in(parent)?,
+                None => tempfile::tempdir()?,
+            };
+            self.tempdir = Some(tmp);
+        }
+        Ok(self
+            .tempdir
+            .as_ref()
+            .expect("scratch tempdir initialized")
+            .path())
```
Pranav2612000, line 138: Nit: Use more "rustic" way by using `if let Some(..)`

src/superfile/fts/builder.rs
```diff
@@ -1,11 +1,10 @@
...
 //!
-//! `FtsBuilder` accumulates posting lists across all FTS-indexed columns
-//! into a `Vec<(u32, u32)>` per `(column, term)`, then `finish()`
-//! emits the on-disk FTS blob laid out as:
+//! `FtsBuilder` accumulates posting records across all FTS-indexed
+//! columns and on `finish_to<W>` emits the on-disk FTS blob:
 //!
 //! ```text
-//!   header (56 bytes)
+//!   header (48 bytes)
```
Pranav2612000, line 7: Why this change? The plan mentioned there should be no change in the underlying format

src/superfile/fts/builder.rs
```diff
@@ -79,34 +249,742 @@ struct ColumnState {
...
+/// Fixed on-disk record size in the spill files: 4 bytes `term_id`
+/// + 4 bytes `doc_id` + 4 bytes `tf`, all little-endian.
+///
+/// This matches vector's "raw f32 bytes" spill strategy in shape:
+/// fixed-size records, no framing, no variable-length payload.
+/// `read_triples` reads N×12 bytes in one syscall-amortised batch
+/// and reinterprets as `&[[u32; 3]]` via `bytemuck` — zero per-
+/// record allocation, no UTF-8 validation, no `Box<str>` round-
+/// trips.
+const TRIPLE_BYTES: usize = 12;
```
Pranav2612000: Is it better to use size::of<Triple> so that we dont need to maintain it?

tests/superfile/fts/uses_spill_builder.rs
```diff
@@ -0,0 +1,142 @@
...
+                column: "title".into(),
+            },
+            FtsConfig {
+                column: "body".into(),
+            },
+        ],
+        Vec::new(),
+        Some(default_tokenizer()),
+    );
+    let mut b = SuperfileBuilder::new(opts).expect("new SuperfileBuilder");
```
Pranav2612000, line 56: Can we set the default spill size to 0MB or lower? I'm not sure if 1024 documents are enough to breach the default 256MB limit

src/superfile/fts/builder.rs
```diff
@@ -148,10 +1130,70 @@ impl FtsBuilder {
             doc_lengths: Vec::new(),
             total_tokens: 0,
         });
-        self.postings.push(FxHashMap::default());
+        self.postings.push(ColumnPostings::new());
+        let _ = column_id;
```
Pranav2612000: Why do we need this line?

src/superfile/fts/builder.rs
```diff
@@ -186,71 +1228,372 @@ impl FtsBuilder {
...
+        let col_post = &mut self.postings[col_idx];
+        match col_post {
+            ColumnPostings::Spilled {
+                partitions,
+                term_to_id,
+                id_to_term,
+                dense_doc_tf,
+                updated_terms,
+                term_arena,
+            } => {
```
Pranav2612000: Let's split Spilled and InRam methods to seperate functions to make code easy to read?

src/superfile/fts/builder.rs
```diff
@@ -186,71 +1228,372 @@ impl FtsBuilder {
...
+                    // VectorBuilder's pre_spill_buffer transition).
+                    let drained = std::mem::take(terms);
+                    let mut partitions = Self::open_partitions_for_column(
+                        self.scratch_dir.path(),
+                        column_id,
+                        self.spill_partitions,
+                    )?;
+                    let term_arena = bumpalo::Bump::new();
+                    let mut term_to_id: TermIdMap = TermIdMap::default();
+                    let mut id_to_term: Vec<&'static str> = Vec::new();
```
Pranav2612000: Nit: initiate with known capacity ( i guess should be equal to the number of terms )

benches/utils/corpus.rs
ekechinwokah: bad merge. restored.

src/superfile/fts/builder.rs
```diff
@@ -743,11 +2737,70 @@ mod tests {
         assert_eq!(fst_off, 48);
     }
 
+    #[test]
+    fn finish_to_matches_finish_byte_for_byte() {
```
Pranav2612000, line 2906: Nit: unnecessary test.

src/superfile/fts/builder.rs
```diff
@@ -760,6 +2813,211 @@ mod tests {
...
+        // And the scratch tempdir under the override must contain no
+        // posting partition files yet (FtsBuilder's scratch tempdir is
+        // a *sub*directory of `parent`).
+        let mut spill_files_found = 0usize;
+        for entry in walkdir_files(parent.path()) {
+            let name = entry.file_name().to_string_lossy().to_string();
+            if name.starts_with("fts_col") && name.ends_with(".bin") {
+                spill_files_found += 1;
+            }
+        }
```
Pranav2612000, line 3010: Lets add the opposite of this in the below spilling test to ensure that the spill files are created

src/superfile/fts/builder.rs
```diff
@@ -760,6 +2813,211 @@ mod tests {
...
+        let mut baseline = FtsBuilder::new(tokenizer());
+        build_corpus(&mut baseline);
+        let baseline_blob = baseline.finish().expect("finish baseline");
+
+        // Tight budget forces external merge. 1 KiB is well below
+        // the dominant partition's on-disk size, so the merge path
+        // is exercised on at least one partition.
+        let mut tight = FtsBuilder::new(tokenizer());
+        tight.set_max_partition_bytes(1024);
+        build_corpus(&mut tight);
```
Pranav2612000, line 3156: Same here. Add a `walkdir` assert to ensure that the individual partition files are created

Pranav2612000:
Also a general review comment is to break down the PRs into smaller reviewable chunks, for example, this PR could have been broken down into
1. SIMD improvements in vector files
2. Migration from stable to unstable sort
3. Improvements to ASCII tokenizer
4. streaming to fts index builds


All of these are independent of each other and 1, 2, 3 would have been much quicker to review and get merged. 
It's only 4 which has a lot of new & complicated logic

Pranav2612000:
Maybe we can address this in a separate PR, but is there a reason why we've fixed the number of partitions ( 128 )?
Would it be faster to start from a small number of partitions and grow them after a certain threshold? Having a fixed 128 size could result in a lot of small partitions causing slowness due to fan-out + merge ( and writes )
The drawback is we may need a more complicated partition choosing algorithm ( t & n-1 - won't work ) and may need to redistribute data.

src/superfile/fts/builder.rs
```diff
@@ -259,269 +1602,939 @@ impl FtsBuilder {
...
+            bump: _,
+        } = self;
+
+        let n_columns = columns.len() as u32;
+        let mut n_terms_total_usize: usize = 0;
+
+        // Decide which finish path to take. Any column in `Spilled`
+        // mode forces the streaming-FST path; otherwise everything
+        // fits in RAM and we use the in-RAM FST builder.
+        let any_spilled = postings.iter().any(|c| c.is_spilled());
```
Pranav2612000: We do a lot of extra initializations even if no column is spilled. Can we early return in !any_spilled?

src/superfile/fts/dict.rs
```diff
@@ -101,6 +102,55 @@ impl DictBuilder {
...
+        Ok(Self {
+            inner: MapBuilder::new(w)?,
+            n_keys: 0,
+        })
+    }
+
+    /// Insert a `(key, value)` pair. `key` must be strictly greater
+    /// than the previously inserted key. Returns an error otherwise.
+    pub fn insert_sorted(&mut self, key: &[u8], value: u64) -> Result<(), fst::Error> {
+        self.inner.insert(key, value)?;
```
Pranav2612000, line 146: Can we also buffer this? Or does fst crate already take care of it?

src/superfile/vector/builder.rs
ekechinwokah: done

src/superfile/fts/builder.rs
ekechinwokah: stale comment. 48 is the correct number.

src/superfile/fts/builder.rs
ekechinwokah: done — replaced the magic 12 with `const TRIPLE_BYTES: usize = std::mem::size_of::<Triple>();` so it tracks the underlying `[u32; 3]` type definition automatically.

tests/superfile/fts/uses_spill_builder.rs
ekechinwokah: good catch. test was misleadingly named. added `SuperfileBuilder::set_fts_spill_threshold_bytes()` (forwards to the inner FtsBuilder) and the integration test now calls it with 1 before the first add_batch, which forces every column to transition to Spilled on the first doc and exercises the real streaming-FST + partition-merge finish path.

src/superfile/fts/builder.rs
ekechinwokah: we don't. removed.

src/superfile/fts/builder.rs
ekechinwokah: done — extracted `FtsBuilder::add_doc_spilled` and `FtsBuilder::add_doc_inram` as separate private methods, each with `#[inline(always)]` so the hot path is unchanged. `add_doc` now downcasts to dispatch on `self.postings[col_idx].is_spilled()` and forwards. Each helper re-destructures its variant internally so split-borrow semantics over self.columns / self.postings / self.bump are preserved.

src/superfile/fts/builder.rs
ekechinwokah: done: `Vec::with_capacity(drained.len())`.

src/superfile/fts/builder.rs
ekechinwokah: prefer to keep. finish() is just a wrapper around finish_to(&mut Vec), so this is really a determinism gate — two builds over the same corpus must emit identical blobs. catches hasher / iteration-order regressions cheaply on the in-RAM path. added a one-liner doc comment explaining it.

src/superfile/fts/builder.rs
ekechinwokah: done.

src/superfile/fts/builder.rs
ekechinwokah: done, but went with a #[cfg(test)] observer rather than a directory walk, since the sorted-chunk files (*_sorted{idx}.bin) live under the builder's scratch tempdir and are reaped by tempdir drop inside finish().

ekechinwokah:
> Also a general review comment is to break down the PRs into smaller reviewable chunks, for example, this PR could have been broken down into
> 
> 1. SIMD improvements in vector files
> 2. Migration from stable to unstable sort
> 3. Improvements to ASCII tokenizer
> 4. streaming to fts index builds
> 
> All of these are independent of each other and 1, 2, 3 would have been much quicker to review and get merged. It's only 4 which has a lot of new & complicated logic

noted. we organized the key PRs to combine a few internal commits so there is a complete corresponding plan for reviewers to follow and we can demonstrate a performance delta as part the PR, which isn't always possible in the individual commits.

src/superfile/fts/builder.rs
ekechinwokah: done.

src/superfile/fts/dict.rs
ekechinwokah: the caller already wraps in BufWriter.

ekechinwokah:
> Maybe we can address this in a separate PR, but is there a reason why we've fixed the number of partitions ( 128 )? Would it be faster to start from a small number of partitions and grow them after a certain threshold? Having a fixed 128 size could result in a lot of small partitions causing slowness due to fan-out + merge ( and writes ) The drawback is we may need a more complicated partition choosing algorithm ( t & n-1 - won't work ) and may need to redistribute data.

done.

src/superfile/fts/builder.rs
Pranav2612000: Works.

Pranav2612000 (approved):

PR #17, opened 2026-05-24
fts: block-max-AND pruning + 2-term flat-merge specialization

vinaykakade:
## Summary

Two follow-on optimizations to the leapfrog AND intersection:

1. **Block-max-AND pruning.** At the top of each leader-loop iteration, once the top-K heap is full, compute an upper bound on the maximum AND score across the leader's current block: `leader.block_max + sum(other.max_block_max_overlapping(leader_block_range))`. If that UB can't beat the heap's kth-best, skip past the leader block without inspecting it. The AND analog of the per-block max pruning we already do for OR.

   A new `TermCursor::block_max_in_range` walks a cursor's blocks intersecting a given doc-id range and returns the max. The single-block `inspect_block_max` underestimates when the leader's block spans multiple cursor blocks with varying `block_max`, which would unsafely skip blocks that could contain top-K matches.

2. **2-term flat-merge specialization.** When `n_cursors == 2`, route to a specialized inner loop that walks both cursors' decoded `block_doc_ids` arrays with two index pointers and scores matches inline. Removes per-doc `skip_to` function-call + within-block linear-scan overhead on the hottest AND case (rare ∧ common). Cross-block transitions stay on the normal `next()` path.

## Results (1M docs, single superfile)

| Shape | Before | After | Speedup |
|---|---|---|---|
| `two_term_and` | 4.44 ms | **113 µs** | **39×** |
| `three_wide_and` | 3.33 ms | 2.45 ms | 1.36× |
| `three_similar_and` | 4.58 ms | 4.32 ms | 1.06× |
| `five_term_and` | 5.11 ms | 4.50 ms | 1.14× |

The 39× headline on `two_term_and` is the 2-term specialization — flat block-merge avoids the per-doc `skip_to` overhead that dominated the previous path on rare ∧ common shapes. Top-of-loop block-max pruning lifts the multi-term shapes by 5-35%.

OR shapes are untouched and continue to use BMM.

## Implementation notes

- `AndHeapEntry` and `and_heap_push` extracted to module-scope free items so the general-n and 2-term paths share heap-insert logic + tie-break rules.
- `block_max_in_range` uses `shallow_advance_block_to` as its starting hint so the inspect-block pointer monotonically advances across the bench's many-leader-block walks — amortized O(1) per call.
- Single-mut-ref-each-of-two-cursors via `split_at_mut(1)` in the 2-term path lets the inner loop read both `block_doc_ids`/`block_tfs` arrays without borrow-checker contortions.

## Test plan

- [x] `cargo test --test superfile -- brute_force_oracle` (20 oracles incl. 6 AND, all pass).
- [x] `cargo test --test supertable -- brute_force_oracle` (14 oracles incl. 4 AND, all pass).
- [x] `INFINO_BENCH_UPDATE_README=1 cargo bench --bench fts -- superfile_fts_search` produces the numbers above; README refreshed in place.

### DIFF
--- a/src/superfile/fts/reader.rs
+++ b/src/superfile/fts/reader.rs
@@ -1097,7 +1097,6 @@ impl FtsReader {
         k: usize,
         heap: &mut BinaryHeap<AndHeapEntry>,
     ) {
-        let n = cursors.len();
         'outer: loop {
             if cursors[0].is_exhausted() {
                 break;
@@ -1132,12 +1131,12 @@ impl FtsReader {
             // catches up the others.
             loop {
                 let mut bumped = false;
-                for i in 0..n {
-                    cursors[i].skip_to(candidate, postings);
-                    if cursors[i].is_exhausted() {
+                for c in cursors.iter_mut() {
+                    c.skip_to(candidate, postings);
+                    if c.is_exhausted() {
                         break 'outer;
                     }
-                    let here = cursors[i].current_doc_id();
+                    let here = c.current_doc_id();
                     if here != candidate {
                         candidate = here;
                         bumped = true;
@@ -1924,11 +1923,11 @@ impl Ord for AndHeapEntry {
 fn and_heap_push(heap: &mut BinaryHeap<AndHeapEntry>, k: usize, score: f32, doc_id: u32) {
     if heap.len() < k {
         heap.push(AndHeapEntry(score, doc_id));
-    } else if let Some(&worst) = heap.peek() {
-        if score > worst.0 || (score == worst.0 && doc_id < worst.1) {
-            heap.pop();
-            heap.push(AndHeapEntry(score, doc_id));
-        }
+    } else if let Some(&worst) = heap.peek()
+        && (score > worst.0 || (score == worst.0 && doc_id < worst.1))
+    {
+        heap.pop();
+        heap.push(AndHeapEntry(score, doc_id));
     }
 }
 
@@ -2197,11 +2196,11 @@ impl TermCursor {
         self.current_block >= self.blocks.len()
     }
 
-    /// Total number of postings (df) for this term. Used by AND
-    /// intersection to pick the rarest cursor as the leader.
-    /// Block count is an exact upper bound on df (df = (n-1)*BLOCK_LEN
-    /// + last_block_n); cursor count comparison via this method gives
-    /// stable smallest-first ordering. Inline cursors return 1.
+    /// Block count, used as a cheap proxy for df when AND intersection
+    /// picks the rarest cursor as the leader. Block count is an exact
+    /// upper bound on df: a term's df is `(blocks - 1) * BLOCK_LEN +
+    /// last_block_n`, so cursors compare in the same order by block
+    /// count as they do by df. Inline cursors return 1.
     #[inline(always)]
     fn block_count(&self) -> usize {
         self.blocks.len()
--- a/src/superfile/fts/reader.rs
+++ b/src/superfile/fts/reader.rs
@@ -686,8 +686,11 @@ impl FtsReader {
         }
 
         // Drain heap → sorted descending by score, ascending by doc_id on ties.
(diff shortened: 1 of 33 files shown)


src/superfile/fts/reader.rs
```diff
@@ -1120,34 +1151,127 @@ impl FtsReader {
...
+                c0.skip_to(c1.current_doc_id(), postings);
+                continue;
+            }
+            // Symmetric advance for c0 if c1 sits above c0's pos.
+            if c1.current_doc_id() > c0.current_doc_id() {
+                c0.skip_to(c1.current_doc_id(), postings);
+                if c0.is_exhausted() {
+                    break 'outer;
+                }
+            }
```
Pranav2612000: Do we need both these conditions? Isn't `c1.current_doc_id() > c0.current_block_last_doc_id() ` a subset of `c1.current_doc_id() > c0.current_doc_id() ` ?

src/superfile/fts/reader.rs
```diff
@@ -1120,34 +1151,127 @@ impl FtsReader {
...
+            if heap.len() >= k {
+                let heap_min = heap.peek().expect("heap len == k").0;
+                let range_start = c0.current_doc_id();
+                let range_end = c0.current_block_last_doc_id();
+                let ub =
+                    c0.current_block_max_bm25() + c1.block_max_in_range(range_start, range_end);
+                if ub <= heap_min {
+                    c0.skip_to(range_end.saturating_add(1), postings);
+                    continue;
                 }
```
Pranav2612000, line 1212: Does doing the same check the other way around also help in making the query even faster?

Pranav2612000 (approved):

Pranav2612000:
@vinaykakade Some conflicts as well

src/superfile/fts/reader.rs
vinaykakade: Good point, done

src/superfile/fts/reader.rs
vinaykakade: Greeat idea, and here is a thought after thinking through. Tl;DR - let us skip this optimization now as it has a chance to be negative, and there are a couple of more optimizations in the pipeline that would make 3+ term and queries even faster.
   
The check would be sound. After c1.skip_to(c0.current_doc_id()) lands c1 in some block, we'd compute ub = c1.current_block_max + c0.block_max_in_range(c1_doc, c1_block_end) and, if ub <= heap_min, skip c0 past c1_block_end. Any AND match in [c1_doc, c1_block_end] is bounded by that UB, so the skip is safe.

However, the the gain is likely small (disclaimer: haven't measured it yet). The current c0-side check already conservatively bounds c1's contribution as the worst c1 block_max over c0's block range — i.e. it uses the strongest c1 sub-block as the UB. The c1-side check would prune c1's individual sub-blocks within that range: finer granularity, not a fundamentally tighter bound. It only fires when:
  - the heap is full (so heap_min is meaningful),
  - c0's block is "strong" enough that the c0-side check didn't already fire,
  - c1's current block is "weak" (low block_max).

For Zipfian shapes that combination is possible but uncommon. Two terms' block_max distributions are largely independent; the rare-term block is usually what drives the bound either way.

Why it might net negative. We already brought two_term_and from 4.4 ms → 113 µs. The dominant win was the flat-merge specialization, not the pruning. At 113 µs we're closer to "index parse + minimum scoring work" than to anything pruning can shrink further. The new check would add one branch + one block_max_in_range walk + one heap.peek() per outer iteration; on a path that's already a tight flat-merge, even small per-iteration overhead can erase the savings. Also, on rare ∧ common shapes the c1-side skip target (c1_block_end + 1) is closer than the c0-side target, so when both checks could fire, the c0-side skips further per hit.

Would rather spend the next pruning effort on generalizing the 2-term flat-merge to the 3-term case since three_wide_and still sits at 2.45 ms, ~20× slower than 2-term, and is the next obvious headroom.

PR #21, opened 2026-05-27
Add rerank codecs for vectors

ekechinwokah:
Adds configurable per-column rerank codecs for vector columns. Each vector column now selects one of four on-disk rerank representations at build time. Default is 8-bit quantization and trades some latency for a significant decrease in memory usage, as we have to do some extra work in reranking to achieve reasonable recall.

- `Fp32`
  - Little-endian fp32
  - `dim * 4` bytes per vector
  - Zero-copy into rerank distance kernels

- `Bf16`
  - Top 16 bits of fp32
  - `dim * 2` bytes per vector
  - SIMD widened via `u16 -> u32 -> fp32` shift before distance compute

- `Sq8`
  - Per-cluster, per-dimension 8-bit scalar quantization
  - `dim * 1` bytes per vector
  - Uses local cluster quantizers rather than global per-column
    quantizers to recover recall on tightly clustered cosine corpora,
    where intra-cluster spread is significantly narrower than
    cross-cluster spread

- `None`
  - No rerank column
  - 1-bit RaBitQ shortlist becomes the final ranking
  - Lowest storage footprint, but reduced recall
  - Shrinks segment size by roughly ~30x at `1M x 384`

## Format changes

The codec discriminator is stored as a single byte at offset `52`
inside the per-column directory entry.

## Sq8 layout

`Sq8` columns now include a sibling `codec_meta` region located between
the `codes` region and the `full[]` region.

The metadata region stores:

- Per-cluster `scale[dim]`
- Per-cluster `offset[dim]`
- Optional per-document decoded norm cache for `L2Sq` columns

The decoded norm cache avoids the per-candidate:
- `u8` widen
- multiply
- dim-wide reduction

during rerank distance evaluation.

## Benchmark Evidence
Metric | main | PR6 (Sq8Residual) | Δ
-- | -- | -- | --
correctness recall@10 | 1.000 | 1.000 | no change
Criterion mean build | 20.691 s | 22.014 s | +6.4% slower
throughput | 48.33 K/s | 45.4 K/s | −6.1%
artifact size | 1515.94 MiB | 790.34 MiB | −48%
peak RSS | 10.47 GiB | 8.06 GiB | −23%
median RSS | 4.89 GiB | 3.50 GiB | −28%
p90 RSS | 7.51 GiB | 5.57 GiB | −26%
cold open | 159.68 ms | 84.31 ms | −47%
first query | 0.61 ms | 0.61 ms | no change

### DIFF
--- a/src/superfile/builder.rs
+++ b/src/superfile/builder.rs
@@ -82,7 +82,10 @@ use crate::superfile::format::footer::{ParquetParts, write_parquet_with_blobs};
 use crate::superfile::format::{self, kv};
 use crate::superfile::fts::builder::FtsBuilder;
 use crate::superfile::fts::tokenize::Tokenizer;
-use crate::superfile::vector::builder::{VectorBuilder, VectorConfig as VecBuildConfig};
+use crate::superfile::vector::builder::VectorBuilder;
+// `VectorConfig` lives in `vector::builder` and is re-exported below
+// (single source of truth post-collapse — see the `pub use` line).
+pub use crate::superfile::vector::builder::VectorConfig;
 use crate::superfile::vector::distance::Metric;
 use arrow_array::{Array, RecordBatch};
 use arrow_schema::{DataType, Schema};
@@ -96,19 +99,11 @@ pub struct FtsConfig {
     pub column: String,
 }
 
-/// Per-column vector configuration. Mirrors the inner
-/// [`VecBuildConfig`] but uses `column` to match the FTS naming for
-/// API consistency at the superfile level.
-#[derive(Clone)]
-pub struct VectorConfig {
-    pub column: String,
-    pub dim: usize,
-    pub n_cent: usize,
-    pub rot_seed: u64,
-    pub metric: Metric,
-    /// On-disk rerank codec. Defaults to [`RerankCodec::Sq8`].
-    pub rerank_codec: crate::superfile::vector::rerank_codec::RerankCodec,
-}
+// `VectorConfig` (the per-column vector config used by
+// `BuilderOptions.vector_columns`) lives in
+// `crate::superfile::vector::builder` and is re-exported at this
+// module path above. Single source of truth — there's no outer
+// wrapper struct.
 
 /// All knobs needed to build a superfile.
 pub struct BuilderOptions {
@@ -336,14 +331,10 @@ impl SuperfileBuilder {
         } else {
             let mut vb = VectorBuilder::new();
             for vc in &opts.vector_columns {
-                vb.register_column(VecBuildConfig {
-                    name: vc.column.clone(),
-                    dim: vc.dim,
-                    n_cent: vc.n_cent,
-                    rot_seed: vc.rot_seed,
-                    metric: vc.metric,
-                    rerank_codec: vc.rerank_codec,
-                })?;
+                // VectorConfig is now the same type at both layers
+                // (re-exported from vector::builder), so the manual
+                // field-by-field bridge is gone — just clone.
+                vb.register_column(vc.clone())?;
             }
             Some(vb)
         };
@@ -548,7 +539,7 @@ fn fts_columns_json(cols: &[FtsConfig]) -> String {
 /// `Serialize` needed.
 ///
 /// Output shape per column:
-/// `{"name":"<escaped>","dim":<u>,"n_cent":<u>,"rot_seed":<u>,"metric":"<l2sq|cosine|negdot>"}`.
+/// `{"column":"<escaped>","dim":<u>,"n_cent":<u>,"rot_seed":<u>,"metric":"<l2sq|cosine|negdot>"}`.
 /// The reader at open time parses this back into
 /// `VectorConfig` to drive distance kernels + IVF probing.
 fn vec_columns_json(cols: &[VectorConfig]) -> String {
@@ -557,7 +548,7 @@ fn vec_columns_json(cols: &[VectorConfig]) -> String {
         if i > 0 {
             s.push(',');
         }
-        s.push_str(r#"{"name":""#);
+        s.push_str(r#"{"column":""#);
         s.push_str(&escape_json(&c.column));
         s.push_str(r#"","dim":"#);
         s.push_str(&c.dim.to_string());
@@ -951,7 +942,7 @@ mod tests {
             rerank_codec: crate::superfile::vector::rerank_codec::RerankCodec::Fp32,
         }];
         let s = vec_columns_json(&cols);
-        assert!(s.contains(r#""name":"emb""#));
+        assert!(s.contains(r#""column":"emb""#));
         assert!(s.contains(r#""dim":384"#));
         assert!(s.contains(r#""n_cent":64"#));
         assert!(s.contains(r#""rot_seed":99"#));
--- a/src/superfile/builder.rs
+++ b/src/superfile/builder.rs
@@ -106,6 +106,8 @@ pub struct VectorConfig {
     pub n_cent: usize,
     pub rot_seed: u64,
     pub metric: Metric,
+    /// On-disk rerank codec. Defaults to [`RerankCodec::Sq8`].
+    pub rerank_codec: crate::superfile::vector::rerank_codec::RerankCodec,
 }
 
 /// All knobs needed to build a superfile.
@@ -340,6 +342,7 @@ impl SuperfileBuilder {
                     n_cent: vc.n_cent,
                     rot_seed: vc.rot_seed,
                     metric: vc.metric,
+                    rerank_codec: vc.rerank_codec,
                 })?;
             }
             Some(vb)
@@ -945,6 +948,7 @@ mod tests {
             n_cent: 64,
             rot_seed: 99,
             metric: Metric::L2Sq,
+            rerank_codec: crate::superfile::vector::rerank_codec::RerankCodec::Fp32,
         }];
         let s = vec_columns_json(&cols);
         assert!(s.contains(r#""name":"emb""#));
--- a/src/superfile/vector/builder.rs
+++ b/src/superfile/vector/builder.rs
@@ -14,6 +14,7 @@ use crate::superfile::format::{self, FST_SEPARATOR, RESERVED_PREFIX};
 use crate::superfile::vector::distance::{Metric, l2_sq};
 use crate::superfile::vector::kmeans::{assign_to_centroids, kmeans};
 use crate::superfile::vector::quant::BitQuantizer;
+use crate::superfile::vector::rerank_codec::RerankCodec;
 use crate::superfile::vector::reservoir::{Reservoir, default_kmeans_sample_size};
 use crate::superfile::vector::rotation::RandomRotation;
 use crate::superfile::vector::spill::{
@@ -45,6 +46,14 @@ fn metric_id(m: Metric) -> u32 {
 }
 
 /// Per-column user-supplied build configuration.
+///
+/// Cloneable — `register_column` takes ownership of one copy
+/// and stores the field-wise data on the corresponding
+/// `ColumnState`; subsequent reads (e.g. for diagnostic
+/// emission) clone from the column to keep the type. Field
+/// additions go behind an `Option` or default so VectorConfig
+/// extensions across minor releases don't force a code change
+/// at the call site.
 #[derive(Debug, Clone)]
 pub struct VectorConfig {
     pub name: String,
@@ -52,6 +61,32 @@ pub struct VectorConfig {
     pub n_cent: usize,
     pub rot_seed: u64,
     pub metric: Metric,
+    /// On-disk rerank codec for this column. See [`RerankCodec`]
+    /// for the supported codecs and their size/recall trade-offs.
+    pub rerank_codec: RerankCodec,
+}
+
+impl VectorConfig {
+    /// Construct a config with the default rerank codec
+    /// ([`RerankCodec::default()`]). Use the `with_*` setters to
+    /// override individual fields.
+    pub fn new(name: String, dim: usize, n_cent: usize, rot_seed: u64, metric: Metric) -> Self {
+        Self {
+            name,
+            dim,
+            n_cent,
+            rot_seed,
+            metric,
+            rerank_codec: RerankCodec::default(),
+        }
+    }
+
+    /// Override the rerank codec.
+    #[must_use]
+    pub fn with_rerank_codec(mut self, codec: RerankCodec) -> Self {
+        self.rerank_codec = codec;
+        self
+    }
 }
 
 /// Default spill threshold: total bytes the in-memory pre-spill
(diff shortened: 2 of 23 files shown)


tests/superfile/vector/reservoir_recall.rs
```diff
@@ -96,6 +97,7 @@ fn build_reader_with_sample_size(
             n_cent,
             rot_seed: 7,
             metric: Metric::Cosine,
+            rerank_codec: RerankCodec::Fp32,
```
Pranav2612000: Should we use this as a test parameter and test for all the different types? Or do you think that this is overkill?

src/superfile/vector/rerank_codec.rs
```diff
@@ -0,0 +1,421 @@
...
+    /// `codec_meta` region; per-vector body is `dim` u8s. The
+    /// distance kernel fuses dequant with the per-candidate
+    /// distance.
+    Sq8,
+    /// No rerank column at all. The shortlist (1-bit RaBitQ
+    /// scores) is the final ranking. Opt-in — recall drops
+    /// 0.05–0.15 on typical normalized-Gaussian / image-
+    /// embedding corpora; trade-off is a ~30× segment-size
+    /// shrink at 1M × 384.
+    None,
```
Pranav2612000: Nit: Looking at this enum, it feels None would be the default case ( but Sq8 is the default ). "None" also conflicts with the Option's None. Possible to rename this to something else?

src/superfile/builder.rs
```diff
@@ -106,6 +106,8 @@ pub struct VectorConfig {
     pub n_cent: usize,
     pub rot_seed: u64,
     pub metric: Metric,
+    /// On-disk rerank codec. Defaults to [`RerankCodec::Sq8`].
+    pub rerank_codec: crate::superfile::vector::rerank_codec::RerankCodec,
```
Pranav2612000: Also needs updates to benchmarks ( both infino & others )

src/superfile/vector/builder.rs
```diff
@@ -45,13 +46,47 @@ fn metric_id(m: Metric) -> u32 {
...
+///
+/// Cloneable — `register_column` takes ownership of one copy
+/// and stores the field-wise data on the corresponding
+/// `ColumnState`; subsequent reads (e.g. for diagnostic
+/// emission) clone from the column to keep the type. Field
+/// additions go behind an `Option` or default so VectorConfig
+/// extensions across minor releases don't force a code change
+/// at the call site.
 #[derive(Debug, Clone)]
 pub struct VectorConfig {
```
Pranav2612000, line 58: Not related to this PR, but why do we have two structs for VectorConfig? Can we not do with one?

src/superfile/vector/builder.rs
```diff
@@ -716,15 +765,24 @@ fn build_subsection_streaming(
 
     // ---- Pass 3: read each bucket sequentially, materialise the
     // cluster-contiguous regions and the cluster index ----
+    //
+    // `RerankCodec::None` segments have no `full[]` region on disk
+    // and pass 2 did not spill the fp32 vectors at all, so we skip the
+    // `full_layout` allocation and the per-row `full_vec` read.
+    let codec = cfg.rerank_codec;
+    let has_full = !matches!(codec, RerankCodec::None);
```
Pranav2612000: Nit: expose this as a method of the `rerank_codec` struct

src/superfile/vector/builder.rs
```diff
@@ -716,15 +765,24 @@ fn build_subsection_streaming(
...
     let mut codes_layout = vec![0u8; n_docs * code_bytes];
-    let mut full_layout = vec![0f32; n_docs * dim];
+    let mut full_layout: Vec<f32> = if has_full {
```
Pranav2612000: Nit: possible to use Option to make the code more "rustic" + reduce allocations

src/superfile/vector/builder.rs
```diff
@@ -746,44 +804,73 @@ fn build_subsection_streaming(
             let dst_code =
                 &mut codes_layout[write_cursor * code_bytes..(write_cursor + 1) * code_bytes];
             reader.read_exact(dst_code)?;
-            let dst_full = &mut full_layout[write_cursor * dim..(write_cursor + 1) * dim];
-            let dst_full_bytes: &mut [u8] = bytemuck::cast_slice_mut(dst_full);
-            reader.read_exact(dst_full_bytes)?;
+            if has_full {
+                let dst_full = &mut full_layout[write_cursor * dim..(write_cursor + 1) * dim];
+                let dst_full_bytes: &mut [u8] = bytemuck::cast_slice_mut(dst_full);
+                reader.read_exact(dst_full_bytes)?;
```
Pranav2612000: Discuss: Instead of copying the whole data, and then reducing it to the reranker format, is it better to copy only the data needed, reducing it as we copy? seems this can also be SIMD'fied but maybe LLVM is already vectorizing it

src/superfile/vector/builder.rs
```diff
@@ -746,44 +804,73 @@ fn build_subsection_streaming(
...
+    let full_size = codec.per_vector_bytes(dim) * n_docs;
     let doc_ids_size = n_docs * 4;
 
     // Offsets relative to subsection start.
     let summary_off = SUB_HEADER_SIZE;
     let centroids_off = summary_off + summary_size;
     let cluster_idx_off = centroids_off + centroids_size;
     let codes_off = cluster_idx_off + cluster_idx_size;
-    let full_off = codes_off + codes_size;
+    // `codec_meta_off` is the on-disk slot in the sub-header for
```
Pranav2612000: Is this a backward incompatible change? Because older files would expect full_off to be present here

src/superfile/vector/builder.rs
```diff
@@ -897,19 +1247,20 @@ fn run_pass2(
...
         for r in 0..actual_rows {
             let cid = asgn[r] as usize;
             let local_doc_id = global_doc_id + r as u32;
             let writer = &mut bucket_writers[cid];
             writer.write_all(&local_doc_id.to_le_bytes())?;
             writer.write_all(&chunk_codes[r * code_bytes..(r + 1) * code_bytes])?;
-            writer.write_all(bytemuck::cast_slice(&chunk[r * dim..(r + 1) * dim]))?;
+            if write_full {
+                writer.write_all(bytemuck::cast_slice(&chunk[r * dim..(r + 1) * dim]))?;
+            }
```
Pranav2612000, line 1259: Here as well, we're writing full data + reading full data and then reducing it based on the rerank format. Can we instead write the reranked data directy? The layers above just need to consolidate and write this data

src/superfile/vector/distance.rs
```diff
@@ -189,6 +191,283 @@ fn l2_sq_le_bytes_unaligned(query: &[f32], bytes: &[u8]) -> f32 {
...
+        Metric::NegDot => -dot_bf16_bytes(query, bytes),
+    }
+}
+
+/// Encode an fp32 value as bf16 with round-to-nearest-even on the
+/// truncated low 16 bits. NaN inputs return a sign-preserving bf16
+/// NaN. Widening back to f32 is exact: the bf16 bits become the top
+/// 16 bits of the fp32 representation and the low 16 bits are zero.
+#[inline]
+pub(crate) fn fp32_to_bf16(x: f32) -> u16 {
```
Pranav2612000, line 358: Similar to an above comment, does SIMDifying provide benefit? LLVM may be auto vectorizing this anyway

src/superfile/vector/builder.rs
```diff
@@ -897,19 +1247,20 @@ fn run_pass2(
...
-        // — BufWriter is !Sync and a per-bucket Mutex would
-        // serialize anyway. The sequential write is dominated
-        // by the kernel-buffered write path (BufWriter
-        // amortises to ~one syscall per 64 KiB / 1 588 B ≈ 41
-        // rows at dim=384), not by the in-process loop body.
+        // Route rows to bucket writers. Sequential per-bucket:
+        // BufWriter is !Sync and a per-bucket Mutex would serialize
+        // anyway. `None` columns skip the fp32 vector spill because
+        // they have no `full[]` region on disk.
+        let write_full = !matches!(codec, RerankCodec::None);
```
Pranav2612000: Is this an expected change? Previously for the Rabit1Q algorithm, which is the one when the codec is `None`, we were writing the full sections and using them to improve the quality of our results.

src/superfile/vector/reader.rs
```diff
@@ -1287,6 +1542,828 @@ mod tests {
...
+        let blob = b.finish().expect("finish");
+        let json =
+            r#"[{"name":"v","dim":128,"n_cent":4,"rot_seed":11,"metric":"l2sq"}]"#.to_string();
+        let r = VectorReader::open(Bytes::from(blob), &json).expect("open");
+
+        let hits = r
+            .search("v", &all[17], 5, n_cent, 5)
+            .expect("None-codec search must succeed");
+        assert!(
+            hits.iter().any(|(did, _)| *did == 17),
```
Pranav2612000: Shouldn't an exactly matching query return the exact vector first, instead of .any() ?

tests/superfile/vector/reservoir_recall.rs
ekechinwokah: done. loop over fp32 / bf16 / sq8 with the same 0.85 floor. left rabitqonly out — different recall regime since it skips rerank; covered by `rabitq_only_self_query_ranks_self_first` in `reader.rs`.

src/superfile/vector/rerank_codec.rs
ekechinwokah: renamed to `rabitqonly`. describes what search actually does, and kills the option::none collision. on-disk discriminator unchanged.

src/superfile/builder.rs
ekechinwokah: infino bench fixtures land in the follow-up pr.

src/superfile/vector/builder.rs
ekechinwokah: done — added `rerankcodec::writes_full()`. replaced both inline matches! checks.

src/superfile/vector/builder.rs
ekechinwokah: done. option<vec<f32>> with as_deref().expect(...) in the codec arms; debug-assert on rabitqonly guards the invariant.

src/superfile/vector/builder.rs
ekechinwokah: backward-compatible. codec_meta_off sits in what was a reserved (zero) slot. pre-pr fp32 segments read back as codec_meta_off = 0 + codec id 0 = fp32 — identical round-trip. comment at the parser site spells this out.

src/superfile/vector/builder.rs
ekechinwokah: yes it is. more vectorization coming in follow up pr.

src/superfile/vector/builder.rs
ekechinwokah: yeah, valid. done: drop fp32 round-trip and stream-encode per codec.

src/superfile/vector/distance.rs
ekechinwokah: not sure in this case — it's a single >> 16, and the call sites are tight loops llvm auto-vectorizes. let me test in follow-on pr.

src/superfile/vector/builder.rs
ekechinwokah: this is the opt-in trade-off. default is sq8, not rabitqonly, so existing call sites without an explicit codec keep rerank. rabitqonly is the explicit "i want the ~30× shrink, i accept the documented 0.05–0.15 recall drop" knob, which IMO needs to be a deliberate user tradeoff given impact to result quality.

src/superfile/vector/reader.rs
ekechinwokah: done.

src/superfile/vector/builder.rs
ekechinwokah: good point. consolidated.

src/superfile/builder.rs
Pranav2612000: Yes. And also the other ones. Things seem to breaking there with this new param

src/superfile/vector/builder.rs
Pranav2612000: Fair. Is there a use case, where I want the raw vectors to be store ( fp32 ) to serve some rare queries. But I want most of my queries to complete quickly, with the accepted tradeoff in recall quality. 
Do we support this right now?

src/superfile/vector/builder.rs
```diff
@@ -714,101 +772,247 @@ fn build_subsection_streaming(
...
+                        let x = (qc as f32) * scale_c[d] + offset_c[d];
+                        acc += (x as f64) * (x as f64);
+                    }
+                }
+                if let Some(norms_off) = sq8_norms_block_off {
+                    let n_off = norms_off + pos * 4;
+                    bytes[n_off..n_off + 4].copy_from_slice(&(acc as f32).to_le_bytes());
+                }
+            }
+        }
```
Pranav2612000: Can this also be written as a part of the cluster's temporary file during pass2? If not lets move this to a separate function ( inlined ) for readability?

Ideally, pass2 creates the temporary file for each cluster

The builder just reads the files and combines them one at a time

src/superfile/vector/builder.rs
```diff
@@ -714,101 +772,247 @@ fn build_subsection_streaming(
...
+        if codec == RerankCodec::Sq8 {
+            sq8_scratch.clear();
+            sq8_scratch.resize(cluster_count * dim, 0.0);
+        }
+
+        for i in 0..cluster_count {
+            let pos = cluster_off + i;
+
+            // doc_id (u32 LE).
+            reader.read_exact(&mut id_buf)?;
```
Pranav2612000: These multiple small reads for each doc_id may be expensive. Is it possible to restructure the code to read all doc_ids, codes, and docs ( if full_mode ) in a single read each ( the same way we do for a write? )

src/superfile/vector/builder.rs
```diff
@@ -714,101 +772,247 @@ fn build_subsection_streaming(
...
+            let did_off = doc_ids_off + pos * 4;
+            bytes[did_off..did_off + 4].copy_from_slice(&id_buf);
+
+            // 1-bit RaBitQ code.
+            let code_off_b = codes_off + pos * code_bytes;
```
Pranav2612000: Is this needed also for the non RaBitQ variants?

Pranav2612000: Ignore. I saw that we use RaBitQ in search to shortlist candidates

src/superfile/vector/builder.rs
ekechinwokah: done.

src/superfile/vector/builder.rs
ekechinwokah: done.

src/superfile/vector/builder.rs
ekechinwokah: there may be, but we do not support this right now. we are able to select at build time whether to go with full fidelity vectors or quantized (16, 8, or 1 bit), but i would suggest we not expose this through the api initially.

Pranav2612000 (approved):

PR #22, opened 2026-05-27
fts: generalize flat-merge AND intersection to n >= 3 terms

vinaykakade:
## Summary

PR #17 specialized the 2-term AND path to a flat sorted-merge over decoded `block_doc_ids` arrays, avoiding the per-doc `skip_to` overhead the leapfrog path incurred. The general n ≥ 3 path still leapfrogged per doc, so 3+ term shapes kept the slower behavior.

This change generalizes the flat-merge to any n ≥ 2:

- Leader walks its decoded `block_doc_ids` with a plain `i++` cursor.
- Every non-leader gets its `pos` advanced with a tight while-scan over its decoded buffer — no `skip_to` in the inner loop.
- When any cursor exhausts its current block, the outer loop crosses blocks via `next()` and re-aligns via the existing alignment step (amortized — alignment fires once per leader block, not per leader doc).
- Block-max-AND pruning at the top of each outer iteration is unchanged.

The 2-term specialization (`run_and_intersect_2term`) stays in place because its hand-named `c0`/`c1` locals stay register-resident across the inner loop, where the general path goes through `others.iter_mut()` slice indirection.

## Results (1M docs, single superfile)

Numbers are infino-only, before this change → after:

| Shape | Before | After | Speedup |
|---|---|---|---|
| `two_term_and` | 113 µs | 116.57 µs | ≈ tied |
| `three_wide_and` | 2.45 ms | **2.03 ms** | 1.21× |
| `three_similar_and` | 4.32 ms | **3.31 ms** | 1.30× |
| `five_term_and` | 4.50 ms | **3.97 ms** | 1.13× |

OR shapes untouched and continue to route through MaxScore+BMM.

## Notes

Closing the gap further between 2-term (116 µs) and n-term (2-4 ms) likely needs either (a) an n=3 specialization with hand-rolled cursor state mirroring the 2-term shape, or (b) hoisting cursor fields out of the `iter_mut` indirection into stack-allocated arrays. Both want profiling first to confirm where time is going — saving for a follow-up.

## Test plan

- [x] `cargo test --test superfile -- brute_force_oracle::oracle_and` (6 AND oracles pass).
- [x] `cargo test --test supertable -- brute_force_oracle` (14 oracles incl. 4 AND, all pass).
- [x] `cargo test --lib` (534 passing).
- [x] `cargo fmt --check` clean.
- [x] `cargo clippy --all-targets -- -D warnings` clean.
- [x] `INFINO_BENCH_UPDATE_README=1 cargo bench --bench fts -- superfile_fts_search` produces the numbers above; README refreshed in place.

### DIFF
--- a/src/superfile/fts/reader.rs
+++ b/src/superfile/fts/reader.rs
@@ -1092,10 +1092,15 @@ impl FtsReader {
         Ok(out)
     }
 
-    /// General `n >= 2`-term AND path. Block-max-AND prunes whole
-    /// leader blocks whose maximum possible AND score sum can't beat
-    /// the current heap floor; for survivors, runs the leapfrog
-    /// convergence over all cursors per leader doc.
+    /// General `n >= 3`-term AND path. Same shape as the 2-term path:
+    /// block-max pruning at the top, then a flat-merge over the
+    /// leader's decoded `block_doc_ids` against each non-leader's
+    /// decoded `block_doc_ids`. For each leader doc, every non-leader's
+    /// `pos` is advanced with a tight `pos += 1` scan instead of
+    /// `skip_to` — no function-call or within-block linear-scan
+    /// overhead per leader doc, just integer comparisons over the
+    /// already-decoded buffers. When any cursor exhausts its block,
+    /// the outer loop crosses blocks via `next()` and re-aligns.
     fn run_and_intersect_general(
         &self,
         cursors: &mut [TermCursor],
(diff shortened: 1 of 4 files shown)


src/superfile/fts/reader.rs
```diff
@@ -1130,44 +1135,110 @@ impl FtsReader {
...
+            // Largest landing-doc becomes the new alignment target if
+            // any cursor jumped past leader. If any cursor crossed
+            // leader's current block, restart the outer loop so pruning
+            // re-fires on leader's new block; otherwise the flat-merge
+            // proceeds in the current decoded blocks.
+            let leader_doc = cursors[0].current_doc_id();
+            let leader_block_end = cursors[0].current_block_last_doc_id();
+            let mut max_other = leader_doc;
+            let mut crossed_block = false;
+            for c in cursors[1..].iter_mut() {
```
Pranav2612000, line 1148: As an optimization, while creating the cursors, can we order them based on the number of docs ( leader is the one with the least docs )?

Pranav2612000 (approved):

src/superfile/fts/reader.rs
vinaykakade: Yes, we already do it. See line 1066 in reader.rs:

cursors.sort_by_key(|c| c.block_count());

PR #24, opened 2026-05-28
Add SIMD fast paths to vectors and remove Bf16 codec

ekechinwokah:
## Summary

This PR adds runtime-dispatched SIMD fast paths for the vector stack. 

It includes three main pieces:

- AVX-512 runtime dispatch for hot vector and bloom kernels.
- AVX2 baseline tightening for Sq8 rerank paths.
- Build-side Sq8 SIMD work: per-cluster min/max folded into pass 2, plus SIMD `f32 → u8` Sq8 encode.
- Removed Bf16 codec - did not produce material advantage over sq8 during tests

## Key results

Tested on AWS `c7i.4xlarge` (Sapphire Rapids).

### Vector distance kernels — ns/call

| kernel | dim | scalar | wide | AVX2 | AVX-512 |
|---------|----:|-------:|-----:|-----:|---------:|
| `distance::dot` (fp32) | 384 | 207.8 | 19.8 | wide | 13.8 |
| `distance::l2_sq` (fp32) | 384 | 217.6 | 23.5 | wide | 17.6 |
| `Sq8Kernel::distance_at` (dot) | 384 | 220.0 | 82.7 | 38.1 | 21.0 |
| `distance::dot` (fp32) | 1536 | 917.6 | 91.5 | wide | 74.1 |
| `distance::l2_sq` (fp32) | 1536 | 929.6 | 106.6 | wide | 79.0 |
| `Sq8Kernel::distance_at` (dot) | 1536 | 931.6 | 355.6 | 191.8 | 89.3 |

### Sq8 f32 → u8 encode — ns/row

| dim | scalar | wide | AVX2 | AVX-512 |
|----:|-------:|-----:|-----:|---------:|
| 384 | 364.0 | 298.4 | 53.4 | 31.4 |
| 1024 | 967.0 | 795.6 | 143.6 | 84.5 |
| 1536 | 1445.4 | 1195.0 | 212.1 | 125.6 |

### RaBitQ `estimate_dot_rotated` — ns/call

| dim | wide | AVX-512 | speedup |
|----:|-----:|---------:|--------:|
| 384 | 33.8 | 25.9 | 1.30× |
| 1024 | 86.2 | 67.2 | 1.28× |
| 1536 | 128.1 | 98.5 | 1.30× |


src/superfile/vector/builder.rs
```diff
@@ -1016,155 +1089,522 @@ fn build_subsection_streaming(
...
+    inv_scale: Vec<f32>,
+    c2: Vec<f32>,
+}
+
+impl Sq8EncodeConsts {
+    fn from_scale_offset(scale: &[f32], offset: &[f32]) -> Self {
+        debug_assert_eq!(scale.len(), offset.len());
+        let dim = scale.len();
+        let mut inv_scale = vec![0.0f32; dim];
+        let mut c2 = vec![0.0f32; dim];
```
Pranav2612000: Nit: unnecessary initialization to 0. We can initialize with capacity and just append.

src/superfile/vector/builder.rs
```diff
@@ -1016,155 +1089,522 @@ fn build_subsection_streaming(
...
+            let q = _mm256_fmadd_ps(r, inv, c);
+            // clamp to [0, 255]
+            let q_clamped = _mm256_max_ps(_mm256_min_ps(q, max255), zero);
+            // truncate-to-i32; lanes are in [0, 255] so the i32
+            // values are non-negative and PACKUSDW interprets them
+            // as unsigned correctly.
+            let q_i32 = _mm256_cvttps_epi32(q_clamped);
+            // Narrow 8 × i32 → 8 × u8 via two saturating packs.
+            let lo = _mm256_castsi256_si128(q_i32);
+            let hi = _mm256_extracti128_si256::<1>(q_i32);
```
Pranav2612000: Does this compile? This documentation https://doc.rust-lang.org/beta/core/arch/x86_64/fn._mm256_extracti128_si256.html says the function takes 2 parameters.

Also, according to the non-simd wide mode fn, we directly copy the bytes of the q_clamped in dst, why do we need this extra packing for SIMD versions? ( I'm sure i'm missing something obvious 😅 )

src/superfile/vector/builder.rs
```diff
@@ -1016,155 +1089,522 @@ fn build_subsection_streaming(
...
+/// is a no-op because we already clamped to `[0, 255]` in float
+/// space; it's the cheapest path that produces packed u8.
+///
+/// # Safety
+///
+/// Callers must ensure the target supports `avx2` and `fma`.
+#[cfg(target_arch = "x86_64")]
+#[target_feature(enable = "avx2,fma")]
+#[allow(clippy::manual_clamp)] // see sq8_encode_row_wide note
+unsafe fn sq8_encode_row_avx2(row: &[f32], inv_scale: &[f32], c2: &[f32], dst: &mut [u8]) {
```
Pranav2612000: Do we have tests for these encode_row SIMD functions?

src/superfile/vector/builder.rs
```diff
@@ -1016,155 +1089,522 @@ fn build_subsection_streaming(
...
+/// where the AVX-512 path picks up its extra factor over AVX2
+/// on top of the 2× lane count.
+///
+/// # Safety
+///
+/// Callers must ensure the target supports `avx512f`.
+#[cfg(target_arch = "x86_64")]
+#[target_feature(enable = "avx512f")]
+#[allow(clippy::manual_clamp)] // see sq8_encode_row_wide note
+unsafe fn sq8_encode_row_avx512(row: &[f32], inv_scale: &[f32], c2: &[f32], dst: &mut [u8]) {
```
Pranav2612000: Thoughts on moving all SIMD operations to a separate file, keeping the `builder.rs` as the file which is responsible for the build functions, and the SIMD file as the implementations for the tools needed to build.

ekechinwokah: makes sense. done.

src/superfile/vector/builder.rs
ekechinwokah: good catch — fixed.

src/superfile/vector/builder.rs
ekechinwokah: yes it does. rust's stable intrinsic moved the imm8 to a const generic, so `::<1>(q_i32)` is one runtime arg + one const — the doc page you linked shows the C-API 2-arg signature; the rust binding uses turbofish for the immediate. added a comment to that effect.

on "why pack at all": the wide path isn't actually copying bytes — `q_clamped[k]` as u8 per lane is eight scalar narrows + eight 1-byte stores after spilling the f32x8 to a stack array. VPACKUSDW + VPACKUSWB is the SIMD-native form of the same narrow, finished by one 8-byte unaligned store. same thing, just vectorized.

src/superfile/vector/builder.rs
ekechinwokah: yes — `sq8_encode_row_simd_paths_match_scalar` sweeps and asserts byte-identical output of wide / avx2 / avx512 against an FMA scalar reference. sibling test `update_min_max_simd_paths_match_scalar` does the same for the pass-2 kernel.

src/superfile/vector/reader.rs
```diff
@@ -455,7 +455,7 @@ impl VectorReader {
             let rerank_codec = RerankCodec::from_codec_id(codec_id).ok_or_else(|| {
                 VectorError::Read(ReadError::MalformedVersion(format!(
                     "column '{}' has unknown rerank-codec id {codec_id} \
-                     (known ids: 0=fp32, 1=bf16, 2=sq8, 3=none)",
+                     (known ids: 0=fp32, 2=sq8, 3=none; id 1 retired)",
```
Pranav2612000: Its a better idea to retire for backward compatibility, but since we're still nascent, should we just move everything by 1, instead of retiring 1?

src/superfile/vector/builder.rs
```diff
@@ -904,17 +928,59 @@ fn build_subsection_streaming(
...
+        Vec::new()
+    };
+    if codec == RerankCodec::Sq8 {
+        for (c, (scale_c, offset_c)) in sq8_quantizers.iter().enumerate() {
+            let sc_off = sq8_scale_block_off + c * dim * 4;
+            bytes[sc_off..sc_off + dim * 4].copy_from_slice(bytemuck::cast_slice(scale_c));
+            let oc_off = sq8_offset_block_off + c * dim * 4;
+            bytes[oc_off..oc_off + dim * 4].copy_from_slice(bytemuck::cast_slice(offset_c));
+        }
+    }
```
Pranav2612000: Not sure about the performance implications, but could this be moved inside the match arm where we copy full bytes? ( Replacing the sq8_quantizers loop, with the loop for each cluster )

IMO, keeps code cleaner by ensuring all writes for a type near each other instead of scattering around the codebase

src/superfile/vector/builder.rs
```diff
@@ -1016,151 +1067,96 @@ fn build_subsection_streaming(
...
 ) {
     debug_assert_eq!(cluster_rows.len(), cluster_count * dim);
 
-    let (scale_c, offset_c) = compute_sq8_quantizer_for_cluster(cluster_rows, dim, cluster_count);
-
-    let sc_off = sq8_scale_block_off + c * dim * 4;
-    bytes[sc_off..sc_off + dim * 4].copy_from_slice(bytemuck::cast_slice(&scale_c));
-    let oc_off = sq8_offset_block_off + c * dim * 4;
-    bytes[oc_off..oc_off + dim * 4].copy_from_slice(bytemuck::cast_slice(&offset_c));
-
```
Pranav2612000, line 1050: Reason for moving this from here to outside? Having it here reduces complexity of its caller, improving readability

src/superfile/vector/builder.rs
```diff
@@ -1016,151 +1067,96 @@ fn build_subsection_streaming(
...
+        // never decoded because no doc was assigned).
+        let span = max[d] - min[d];
+        if span > 0.0 && span.is_finite() {
+            offset[d] = min[d];
+            scale[d] = span / 255.0;
+        } else {
+            offset[d] = if min[d].is_finite() { min[d] } else { 0.0 };
+            scale[d] = 1.0;
+        }
     }
```
Pranav2612000, line 1147: should this also be simd'ified? ( Maybe in a separate PR )

src/superfile/vector/distance.rs
```diff
@@ -12,6 +12,7 @@
 use wide::f32x8;
 
 use crate::superfile::vector::rerank_codec::RerankCodec;
+use crate::superfile::vector::simd_dispatch::{avx2_enabled, avx512_enabled};
```
Pranav2612000, line 16: Only import these with the appropriate target architecture cfg. Throws  unused imports warnings on mac

src/superfile/vector/distance.rs
```diff
@@ -281,45 +418,28 @@ impl<'a> Sq8Kernel<'a> {
...
-        let mut cross = acc.reduce_add();
-        while i < self.dim {
-            cross += self.q_prime[i] * (code_bytes[i] as f32);
-            i += 1;
-        }
+        // Per-doc inner reduction: Σ_d q_prime[d] * code[d] as f32.
+        // Dispatches to AVX-512 (16-lane FMA with VPMOVZXBD widen)
+        // when the runtime gate passes; otherwise the f32x8 widen-
+        // and-FMA kernel that has shipped since 012.
+        let cross = sq8_cross_product(&self.q_prime, code_bytes, self.dim);
```
Pranav2612000: Is this cross product or dot product? shouldn't cross product return a vector back?

src/superfile/vector/distance.rs
```diff
@@ -330,182 +450,178 @@ impl<'a> Sq8Kernel<'a> {
...
-            let lane = f32x8::from(
-                <[f32; 8]>::try_from(c).expect("chunks_exact(8) yields slices of length 8"),
-            );
-            acc += lane * lane;
-        }
-        for &x in tail {
-            tail_acc += x * x;
-        }
-        (acc.reduce_add() + tail_acc).sqrt()
-    };
```
Pranav2612000, line 496: Did this not have performance improvements? If this function is only being used in tests, add #[cfg(test)]

Pranav2612000 (approved):

src/superfile/vector/reader.rs
ekechinwokah: agreed. fixed.

src/superfile/vector/builder.rs
ekechinwokah: done. dropped the up-front `Vec<Sq8EncodeConsts>` entirely; `Sq8EncodeConsts::from_scale_offset(scale_c, offset_c)` is now derived inline in the `RerankCodec::Sq8` arm of the per-cluster loop.

src/superfile/vector/builder.rs
ekechinwokah: well the original reason was that the helper computed (scale, offset) from `cluster_rows`. once we folded min/max accumulation into pass 2, the quantizers are known up front — so there was nothing left to compute per-cluster, only bytes to copy.

per your suggestion I moved the (scale, offset) `codec_meta` write back into the per-cluster loop, placed just above the empty-cluster continue so empty clusters still get their sentinel bytes (on-disk byte pattern is preserved). all Sq8 per-cluster writes — codec_meta + per-row encode — now live together in one place.

src/superfile/vector/distance.rs
ekechinwokah: gated on `#[cfg(target_arch = "x86_64")]`. also fixed the same pattern in three other files that imported `simd_dispatch` gates at module level: `sq8_simd.rs`, `quant.rs`, and `supertable/manifest/bloom.rs`.

src/superfile/vector/distance.rs
ekechinwokah: bad naming. it's a dot product. fixed naming across the code.

src/superfile/vector/builder.rs
ekechinwokah: not on the hot path. probably not worth it.

src/superfile/vector/distance.rs
ekechinwokah: good catch. restored.

PR #27, opened 2026-05-31
WAL-driven updates + deletes

vinaykakade:
## Summary

End-to-end WAL-driven `update()` + `delete()` for the supertable, plus
the supporting recovery / GC / tombstone-aware-compaction infrastructure.
Lands milestones M1–M8 of the update/delete design on a single long-lived
`updates` branch.

Public API additions (`SupertableWriter`):

- `update(predicate, new_rows) -> Result<PendingUpdate, MutationError>`
  buffers a 1:1-cardinality replacement.
- `delete(predicate) -> Result<PendingDelete, MutationError>` buffers a
  predicate-driven delete.
- `commit() -> Result<CommitResult, CommitError>` drains the pending-
  appends buffer, then drives buffered updates + deletes through the
  WAL pipeline in order. Surfaces `PartialCommit { committed_wal_ids, … }`
  on mid-flush failure; the recovery sweep on the next supertable open
  completes any WAL that landed durably before the failure.

`Supertable::create` is now create-or-open: when storage already carries
a committed pointer, `create` transparently delegates to `open` instead
of silently shadowing the prior state with an empty manifest. Stays sync
(returns `Result<Self, OpenError>`); internally bridges to async I/O for
the pointer probe.

## What's in each milestone

- **M1** — `wal::state_doc` (state machine + JSON schema), `wal::persistence`
  (etag-CAS WalStore + arrow sidecar PUT/GET with blake3 verify),
  `wal::tombstones_codec` (hand-rolled binary framing for the per-superfile
  bitmap sidecar).
- **M2** — `wal::pipeline::run_append_phase`: `_id`-prefix the user batch,
  build the superfile, swap the manifest via the writer's `persist_commit`,
  advance WAL state to Appended. Idempotent + replay-deterministic.
- **M3** — `wal::pipeline::run_tombstone_phase`: per-target resolve against
  the manifest's `[id_min, id_max]` prune, per-superfile sidecar CAS-PUT
  with seal detection, per-target + final WAL state CAS. Reader-side
  `tombstones::SidecarCache` (DashMap + TTL gate) + filter hooks in FTS,
  vector, and SQL query paths.
- **M4** — `wal::lease` (CAS-only acquire / heartbeat / release + a
  background heartbeat task with progress-check stuck-worker detection)
  and `wal::recovery::scan_and_recover` (on-demand sweep; runs at
  `Supertable::create` / `open` time and via the explicit
  `run_recovery_sweep_once()` operator hatch).
- **M5** — `Supertable::scan_ids_matching(expr)` predicate-resolution
  primitive over DataFusion, plus the public `update` / `delete` /
  `commit` entry points on `SupertableWriter`.
- **M6** — `wal::tombstones_admin::{seal, live_rows}` helpers for a
  future tombstone-aware compactor; race-window test (writer-then-seal
  leaves the bit visible to the compactor).
- **M7** — `wal::gc::run_sweep`: reaps Complete WALs past `T_wal_grace`
  + orphan arrow sidecars past `T_sidecar_grace`. Wired into the
  open-time sweep alongside recovery.
- **M8** — `benches/supertable_update` throughput bench + a
  `tests/supertable/update_crash_property.rs` harness that drives
  recovery against every legal intermediate WAL shape and pins the
  end-state invariant.

## Notable design points

- **Append-first-then-tombstone ordering** at commit time: during the
  transient window between the new rows landing in the manifest and the
  per-target tombstones taking effect, queries see BOTH old and new
  rows. Eventually consistent, never *eventually empty*.
- **Captured-at-call semantics**: `update`/`delete` resolve their
  predicates against the manifest snapshot at the call site; rows that
  newly match between the call and the commit are NOT tombstoned.
- **CAS layer is the safety boundary**: leases are advisory; every
  per-step state transition surfaces through `put_if_match`. Two-process
  contention is correct by construction.
- **Sync API everywhere**: writer paths stay sync via the established
  `Handle::try_current + block_in_place` pattern; async callers are an
  opt-in.

## Diff size

- +9.5K / -150 lines across 48 files.
- New tests: ~25 unit tests + 4 integration test files
  (`writer_mutations`, `wal_recovery`, `update_crash_property`,
  `query/tombstone_filter`).
- 956 tests passing (737 lib + 219 integration); coverage at 91.41 %
  lines (87.09 % functions, 90.46 % regions); CI gate is green.

## Test plan

- [x] `cargo test --lib` — 737 passing
- [x] `cargo test` (full matrix) — 956 passing
- [x] `make ci` — fmt-check ✓, clippy `-D warnings` ✓, coverage gate ✓
- [x] New integration tests covering: predicate-driven delete +
  follow-up SQL/FTS exclusion; predicate-driven update + 1:1
  cardinality enforcement; open-time recovery sweep against
  pre-seeded Intent WALs; expired-lease ("process A died") preemption;
  create-against-existing-pointer delegates to open
- [x] Crash-property test sweeps 6 legal intermediate WAL shapes
  through recovery
- [ ] Manual smoke against an S3 backend (LocalFs only in CI)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/supertable/wal/persistence.rs
```diff
@@ -0,0 +1,881 @@
...
+            }
+        };
+        let bytes = self
+            .storage
+            .get(&path)
+            .await
+            .map_err(|source| WalStoreError::Storage {
+                path: path.clone(),
+                source,
+            })?;
```
Pranav2612000: The underlying object_store get api also includes the result etag. Can we use that directly instead of a separate head request?

src/supertable/wal/persistence.rs
```diff
@@ -0,0 +1,881 @@
...
+                return Err(WalStoreError::CasFailed { path });
+            }
+            Err(other) => {
+                return Err(WalStoreError::Storage {
+                    path,
+                    source: other,
+                });
+            }
+        }
+        let meta = self
```
Pranav2612000: Avoid this head call by using the e_tag returned by the put api

src/supertable/wal/persistence.rs
```diff
@@ -0,0 +1,881 @@
...
+        let meta = match self.storage.head(&path).await {
+            Ok(m) => m,
+            Err(StorageError::NotFound { .. }) => return Ok(None),
+            Err(other) => {
+                return Err(WalStoreError::Storage {
+                    path,
+                    source: other,
+                });
+            }
+        };
```
Pranav2612000: avoid this head call by using the e_tag that the `get` api returns

Cargo.toml
```diff
@@ -68,6 +68,11 @@ uuid = { version = "1", features = ["v4"] }
 xxhash-rust = { version = "0.8", features = ["xxh3"] }
 num_cpus = "1"
 
+# Tombstone bitmaps.
```
Pranav2612000, line 71: Nit: Also check in updates to the Cargo.lock file to lock the nested dependencies

src/supertable/wal/state_doc.rs
```diff
@@ -0,0 +1,525 @@
...
+    pub predicate_repr: String,
+
+    /// The resolved set the predicate matched at call-time
+    /// (against the supertable's current manifest snapshot).
+    /// Recovery replays from this list directly — the predicate
+    /// itself is never re-evaluated, so the resolution is
+    /// stable across whatever manifest churn happens between
+    /// the original call and a later recovery. Length is
+    /// bounded by the per-mutation target cap.
+    pub target_ids: Vec<WalId>,
```
Pranav2612000: Should this be WalId?

src/supertable/wal/state_doc.rs
```diff
@@ -0,0 +1,525 @@
...
+    /// by what `IdGenerator` happens to produce at replay time.
+    #[serde(default)]
+    pub minted_id_spans: Vec<IdSpan>,
+
+    /// Range of `(superfile_id, doc_id)` the append phase
+    /// produced. Filled in once the manifest swap that
+    /// publishes the new superfile has committed; `None`
+    /// until then.
+    #[serde(default)]
+    pub appended_pair_range: Option<AppendedPairRange>,
```
Pranav2612000: What is this used for? Couldn't find its use mentioned in the plan

src/supertable/wal/state_doc.rs
```diff
@@ -0,0 +1,525 @@
...
+    /// corruption.
+    #[serde(default)]
+    pub new_row_content_hash: Option<String>,
+
+    /// UUID v4 minted at the caller's `update()` entry; the
+    /// dedicated superfile the UPDATE's new rows will land in.
+    /// Stored before any I/O so recovery produces bit-identical
+    /// superfile bytes.
+    #[serde(default)]
+    pub preallocated_superfile_id: Option<Uuid>,
```
Pranav2612000, line 333: Is this the same superfile id to whom the appends during this commit go to?

src/supertable/wal/tombstones_codec.rs
```diff
@@ -0,0 +1,427 @@
...
+//!
+//! ## Layout
+//!
+//! ```text
+//! offset  size  field
+//! ----------------------------------------------------------------
+//! 0       8     magic: "INFTOMB\0"                       (literal)
+//! 8       4     schema_version: u32 LE                   (= 1)
+//! 12      1     seal_flag: u8                            (0|1)
+//! 13      24    seal_payload, present iff seal_flag == 1:
```
Pranav2612000, line 19: Should we keep some space empty and start seal_payload at 16 idx

src/supertable/wal/pipeline.rs
```diff
@@ -0,0 +1,2024 @@
...
+
+    // ---- Step 6: Advance WAL state to Appended ----
+    advance_to_appended_if_needed(wal_store, wal_doc, wal_etag, preallocated_superfile_id).await
+}
+
+/// Flatten a `Vec<IdSpan>` into the implied sequence of `i128`
+/// ids in order. Total cost is O(n) in the flattened count;
+/// allocated once per append-phase invocation.
+fn flatten_spans(spans: &[IdSpan]) -> Vec<i128> {
+    let total: usize = spans.iter().map(|s| s.len() as usize).sum();
```
Pranav2612000, line 463: Is it worth iterating over the spans, just to get the capity?

src/supertable/wal/pipeline.rs
```diff
@@ -0,0 +1,2024 @@
...
+        });
+    }
+    if user_batch.num_rows() != flat_ids.len() {
+        return Err(AppendPhaseError::IdSpansLengthMismatch {
+            wal_id: wal_doc.wal_id.to_hex(),
+            flat_len: flat_ids.len(),
+            expected: user_batch.num_rows() as u32,
+        });
+    }
+    let scalar_with_id = prepend_id_column(&user_batch, &flat_ids, &inner.options)?;
```
Pranav2612000: This internally calls, split_vectors and we ignore its vector results. We again call split_vectors below. Is it possible to combine the results so that we only have one expensive `split_vectors` call

src/supertable/wal/pipeline.rs
```diff
@@ -0,0 +1,2024 @@
...
+            })?;
+        let raw = builder
+            .finish()
+            .map_err(|e| AppendPhaseError::SuperfileBuild {
+                message: format!("finish: {e}"),
+            })?;
+        Bytes::from(raw)
+    };
+
+    // ---- Step 6: Per-superfile summaries + SuperfileEntry ----
```
Pranav2612000: Nit: Step 6 comment before Step 5 below

src/supertable/wal/pipeline.rs
```diff
@@ -777,18 +803,312 @@ fn count_outcomes(
...
-    _wal_store: &WalStore,
-    _wal_doc: &WalStateDoc,
-    _wal_etag: &Etag,
+    supertable: &Supertable,
+    wal_store: &WalStore,
+    wal_doc: &WalStateDoc,
+    wal_etag: &Etag,
 ) -> Result<(TombstonePhaseOutcome, WalStateDoc, Etag), TombstonePhaseError> {
-    unimplemented!("tombstone phase do_apply not yet wired up");
+    use crate::supertable::wal::state_doc::TombstoneOutcome;
```
Pranav2612000: Nit: move all imports to the top of the file

src/supertable/wal/pipeline.rs
```diff
@@ -0,0 +1,2024 @@
...
+                    message: format!(
+                        "open superfile {} (storage fallback): {message}",
+                        entry.uri.0
+                    ),
+                }
+            })?,
+        };
+
+        if let Some(doc_id) =
+            scan_id_column_for_match(parquet_bytes, &id_column, target).map_err(|message| {
```
Pranav2612000: We extract the "parquet_bytes" from the reader, and then convert them again to a reader in `scan_id_column_for_match`. 
Would it be better to expose a method for search over parquet_bytes in Superfile, and call that method directly, instead of these conversion and passing around of bytes?

src/supertable/tombstones/cache.rs
```diff
@@ -0,0 +1,303 @@
...
+
+    /// Refresh `superfile_id` from storage. Caches the new
+    /// bitmap + etag (or the "known 404" sentinel) and returns
+    /// the bitmap.
+    fn refresh(&self, superfile_id: Uuid) -> Result<Arc<RoaringBitmap>, SidecarCacheError> {
+        let wal_store = self.wal_store.clone();
+        let drive = async move { wal_store.get_tombstones(superfile_id).await };
+
+        let result = match tokio::runtime::Handle::try_current() {
+            Ok(handle) => tokio::task::block_in_place(|| handle.block_on(drive)),
```
Pranav2612000: Not related to this PR, but we have decided to use tokio::block at a couple of places. 

For consumers of this crate, this will cause their programs to stall/block. Not sure what we feel about this behaviour.

src/supertable/tombstones/cache.rs
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    /// Refresh `superfile_id` from storage. Caches the new
+    /// bitmap + etag (or the "known 404" sentinel) and returns
+    /// the bitmap.
+    fn refresh(&self, superfile_id: Uuid) -> Result<Arc<RoaringBitmap>, SidecarCacheError> {
+        let wal_store = self.wal_store.clone();
+        let drive = async move { wal_store.get_tombstones(superfile_id).await };
+
+        let result = match tokio::runtime::Handle::try_current() {
```
Pranav2612000: Not a blocker for this PR, but we're using this pattern of tokio runtime use if exists otherwise create at a couple of places. Maybe we should abstract this away to a separate function.

src/supertable/wal/recovery.rs
```diff
@@ -0,0 +1,493 @@
...
+    if inner.options.storage.is_none() {
+        return Err(RecoveryError::NoStorageAttached);
+    }
+    let storage = inner
+        .options
+        .storage
+        .as_ref()
+        .expect("checked above")
+        .clone();
+    let wal_store = WalStore::new(Arc::clone(&storage));
```
Pranav2612000, line 132: Will this initialise WAL with object store as the storage? Is this the expected behaviour?

src/supertable/wal/recovery.rs
```diff
@@ -0,0 +1,493 @@
...
+        Err(WalStoreError::Storage {
+            source: crate::storage::StorageError::NotFound { .. },
+            ..
+        }) => return Err(SweepStep::Vanished),
+        Err(_) => return Err(SweepStep::Vanished),
+    };
+
+    // Complete WALs need no work.
+    if doc.state == WalState::Complete {
+        return Ok(OneWalOutcome::AlreadyComplete);
```
Pranav2612000, line 230: If we get a WAL which is Complete, should we delete it so we don't see it again the next time we run

src/supertable/wal/tombstones_admin.rs
```diff
@@ -0,0 +1,310 @@
...
+/// - Sidecar present + sealed by **the same `compaction_id`** →
+///   no-op success; return the existing sidecar. Lets the
+///   compactor's recovery path replay seal calls safely.
+/// - Sidecar present + sealed by a different compaction →
+///   `AlreadySealed` so the compactor knows to recover that
+///   work first.
+/// - CAS-loss on the PUT → `CasLost` so the compactor can
+///   re-read + decide whether to retry (typically yes, after
+///   re-reading the manifest).
+pub async fn seal(
```
Pranav2612000, line 78: When is this called right now?

src/supertable/writer.rs
```diff
@@ -313,7 +356,466 @@ impl SupertableWriter {
...
+            .map_err(MutationError::PredicateEval)?;
+        let matched = target_ids.len();
+        if matched > MAX_TARGETS_PER_MUTATION {
+            return Err(MutationError::MatchCountExceedsCap {
+                matched,
+                cap: MAX_TARGETS_PER_MUTATION,
+            });
+        }
+        let new_row_count = new_rows.num_rows();
+        if matched != new_row_count {
```
Pranav2612000, line 483: This expects the user to pass in new_rows with entries equal to the number of existing entries? Is this a good api? Should the user just pass one entry which we then reuse for all rows? Similar to how `UPDATE .. WHERE` works

Pranav2612000:
I'm also not sure about certain update scenarios, where the user may expect a different behaviour from what we current ly have 
- Data duplication during updates ( user needs to prevent this )
- Updates and deletes have no effect on uncommitted data. This is unlike the behaviour of turbopuffer and lancedb ( although they don't have a "commit" behaviour )


commits are also slower now, because of the retries that we have during updates / deletes. A comment mentions, to ask the user to have their own mechanisms to tackle commit taking a long time. Is that a good api?


IMO, we should also finalize if we're planning to support multiple writers. Supporting multiple writers needs more sophisticated sync mechanisms, and our experience has shown that S3 is not the best for this. If we want to support a single writer, a lot of code can be made simpler and faster.

src/supertable/writer.rs
vinaykakade: The mental model for the update here is append + delete -- in order to prevent users from creating a mistake, would suggest let us go with this check and error. If this is too restrictive, or we want to relax in some other manner, we can revisit based on the actual usage.

src/supertable/wal/tombstones_admin.rs
vinaykakade: This is added for the compactor - but there isn't a production caller yet. WIll remove it from this PR and we can add it back for the compactor.

Pranav2612000: I dont mind keeping it as well. Just wanted to confirm that this was the expected behaviour.

src/supertable/wal/persistence.rs
vinaykakade: Done

src/supertable/wal/persistence.rs
vinaykakade: Done

src/supertable/wal/persistence.rs
vinaykakade: Done

src/supertable/wal/tombstones_admin.rs
vinaykakade: Keeping it in PR right now as compaction is coming up shortly.

Cargo.toml
vinaykakade: Done - cargo.lock is no longer git ignored.

src/supertable/tombstones/cache.rs
vinaykakade: Fine for multi_thread tokio, but would be a problem for current_thread tokio. Let us revisit this in the context of python client libraries. (Many other tools have both sync/async versions of the API so that clients can choose based on their use-case; we may need to start with sync-only to limit the surface area and then add async.) In either case, let us address this in another PR.

src/supertable/tombstones/cache.rs
vinaykakade: There was a separate helper already at supertable/handle.rs -- pulled it out and have started using it.

src/supertable/wal/recovery.rs
vinaykakade: yes - wal lives in the object store at wal/mutations/<wal_id>.json under the same object-store root as manifests + segments

src/supertable/wal/recovery.rs
vinaykakade: Keeping the deletion only as part of gc deletion sweep: Writer.rs:720-721 (UPDATE) and :799 (DELETE) inline-delete the state doc + arrow sidecar the moment the writer's own pipeline hits Complete. That's the steady-state path. wal::gc::run_sweep (gc.rs:111) is the safety net for the rare case where the inline delete failed (transient storage error, or writer crashed between the COMPLETE CAS and the inline delete). Trying to not intentionally overload startup/recovery with deletions.

src/supertable/wal/tombstones_codec.rs
vinaykakade: Today, we aren't doing alignment sensitive reads. In the future, if/when we do them, we can do a schema version bump and change the format. Happy to change if you feel strongly though.

src/supertable/wal/pipeline.rs
vinaykakade: Great point. Did some benchmarking and changing to optimal implementation (replacing "push" below with extend).

vinaykakade: ┌───────────┬────────────────────┬───────────────────────┬─────────┐
  │ Total ids │ push loop (before) │ extend(Range) (after) │ Speedup │
  ├───────────┼────────────────────┼───────────────────────┼─────────┤
  │        1K │            2.32 µs │               0.31 µs │     ~7× │
  ├───────────┼────────────────────┼───────────────────────┼─────────┤
  │      100K │             188 µs │                 64 µs │     ~3× │
  ├───────────┼────────────────────┼───────────────────────┼─────────┤
  │        1M │            1.35 ms │               0.21 ms │     ~6× │
  └───────────┴────────────────────┴───────────────────────┴─────────┘

src/supertable/wal/pipeline.rs
vinaykakade: Done

src/supertable/wal/pipeline.rs
vinaykakade: Done

src/supertable/wal/pipeline.rs
vinaykakade: Done - also filed #29 to improve caching.

src/supertable/wal/state_doc.rs
vinaykakade: Unused - removed

src/supertable/wal/state_doc.rs
vinaykakade: No, updates go to a separate superfile to preserve crahs safety. See comment added in this file ~line 280.

src/runtime_bridge.rs
```diff
@@ -0,0 +1,55 @@
...
+                .enable_all()
+                .build()
+                .expect(
+                    "invariant: tokio Runtime build only fails on \
+                     catastrophic OS resource exhaustion",
+                );
+            rt.block_on(fut)
+        }
+    }
+}
```
Pranav2612000, line 55: Not related to this PR, but there's more places in the code where we do similar logic.
There's also a `sql_runtime` which starts a multi threaded tokio with 1 worker thread.

Let's consolidate all this logic in this file.
Also, when do we use `new_current_thread()` vs `new_multi_threaded()` vs `new_multithreaded()` with 1 worker thread?

src/supertable/wal/state_doc.rs
vinaykakade: Changed to a new type RowId - also added SupertableHandleId as a type, so that the the type name is descriptive.

src/supertable/wal/pipeline.rs
```diff
@@ -1056,39 +1054,50 @@ fn resolve_target_id_in_manifest(
...
+                Arc::new(SuperfileReader::open(bytes).map_err(|e| {
+                    TombstonePhaseError::IdLookupFailed {
+                        target_id: target_id.to_hex(),
+                        message: format!(
+                            "SuperfileReader::open {} (storage fallback): {e}",
+                            entry.uri.0
+                        ),
+                    }
+                })?)
+            }
```
Pranav2612000, line 1067: Unrelated to this PR, but let's also unify this in a common function and reuse whenever we need to get reader?

src/superfile/reader.rs
```diff
@@ -275,6 +276,63 @@ impl SuperfileReader {
...
+    /// fixed precision/scale; we decode each value as `i128`.
+    ///
+    /// Used by the WAL recovery sweep's `resolve_target_id`
+    /// path: given a `target_id`, scan the candidate superfile to
+    /// find where the row lives so the tombstone phase can mark
+    /// its bit. Currently rebuilds a `ParquetRecordBatchReader`
+    /// each call — opportunity for a follow-up that caches
+    /// parquet metadata on the reader (see TODO in module-level
+    /// comments).
+    pub fn id_lookup(&self, target: i128) -> Result<Option<u32>, ReadError> {
```
Pranav2612000, line 294: We can also address this separately, but I guess there should be a way to read and iterate through just the `_id` column instead of the whole row.

src/runtime_bridge.rs
vinaykakade: Tracking in #32

src/supertable/wal/pipeline.rs
vinaykakade: Tracking in #32

src/superfile/reader.rs
vinaykakade: Tracking in #32

Pranav2612000 (approved):

PR #28, opened 2026-06-01
examples: end-to-end demo across superfile, open-format, supertable

muralikpbhat:
Adds `examples/demo.rs` — a small runnable tour of the stack:

- **A. superfile** — build a one-doc superfile and run BM25 + vector kNN against the indexes embedded in the same bytes.
- **B. open-format** — read those same bytes back as a plain Parquet table via DataFusion (the embedded blobs and `inf.*` KV are ignored per spec).
- **C. supertable** — two commits → two segments, cross-segment BM25 fan-out, and SQL over the auto-injected `_id` column.

Run with `cargo run --example demo`.

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -1,0 +1,179 @@
+//! Smallest possible end-to-end tour of the stack:
+//!
+//!   A. superfile  — feed one document, then BM25 *and* vector kNN
+//!      against the indexes embedded in the same bytes.
+//!   B. open-format — those same bytes are a valid Parquet file;
+//!      vanilla DataFusion reads them as a plain table.
+//!   C. supertable — the cross-segment layer that auto-injects a
+//!      real `_id`, queried across two committed segments.
+//!
+//! Run with:
+//! ```text
+//! cargo run --example demo
+//! ```
+
+use std::sync::Arc;
+
+use arrow::util::pretty::pretty_format_batches;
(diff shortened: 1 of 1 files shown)


examples/demo.rs
```diff
@@ -0,0 +1,179 @@
...
+    // `emb` is a vector column: a logical name only — its f32s are
+    // passed to add_batch separately and never enter the Parquet
+    // schema. n_cent=1 because a single doc yields one IVF cluster.
+    let opts = BuilderOptions::new(
+        schema.clone(),
+        "doc_id",
+        vec![FtsConfig {
+            column: "title".into(),
+        }],
+        vec![VectorConfig::new("emb".into(), 16, 1, 7, Metric::Cosine)],
```
savannahar68, line 60: instead of hardcoding 16, can we have this `const EMB_DIM: usize = 16;`
and reuse the same on line 75

or rewrite it as 
```
 vec![VectorConfig::new("emb".into(), 16, 1, 7, Metric::Cosine)], // dim=16, n_cent=1 (one cluster per single doc), rot_seed=7
```

savannahar68 (approved):
lgtm
minor nit, can be picked up later

muralikpbhat:
Thanks for merging this in. I was about to respond to Savan's suggestions and opened a new [PR #30](https://github.com/infino-ai/infino/pull/30) since this is closed.

PR #34, opened 2026-06-02
Add object storage support

ekechinwokah:
## Summary
This branch adds and optimizes the lazy object-storage read path for superfiles and supertables. The main change is that readers can open indexes backed by S3/object storage, fetch only the byte ranges needed for a query, promote warmed data into the disk cache/mmap path, and run hot/warm/cold search tiers through the same production reader code.

The vector path gets the largest implementation changes: compressed rerank data, lower-round-trip cold reads, survivor-only reranking, distance/read-path optimizations, and calibrated supertable vector search across recall targets. The FTS path is also moved onto the lazy object-store reader path so postings/dictionaries can be queried through the same cache-backed storage layer.

The branch also adds real-S3 storage/config support, new disk-cache behavior/tests, supertable manifest/query plumbing needed for lazy readers, and a much larger benchmark harness to measure the new object-storage behavior.

## Major Areas Changed
- Object-storage-backed superfile/supertable reads:
  Adds lazy byte sources, range-read plumbing, S3-backed storage smoke coverage, and disk-cache-backed reader paths.

- Disk cache and warm promotion:
  Adds the cache path used for cold reads, warm reads, mmap promotion, cache sweeping, and hot/warm/cold tier separation.

- Vector search:
  Adds compressed rerank support, survivor-only reranking, object-store cold-path reductions, distance/read optimizations, and recall-calibrated supertable vector search.

- FTS search:
  Moves FTS readers through the lazy object-store path and measures hot/warm/cold behavior on the shared 10M-doc supertable.

- Supertable query/manifest plumbing:
  Updates manifest partition metadata, query readers, vector query execution, FTS query execution, and superfile reader integration so segmented supertables can be searched through lazy object-store-backed readers.

- Bench/test/docs:
  Replaces the older supertable bench split with `supertable_all`, adds real-S3 hot/warm/cold benchmark reporting, adds recall/diagnostic benches, expands disk-cache/lazy-source tests, and updates architecture docs.

## Benchmark Evidence - See benches/README.md
- Vector-only 10M build: 405.20s, 24.7K docs/s, 5.18 GiB peak RSS.
- Combined FTS + vector 10M build: 494.81s, 20.2K docs/s, 7.47 GiB peak RSS.
- Vector cold search on real S3: 929.70ms / 990.25ms / 1.08s at recall ≥0.90 / 0.95 / 0.99.
- FTS cold search on real S3: 

Query | low | p50 | high
-- | -- | -- | --
single_rare | 324ms | 350ms | 376ms
single_common | 393ms | 421ms | 450ms
two_term_or | 406ms | 420ms | 435ms
three_wide_or | 383ms | 1315ms | 2738ms
three_similar_or | 296ms | 322ms | 346ms
five_term_or | 346ms | 408ms | 491ms
ten_term_or | 608ms | 621ms | 635ms
prefix | 496ms | 542ms | 582ms

## Caveat
This branch establishes the lazy object-storage search architecture and measurement harness, but it does not yet hit the target performance goals.


benches/scale/vector_recall.rs
```diff
@@ -0,0 +1,247 @@
...
+                c + (s as f32) * 0.3
+            })
+            .collect();
+        if normalize_each {
+            normalize(&mut v);
+        }
+        corpus.push(v);
+    }
+    corpus
+}
```
Pranav2612000: Should be okay for now. But worth pointing out that we've generated the corpus in a way optimized for our centroid clustering vector search algorithm. This may not be representative of real workloads

benches/fts/main_superfile.rs
```diff
@@ -1,14 +0,0 @@
-//! Superfile FTS bench (1M docs). Standalone binary.
```
Pranav2612000, line 1: Reason for deleting this bench? It allowed us to run only the superfile benches

benches/tombstone_overhead/main.rs
```diff
@@ -227,8 +227,7 @@ fn bench_fts(c: &mut Criterion) {
     g.bench_function("clean", |b| {
         let st = fixture_clean();
         b.iter(|| {
-            let r = st.reader();
-            let hits = r
+            let hits = st
```
Pranav2612000, line 230: So with this new API, user can directly search on supertable, without getting a reader first.
- What was the reasons of having a separate reader @vinaykakade 
- What's the reasoning behind removing it now and allowing search directly? @ekechinwokah

Pranav2612000: If its the sync/async distinction? should we expose both from the reader? The async ones with a separate async prefix.
The current logic ( sync through supertable, async through reader ) isn't intuitive.

benches/scale/vector_recall.rs
```diff
@@ -0,0 +1,247 @@
...
+fn search_blocking(
+    reader: &SuperfileReader,
+    query: &[f32],
+    k: usize,
+    opts: VectorSearchOptions,
+) -> Vec<(u32, f32)> {
+    futures::executor::block_on(reader.vector_search("emb", query, k, opts)).expect("vector_search")
+}
+
+fn generate_planted_corpus(seed: u64, normalize_each: bool) -> Vec<Vec<f32>> {
```
Pranav2612000: Can we reuse the methods in `corpus` for creating the corpus and queries?

benches/utils/unified_object_store.rs
```diff
@@ -0,0 +1,2125 @@
...
+            self.range_total_us.fetch_add(us as u64, Ordering::Relaxed);
+            self.range_log.lock().unwrap().push(RequestEvent {
+                len,
+                start_us,
+                end_us,
+            });
+            r
+        }
+
+        // Plan 013 M5 — must forward to `self.inner.tail` rather
```
Pranav2612000: Nit: Do a cleanup to remove all occurrences referrencing plans/internal documents

docs/architecture/superfile.md
Pranav2612000, line 1: The newer files ( for superfile and supertable ) are much smaller, and less verbose than the older ones. Is there a specific reason behind this change?

src/config/config.yaml
```diff
@@ -71,3 +71,30 @@ supertable:
...
   verify_crc_on_open: true
+
+# Persistent storage wiring. Defaults to in-memory-only. Set
+# `backend: s3` plus `s3_bucket` / `s3_prefix` to make
+# SupertableOptions::apply_config attach an S3StorageProvider.
+#
+# Env overrides:
+#   INFINO_STORAGE__BACKEND=s3
+#   INFINO_STORAGE__S3_BUCKET=my-bucket
+#   INFINO_STORAGE__S3_PREFIX=tables/my-table
```
Pranav2612000: Should the env names and other identifiers throughout the code have 'S3' prefix? We want it to not be tightly coupled with S3, but generic for all object stores

src/config/mod.rs
```diff
@@ -124,6 +129,84 @@ impl Default for SupertableSettings {
...
+/// Config-side spelling for disk-cache cold-fetch mode. Kept
+/// separate from the runtime enum so serde naming stays stable
+/// without coupling config format to internal module layout.
+#[derive(Debug, Clone, Copy, Deserialize, Serialize, PartialEq, Eq, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum StorageColdFetchMode {
+    HybridWithPrefetch,
+    RangeOnly,
+    #[default]
+    LazyForegroundWithBackgroundFill,
```
Pranav2612000, line 169: Can you add comments on how each mode works? Also provides automatic documenation for rustdoc

src/config/mod.rs
```diff
@@ -124,6 +129,84 @@ impl Default for SupertableSettings {
...
+    /// in-memory-only behavior.
+    pub backend: StorageBackend,
+    /// Local filesystem root when `backend: local_fs`.
+    pub local_root: Option<PathBuf>,
+    /// S3 bucket when `backend: s3`. Credentials and region use
+    /// the standard AWS provider chain (`AWS_*`, profile,
+    /// instance role, etc.).
+    pub s3_bucket: Option<String>,
+    /// Logical table prefix inside the S3 bucket.
+    pub s3_prefix: String,
```
Pranav2612000: Nit: change to 'bucket' & 'prefix' to make it generic for all object stores

src/config/mod.rs
```diff
@@ -124,6 +129,84 @@ impl Default for SupertableSettings {
...
+    pub disk_cache_root: Option<PathBuf>,
+    pub disk_budget_bytes: u64,
+    pub cold_fetch_mode: StorageColdFetchMode,
+    pub cold_fetch_streams: usize,
+    pub cold_fetch_chunk_bytes: u64,
+    /// Global cap on concurrent background segment fills. See
+    /// [`crate::supertable::reader_cache::DiskCacheConfig::prefetch_concurrency`].
+    pub prefetch_concurrency: usize,
+    pub mmap_cold_threshold_secs: u64,
+    pub mmap_sweep_interval_secs: u64,
```
Pranav2612000, line 210: Can you add comments on what do these control?

src/config/mod.rs
Pranav2612000: Should the `prefix` also be an option? What would its value be when the filesystem is local?
Can we merge `local_root` & `prefix`?

src/storage/local_fs.rs
```diff
@@ -306,22 +306,22 @@ mod tests {
     async fn put_then_get_roundtrip() {
         let (_dir, p) = provider();
         let payload = Bytes::from_static(b"hello supertable storage");
-        p.put_atomic("data/seg-abc.sf", payload.clone())
+        p.put_atomic("data/seg-abc.sf.parquet", payload.clone())
```
Pranav2612000, line 309: Reason for adding the parquet prefix?

src/superfile/lazy_source.rs
```diff
@@ -149,6 +455,168 @@ impl LazyByteSource for BytesLazyByteSource {
...
+impl LazyByteSource for PrefetchedSource {
+    fn size(&self) -> u64 {
+        self.inner.size()
+    }
+
+    async fn range(&self, start: u64, len: u64) -> Result<Bytes, LazyByteSourceError> {
+        if let Some(b) = self.lookup(start, len) {
+            return Ok(b);
+        }
+        self.inner.range(start, len).await
```
Pranav2612000, line 550: Once we get this range, should we add it to the `prefetched` list so that future calls can directly refer it, instead of calling it again?

src/superfile/fts/fst_value.rs
```diff
@@ -58,15 +67,19 @@ impl FstValue {
         }
     }
 
-    /// Pack a metadata_offset into the PFOR-form FST value. The low
-    /// bit is always 0; the offset occupies bits 1..N.
+    /// Pack `(metadata_offset, postings_length)` into the PFOR-form
+    /// FST value. The low bit is always 0.
     #[inline]
-    pub(crate) fn pack_pfor(metadata_offset: u64) -> u64 {
+    pub(crate) fn pack_pfor(metadata_offset: u64, postings_length: u32) -> u64 {
```
Pranav2612000, line 73: Can a u64 and a u32 together be safely packed to a u64? Could it cause issues when the metadata_offset is high? What guarantees that the metadata_offset is always less than PFOR_OFFSET_MAX. Is it because our superfile sizes are much smaller than this?

src/superfile/fts/reader.rs
```diff
@@ -702,23 +890,53 @@ impl FtsReader {
...
-    fn build_term_cursors(
+    async fn build_term_cursors(
         &self,
         column_id: u32,
         terms: &[&str],
     ) -> Result<Vec<TermCursor>, FtsError> {
-        let dict = self.dict();
+        let fst_bytes = self.dict_bytes_async().await?;
+        let dict = DictReader::open(&fst_bytes)
+            .expect("FST CRC verified at open(); bytes must be a valid FST");
```
Pranav2612000: Could this fail due to other issues? Is it better to propogate the error instead of crashing

src/superfile/vector/builder.rs
```diff
@@ -479,18 +475,17 @@ impl VectorBuilder {
...
+        //    subsection so lazy open can fetch subsection headers and
+        //    Sq8 metadata in the same network batch.
         let mut directory: Vec<u8> = Vec::with_capacity(directory_size);
         for (i, sub) in subsections.iter().enumerate() {
             let (cfg, _) = &column_configs[i];
             let summary_offset_abs = subsection_start_off + sub.summary_offset_in_sub as u64;
             directory.extend_from_slice(&(i as u32).to_le_bytes()); // column_id
             directory.extend_from_slice(&(cfg.dim as u32).to_le_bytes()); // dim
-            directory.extend_from_slice(&(cfg.n_cent as u32).to_le_bytes()); // n_cent
+            directory.extend_from_slice(&(sub.n_cent as u32).to_le_bytes()); // physical n_cent
```
Pranav2612000, line 488: How does cfg.n_cent differ from sub.n_cent ?

src/superfile/vector/builder.rs
```diff
@@ -816,174 +814,161 @@ fn build_subsection_streaming(
...
-        // quantizer bytes — keeps the on-disk byte pattern stable
-        // regardless of pass-2 assignment outcomes.
-        if codec == RerankCodec::Sq8 {
-            let (scale_c, offset_c) = &sq8_quantizers[c];
-            let sc_off = sq8_scale_block_off + c * dim * 4;
-            bytes[sc_off..sc_off + dim * 4].copy_from_slice(bytemuck::cast_slice(scale_c));
-            let oc_off = sq8_offset_block_off + c * dim * 4;
-            bytes[oc_off..oc_off + dim * 4].copy_from_slice(bytemuck::cast_slice(offset_c));
-        }
-
```
Pranav2612000, line 971: Reason for moving this to the top? Keeping it here prevents one extra iteration, and keeps the code closer

src/superfile/vector/reader.rs
```diff
@@ -294,34 +264,246 @@ impl VectorReader {
...
+            ));
+            subheader_ranges.push((i, subsection_off));
+        }
+
+        let subheaders_fut =
+            futures::future::try_join_all(subheader_ranges.iter().map(|&(i, subsection_off)| {
+                let source = Arc::clone(&source);
+                async move {
+                    let bytes = source
+                        .range(subsection_off as u64, SUB_HEADER_SIZE as u64)
```
Pranav2612000, line 418: Are all sections one after the other? In that case, does it make sense to just do one range call for all subsections instead of multiple calls for each subsection? Could reduce the number of GET calls reducing calls

src/superfile/vector/reader.rs
```diff
@@ -294,34 +264,246 @@ impl VectorReader {
...
+                                "lazy open: subsection {i} sub-header fetch: {e}"
+                            )))
+                        })?;
+                    Ok::<_, VectorError>((i, subsection_off, bytes))
+                }
+            }));
+        let subheaders = subheaders_fut.await?;
+
+        for (i, subsection_off, sub_header) in subheaders {
+            overlay.install(subsection_off as u64, sub_header.clone());
```
Pranav2612000: Let's install after verifying first that the header is valid ( next line )?

src/superfile/vector/reader.rs
```diff
@@ -1040,92 +1633,393 @@ fn read_cluster_entry(cluster_idx_slice: &[u8], c: usize) -> (u32, u32) {
...
+                            )
+                        },
+                    )
+                }
+                Sq8ColumnMeta::Lazy {
+                    scale_abs_off,
+                    offset_abs_off,
+                    norms_abs_off,
+                } => {
+                    if let Some(meta_bytes) = lazy_sq8_meta_bytes {
```
Pranav2612000, line 1706: IMO, the eager and Lazy parts should be the same, with the exception that the Lazy one needs to fetch the metadata first. Can we use a common function for the part once the metadata is fetched?

Pranav2612000 (commented):
Review 1/n

src/superfile/lazy_source.rs
```diff
@@ -100,6 +136,276 @@ pub enum LazyByteSourceError {
...
+                            LazyByteSourceError::Storage(crate::storage::StorageError::Permanent {
+                                uri: "lazy-source://superfile-reader".to_string(),
+                                source: Box::new(std::io::Error::other(format!(
+                                    "tokio runtime build for lazy source fetch: {e}"
+                                ))),
+                            })
+                        })?;
+                    rt.block_on(src.range(start, len))
+                })
+                .join()
```
Pranav2612000: We have a helper `bridge_sync_to_async` function in src/runtime_bridge.rs which does this tokio initialization if needed. Let's re-use this here

src/superfile/lazy_source.rs
```diff
@@ -100,6 +136,276 @@ pub enum LazyByteSourceError {
...
+                {
+                    tokio::task::block_in_place(|| handle.block_on(fut))?
+                }
+                _ => std::thread::spawn(move || {
+                    let rt = tokio::runtime::Builder::new_current_thread()
+                        .enable_all()
+                        .build()
+                        .map_err(|e| {
+                            LazyByteSourceError::Storage(crate::storage::StorageError::Permanent {
+                                uri: "lazy-source://superfile-reader-parallel".to_string(),
```
Pranav2612000: We have a helper bridge_sync_to_async function in src/runtime_bridge.rs which does this tokio initialization if needed. Let's re-use this here

src/superfile/reader.rs
```diff
@@ -115,21 +122,138 @@ impl SuperfileReader {
...
-        source: &dyn crate::superfile::LazyByteSource,
-        opts: OpenOptions,
+        source: Arc<dyn crate::superfile::LazyByteSource>,
+        _opts: OpenOptions,
     ) -> Result<Self, ReadError> {
-        let size = source.size();
-        let bytes = source.range(0, size).await.map_err(|e| match e {
-            crate::superfile::LazyByteSourceError::Storage(se) => {
-                ReadError::MalformedKv(format!("lazy source storage: {se}"))
+        use parquet::arrow::parquet_to_arrow_schema;
```
Pranav2612000, line 137: Nit: do a cleanup and hoist all dependencies in all files to the top

src/superfile/reader.rs
```diff
@@ -292,7 +425,13 @@ impl SuperfileReader {
...
     /// comments).
     pub fn id_lookup(&self, target: i128) -> Result<Option<u32>, ReadError> {
-        let builder = ParquetRecordBatchReaderBuilder::try_new(self.bytes.clone())
+        let bytes = self.bytes.clone().ok_or_else(|| {
+            ReadError::Io(std::io::Error::other(
+                "id_lookup requires an eager-opened superfile; this reader was opened via \
+                 the lazy path and does not hold the full segment bytes",
+            ))
+        })?;
+        let builder = ParquetRecordBatchReaderBuilder::try_new(bytes)
```
Pranav2612000, line 443: This function is used for updates, and updates may fail due to this error. Should we
- ensure that we always eager open the superfile during updates? 
OR
- Modify this function to lazily fetch ids columns and run a query on them

src/superfile/reader.rs
```diff
@@ -540,33 +686,29 @@ impl SuperfileReader {
...
 impl VectorSearchOptions {
-    /// Builder default: `nprobe = 8`, `rerank_mult = 20`.
     pub const DEFAULT_NPROBE: usize = 8;
-    pub const DEFAULT_RERANK_MULT: usize = 20;
 
-    /// Construct with both defaults applied.
+    /// Internal rerank multiplier. `k * RERANK_MULT` candidates
+    /// from the 1-bit RaBitQ shortlist enter Sq8/residual rerank.
+    /// Bench-validated: recall saturates at 4 on 10M×384 cosine.
+    pub const RERANK_MULT: usize = 4;
```
Pranav2612000, line 715: 4 seems to be the standard, but are we sure about this change from 20 to 4? Are we over optimizing for the benchmarks?

src/supertable/query/fts.rs
```diff
@@ -299,12 +326,55 @@ impl SupertableReader {
...
+    /// [`Supertable::query_sql`], using the same sync→async bridge
+    /// ([`Supertable::block_on_query`]).
+    ///
+    /// Pins a reader at call entry (one `ArcSwap::load_full`, like
+    /// `query_sql`) and drives the async
+    /// [`SupertableReader::bm25_search`] kernel to completion. Returns
+    /// up to `k` hits sorted by BM25 score *descending*. Use the
+    /// `SupertableReader` method directly when you are already in async
+    /// context and want to `.await`.
+    pub fn bm25_search(
```
Pranav2612000, line 364: Thoughts on exposing both the sync and async apis either on the reader or on Supertable ( instead of sync on Supertable and async on reader )

src/supertable/query/sql.rs
```diff
@@ -102,10 +102,12 @@ impl Supertable {
         let scalar_schema = self.options().scalar_schema();
         let tombstone_cache = reader.tombstone_cache.clone();
 
-        let table = build_mem_table(scalar_schema, manifest, store, disk_cache, tombstone_cache)?;
         let sql = sql.to_owned();
 
         let drive = async move {
+            let table =
+                build_mem_table(scalar_schema, manifest, store, disk_cache, tombstone_cache)
+                    .await?;
```
Pranav2612000: Will these sql queries work for data on S3?

src/supertable/wal/pipeline.rs
```diff
@@ -1038,10 +1043,13 @@ fn resolve_target_id_in_manifest(
...
-            &inner.options.store,
-            inner.options.disk_cache.as_ref(),
-            &entry.uri,
+        let reader = match bridge_sync_to_async(
+            crate::supertable::query::superfile_reader::superfile_reader(
+                &inner.options.store,
+                inner.options.disk_cache.as_ref(),
+                &entry.uri,
+                entry.subsection_offsets.as_ref(),
+            ),
```
Pranav2612000, line 1055: I think this will open the reader in LazyForegroundWithBackgroundFill mode, and the corresponding get_ids fn may fail. Right? Updates may not work

src/supertable/handle.rs
```diff
@@ -519,6 +525,60 @@ impl Supertable {
...
+    /// `sql_runtime`. Lets `vector_search` / `bm25_search` /
+    /// `bm25_search_prefix` present a sync public API over the async
+    /// `SupertableReader` kernels without spinning a throwaway runtime
+    /// per call.
+    pub(crate) fn block_on_query<F: Future>(&self, fut: F) -> F::Output {
+        match tokio::runtime::Handle::try_current() {
+            Ok(handle) => tokio::task::block_in_place(|| handle.block_on(fut)),
+            Err(_) => self.sql_runtime().block_on(fut),
+        }
+    }
```
Pranav2612000: Can we reuse the existing bridging function that we have? If not, lets move all bridging code to the runtime_bridge.rs file

src/superfile/lazy_source.rs
```diff
@@ -149,6 +455,168 @@ impl LazyByteSource for BytesLazyByteSource {
...
+/// and resolves zero-copy via [`Self::try_get_range_sync`] —
+/// no extra GETs against the underlying source.
+///
+/// Concurrency: `install` is non-mutating (`&self`) and uses
+/// an interior `Vec` to allow staged construction during
+/// `open_lazy`, then the overlay is wrapped in `Arc` and
+/// shared by every subsequent read. The Vec is read-only after
+/// the open completes, so the racy-read-during-install pattern
+/// the M2 open path actually uses is single-threaded.
+pub struct PrefetchedSource {
```
Pranav2612000, line 484: Is it necessary for the LazySubSource and PrefetchedSource to also implement LazyByteSource? They already have an inner which implements it.

I guess the confusing part is the same trait ( LazyByteSource ) is implemented by some structs at the Supertable level and some at the Superfile level.

Pranav2612000 (commented):
We already have it somewhere, but can you point me to the code for eviction of segments when we observe that the cache is already full, when segments are being loaded in the background.

Review 2/n

src/supertable/query/fts.rs
ekechinwokah: thx. just caught it as i was reviewing. all public methods should be sync. fixed.

src/superfile/reader.rs
ekechinwokah: fair point. let's revisit when we start adding more corpus tests. the intent is to keep the user interface as simple as possible.

benches/fts/main_superfile.rs
ekechinwokah: consolidated into the unified harness; superfile-only FTS is still reachable via a criterion filter (cargo bench --bench supertable_all -- <filter>).

benches/scale/vector_recall.rs
ekechinwokah: yes, routing it through the benches/utils/corpus generators to dedupe corpus + query creation.

benches/utils/unified_object_store.rs
ekechinwokah: agree — it's all over the code base though. i have a cleanup PR 2 positions down in the queue.

docs/architecture/superfile.md
ekechinwokah: yes - we should discuss with vinay, but 2 reasons:

- i think our real IP here is not the code itself but the thinking behind (claude plans). we should protect that vigorously and perhaps less verbose.
- i could be convinced otherwise, but i think the current files are too verbose in general for good UX.

src/config/config.yaml
ekechinwokah: agree — renaming to generic bucket/prefix (+ env vars) so it's object-store-agnostic, as a focused rename commit.

src/config/mod.rs
ekechinwokah: done.

src/config/mod.rs
ekechinwokah: renaming to generic bucket/prefix. on merging: local_root is the filesystem root for the local provider; prefix is a logical key prefix inside a bucket (empty for local) — different layers; I'll document the distinction.

src/config/mod.rs
ekechinwokah: done.

src/storage/local_fs.rs
ekechinwokah: yes. more likely to not have issues with ecosystem tools, and also more comfort for adopters that this is a real parquet file.

src/superfile/lazy_source.rs
ekechinwokah: PrefetchedSource is intentionally an open-time, read-only overlay — runtime range caching is the disk-cache/mmap promotion layer's job, so a second mutable cache here would duplicate that and add locking. we can revisit if profiling shows repeat GETs.

src/superfile/fts/fst_value.rs
ekechinwokah: layout is 1 flag + 42 offset bits (4 TiB ceiling) + 21 length bits, so the invariant holds for any real superfile. but the guard is debug_assert! only — promoting it to a hard assert on the builder path so a release-mode overflow can't silently corrupt.

src/superfile/fts/reader.rs
ekechinwokah: agree — converting to a propagated ReadError.

src/superfile/vector/builder.rs
ekechinwokah: cfg.n_cent = requested centroid count; sub.n_cent = physical count actually built (clamped when a subsection has fewer docs than requested centroids). both stored so the reader knows logical vs physical. adding a comment.

src/superfile/vector/builder.rs
ekechinwokah: moved because the Sq8 meta is now one contiguous region so the lazy reader fetches it in a single range GET; the old interleaved per-cluster write doesn't match that layout. the extra pass is negligible compared to the cost of GETs.

src/superfile/vector/reader.rs
ekechinwokah: they're already fetched concurrently (`try_join_all`); a single span would overfetch the whole blob since subheaders are scattered. maybe we could do it if profiling supports.

src/superfile/vector/reader.rs
ekechinwokah: good catch — reordering so we only install after the header validates.

src/superfile/vector/reader.rs
ekechinwokah: done.

ekechinwokah:
> We already have it somewhere, but can you point me to the code for eviction of segments when we observe that the cache is already full, when segments are being loaded in the background.
> 
> Review 2/n

eviction is in `src/supertable/reader_cache/disk.rs` — budget-aware via `sweep_for_budget`/`sweep_once` (budget_bytes, n_evictions), with a pinned-URI set so segments actively loading in the background aren't evicted. with no eligible victims it returns `DiskCacheBudgetExceeded` and the query bypasses the cache.

src/superfile/lazy_source.rs
ekechinwokah: done.

src/superfile/lazy_source.rs
ekechinwokah: done

src/superfile/reader.rs
ekechinwokah: done

src/superfile/reader.rs
ekechinwokah: you're right — fixed. same root cause as the wal/pipeline.rs issue.

src/supertable/query/sql.rs
ekechinwokah: yes — `build_mem_table` goes through the same lazy `superfile_reader` (store + disk cache), so S3-backed segments work; it projects columns on demand rather than pulling whole segments. added smoke test.

ekechinwokah: also the next PR is for datafusion which integrates and tests sql end to end.

src/supertable/wal/pipeline.rs
ekechinwokah: correct. same issue as earlier. fixed.

src/supertable/handle.rs
ekechinwokah: yes, in general this was somewhat broken in the current PR. fixed.

src/superfile/lazy_source.rs
ekechinwokah: they're decorators — a sub-range view and an open-time prefetch overlay — so they implement `LazyByteSource` to be substitutable as a Source. the "same trait at superfile and supertable levels" is the composition seam. can also rename if helpful. adding some documentation for clarity.

benches/tombstone_overhead/main.rs
ekechinwokah: yeah confusing because public api is not defined.

defined a v1 api, with reader(), writer(), and stats() as the external top level functions, all sync, no async. so st.reader().bm25(), st.writer().append(), and st.stats().manifest() for example. documented in supertable.md.

benches/scale/vector_recall.rs
ekechinwokah: fair. more entropy should actually work better though, at least for ranking.

benches/tombstone_overhead/main.rs
Pranav2612000: Sure. cc: @vinaykakade as he's also working on database level apis

src/storage/local_fs.rs
Pranav2612000: Good point

src/superfile/vector/builder.rs
```diff
@@ -816,174 +814,161 @@ fn build_subsection_streaming(
...
+    let cluster_order = centroid_storage_order(&centroids, n_cent, dim);
+
+    // 6. Build the subsection bytes.
+    //    subsection layout
+    //    (see `format::vec::SUBSECTION_VERSION` for the spec):
+    //
+    //      [sub_header]
+    //      [summary_centroid][centroids][cluster_idx][codec_meta]   ← open-time region
+    //      [per-cluster blocks: each = codes_chunk + doc_ids_chunk]
+    //      [full]                                                   ← rerank column
```
Pranav2612000: Does the `codes_chunk` mean the 1 bit RabitQ values?

src/config/mod.rs
```diff
@@ -124,6 +129,105 @@ impl Default for SupertableSettings {
...
+/// Config-side spelling for disk-cache cold-fetch mode. Kept
+/// separate from the runtime enum so serde naming stays stable
+/// without coupling config format to internal module layout.
+#[derive(Debug, Clone, Copy, Deserialize, Serialize, PartialEq, Eq, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum StorageColdFetchMode {
+    /// Multi-stream parallel prefetch into the disk cache before
+    /// handing the reader to the caller. Lowest cold-search latency
+    /// at the cost of higher cold-open latency.
+    HybridWithPrefetch,
```
Pranav2612000: I guess this description is different from the one in ColdFetchMode enum

src/superfile/fts/reader.rs
```diff
@@ -135,33 +143,110 @@ impl FtsReader {
...
+        // the FTS open spends 2 serial RTTs (header, then this parallel
+        // pair) instead of 3. On a warm/in-memory source both resolve
+        // through the sync zero-copy path at no cost. The doc-length
+        // tail is fetched whole (one range) rather than dir-then-arrays,
+        // keeping the open-time GET count minimal.
+        let (fst_region, doc_lengths_region) = futures::try_join!(
+            fetch_lazy_range(source.as_ref(), 48..postings_offset, "fts/dict"),
+            fetch_lazy_range(
+                source.as_ref(),
+                doc_lengths_table_offset..source_len,
```
Pranav2612000, line 203: Is the source_len here? The length of the whole FTS Blob? Does this mean we're fetching the whole FTS blob, instead of just the doc_lengths_table ?

src/superfile/fts/reader.rs
```diff
@@ -236,11 +332,20 @@ impl FtsReader {
...
         // Read doc-lengths directory: n_columns × 16-byte entries + 4-byte CRC.
         let dir_size = n_columns * 16;
-        let dir_bytes = &blob[doc_lengths_table_offset..doc_lengths_table_offset + dir_size];
+        let dir_end = doc_lengths_table_offset + dir_size;
+        if dir_end + 4 > source_len {
+            return Err(FtsError::Read(ReadError::MalformedVersion(
+                "doc-lengths directory runs past blob end".into(),
+            )));
+        }
+        let dir_region = fetch_source_range(
```
Pranav2612000, line 345: Related to the above comment
Can we also do a GET for the `doc_lengths_array` and `overlay.install` it? Otherwise there'll be a lot of range calls ( proportional to n cols )
Or are we already doing it?

src/supertable/reader_cache/disk.rs
```diff
@@ -801,22 +910,359 @@ impl DiskCacheStore {
...
+            }
+        };
+
+        // Fire the footer tail and fixed subsection headers concurrently.
+        // Larger vector / FTS metadata ranges are discovered and fetched
+        // by the inner readers after this seeded open.
+        let storage = Arc::clone(&self.storage);
+        let storage_uri_a = storage_uri.clone();
+        let storage_uri_b = storage_uri.clone();
+        let storage_uri_c = storage_uri.clone();
```
Pranav2612000: Nit: better variable names instead

src/supertable/reader_cache/disk.rs
```diff
@@ -801,22 +910,359 @@ impl DiskCacheStore {
...
+            let start = parquet_tail_start;
+            if end == start {
+                return Ok::<_, StorageError>(Bytes::new());
+            }
+            storage.get_range(&storage_uri_a, start..end).await
+        };
+        let vec_fut =
+            async move { fetch_hint_ranges(storage_for_vec, storage_uri_b, vec_ranges).await };
+        let fts_fut =
+            async move { fetch_hint_ranges(storage_for_fts, storage_uri_c, fts_ranges).await };
```
Pranav2612000: Move these inside the else condition of `!offsets.open_blob.is_empty()` so that we dont even need to set up these futures if we already have blob data

src/supertable/reader_cache/disk.rs
```diff
@@ -801,22 +910,359 @@ impl DiskCacheStore {
...
+        // open on the real S3 wire (~25-50 ms in-region).
+        let range_src: Arc<dyn crate::superfile::LazyByteSource> =
+            Arc::new(crate::supertable::StorageRangeSource::with_unknown_size(
+                Arc::clone(&self.storage),
+                storage_uri.clone(),
+            ));
+        let lazy_reader = SuperfileReader::open_lazy_with(
+            Arc::clone(&range_src),
+            OpenOptions {
+                verify_crc: self.config.verify_crc_on_open,
```
Pranav2612000: Should this always be false for lazy loads? Otherwise we may be forced to read the whole segment, and its easier to do a normal read instead of ranged reads ( reduces number of calls which reduces costs ). Same for the lazy reader with hints

src/superfile/vector/builder.rs
ekechinwokah: codes_chunk means the 1-bit RaBitQ estimate-code bytes. updated the layout comment to say that explicitly and fixed the diagram: the current layout stores codes_chunk + doc_ids_chunk + full_chunk per cluster.

src/config/mod.rs
ekechinwokah: ah. the config-side description was stale relative to the runtime ColdFetchMode docs. updated it to describe the actual hybrid path: parallel range GETs feed both the foreground reader and disk-cache fill.

src/superfile/fts/reader.rs
ekechinwokah: `source_len` is the FTS subsection length, not the whole superfile. that range starts at the doc-lengths table offset and fetches the trailing doc-lengths region, not the entire FTS blob. renamed/commented it as doc_lengths_tail for clarity.

src/superfile/fts/reader.rs
ekechinwokah: we are already doing that: the lazy open fetches the doc-lengths tail once and installs it into the overlay, so metadata decode does not issue per-column range calls. clarified the comment to call this out explicitly.

src/supertable/reader_cache/disk.rs
ekechinwokah: done.

src/supertable/reader_cache/disk.rs
ekechinwokah: done

src/supertable/reader_cache/disk.rs
ekechinwokah: yes, latency for CRC check is really high. for lazy opens it should skip full CRC scans by default. made explicit at the disk-cache call site with `verify_crc`: false and added a comment explaining that full verification would force whole-segment range reads; in any case eager cache promotion can still verify after materializing the full segment.

src/supertable/reader_cache/disk.rs
```diff
@@ -1143,6 +1614,260 @@ async fn finalize_to_mmap(params: FinalizationParams) -> Result<(), DiskCacheErr
...
+        while next_chunk < n_chunks && in_flight.len() < n_streams {
+            if background_store_abandoned(store) {
+                return Ok(false);
+            }
+            let start = next_chunk * chunk_size;
+            let end = (start + chunk_size).min(size);
+            let storage = Arc::clone(&store.storage);
+            let file = Arc::clone(&file);
+            let uri = storage_uri.to_string();
+            in_flight.push(async move {
```
Pranav2612000, line 1591: Since we want to get the whole file, thoughts on using a get on the whole file instead of doing it in chunks?

src/supertable/reader_cache/disk.rs
```diff
@@ -1143,6 +1614,260 @@ async fn finalize_to_mmap(params: FinalizationParams) -> Result<(), DiskCacheErr
...
+        }
+        if background_store_abandoned(store) {
+            return Ok(false);
+        }
+    }
+
+    if background_store_abandoned(store) {
+        return Ok(false);
+    }
+    tokio::task::spawn_blocking(move || file.sync_all())
```
Pranav2612000, line 1618: do we need a file.flush() before?

src/supertable/reader_cache/disk.rs
```diff
@@ -801,22 +910,359 @@ impl DiskCacheStore {
...
+    /// Initial hinted bandwidth shape:
+    ///   - parquet tail: 64 KiB
+    ///   - vec fixed header: 32 B
+    ///   - FTS fixed header: 48 B
+    async fn cold_fetch_lazy_with_hints(
+        self: &Arc<Self>,
+        uri: &SuperfileUri,
+        offsets: &SubsectionOffsets,
+    ) -> Result<Arc<CachedEntry>, DiskCacheError> {
+        use crate::superfile::PrefetchedSource;
```
Pranav2612000: Nit: hoist to top

src/supertable/reader_cache/disk.rs
```diff
@@ -801,22 +910,359 @@ impl DiskCacheStore {
...
+    /// HEAD-discovery dance entirely — the
+    /// [`crate::supertable::StorageRangeSource`] inner source
+    /// starts with a known size, and the parquet tail GET fires
+    /// as a plain `get_range` instead of a suffix range.
+    ///
+    /// Initial hinted bandwidth shape:
+    ///   - parquet tail: 64 KiB
+    ///   - vec fixed header: 32 B
+    ///   - FTS fixed header: 48 B
+    async fn cold_fetch_lazy_with_hints(
```
Pranav2612000: Can `cold_fetch_lazy` and `cold_fetch_lazy_with_hints` be the same function? The function accepts and optional `offsets`. If the offsets has value, we do the preprocessing for parquet, fts and vectors. 
StorageRangeSource can also have a single function instead of known and unknown to make it even cleaner?

src/supertable/reader_cache/disk.rs
```diff
@@ -302,13 +337,57 @@ impl DiskCacheStore {
...
+            }
+            _ => Arc::new(crate::supertable::StorageRangeSource::with_unknown_size(
+                Arc::clone(&self.storage),
+                storage_uri,
+            )),
+        };
+        let reader = SuperfileReader::open_lazy_with(
+            range_src,
+            OpenOptions {
+                verify_crc: self.config.verify_crc_on_open,
```
Pranav2612000: Here as well, should it be false? ( The function comment mentions opening a range against object store )

src/superfile/fts/reader.rs
Pranav2612000: Still see it as `source_len` here. https://github.com/infino-ai/infino/blob/732a2dc4324de00bc50e3c067934d99c912e1980/src/superfile/fts/reader.rs#L203

Also, `source_len` is set to the value of the `kv::FTS_LENGTH` [here](https://github.com/infino-ai/infino/blob/732a2dc4324de00bc50e3c067934d99c912e1980/src/superfile/reader.rs#L242) which I guess is the value of the whole FTS blob?

src/config/mod.rs
```diff
@@ -124,6 +129,106 @@ impl Default for SupertableSettings {
...
+    /// Parallel range GETs serve both the foreground reader and the
+    /// disk-cache fill. Foreground returns after the range fetches;
+    /// pwrite, mmap, and cache registration finish in the background.
+    /// Uses one copy of segment bandwidth per cold miss.
+    HybridWithPrefetch,
+    /// Single-range sequential fetches (no background fill). Useful
+    /// for constrained environments where parallelism is undesirable.
+    RangeOnly,
+    /// Foreground returns immediately with a lazy reader (zero S3
+    /// GETs at open time); a background task fills the disk cache
```
Pranav2612000: Is this wrong - zero S3 GETs at open time?
Do we not make GET calls to fetch the vec and fts ranges in `cold_fetch_lazy_with_hints` ?

src/supertable/options.rs
```diff
@@ -743,9 +821,63 @@ impl SupertableOptions {
...
+                let root = cfg.storage.local_root.as_ref().ok_or_else(|| {
+                    BuildError::Store("storage.backend=local_fs requires storage.local_root".into())
+                })?;
+                Some(Arc::new(LocalFsStorageProvider::new(root)?) as Arc<dyn StorageProvider>)
+            }
+            StorageBackend::S3 => {
+                let bucket = cfg.storage.bucket.as_ref().ok_or_else(|| {
+                    BuildError::Store("storage.backend=s3 requires storage.bucket".into())
+                })?;
+                Some(Arc::new(S3StorageProvider::new_with_prefix(
```
Pranav2612000, line 841: The `new_with_prefix` method does not accept & use access & secret keys. Is this expected?
I think we should allow the user to pass them through code as well, instead of relying on the features of the object_store crate for direct authentication ( env variables, imds etc )

src/superfile/fts/reader.rs
Pranav2612000: Oh! Tagged the incorrect section. I meant to tag the `doc_lengths_array` here 
https://github.com/infino-ai/infino/blob/732a2dc4324de00bc50e3c067934d99c912e1980/src/superfile/fts/reader.rs#L392
Is this also included in this install
https://github.com/infino-ai/infino/blob/732a2dc4324de00bc50e3c067934d99c912e1980/src/superfile/fts/reader.rs#L210

src/supertable/reader_cache/disk.rs
ekechinwokah: get() would materialize the whole segment in memory before writing it to disk. the chunked range path is intentional: it bounds memory by cold_fetch_streams * `cold_fetch_chunk_bytes`, supports parallelism, and writes directly into the cache file.

src/supertable/reader_cache/disk.rs
ekechinwokah: i don’t think so. we are not using a buffered writer; the write tasks complete their file writes before sync_all(), and sync_all() flushes the file contents/metadata to storage.

src/supertable/reader_cache/disk.rs
ekechinwokah: done

src/supertable/reader_cache/disk.rs
ekechinwokah: yes, same answer as the other lazy path. lazy object-store opens should use `verify_crc: false`; otherwise CRC verification forces whole-segment reads and eats up memory.

src/config/mod.rs
ekechinwokah: fixed. it is only zero GETs when open_blob is present from manifest-part prefetch. otherwise the hinted lazy path does open-time GETs for parquet tail / vec / FTS ranges.

src/supertable/options.rs
ekechinwokah: yeah i haven't gotten to managing creds yet. i'm currently relying on instance creds/IAM.

src/supertable/reader_cache/disk.rs
ekechinwokah: done.

src/supertable/reader_cache/disk.rs
Pranav2612000: Makes sense.

src/supertable/reader_cache/disk.rs
Pranav2612000: Fair. But its worth noting that this increases the number of S3 get calls, which can affect user's cost

Pranav2612000 (approved):

src/supertable/reader_cache/disk.rs
ekechinwokah: yeah there's a balance of cost and latency that we need to continually make. run benches with `--object-store `to see estimated S3 costs, memory usage, and latency on 10m docs; it's helpful for making tradeoffs in short term.

src/superfile/fts/reader.rs
ekechinwokah: correct — the rename only hit the GET label, not the var. fixed: it's `fts_blob_len` now, and yes it's the full FTS subsection length (≈ `kv::FTS_LENGTH`), not the superfile. the range is still just the tail though: `doc-lengths `is the last region (header → fst → postings → doc-lengths), so `doc_lengths_table_offset..fts_blob_len` is dir + per-col arrays + crcs in one GET. `fst` is its own range, postings stay lazy — we never pull the whole blob at open.

src/superfile/fts/reader.rs
ekechinwokah: yep. the `overlay.install(doc_lengths_table_offset, doc_lengths_tail)` covers the whole tail — directory + every per-column array+crc — so each per-col array_region fetch is an overlay hit. open-time GETs stay O(1), not O(n_cols). added comments at both fetch sites.

PR #36, opened 2026-06-03
bench/supertable-update: correctly update using mutation counter across criterion calls

vinaykakade:
## Problem

The `supertable-update` bench panics partway through the `single_row_predicate_updates` group:

```
thread 'main' panicked at benches/supertable_update/main.rs:
update: CardinalityMismatch { matched: 0, new_rows: 1 }
```

## Original root cause

The bench updated `row{i}` → `row{i}-prime` against a **single shared supertable** inside `b.iter`. Criterion runs that closure many times (warm-up + measurement). The first run renames the matched rows to `…-prime`; every subsequent run replays the same predicate `row{i}`, which now matches **zero** rows — so `update` fails its cardinality contract (`matched: 0, new_rows: 1`) and panics. (The delete group had the same shape but didn't panic — a delete that matches nothing is a tolerated no-op, so it silently measured empty commits after the first run.)

In short: the mutation was destructive to its own match set, but the bench replayed an identical, now-stale predicate on every iteration.

## Fix

Rework both mutation benches to advance a **generation counter** against a reused supertable, so each iteration targets a fresh, still-present row:

- updates roll one row's title forward — `target-{i}` → `target-{i+1}`
- deletes consume distinct rows — `row{i:08}`

Each step commits, because `update` resolves its predicate against the **committed** manifest snapshot (an uncommitted rename isn't visible to the next predicate).

### Subtlety hit while implementing the fix

With the rolling counter, the panic initially persisted — because the counter `i` was declared **inside** the `|b|` closure. Criterion invokes that closure multiple times (estimate → warm-up → measurement) while the chain's state lives in the reused supertable, so on a later invocation `i` reset to `0` and the first update looked for `target-0` — long since rewritten — reproducing `matched: 0`. The counter is therefore hoisted **out** of the closure so it runs monotonically across every invocation.

Also adds a row-count invariant to each bench (update preserves the count; delete shrinks it) to guard against a future regression where mutations silently match nothing.

## Verification

Ran the bench exactly as the nightly does — `cargo bench --bench supertable-update`, default scale (10k docs / 20 mutations), default criterion timing. All three groups now complete, no panic:

| group | time | throughput |
|---|---|---|
| ingest | 6.63 ms | 1.51 M elem/s |
| delete | 29.5 ms | 678 elem/s |
| update | 1.24 s | 16 elem/s |

`cargo fmt --check` and `cargo clippy --bench supertable-update` are clean.

### Known follow-up (not addressed here)

The update group is slow (~16 elem/s) because the reuse-chain mints a new superfile per update and never compacts, so predicate resolution rescans a set that grows across the sampled iterations — the reported number trends slower the longer criterion samples. Worth a separate pass to make it a steady-state metric.

### DIFF
--- a/benches/supertable_update/main.rs
+++ b/benches/supertable_update/main.rs
@@ -129,12 +129,18 @@ fn bench_deletes(c: &mut Criterion) {
     });
     g.finish();
 
-    // Delete shrinks the table: the reused supertable must have
-    // fewer rows than the baseline — guards against a regression
-    // where deletes silently match nothing.
-    assert!(
-        count_rows(&st) < n as i64,
-        "deletes left the row count at the baseline {n} — nothing was removed"
+    // Exact end-state: each of the `i` delete attempts removed a
+    // distinct row while one was still present, so exactly
+    // `min(i, n)` rows are gone and `n - min(i, n)` remain. Asserting
+    // the exact count (not just `< n`) catches both under- and
+    // over-deletion — e.g. a predicate that silently matched nothing
+    // or matched more than one row.
+    let expected = (n as u64).saturating_sub(i) as i64;
+    assert_eq!(
+        count_rows(&st),
+        expected,
+        "after {i} single-row deletes over an {n}-row table, \
+         expected {expected} rows to remain"
     );
 }
 
--- a/benches/supertable_update/main.rs
+++ b/benches/supertable_update/main.rs
@@ -7,8 +7,9 @@
 //! - **Mutation throughput** — updates/sec + deletes/sec
 //!   measured over the full WAL pipeline (resolve + append +
 //!   tombstone + state-doc CAS + cleanup).
-//! - **End-state correctness** — a closing query asserts the
-//!   row count + presence/absence shape the bench expects.
+//! - **End-state correctness** — a closing assertion checks the
+//!   row-count invariant each mutation implies: updates preserve
+//!   the count, deletes shrink it.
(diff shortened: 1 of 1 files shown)


benches/supertable_update/main.rs
```diff
@@ -88,46 +102,39 @@ fn bench_deletes(c: &mut Criterion) {
...
-    // appeared) and is below the baseline by at least m on
-    // the first iteration. Use ≤ baseline to stay
-    // iteration-count agnostic.
+    // Delete shrinks the table: the reused supertable must have
+    // fewer rows than the baseline — guards against a regression
+    // where deletes silently match nothing.
     assert!(
-        total <= n as i64,
-        "row count {total} exceeded baseline {n} — mutations leaked rows"
+        count_rows(&st) < n as i64,
```
savannahar68: can we have exact assertion over here? 
comparator based assertions leads to missing out on bugs/issues
not relevant to benches but ideally throughout the repo we should have exact assertions. 

AI generates a lot of comparator based assertion which we should avoid.

wdyt?

savannahar68 (approved):
minior comment
please check

benches/supertable_update/main.rs
vinaykakade: Done - https://github.com/infino-ai/infino/pull/36/changes/1fd38e70852a3f7206c4b4853ad88eff048d048c

Issue #32, opened 2026-06-02
Follow-ups from #27 review: consolidate runtime bridging & reader opening, _id-only column scan


vinaykakade:
Follow-up items raised by @Pranav2612000 while reviewing #27. None block that PR — tracking them here.

## 1. Consolidate sync→async runtime bridging

`bridge_sync_to_async` in `src/runtime_bridge.rs` (around lines 38–55) is not the only place we bridge sync to async. `sql_runtime` also spins up a multi-threaded tokio runtime with a single worker thread, and there are other similar spots scattered in the code.

- [x] Consolidate all sync→async bridging logic into `src/runtime_bridge.rs`. — done in #38 (one `bridge_sync_to_async` + one process-wide `multi_thread(1)` runtime; `sql_runtime` and the ~8 inline `Handle::try_current()` blocks removed).
- [x] Document and decide the policy for when to use `new_current_thread()` vs `new_multi_thread()` vs `new_multi_thread()` pinned to 1 worker thread. — done in #38 (module docs: owned runtime is `multi_thread` pinned to 1 worker; `current_thread` never used for an owned runtime).

## 2. Unify superfile reader opening

In `src/supertable/wal/pipeline.rs` (around lines 1057–1083), the "cache hit, else storage fallback" logic for obtaining a `SuperfileReader` is open-coded.

- [ ] Extract a common helper for "get a reader for this superfile URI (try cache, else storage fallback)" and reuse it everywhere we need a reader.

## 3. `_id`-only column scan in `id_lookup`

`SuperfileReader::id_lookup` in `src/superfile/reader.rs` currently materializes whole rows.

- [x] Provide a way to read/iterate just the `_id` column instead of the whole row. — done in #33 (projects leaf column 0 via `ProjectionMask::leaves`).
- [x] (Related) Cache parquet metadata on the reader so it isn't re-read on each call — see the existing module-level TODO. Still open: `id_lookup` rebuilds the `ParquetRecordBatchReaderBuilder` from `self.bytes` on every call. Tracked in detail by #29.

Pranav2612000:
@vinaykakade starting work on 3

ekechinwokah:
each of the reported issues was addressed in #39: unified runtime, unified reader, and optimizing id scans.

PR #38, opened 2026-06-04
runtime: consolidate sync→async bridging into one bridge + runtime

vinaykakade:
## What

Consolidate every sync→async bridge in the supertable into a single `runtime_bridge::bridge_sync_to_async`, backed by one process-wide runtime.

## Why

There were **three** divergent sync→async bridging patterns:

1. `bridge_sync_to_async` (built a throwaway `current_thread` runtime per call for the no-ambient case);
2. a per-supertable `sql_runtime` (`multi_thread`, 1 worker);
3. ~8 copy-pasted `match tokio::runtime::Handle::try_current() { … }` blocks across `query` / `writer` / `recovery`.

Two of them even disagreed on the no-ambient path (throwaway `current_thread` vs the owned `multi_thread(1)`). This is the cleanup tracked in #32 (item 1).

## How

- One `bridge_sync_to_async`, backed by a single process-wide `multi_thread().worker_threads(1)` runtime (lazy `OnceLock`).
- Routed every call site through it; deleted `sql_runtime` (field, accessor, lazy-init) and the dead `Runtime`/`OnceLock` imports.
- Documented the runtime-flavor policy: `multi_thread(1)` for the owned runtime (a real runtime so `spawn`/`block_in_place` work, one worker since the work is IO-bound), `current_thread` never used for an owned runtime.
- The `multi_thread`-ambient requirement is unchanged; lifting the `current_thread`-ambient panic needs `Send + 'static` futures and is deferred.

## Scope / not in this PR

No public API change; the user-facing read/write surface stays sync. The async public surface (async `Connection`/`Table` facade) is separate, deferred work — benchmarking showed an async read path costs ~10–16% on sub-millisecond queries with no consumer today, so async will be added uniformly later when there's a real async caller.

## Testing

Behavior-preserving (the owned runtime was already `multi_thread(1)`). `make ci` green — full suite incl. multi-process + crash-injection. `cargo fmt --check` and `cargo clippy --all-targets -D warnings` clean.

### DIFF
--- a/src/supertable/handle.rs
+++ b/src/supertable/handle.rs
@@ -12,10 +12,9 @@
 //! and (via `Drop`) flips it false on release.
 
 use std::sync::atomic::AtomicBool;
-use std::sync::{Arc, Mutex, OnceLock};
+use std::sync::{Arc, Mutex};
 
 use arc_swap::ArcSwap;
-use tokio::runtime::Runtime;
 
 use super::error::{BuildError, OpenError};
 use super::manifest::Manifest;
@@ -65,14 +64,6 @@ pub(super) struct SupertableInner {
     /// supertable, constructed fresh on `create()` /
     /// `open()` with a 40-bit random worker_id.
     pub(super) id_generator: Mutex<crate::supertable::utils::idgen::IdGenerator>,
-    /// Lazily-initialized tokio Runtime that drives DataFusion
-    /// plans for `query_sql`. Tokio is single-worker here — it
-    /// runs the async I/O state machine, not CPU-bound work
-    /// (that lives on `options.reader_pool`). One Runtime per
-    /// supertable, shared across all SQL queries; allocated on
-    /// first use rather than at `create()` so supertables that
-    /// never run SQL don't pay the runtime cost.
-    pub(super) sql_runtime: OnceLock<Arc<Runtime>>,
     /// Per-process reader-side cache of per-superfile tombstone
     /// bitmaps. `Some` when storage is attached (the cache
     /// fetches sidecars from `superfiles/<id>.tombstones`);
@@ -89,25 +80,6 @@ pub(super) struct SupertableInner {
     /// opens five supertables holds five distinct ids). Minted
     /// via `IdGenerator::next_id()` once at create / open.
     pub(super) handle_id: crate::supertable::wal::state_doc::SupertableHandleId,
-}
-
-impl SupertableInner {
-    /// Get (or lazily build) the SQL Runtime.
-    pub(super) fn sql_runtime(&self) -> Arc<Runtime> {
-        Arc::clone(self.sql_runtime.get_or_init(|| {
-            Arc::new(
-                tokio::runtime::Builder::new_multi_thread()
-                    .worker_threads(1)
-                    .enable_all()
-                    .thread_name("supertable-sql")
-                    .build()
-                    .expect(
-                        "invariant: tokio Runtime build only fails on \
-                         catastrophic OS resource exhaustion",
-                    ),
-            )
-        }))
-    }
 }
 
 impl Supertable {
@@ -127,9 +99,8 @@ impl Supertable {
     ///   shadows existing committed state" footgun.
     ///
     /// Sync API. Internally bridges to async I/O for the
-    /// pointer probe + the open delegation via the same
-    /// `Handle::try_current() + block_in_place` pattern the
-    /// rest of the supertable's sync paths use. Works from
+    /// pointer probe + the open delegation via the shared
+    /// `runtime_bridge::bridge_sync_to_async`. Works from
     /// sync `#[test]` contexts and from multi-thread
     /// `#[tokio::test]` contexts; calling from a
     /// `current_thread` runtime panics (use the async
@@ -173,7 +144,6 @@ impl Supertable {
             manifest: ArcSwap::new(Arc::new(initial)),
             writer_outstanding: AtomicBool::new(false),
             id_generator: Mutex::new(id_generator),
-            sql_runtime: OnceLock::new(),
             tombstone_cache,
             handle_id,
         });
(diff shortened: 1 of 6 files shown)


src/supertable/query/superfile_reader.rs
```diff
@@ -61,36 +62,8 @@ pub fn superfile_reader(
...
-            {
-                Ok(rt) => rt,
-                Err(e) => {
-                    return Err(ReaderCacheError::OpenFailed {
-                        source: crate::superfile::ReadError::Io(std::io::Error::other(format!(
-                            "tokio runtime build for disk cache fetch: {e}"
-                        ))),
-                    });
-                }
-            };
```
Pranav2612000, line 90: This gracefully handles an error when tokio runtime failed to build. We crash in the `bridge_sync_to_async` fn. What's the best way here?

src/supertable/handle.rs
```diff
@@ -91,25 +82,6 @@ pub(super) struct SupertableInner {
...
-                    .worker_threads(1)
-                    .enable_all()
-                    .thread_name("supertable-sql")
-                    .build()
-                    .expect(
-                        "invariant: tokio Runtime build only fails on \
-                         catastrophic OS resource exhaustion",
-                    ),
-            )
-        }))
```
Pranav2612000, line 109: Not sure how important this is, but here we
- have a constant name to the thread
we lose out on this when we build from the generic function

src/supertable/query/hierarchical_iter.rs
```diff
@@ -69,18 +64,7 @@ pub fn load_kept_parts(
...
-            let rt = tokio::runtime::Builder::new_current_thread()
-                .enable_all()
-                .build()
-                .map_err(|e| {
-                    QueryError::Store(format!("tokio runtime build for hierarchical_iter: {e}"))
-                })?;
-            rt.block_on(drive)
-        }
-    }
+    bridge_sync_to_async(drive)
```
Pranav2612000, line 67: If uninitialized, this will initialize the runtime with 1 worker thread ( previously this fn did not have this restriction ). Can that cause issues?

src/supertable/handle.rs
```diff
@@ -575,11 +544,7 @@ impl Supertable {
         crate::supertable::wal::recovery::RecoveryReport,
         crate::supertable::wal::recovery::RecoveryError,
     > {
-        let drive = self.run_recovery_sweep_once();
-        match tokio::runtime::Handle::try_current() {
-            Ok(handle) => tokio::task::block_in_place(|| handle.block_on(drive)),
-            Err(_) => self.inner.sql_runtime().block_on(drive),
-        }
+        bridge_sync_to_async(self.run_recovery_sweep_once())
```
savannahar68, line 547: what is the logic of making only create method as async and rest all are async?
how does python binding works?

savannahar68 (approved):
lgtm
one minor comment for my understanding as i dont see us having a better api which will be exposed in future to pthon bindings

src/supertable/query/superfile_reader.rs
vinaykakade: In the graceful handling (current main), runtime is generated once per call - no we are doing it once using OnceLock. If we can't get a runtime from bridge_sync_to_async - would generally be the case when OS resources are exhausted - there isn't really a way to recover from the higher level, so I feel it is right to not have a fallback here.

[Another option is to build a runtime at supertable open (or, rather infino::init that I am planning to add soon) and use that, we can explore that if this becomes an issue.]

src/supertable/handle.rs
vinaykakade: Useful for debugging, good catch. Added.

src/supertable/handle.rs
vinaykakade: This PR is only standardizing our current usage of creating runtimes, to reduce duplicate code. Unification of APIs will be part of one of the next PRs.

src/supertable/query/hierarchical_iter.rs
vinaykakade: Good question — it doesn't cause issues, and the old N-worker runtime didn't actually help this path either.

load_kept_parts uses join_all, which is single-task concurrency, not multi-thread parallelism: it polls all N part futures on one task and never fans them across workers (no spawn). So worker count is irrelevant to how it schedules , the old Runtime::new() gave it no parallelism advantage.

The actual GET parallelism lives below the worker pool and isn't bounded by worker_threads(1): object_store's remote backends are reactor-async (one worker drives many in-flight requests — the whole point of async I/O), and the local backend offloads reads to spawn_blocking (the separate blocking-thread pool). So N GETs still overlap; wall-clock stays ~max(latency).

The 1 worker only bottlenecks CPU-bound spawned async tasks, and we keep CPU work on rayon. I added a note to the runtime_bridge module docs (56d620f) so this is spelled out for the next reader.

vinaykakade: Actually, you were right. There was a perf regression in my prior approach, switch to the thread local runtime approach which keeps performance comparable. See below for supertable fts bench:


  | query | main | thread-local fix | Δ | multi_thread(1) was |
  |---|---|---|---|---|
  | single_rare | 182.0 µs | 182.9 µs | +0.3% (ns) | −0% |
  | single_common | 162.1 µs | 161.6 µs | −0.8% | −0.7% |
  | two_term_or | 281.5 µs | 281.7 µs | +0.2% (ns) | **+6.4%** |
  | three_wide_or | 1.797 ms | 1.795 ms | −0.2% (ns) | **+11.6%** |
  | three_similar_or | — | 5.798 ms | **−1.8%** | **+9.0%** |
  | five_term_or | 11.94 ms | 11.85 ms | −0.7% (ns) | **+15.6%** |
  | prefix | 28.82 ms | 28.65 ms | −0.3% (ns) | **+17.4%** |

savannahar68: imo we should'nt do `switch to the thread local runtime` 
the perf number for current_thread runtime will be better than setting up multi_thread` runtime
but comes with a drawback that we cannot spin background thread (block in place threads)
and if we do blocking operation on same current_thread runtime then it takes a hit on concurrency

so yes perf will take some hit, but overall we should go ahead with mutli_thread runtime with single worker thread approach considering larger picture of being used in python binding etc

PR #39, opened 2026-06-04
Add SQL-native search via DataFusion

ekechinwokah:
# **Summary**

This PR brings SQL-native search to the object-store read path and unifies the query and build substrates across full-text, vector, and SQL. Users can run full-text, vector, and hybrid retrieval directly from SQL while keeping the same snapshot-consistent execution model and skip-pruned fan-out the typed query APIs use — and that fan-out and pruning are now genuinely *one* code path shared by every modality.

### **SQL Search Table-Valued Functions**

Search is exposed as DataFusion table-valued functions:

- `vector_search(column, query, k)` — vector kNN as a relation.
- `bm25_search(column, query, k)` — full-text BM25 search.
- `bm25_search_prefix(column, prefix, k)` — prefix BM25 (last token expanded).

A shared TVF/exec framework (`query/exec/`) handles argument parsing, the common output schema (`_id`, projected scalar columns, `score`), and hit resolution. Each TVF lowers to a custom `ExecutionPlan` that calls the existing async search kernels.

### **DataFusion Integration**

- `SupertableProvider` exposes a supertable as a DataFusion `TableProvider`.
- **Unified two-tier pruning** on the SQL scan path (`query/prune.rs`): the same `select_segments` substrate FTS uses now drives SQL. A part-tier and segment-tier pass drops segments using persisted per-segment min/max from the manifest (`scalar_skip`), then DataFusion's `ParquetSource` prunes row groups / pages and pushes projection + limit over the survivors.
- **SQL equality prunes through the FTS index.** A `WHERE col = 'x'` on an FTS-indexed column lowers to a `TermPresence` leaf and is dropped against a segment's term bloom — if the tokens can't be present, the segment never reaches DataFusion. (Correctness-preserving: presence is necessary, not sufficient, so the scan still filters.)
- Deletes are honored lazily during the scan: a segment's tombstone bitmap is translated into a `ParquetAccessPlan` row selection, so deleted rows are skipped at decode rather than materialized then filtered.
- `query_sql` returns `Vec<RecordBatch>` and caches the `SessionContext` (provider + TVFs) keyed on the manifest snapshot; `scan_ids_matching` reuses the same provider for writer-side delete/update id-capture.
- `ensure_fresh()` is applied on every read entry point so SQL honors the same consistency policy (Strong / BoundedStaleness / Snapshot) as the search APIs.
- `SuperfileReader` gains `take_by_local_doc_ids` (column-projected, row-selected resolve of hits → `_id` + scalars) and a lazy-open path; metadata is parsed once at open and reused per resolve.

### **Shared query fan-out**

Vector, BM25, and prefix search now share one cross-segment fan-out (`query/dispatch.rs`): async I/O (segment opens, cold range GETs) on the shared query runtime, CPU work (scoring, rerank) on the rayon pool. FTS previously hand-rolled its own fan-out and duplicated the dispatch helpers — that copy is deleted, so the "shared dispatch" contract the code already documented is now real. Vector search gained an async sibling so the supertable fan-out drives every segment concurrently instead of serializing cold GETs.

### **Global vector cluster pruning**

Vector search no longer probes `nprobe` clusters in *every* segment. The manifest's `VectorSummary` now carries Sq8-quantized per-cluster IVF centroids (`ClusterCentroids`: codes + per-cluster mins/scales + counts), so search ranks **all clusters across all segments globally** and fans out only to the winning `(segment, cluster)` pairs. This is an additive, versioned manifest field with round-trip encode/decode; segments written without it decode to empty clusters and fall back to per-segment probing. New reader hooks: `VectorReader::cluster_centroids` / `search_clusters_async` and `SuperfileReader::vector_search_clusters`.

### **Shared build fan-out**

The commit path mirrors the query side: segment builds fan out across the writer pool through a shared `build::fanout_shards`. `SuperfileBuilder::finish()` overlaps Parquet body encode with index finalization when there's no vector index, and serializes body encode after the index blobs when a vector index is present (avoiding rayon oversubscription). The `max_concurrent_vector_builds` knob is gone.

### **Benchmarks and Tests**

**Benchmarks** — unified real-S3 (`INFINO_REAL_S3_BUCKET`) supertable harness covering FTS, vector, and combined FTS+vector ingest plus hot and cold search. Supertable ingest improves +17–26% vs `main`; recall@10 stays ≥ 0.90 at default options. See `benches/README.md` for the full, updated numbers.

**Tests**

- `smoke_s3` — TVF-through-`query_sql` over the real S3 wire.
- `hybrid_search` SQL integration tests (schema/projection, set-union property, ranking).
- `fanout_concurrency` — concurrency stress over the shared tokio/rayon fan-out (bm25 / vector / SQL / COUNT(*)) asserting liveness + determinism.
- SQL-equality-on-FTS bloom-prune tests (prune never drops a true match).
- Manifest `ClusterCentroids` round-trip + writer-populates-clusters coverage.


benches/utils/unified_object_store.rs
```diff
@@ -579,6 +579,10 @@ fn bench(c: &mut Criterion) {
         diag::diagnose_real_s3_supertable_e2e();
         return;
     }
+    if std::env::var("INFINO_DIAG_QUERY_SQL_OVERHEAD").is_ok() {
```
Pranav2612000, line 562: How are we tracking all the different ENV variables and config knobs that we've exposed. We have so many that its possible to forget / miss out on some

src/storage/s3.rs
Pranav2612000, line 1: Let's move these S3 improvements to a separate PR

src/superfile/builder.rs
```diff
@@ -179,8 +179,22 @@ pub struct BuilderOptions {
...
+    /// cost. Compression stays on; the only cost is a few extra
+    /// page headers + offset-index entries for the id column.
+    pub id_page_size_limit: usize,
 }
 
+/// Default per-column data-page size limit for the id column
+/// (uncompressed bytes). At 16 bytes/row (`Decimal128`) this is
+/// ~512 rows/page, vs the ~65 536-row single page a default
+/// (1 MiB) limit produces for a full row group.
+pub const DEFAULT_ID_PAGE_SIZE_LIMIT: usize = 8 * 1024;
```
Pranav2612000, line 196: What was this value before? 
Will some queries be slower now, because we need to "read" more pages

src/superfile/reader.rs
```diff
@@ -418,6 +438,169 @@ impl SuperfileReader {
         self.bytes.as_ref()
     }
 
+    /// Resolve in-segment row offsets to their durable identity
+    /// (007 D2 / Phase 3).
```
Pranav2612000: Nit: Remove references

src/superfile/reader.rs
```diff
@@ -418,6 +438,169 @@ impl SuperfileReader {
...
+        }
+        let selection = RowSelection::from(selectors);
+
+        // Metadata-cached read: reuse the `ArrowReaderMetadata` parsed
+        // at open (no per-call footer parse), so this targeted read
+        // only pays the projected-column page decode. The page index
+        // (when present) lets `RowSelection` seek to the relevant
+        // pages. CPU-bound over in-memory bytes — callers fan these
+        // across `options.reader_pool` for cross-segment parallelism.
+        let builder = ParquetRecordBatchReaderBuilder::new_with_metadata(bytes, arrow_meta);
```
Pranav2612000, line 553: In the `id_lookup` fn in reader.rs, we build a ParquetRecordBatchReaderBuilder without using the arrow_meta. Should we do the same here, so that we don't need to store the arrow_meta?

src/supertable/lazy_source.rs
Pranav2612000, line 1: Let's move this to a separate PR?

src/supertable/tombstones/cache.rs
Pranav2612000, line 1: Lets move these changes to a separate PR

src/supertable/query/sql.rs
```diff
@@ -95,30 +74,77 @@ impl Supertable {
...
-        let scalar_schema = self.options().scalar_schema();
-        let tombstone_cache = reader.tombstone_cache.clone();
 
-        let sql = sql.to_owned();
+        // Cache hit: reuse SessionContext (provider + 3 TVFs
+        // already registered) on the same manifest. Skips the
+        // ~1.5 ms `SessionContext::new()` + register_* setup.
+        let ctx = {
+            let mut guard = self
+                .sql_session_cache()
```
Pranav2612000: Should sql_session_cache be a RwLock instead of a Mutex, since we'll be receiving read heavy workloads

src/supertable/query/sql.rs
Pranav2612000, line 190: Can this fn also be modified to make use of the context, instead of having to build a table in memory?

Pranav2612000 (commented):
1/n

docs/architecture/supertable.md
```diff
@@ -203,6 +224,17 @@ into fixed-size chunks driven in parallel, which lowers peak memory
...
+Range reads are exact-or-error. A read for a byte range returns
+exactly that many bytes or a typed error — never a short buffer. If the
+backend returns a truncated or clamped range (a body cut short by a
+transport hiccup, or a range past the object's real end), the missing
+tail is re-fetched to complete the range, and a genuinely unsatisfiable
+range surfaces as an error rather than silently feeding a partial slice
+to a reader. Transient transport failures on an idempotent range GET
+(for example a send failure on a pooled connection the backend closed)
+are retried with backoff, so the burst of concurrent range GETs a cold
+query fans out does not surface a flaky connection as a read failure.
```
Pranav2612000: Should this be here? Seems like more of an implementation detail, rather than the architecture

benches/utils/unified_object_store.rs
ekechinwokah: upcoming cleanup PR moves most of these to constants.

src/supertable/lazy_source.rs
ekechinwokah: would prefer not to break up the PR.

src/supertable/tombstones/cache.rs
ekechinwokah: would prefer not to break up the PR.

src/superfile/builder.rs
ekechinwokah: no per-column value before; `_id` just used the global 1 MiB default (one ~65k-row page). this only shrinks the `_id` column and makes lookups faster: `take_by_local_doc_ids` uses a `RowSelection`, so a point lookup decompresses a ~512-row page instead of the ~1MB one (the 989us->171us win). only `id_lookup` (cold recovery) reads all pages.

src/superfile/reader.rs
ekechinwokah: done

src/superfile/reader.rs
ekechinwokah: intentional — hot rerank path, caching the meta skips a per-call footer+page-index parse. `id_lookup` is cold recovery so it just rebuilds the reader.

src/superfile/vector/reader.rs
```diff
@@ -1381,6 +1373,227 @@ impl VectorReader {
...
+    /// region, per-cluster code prefixes + Sq8 meta, survivor rerank
+    /// rows) are `await`ed on the caller's runtime instead of bridged
+    /// through a per-call throwaway runtime. This is what lets the
+    /// supertable vector fan-out drive every segment concurrently on
+    /// the shared query runtime — mirroring the FTS
+    /// `bm25_search_pretokenized` path — rather than serializing cold
+    /// object-store GETs. The CPU steps (centroid/code scoring,
+    /// rerank) call the same helpers as the sync path and parallelize
+    /// on the global rayon pool; warm/in-memory ranges still resolve
+    /// sync/zero-copy via `try_get_range_sync` with no `await`.
```
Pranav2612000, line 1308: Is it possible to refactor this so that there's a lot less duplicate code between the two version?

src/supertable/query/sql.rs
ekechinwokah: critical section is just an Arc::ptr_eq + cheap clone; guard drops before the query runs, so no read contention to relieve. mutex is fine here.

src/superfile/vector/reader.rs
```diff
@@ -2085,77 +2305,94 @@ fn get_cluster_ranges_coalesced_with_extra(
...
 
-    let mut fetch_ranges: Vec<Range<usize>> = groups.iter().map(|(r, _, _)| r.clone()).collect();
-    fetch_ranges.push(extra);
-    let mut fetched = source.get_ranges_parallel(&fetch_ranges)?;
-    let extra_bytes = fetched.pop();
-
-    let mut out: Vec<Option<Bytes>> = vec![None; ranges.len()];
-    for ((merged_range, _, members), bytes) in groups.into_iter().zip(fetched) {
+    let fetch_ranges: Vec<Range<usize>> = groups.iter().map(|(r, _, _)| r.clone()).collect();
+    let mut scatter: Vec<(usize, usize, usize)> = vec![(0, 0, 0); ranges.len()];
```
Pranav2612000, line 2376: Nit: initialize using `with_capacity()` and push later

src/supertable/query/sql.rs
ekechinwokah: good call — done. routed it through the `SupertableProvider` so delete/update id-capture gets the same segment-skip + row-group pruning + lazy tombstone filtering as query_sql, instead of building a `MemTable`. dropped the eager path.

src/storage/s3.rs
ekechinwokah: assuming you are talking about the s3 fix, that's in main. but otherwise would prefer not to break up PR.

src/supertable/tombstones/cache.rs
Pranav2612000: Any reason why? The negative TTL change seems to be unrelated to "Add SQL support via Datafusion". Breaking up
- helps catch regressions ( easier to pin-point which commit and which change increased latencies )
- makes it easier to review and faster to get things merged into main ( reducing merge conflicts )
- makes rollbacks easier

0kay if we don't break it down for now. But I would appreciate if we start breaking PRs down, especially as we make the project open source. Its a proven engineering strategy. Would require a slight change of mindset where every time we're add something to an existing PR, we first think if it can be done in a separate PR. 
We'll be running pre-commit checks for each PR which would track latency & cost to ensure that we stick to our goal.

src/supertable/query/provider.rs
```diff
@@ -0,0 +1,698 @@
...
+            let path = entry.uri.storage_path();
+            let size = bytes.len() as u64;
+
+            // Lazy delete path: translate this segment's tombstone
+            // bitmap into a Parquet row selection so deleted rows are
+            // never decoded. Absent/empty overlay → full scan, zero
+            // overhead. The `local_doc_id` in the bitmap is the row's
+            // global position within the segment's Parquet body, which
+            // is exactly the coordinate `ParquetAccessPlan` selects on.
+            let access_plan = match self.tombstone_cache.as_ref() {
```
Pranav2612000, line 321: Should we `prefetch` to tombstone_cache like we do for bm25_search?

src/supertable/query/provider.rs
```diff
@@ -0,0 +1,698 @@
...
+
+        // Expose the surviving segments' bytes to DataFusion via an
+        // in-memory object store. `parquet_bytes()` is Arc-backed,
+        // so `clone()` / `PutPayload::from` are refcount bumps.
+        //
+        // One `Instant::now()` for the whole scan so every per-segment
+        // tombstone lookup shares the same `SidecarCache` TTL
+        // reference (mirrors the eager `build_mem_table` path).
+        let now = std::time::Instant::now();
+        let object_store = Arc::new(InMemory::new());
```
Pranav2612000: Nit: variable name `object_store` is misleading

src/supertable/tombstones/cache.rs
ekechinwokah: understood. PRs will be smaller going forward once the big items are merged. sent you a message on slack.

docs/architecture/supertable.md
ekechinwokah: agreed — fixed.

src/superfile/vector/reader.rs
ekechinwokah: good catch — pulled the whole shortlist+candidate stage into a shared build_shortlist; the two now only differ on the sync-vs-async fetch calls, so scoring/recall can't drift between them.

src/superfile/vector/reader.rs
ekechinwokah: with_capacity+push would reorder the outputs.  scatter[idx] is a scatter-write keyed by the original range position, not a sequential append, so vec![default; n] is the right shape;

src/supertable/query/provider.rs
ekechinwokah: done

src/supertable/query/provider.rs
ekechinwokah: agreed. renamed to `mem_store`.

Pranav2612000 (approved):

PR #43, opened 2026-06-05
benches: selectable object-store backend (Azure + S3) with ephemeral cleanup

pratyushlokhande:
## What

Lets the supertable/superfile warm/cold benches run against **Azure Blob** (on par with S3), with the backend chosen **explicitly**, and makes every real-backend run clean up after itself.

## Why

`benches/utils/tiers.rs` picked the warm/cold backing store by **inferring** from which credential env happened to be set (real S3 if `INFINO_REAL_S3_BUCKET`, else the in-process s3s-fs emulator). Problems:
- no Azure option (`object_store` enabled `["aws"]` only);
- implicit selection — exporting two credential sets silently picked one;
- a real-backend run wrote a throwaway table under a unique prefix and **never deleted it**, leaking an index on the object store every run.

## Changes

**Explicit backend selection.** `INFINO_BENCH_STORE` = `s3s_fs` (default) | `s3` | `azure` — never inferred. The chosen backend must have its location env set or it errors. `BenchStore { S3sFs, Remote(RemoteBackend) }` with a pure, unit-tested `parse`. `RemoteBackend { S3, Azure }` is the single per-backend construction site (provider / label / prefix_root) — adding a backend is one arm.

| `INFINO_BENCH_STORE` | Store | Extra env |
|---|---|---|
| _unset_ / `s3s_fs` | in-process s3s-fs emulator | — |
| `s3` | AWS S3 | `INFINO_REAL_S3_BUCKET` + `AWS_*` |
| `azure` | Azure Blob | `INFINO_REAL_AZURE_CONTAINER` + `AZURE_STORAGE_ACCOUNT_NAME` + `AZURE_STORAGE_ACCOUNT_KEY` |

**Honest, single-sourced labels.** Storage labels are `s3` / `azure` / `s3s_fs` — dropped the misleading `real_` prefix (the emulator already has its own distinct name, so `real_azure` implied a fake counterpart). The set lives once in `tiers::STORAGE_LABELS` (guarded by a test against `RemoteBackend::label`); markdown report-gen iterates it, and the redundant hardcoded group-name lists in the FTS/vector search benches are gone (the family alias already prefixes every group name).

**Ephemeral cleanup.** A real-backend run registers its unique prefix; each real-backend bench `main` drains it (`tiers::cleanup_ephemeral`, list + delete) after the run — statics don't `Drop` at process exit, so teardown is explicit, and a failed cleanup list is surfaced, not swallowed. Removed the now-dead `StorageFixture.remote` and `SuperfileCommitted.{remote, cleanup_path, object_path, object_size}` (no readers; the prefix-level registry supersedes the never-consumed per-object cleanup).

**Docs.** `benches/README.md` gains an "Object-store backends" section: selection table, run examples, group labels, cleanup, and the emulator caveat.

## How to run

```sh
# Default — emulator, no credentials
cargo bench --bench supertable_all -- supertable_fts

# Real AWS S3
INFINO_BENCH_STORE=s3 INFINO_REAL_S3_BUCKET=my-bucket \
  cargo bench --bench supertable_all -- supertable_fts

# Azure Blob
INFINO_BENCH_STORE=azure INFINO_REAL_AZURE_CONTAINER=my-container \
  AZURE_STORAGE_ACCOUNT_NAME=... AZURE_STORAGE_ACCOUNT_KEY=... \
  cargo bench --bench supertable_all -- supertable_fts
```

## Testing

- 7 `tiers` unit tests (selector defaults / required-env / unknown-store error / no-inference-from-creds / label / `STORAGE_LABELS` covers every backend).
- `supertable_all`, `superfile_fts`, `superfile_vector`, `object-store` benches compile; clippy + fmt clean.
- Verified end-to-end on a live Azure account at reduced scale: `INFINO_BENCH_STORE=azure` → ingest → open → BM25 correctness → `cleaned 193 objects … (azure)`, exit 0. The ephemeral run self-cleans.

## Notes for reviewers

- **Scope** is backend selection + Azure + cleanup. Building a table once and scoring it on later runs without re-ingesting is a follow-up.
- **Pre-existing, not introduced here:** the in-process s3s-fs emulator is unstable during the supertable's concurrent multi-commit ingest (dropped connections / commit-retry exhaustion). Reproduced on `main` too. Real backends (`s3` / `azure`) are the reliable object-store path; the emulator is a credential-free wire-path check, best for the single-commit superfile benches. A dedicated fix is its own task.
- This branch is benches-only; no library code changes.

### DIFF
--- a/benches/utils/tiers.rs
+++ b/benches/utils/tiers.rs
@@ -49,10 +49,11 @@ impl Tier {
 }
 
 /// Storage labels that can appear in a warm/cold search group name
-/// (`{family}_{tier}_search_{label}`). The emulator is `s3s_fs`; real
-/// backends use [`RemoteBackend::label`]. Markdown report generation
-/// iterates this set; a unit test guards it against `RemoteBackend::label`.
-pub const STORAGE_LABELS: &[&str] = &["s3s_fs", "s3", "azure", "azurite"];
+/// (`{family}_{tier}_search_{label}`) — one per (store, mode): `s3_local`
+/// (in-process s3s-fs), `s3_remote`, `azure_local` (Azurite), `azure_remote`.
+/// Markdown report generation iterates this set; a unit test guards it
+/// against the backend `label()` methods.
+pub const STORAGE_LABELS: &[&str] = &["s3_local", "s3_remote", "azure_local", "azure_remote"];
 
 /// Default Azurite container (must already exist in the emulator).
 const AZURITE_CONTAINER_DEFAULT: &str = "infino-bench";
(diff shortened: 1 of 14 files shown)


benches/utils/tiers.rs
```diff
@@ -102,6 +103,56 @@ pub fn block_on<F: std::future::Future>(fut: F) -> F::Output {
...
+}
+
+static CLEANUP: Mutex<Vec<CleanupTarget>> = Mutex::new(Vec::new());
+
+/// Delete every remote prefix registered this run. Call once from the
+/// `main` of any bench that goes through `backing_store` against a real
+/// store, after the criterion groups finish — statics don't run `Drop`
+/// at process exit, so teardown is explicit. No-op for s3s-fs (its
+/// tempdir self-cleans) and for persisted runs (nothing is registered).
+pub fn cleanup_ephemeral() {
```
Pranav2612000, line 124: Nit: can this be configurable? There may be use cases where I want the data to not be cleaned up

Pranav2612000 (approved):
Minor Nit/suggestion. Everthing else looks good

benches/utils/tiers.rs
pratyushlokhande: Yes, already in plan. In 2-step flow (prepare + benchmark). we will have control whether to delete or not.

PR #59, opened 2026-06-08
Restore reader method on supertable

ekechinwokah:
#34 had a regression, removing the reader method from the supertable API

the code needs to match the intended public contract [here](https://github.com/infino-ai/infino/blob/main/docs/architecture/overview.md). 

restoring.

sanity check shows no regression on performance.

**Supertable FTS reader — real S3 (100K docs × dim 384, 8 superfiles)**

| **Operation** | **Latency** |
| --- | --- |
| producer build + commit | 6.68 s |
| cold_open | 543 ms |
| cold `reader().vector_search` (nprobe=8) | 385 ms |
| cold `reader().bm25_search` | 48.4 ms |
| **warm `reader().vector_search`** | **2.6 ms** |
| **warm `reader().bm25_search`** | **1.1 ms** |


src/supertable/handle.rs
```diff
@@ -600,10 +600,18 @@ impl Supertable {
     /// Pinned reader. Captures the current manifest at construction
     /// and holds it for its lifetime. New commits don't affect a
```
muralikpbhat, line 601: is this still true given we are refreshing every time?

muralikpbhat: > is this still true given we are refreshing every time?

Please ignore. I was confused a bit by the docs in table and reader.

muralikpbhat (approved):

PR #60, opened 2026-06-08
feat superfile writer: add fn to create batch from another superfile

Pranav2612000:
### What does this PR do?
- Adds a `add_batch_from_reader` fn to create a new batch from an existing reader

- Building blocks needed for compaction. This change exposes an api to build a superfile from existing superfiles

### How do we do this change?
- From the reader, we read the parquet data to get the RecordBatches
- For vectors, we read the raw vectors written ( using the new `get_vectors_fp32` fn )
- Both these are passed to the existing `add_batch` fn

### Gotchas
- Merging for non fp32 vector config ( e.g sp8 ) is non-trivial. This will be handled separately
- FTS's may also be merged directly, but would require some modifications with the doc_ids. This will also be handled separately 

### How has this been tested?
- Unit tests added for both new fns

### DIFF
--- a/src/superfile/builder.rs
+++ b/src/superfile/builder.rs
@@ -76,13 +76,13 @@
 //! `inf.fts.columns` JSON already carries a `"tokenizer"` field on
 //! each entry (currently always `"ascii_lower"`), so the on-disk
 //! format is forward-compatible without a file rewrite.
-
-use crate::superfile::BuildError;
 use crate::superfile::format::footer::{encode_parquet_body, splice_index_blobs};
 use crate::superfile::format::{self, kv};
 use crate::superfile::fts::builder::FtsBuilder;
 use crate::superfile::fts::tokenize::Tokenizer;
 use crate::superfile::vector::builder::VectorBuilder;
+use crate::superfile::vector::reader::ColumnReader;
+use crate::superfile::{BuildError, SuperfileReader};
 // `VectorConfig` lives in `vector::builder` and is re-exported below
 // (single source of truth post-collapse — see the `pub use` line).
 pub use crate::superfile::vector::builder::VectorConfig;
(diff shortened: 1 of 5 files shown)


src/superfile/vector/reader.rs
```diff
@@ -1569,6 +1503,127 @@ impl VectorReader {
...
+        // Collect all cluster ranges needed for fetching
+        let mut cluster_ranges: Vec<Range<usize>> = Vec::new();
+        let mut cluster_meta: Vec<(usize, u32, u32)> = Vec::new();
+
+        for c in 0..col.n_cent as usize {
+            let (off, cnt) = read_cluster_entry(&cluster_idx, c);
+            if cnt == 0 {
+                continue;
+            }
+            cluster_ranges.push(col.cluster_block_range(off, cnt));
```
savannahar68, line 1546: can you remove dead_code annotation from `cluster_block_range`

src/superfile/builder.rs
```diff
@@ -447,6 +444,79 @@ impl SuperfileBuilder {
...
     }
 
+    /// Add all data (Parquet + fts + vectors) from another [`SuperfileReader`] to this builder.
+    ///
+    /// Extracts the record batch and vectors from the reader and adds them via
+    /// [`Self::add_batch`]. This is useful for merging superfiles or copying data
+    /// between builders.
+    ///
+    /// **Requirements:**
+    /// - The reader's vector columns must use the **Fp32 codec**. Other codecs
```
savannahar68, line 528: should we also add that schema should be same between superfiles as a requirement? 
or should that be done outside at compaction layer function?

savannahar68 (approved):
lgtm
minor comment about schema check which is TBD

src/superfile/reader.rs
```diff
@@ -440,6 +437,31 @@ impl SuperfileReader {
     pub fn parquet_bytes(&self) -> Option<&Bytes> {
         self.bytes.as_ref()
     }
+    /// Returns a record batch containing all documents with all columns
+    pub fn get_record_batch(&self) -> Result<RecordBatch, ReadError> {
```
savannahar68, line 441: lets skip deleted docs in this funcrtion

PR #62, opened 2026-06-08
ci: on-demand Azure supertable benchmark (in-region VM + sccache blob cache)

pratyushlokhande:
## What

A `workflow_dispatch` workflow (`.github/workflows/supertable-bench-azure.yml`) that benchmarks **supertable ingest against real Azure Blob storage from in-region compute**, building on the VM with an sccache compile cache backed by Azure Blob.

Steps (single-responsibility):
1. OIDC login to Azure.
2. Provision an ephemeral VM in the storage account's region (per-run RG `rg-infino-bench-<run_id>`, tagged; NSG with no default inbound rule, SSH locked to the runner's `/32`).
3. Wait for SSH.
4. **Clone + toolchain + sccache** — clone `@ref` via the job token, install rustup + a prebuilt sccache.
5. **Build (sccache)** — compile with `RUSTC_WRAPPER=sccache` + `CARGO_INCREMENTAL=0`, print cache stats, record the binary path.
6. **Run** — exec that binary directly (no `cargo`) against Azure.
7. Summarize → step summary (results table, errors/panics, warnings) + log artifact.
8. Teardown — `az group delete` on success/failure/timeout/cancellation.

## Why in-region

The supertable bench moves multi-GB to object storage and reads it back during commits. Over a non-colocated network path that's slow and unreliable (single-stream transfers collapse; large reads time out). Co-locating compute with the bucket runs it over the Azure backbone. This is the supported way to get representative supertable numbers.

## Why sccache over blob (build on the VM)

- **Warm builds across branches.** sccache is content-addressed (not branch-keyed), so the dependency graph compiles once and is reused on every branch's ephemeral VM; only changed crates recompile.
- **No portability contract.** Building on the VM means build env == run env — no glibc/arch/CPU concerns from shipping a binary across hosts.
- **No new infra.** The cache reuses the bench's own blob container under a `sccache` key prefix (separate from bench data); the connection string is derived from the existing account creds — no new secret, no dedicated container.
- VM stays ephemeral/cattle; the durable cache lives in blob.

## Build/run split

Build and run are separate steps. To avoid a cargo re-fingerprint/rebuild between them, the run step **execs the recorded binary directly** rather than re-invoking `cargo bench`.

## Safety / cost

- **Hard timeout** (`timeout_minutes`, default 30) caps VM billing; teardown runs on timeout.
- **Serialized** (`concurrency`) — one VM billing at a time; queued, never cancelled.
- **SSH** locked to the runner's egress `/32`; NSG has no default inbound rule.
- **OIDC** — no static cloud credentials in the repo.
- **Cost tags** on RG + VM → Cost Management attributes spend by `purpose` / `run-id` after teardown.

## Inputs

`docs` (default `1000000`, plain integer), `vm_size` (default `Standard_D8s_v7` — current gen, available in eastus), `location`, `ref`, `keep_table`, `timeout_minutes`.

## Prerequisites (reviewer / operator)

- Secrets: `AZURE_CLIENT_ID` / `AZURE_TENANT_ID` / `AZURE_SUBSCRIPTION_ID`, `AZURE_STORAGE_ACCOUNT_NAME` / `AZURE_STORAGE_ACCOUNT_KEY`, `INFINO_REAL_AZURE_CONTAINER`.
- OIDC service principal: Contributor on a dedicated bench resource group/subscription (not prod-wide); federated credential scoped to this repo.
- Optionally set a blob lifecycle policy on the `sccache/` prefix for cache eviction.

## Validation

`actionlint` clean (schema + expressions + shellcheck on every `run:` block); heredocs terminate at column 0. Not yet run end-to-end — a first dispatch validates the standalone-binary run + OIDC scope + sccache blob auth.

- Scheduled sweep to delete stray `rg-infino-bench-*` groups if a runner is hard-killed (the one path teardown can't cover).
- Pass secrets via stdin rather than the ssh command line.

### DIFF
--- a/.github/workflows/supertable-bench-azure.yml
+++ b/.github/workflows/supertable-bench-azure.yml
@@ -3,7 +3,9 @@ name: Supertable Benchmark (Azure)
 # Supertable ingest benchmark on a per-run ephemeral Azure VM, in the
 # storage account's region. Compute is co-located with the bucket so
 # object-store transfers run over the Azure backbone, not the public
-# internet. The VM clones, builds, runs, and is destroyed each run.
+# internet. The VM clones and builds with an sccache compile cache backed
+# by Azure Blob (shared across branches), runs the bench, and is destroyed
+# each run.
 
 on:
   workflow_dispatch:
@@ -52,6 +54,7 @@ env:
   RG: rg-infino-bench-${{ github.run_id }}
   VM: bench-vm
   ADMIN: azureuser
+  SCCACHE_VERSION: v0.8.2
 
 jobs:
   benchmark:
@@ -66,7 +69,7 @@ jobs:
           tenant-id: ${{ secrets.AZURE_TENANT_ID }}
           subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
 
-      - name: Provision ephemeral VM (East US)
+      - name: Provision ephemeral VM
         id: provision
         run: |
           set -euo pipefail
@@ -117,7 +120,7 @@ jobs:
           done
           echo "::error::SSH never came up"; exit 1
 
-      - name: Clone repo + toolchain (on VM)
+      - name: Clone repo + toolchain + sccache (on VM)
         run: |
           set -eo pipefail
           VM_IP="${{ steps.provision.outputs.vm_ip }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/supertable-bench-azure.yml
```diff
@@ -0,0 +1,264 @@
...
+      docs:
+        description: "Doc count (plain integer, NOT 1M/100K — those do not parse)"
+        required: false
+        default: "1000000"
+        type: string
+      vm_size:
+        description: "Azure VM size."
+        required: false
+        default: "Standard_D8s_v5"
+        type: string
```
naman-modi, line 22: Hey Pratyush, please check the availability & quota limits for our az subscription.

pratyushlokhande: Checked.. `Standard_D8s_v7` is available.. updated as default.

naman-modi (approved):

PR #73, opened 2026-06-09
ci(bench): compare PR benchmarks against main + track history

pratyushlokhande:
## Goal

Make the informational PR benchmark show **how this PR moved performance vs `main`** instead of `(new)` on every cell — and keep a per-commit **history of main** so we can see the trend over time.

## What changes

Two workflow files; **no Rust change** — the report already computes per-cell deltas from a baseline JSON, it was just never fed main's numbers.

- **PR run** → downloads main's metrics from Azure Blob into the report's baseline path, so the run renders `+x% better / -y% worse` vs main.
- **Merge to main** → publishes the run as:
  - `bench/<bench>/baseline/<vm>-<docs>.json` — mutable "latest main" pointer (what PRs diff against),
  - `bench/<bench>/history/<vm>-<docs>-<sha>.json` — immutable per-commit record (sha-keyed filename, commit-time in blob metadata) = the trend.
- One `bench` job handles both, switching on `github.event_name` (PR diffs, push publishes).

## Behavior

| Event | Result |
| --- | --- |
| push to `main` | bench → publish baseline + history |
| PR → `main` (same repo) | restore main baseline → bench → diff, no publish |
| PR from a fork | skipped (no OIDC) |
| push to any other branch | nothing (workflow only triggers on `main`) |

Still informational and out of the required-checks set — a red bench never blocks merge.

## Reviewer notes

- **Verified against real Blob** locally (mimicking the workflow's `az` calls, namespaced + cleaned up): publish, list-with-metadata, **immutability** (same-sha re-upload is blocked, no `--overwrite`), and a byte-identical restore round-trip.
- **Apples-to-apples**: baseline keyed by `vm_size` + `docs`; a shape/doc-count mismatch falls back to `(new)`, never a false delta.
- **Safety**: publish is gated on `success()` — a failed/partial run never publishes. Restore is best-effort — a transient `scp` flake warns and continues rather than failing the PR bench.
- **Deltas are informational**: shared-tenant VM jitter means treat <~5–10% as noise.
- **Accepted trade-offs**: (1) benches now also run on every merge to `main` (the baseline producer — 2 VMs/merge, serialized per bench); (2) rapid back-to-back merges can skip a history point for the superseded middle commit (`cancel-in-progress: false` keeps one running + one pending).

## Testing

Local round-trip of the blob logic confirmed end-to-end (publish → list → immutability → restore → delta render). Full VM path validates on this PR's own bench run.

### DIFF
--- a/.github/workflows/supertable-bench-azure.yml
+++ b/.github/workflows/supertable-bench-azure.yml
@@ -45,6 +45,11 @@ on:
         required: false
         default: "30"
         type: string
+      update_baseline:
+        description: "Publish this run as the main baseline + history (main only)."
+        required: false
+        default: false
+        type: boolean
   # Callable from other workflows (e.g. the PR group in ci.yml). Same
   # inputs; `choice` isn't valid here so `bench` is a string.
   workflow_call:
@@ -59,6 +64,9 @@ on:
       # Discriminator appended to resource names so parallel callers
       # (e.g. matrix legs sharing one run_id) don't collide / cross-delete.
       name_suffix:     { type: string, required: false, default: "" }
+      # On main: publish the run as the baseline pointer + per-commit
+      # history. On a PR: false — restore main's baseline and diff against it.
+      update_baseline: { type: boolean, required: false, default: false }
 
 permissions:
   id-token: write   # Azure OIDC federated login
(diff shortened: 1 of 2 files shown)


.github/workflows/supertable-bench-azure.yml
```diff
@@ -282,6 +327,46 @@ jobs:
...
+              "${ADMIN}@${VM_IP}:infino-src/target/infino-bench/${j}.json" "${j}.json" \
+              || { echo "::warning::no metrics for $j — skipping"; continue; }
+            az storage blob upload --overwrite --account-name "$ACC" --account-key "$KEY" \
+              -c "$CONTAINER" -n "bench/${j}/baseline/${VM_SIZE}-${DOCS}.json" -f "${j}.json"
+            # No --overwrite: a commit's record is immutable once written.
+            az storage blob upload --account-name "$ACC" --account-key "$KEY" \
+              -c "$CONTAINER" -n "bench/${j}/history/${VM_SIZE}-${DOCS}-${SHA}.json" \
+              -f "${j}.json" --metadata "sha=${SHA}" "ts=${TS}" "ref=main" \
+              || echo "::notice::history for ${SHA} already exists"
+            echo "published: $j ($SHA)"
```
naman-modi, line 379: Please handle the `az cli` error scenarios.

naman-modi (approved):

.github/workflows/supertable-bench-azure.yml
pratyushlokhande: Handled — baseline + history uploads now warn on genuine az errors instead of masking them as 'already exists'; existence is checked explicitly. Kept it non-failing since the bench is informational.

PR #76, opened 2026-06-10
Refactor benches and extend public API

ekechinwokah:
### Search API

- `bm25_search` / `vector_search` now return Arrow rows (`Vec<RecordBatch>`: `_id`, scalar columns, trailing `score`) and take a column `projection` plus a `materialize: Option<bool>` knob. `None` picks the per-method default: **BM25 materializes by default; vector search returns just `_id` + `score`** (kNN is usually a retrieval step feeding a follow-up fetch). `materialize = false` skips scalar decode entirely.
- The unranked `token_match` / `exact_match` are public on `Supertable`(returning `SearchHit`: `_id` + score) **and** registered as catalog-level table functions, so they compose in SQL (`token_match('docs','title','rust') EXCEPT token_match(...)`).
- Async kernels and the segment-local hit representation stay internal on `SupertableReader`.

### Query engine (from test-support)

- SQL scans read through `SuperfileObjectStore`, a thin `object_store::ObjectStore` over each segment's existing byte source from the Superfile reader: warm/mmap segments resolve ranges as zero-copy slices, cold segments range-fetch — no more copying into a DataFusion `InMemory` store.
- Shared row materializer eagerly reopens lazy readers only when scalar columns are actually projected.
- Kept the per-provider object-store registry key on top of this, so multi-table catalog SQL doesn't collide across providers.

### Catalog hardening (review findings)

- `drop_table` + re-`create_table` of the same name now yields a fresh, empty table (unique per-creation storage subtree) instead of silently resurrecting the dropped table's data.
- `_`prefixed table names rejected (reserved namespace), explicit metric encoding, `SearchHit` is `#[non_exhaustive]`.
- Python bindings release the GIL around all blocking calls and mirror the new search API (`materialize=` kwarg, pyarrow `Table` results).

### Bench harness

- One `bench` target, positional selection:
`cargo bench -- [superfile|supertable] [fts|vector|sql] [build|warm|cold]`, with diagnostics as tokens (`scale`, `tombstone`, `update`, `sql-diag`, `object-store`). Per-tier scale knobs (`INFINO_BENCH_SUPERFILE_DOCS` / `INFINO_BENCH_SUPERTABLE_DOCS`).
- `benches/README.md` rewritten against the actual harness and populated with a full run; the Azure bench CI workflow is adapted to the unified binary.
- changes github CI workflow to use 1M docs instead of 10M for supertable tests.

### Docs

- README: "Search is an access path, not just an API" — indexed text predicates resolve to candidate row sets before any column decode, and search TVFs are relations feeding joins/aggregates.
- Architecture overview documents the `connect`/`Connection` entry point; [supertable.md](http://supertable.md/)'s public-API section matches the shipped surface.

## Verification

- Full suite: **1,148 tests passing** (lib + superfile + supertable + standalone + doctests), `cargo check --all-targets` clean in both feature configs, Python crate clean.
- **Vector recall@10 = 0.995** on the supertable bench (≥ 0.99 bar), 0.996 at the calibrated 0.99-target point.
- Bench suite run end-to-end against real S3 (us-east-1, 1M docs/tier); numbers recorded in `benches/README.md`. Highlight: aggregates over FTS candidates via token-index pushdown run ~12× faster than the full scan (23 ms → 1.7–2 ms warm). No prior baseline existed on this host, so deltas start from this run.

## Notes for reviewers

- `public-api.txt` was updated by hand alongside the signature changes; run `make public-api-update` (nightly + cargo-public-api) to regenerate and confirm before merge.
- The 10M-doc supertable run is deferred; this PR's recorded numbers are the 1M configuration.
- Known follow-up: the FTS-pushdown selectivity gate looks mis-tuned at supertable scale for non-selective predicates (67 ms vs 30 ms full scan on `SUM ... IN all`); tracked separately.


benches/bench/main.rs
```diff
@@ -0,0 +1,273 @@
...
+//! cargo bench -- sql-diag
+//! cargo bench -- object-store
+//! ```
+//!
+//! Token vocabulary:
+//!   tier        : `superfile` | `supertable`        (omitted => both)
+//!   modality    : `fts` | `vector` | `sql`          (omitted => all three)
+//!   phase       : `build` | `warm` | `cold` | `search` (= warm+cold)
+//!                 (omitted => all three phases)
+//!   `all`       : explicit "every tier × modality" (the default)
```
muralikpbhat: ... x every phase

benches/bench/main.rs
```diff
@@ -0,0 +1,273 @@
...
+//!   diagnostic  : `scale` | `tombstone` | `update` | `sql-diag` | `object-store`
```
muralikpbhat: this is still 'tool overloading'...for example 'all' can mean 'diagnostic' is also included, which is not the case. May be document that in 'all'?

another option is to split at high-level. say, include '-- diagnostic' or '-- standalone' as a tier? and replace 'all' with 'super' which indicates both superfile and supertable only. modality is used only when tier is super*. For standalone, we run it as 'cargo bench -- diagnostic [scale tombstone update sql-diag object-store]'

benches/bench/main.rs
```diff
@@ -0,0 +1,273 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Infino Authors
+
+//! The single Infino benchmark binary.
+//!
+//! Everything is selected by positional tokens after `cargo bench --`.
```
muralikpbhat, line 6: thx, I like this clean up

benches/utils/ingest/supertable.rs
```diff
@@ -14,19 +16,23 @@ use infino::supertable::storage::StorageProvider;
...
-use crate::corpus::{self, DIM, SequentialSyntheticCorpus};
+use crate::corpus::{self, DIM, MmapTextCorpus, SequentialSyntheticCorpus};
+use crate::harness::{emb_for, scatter_key, sql_options, sql_schema};
+use crate::markdown::fmt_count;
 use crate::tiers;
 
 /// Supertable-shape document count — the supplied parameter. Default 10M
 /// ([`crate::corpus::supertable_docs`]); override with
-/// `INFINO_BENCH_DOC_COUNT`.
+/// `INFINO_BENCH_SUPERTABLE_DOCS`.
```
muralikpbhat, line 26: Thanks for fixing this, it was confusing earlier

docs/architecture/overview.md
```diff
@@ -37,19 +37,43 @@ Think of three ideas:
...
+            │                                      │             │
+            └──────────── byte-range reads ────────┴─────────────┘
                                  ▼
                       Object storage (S3) — cheap, durable,
                               the source of truth
 ```
 
+## Opening Infino
+
+There is no server to run. You open a **connection** to a root location
```
muralikpbhat: open connection from where? how do we install the Infino library there? are we alluding to any standard sql clients? Are we planing a 'getting started for sql' doc?

benches/bench/main.rs
ekechinwokah: fixed - now reads "every tier × modality × phase"

docs/architecture/overview.md
ekechinwokah: fair point. added some clarity in the doc. no wire protocol so external SQL clients can't attach today. please discuss with vinay on getting started guide.

benches/bench/main.rs
ekechinwokah: makes sense. done.

muralikpbhat (approved):
Looks great, approving for the benches section. For the rest (especially public api), would be great for someone else to take a look as I am not authoritative yet :)

README.md
```diff
@@ -62,8 +62,10 @@ let batch = RecordBatch::try_new(
 docs.append(&batch)?;
 
 // Keyword search (BM25): hits carry the auto-injected `_id` + score.
-let hits = docs.bm25_search("title", "fox", 10, BoolMode::Or)?;
-assert_eq!(hits.len(), 1);
+// Returns Arrow rows. projection `None` = all columns; materialize `None`
+// = the method default (BM25 materializes; vector search does not).
+let batches = docs.bm25_search("title", "fox", 10, BoolMode::Or, None, None)?;
```
vinaykakade: Instead of "None", let us add an actual column. Or, if you want to keep the example to return all columns, let us make None as the default and have this call to be docs.bm25_search("title", "fox", 10, BoolMode::Or)

src/supertable/query/fts.rs
```diff
@@ -579,16 +657,58 @@ impl Supertable {
         query: &str,
         k: usize,
         mode: BoolMode,
+        projection: Option<&[&str]>,
+        materialize: Option<bool>,
```
vinaykakade: We shouldn't need both projection and materialize. What if we just have projection, and if it is empty, we won't materialize, but otherwise the user has to explicitly list the names of columns to be materialized? 

The default for projection would be [] and type would be &[&str] - so don't materialize any column by default. Or, it could be None, in which case projection should be Option<&[&str]>, and that would mean return all scalar columns. Either of this is fine.

infino-python/src/lib.rs
```diff
@@ -212,58 +215,89 @@ impl Table {
...
-            other => {
-                return Err(PyValueError::new_err(format!(
-                    "mode must be 'or' or 'and', got {other:?}"
-                )));
-            }
-        };
-        let hits = self
-            .inner
-            .bm25_search(column, query, k, mode)
+        materialize: Option<bool>,
```
vinaykakade: There is no projection param here (which exists in Rust api). Let us maintain consistency and add projection here.

infino-python/src/lib.rs
```diff
@@ -212,58 +215,89 @@ impl Table {
...
+    #[pyo3(signature = (column, query, k, nprobe=None, materialize=None))]
     fn vector_search<'py>(
         &self,
         py: Python<'py>,
         column: &str,
         query: Vec<f32>,
         k: usize,
         nprobe: Option<usize>,
-    ) -> PyResult<Bound<'py, PyList>> {
+        materialize: Option<bool>,
```
vinaykakade: Same comment as above about missing projection param.

src/catalog/mod.rs
```diff
@@ -489,26 +504,54 @@ fn build_disk_cache(
...
     }
 }
 
+/// A unique-per-creation physical subtree for a table. The catalog name
+/// is the stable identity; this is only the storage location, made
+/// unique so a `drop_table` (logical — it leaves the bytes in place)
+/// followed by a re-create of the same name lands on a fresh subtree
+/// rather than re-opening the dropped table's committed data. Stays a
+/// single path segment (same depth as the old `<root>/<name>`).
+fn unique_location(name: &str) -> String {
```
vinaykakade, line 561: Good idea. How would we eventually reclaim the storage of old tables that are no longer needed? Could be just behind the to-be-added compact() method that does this cleanup too - but would be good think through how does the space get reclaimed.

src/catalog/mod.rs
```diff
@@ -489,26 +504,54 @@ fn build_disk_cache(
...
+    use std::sync::atomic::{AtomicU64, Ordering};
```
vinaykakade: Let us move inline imports to the top (perhaps need AGENTS.md update too).

src/supertable/query/df_object_store.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+#[async_trait]
+impl ObjectStore for SuperfileObjectStore {
+    async fn get_opts(&self, location: &ObjPath, options: GetOptions) -> OsResult<GetResult> {
+        let source = self.source(location)?;
+        let size = source.size();
+        let meta = ObjectMeta {
+            location: location.clone(),
+            last_modified: chrono::Utc::now(),
```
vinaykakade: get_opts call should not be adding a new value for last_modified. Where is this used?

src/supertable/query/df_object_store.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+/// Resolve a [`GetRange`] (or its absence) into a concrete, clamped
+/// `[start, end)` over an object of `size` bytes.
+fn resolve_range(range: Option<GetRange>, size: u64) -> Range<u64> {
+    match range {
+        None => 0..size,
+        Some(GetRange::Bounded(r)) => r.start.min(size)..r.end.min(size),
+        Some(GetRange::Offset(start)) => start.min(size)..size,
+        Some(GetRange::Suffix(n)) => size.saturating_sub(n)..size,
```
vinaykakade, line 103: There is only a test for GetRange::Bounded. Let us add for GetRange::Offset and GetRange::Suffix as well (unless they are already in I missed it).

src/supertable/query/fts.rs
```diff
@@ -579,16 +657,58 @@ impl Supertable {
...
+    /// Unranked token match over one FTS column: every row whose
+    /// `column` matches `query`'s tokens under `mode` (`Or` = any token,
+    /// `And` = every token). Returns [`SearchHit`]s with `score == 0.0`
+    /// and unspecified order — a candidate set, not a ranking.
+    pub fn token_match(
+        &self,
+        column: &str,
+        query: &str,
+        mode: BoolMode,
     ) -> Result<Vec<SearchHit>, crate::InfinoError> {
```
vinaykakade: To be consistent, this should also take projection argument, and return Vec<RecordBatch> - similar to bm25_search?

src/supertable/query/fts.rs
```diff
@@ -579,16 +657,58 @@ impl Supertable {
...
+    }
+
+    /// Unranked exact match: rows whose `column` value equals `value`
+    /// exactly (index-pruned, then text-verified). Returns
+    /// [`SearchHit`]s with `score == 0.0` and unspecified order.
+    pub fn exact_match(
+        &self,
+        column: &str,
+        value: &str,
+    ) -> Result<Vec<SearchHit>, crate::InfinoError> {
```
vinaykakade: Same as above - to be consistent, this should also take projection argument, and return Vec - similar to bm25_search?

src/supertable/query/vector.rs
```diff
@@ -326,16 +400,22 @@ impl Supertable {
         query: &[f32],
         k: usize,
         options: VectorSearchOptions,
-    ) -> Result<Vec<SearchHit>, crate::InfinoError> {
+        projection: Option<&[&str]>,
+        materialize: Option<bool>,
```
vinaykakade: Same comment as FTS: I think we can get rid of materialize flag.

src/supertable/query/exec/common.rs
```diff
@@ -230,30 +229,60 @@ async fn resolve_columns(
...
+    // a reader holding resident bytes (in-memory tier / freshly written)
+    // decodes locally via `take_by_local_doc_ids`; a lazy object-store
+    // reader streams ONLY the projected hit rows through parquet's async
+    // `ParquetObjectReader` (footer + projected column pages via range
+    // GETs), so a cold read never materializes the whole segment.
+    //
+    // Segment count here is bounded by the global top-k (one entry per
+    // distinct hit-bearing segment), so the per-segment fan-out is small
+    // and the concurrent range-GET reads overlap on the query runtime.
+    let per_segment: Vec<RecordBatch> = futures::future::try_join_all(
```
vinaykakade: rayon was replaced by try_join_all here. If all segments are in memory, is the rayon approach better / are there any perf advantages of not using rayon here?

src/supertable/query/exec/common.rs
```diff
@@ -278,6 +307,117 @@ async fn resolve_columns(
...
+    file_size: Option<u64>,
+    file_schema: &SchemaRef,
+    n_docs: u64,
+    local_doc_ids: &[u32],
+    names: &[&str],
+) -> DfResult<RecordBatch> {
+    use futures::TryStreamExt;
+    use parquet::arrow::ProjectionMask;
+    use parquet::arrow::arrow_reader::{RowSelection, RowSelector};
+    use parquet::arrow::async_reader::{ParquetObjectReader, ParquetRecordBatchStreamBuilder};
```
vinaykakade: You can move the imports to the top of the file

src/supertable/query/exec/common.rs
```diff
@@ -278,6 +307,117 @@ async fn resolve_columns(
...
+/// [`SuperfileReader::take_by_local_doc_ids`]: crate::superfile::SuperfileReader::take_by_local_doc_ids
+async fn take_rows_object_store(
+    store: Arc<dyn object_store::ObjectStore>,
+    path: object_store::path::Path,
+    file_size: Option<u64>,
+    file_schema: &SchemaRef,
+    n_docs: u64,
+    local_doc_ids: &[u32],
+    names: &[&str],
+) -> DfResult<RecordBatch> {
```
vinaykakade, line 376: nit: any way to reduce the code duplication across this and the SuperFileReader::take_by_local_doc_ids?

vinaykakade (approved):

vinaykakade:
LGTM with comments added. Most important one is the comment around not needing explicit materialize in the public api (projection should be sufficient), and have python public api to also get projection parameter. Fine to make some of the follow-up items as GH issues too.

benches/utils/supertable.rs
```diff
@@ -0,0 +1,923 @@
...
+                .vector_search(
+                    supertable::VEC_COLUMN,
+                    &q_cal[0],
+                    TOP_K,
+                    exec_vec::search_opts(DEFAULT_NPROBE, DEFAULT_RERANK_MULT),
+                    None,
+                )
+                .expect("warm prewarm vector_search");
+            consumer
+                .wait_until_warm(Duration::from_secs(600))
```
muralikpbhat: Found an issue with this while testing cold with s3 + 10M. This wait_until_warm should be called only when phases.warm is true, but it is currently happening for cold also. Because of that, it is failing to load 256 large segments into cache and I had to test with INFINO_DISABLE_BG_FILL=1 to bypass that. However, the 10M test in running forever with that and I need to debug it further. 

Anyways, wanted to flag this missing phases.warm gate issue (it is handled properly for fts and sql though) so that you can either fix it before merging this, or I can take it up as a follow up.

ekechinwokah: agreed. fixed.

src/supertable/query/exec/common.rs
ekechinwokah: see comment on slack. in general, based on testing we are using rayon for cpu waves and tokio for I/O, bridged via oneshot so tokio workers don't blocked. will fix regression here.

src/supertable/query/vector.rs
ekechinwokah: done.

src/supertable/query/fts.rs
ekechinwokah: done.

src/supertable/query/fts.rs
ekechinwokah: done

src/supertable/query/df_object_store.rs
ekechinwokah: done.

src/catalog/mod.rs
ekechinwokah: done.

infino-python/src/lib.rs
ekechinwokah: fixed

infino-python/src/lib.rs
ekechinwokah: fixed

src/supertable/query/fts.rs
ekechinwokah: agreed. changed.

README.md
ekechinwokah: done.

src/supertable/query/exec/common.rs
ekechinwokah: done

src/supertable/query/df_object_store.rs
ekechinwokah: nothing consumes the timestamp today — `ObjectMeta` just needs a value for the parquet reader plumbing. but you are right in principle - pinned this to a fixed epoch constant since segments are immutable.

src/supertable/query/exec/common.rs
ekechinwokah: not sure — they share the row-selection logic (sorted/dedup → RowSelection runs → rank-back), but the I/O models are fundamentally different: `take_by_local_doc_ids` is a sync CPU decode over resident bytes (warm tier, rayon pool), while `take_rows_object_store` streams only the footer + projected column pages. unifying them would force one abstraction over sync-resident vs async-streaming reads. 

shared selection/rank-back piece is a small helper worth extracting though. done.

src/catalog/mod.rs
ekechinwokah: makes sense directionally. added `drop_table(name, purge)` — unregisters first, then deletes the subtree, so a crash only leaves orphans. compact() will reclaim its own superseded segments; we could implement a future `Connection::vacuum()` or something to sweep the rest?

PR #77, opened 2026-06-10
feat superfile writer: ignore deleted documents when creating batch from superfile

Pranav2612000:
### What does this PR do?
- Modifies the `add_batch_from_reader` to accept a list of deleted doc ids, to be ignored and not added in the new batch

- This allows us to remove the deleted documents from the superfiles during compaction

### How do we do this change?
- For parquet, we rely on the `with_row_groups` and `with_row_selection` methods to speed up retrieval by ignoring deleted rows
- For vector, we use bitmap's membership method to ignore rows which have been deleted


savannahar68 (approved):
lgtm

PR #82, opened 2026-06-10
feat fts: `-term` negation for bm25_search

naman-modi:
### What does this PR do?
Adds negation to `bm25_search`: `"rust -python"` scores docs containing `rust` and drops any that also contain `python`.

Exclusion queries are a standard FTS capability we lacked, and there was no way to say "match X but not Y".

### How do we do this change?
- `fts/parser.rs` splits the leading `-` sigil before tokenizing (the tokenizer eats `-`).
- Negated terms become `TermCursor`s; `ExcludeFilter` advances them with `skip_to` as the positive kernels walk candidates, gating heap admission — a negated posting list is never fully decoded.
- Kernels take `Option<&mut ExcludeFilter>`; `None` (no negation) costs nothing.
- The supertable orchestrator parses once per query; only positives feed the bloom prune; negation queries run one un-ranged unit per segment.
- Negation-only queries error (`FtsError::NegationOnly`).

### Performance
- Benchmarked: new negation query family in `benches/superfile_fts` (with a correctness gate before timing); numbers recorded in `benches/README.md`: 1M docs, k=10 hot: rare-neg 26.8 µs, common-neg 1.47 ms, two-term-OR-neg 4.58 ms, two-term-AND-neg 4.97 ms.
- No regression on regular (no `-term`) searches.

### Notes
- A generic `Admitter` gate measured 25–30% slower on the OR kernel (the monomorphized copy compiles worse than the plain fn), hence `Option<&mut ExcludeFilter>`.
- Negated terms must never prune segments: under `And` the bloom check would drop every segment *lacking* the negated term.
- Out of scope (deferred): prefix search, `search_multi`, phrases (needs positions; format change), ranged fan-out with negation, typed supertable error variant.


src/superfile/fts/reader.rs
```diff
@@ -633,39 +633,83 @@ impl FtsReader {
...
 
-        // Single-term fast path: BlockMaxWAND-driven block skipping.
-        // Walks blocks in order, populating a top-k min-heap. Once the
-        // heap is full, blocks whose skip-table-recorded `max_bm25`
-        // can't beat the heap's worst score are skipped without
-        // decoding.
+    /// Shared dispatch for [`Self::search`] and [`Self::search_excluding`]:
+    /// routes positives to the single-term / OR / AND kernel and threads
+    /// `filter` through to the heap-admission sites.
+    async fn search_positive(
```
savannahar68: can we have a better name over here? beacuse felt that we will only use this to search positive terms
but we are also passing negative filters over here

maybe 

search_with_filters

benches/utils/fts_superfile.rs
```diff
@@ -359,18 +401,41 @@ fn to_infino_mode(mode: BoolMode) -> InfinoBoolMode {
     }
 }
 
+/// One warmup call, then `HOT_ITERS` timed calls of `run`; returns the
+/// p50. Shared scaffold for every manual hot-timing path.
+fn hot_p50<T>(mut run: impl FnMut() -> T) -> Duration {
+    std::hint::black_box(run());
```
vinaykakade: nit: move the inline imports to the top of the file

benches/README.md
```diff
@@ -126,7 +126,52 @@ _Run `INFINO_BENCH_UPDATE_README=1 cargo bench --bench superfile_fts` to populat
 <!-- END: bench/fts/superfile/ingest -->
 
 <!-- BEGIN: bench/fts/superfile/search -->
-_Run `INFINO_BENCH_UPDATE_README=1 cargo bench --bench superfile_fts` to populate._
+### Superfile FTS — search, single-segment / in-memory (1M docs)
```
vinaykakade, line 169: We need a better way to modify this, but let us not remove the comment on line 129. The exact command there is changing with #76

src/superfile/fts/reader.rs
naman-modi: Makes sense, done.

benches/utils/fts_superfile.rs
naman-modi: Updated, Vinay.

benches/README.md
naman-modi: I have solved the merge conflicts, and re-run the benches with the command.

src/superfile/fts/parser.rs
```diff
@@ -0,0 +1,127 @@
...
+//!
+//! This must run before tokenizing: the tokenizer treats `-` as
+//! punctuation, so `"rust -python"` would tokenize to `["rust",
+//! "python"]` and lose the marker. So we split on whitespace, decide
+//! each run's polarity, then tokenize each side.
+//!
+//! Polarity only — no validation. A query with no positives is returned
+//! as-is; the caller decides that's an error.
+
+use crate::superfile::fts::tokenize::Tokenizer;
```
vinaykakade: We shouldn't have a separate parser for negations. The primary issue is performance - see the to_owned() strings in the below code.

Instead, modify the AsciiLowercaseTokenizer - which tokenizes the input without making string copies. Something like the below there may work:


  ```rust
  // impl AsciiLowerTokenizer  (reuses simd_skip_non_token / simd_scan_token_run)
  pub fn tokenize_query_each<'a, F>(&self, query: &'a str, mut f: F)
  where
      F: FnMut(bool, std::borrow::Cow<'a, str>),
  {
      use std::borrow::Cow;
      let mut buf: Vec<u8> = Vec::new(); // scratch, only touched on the uppercase path
      for run in query.split_whitespace() {
          let (negative, rest): (bool, &'a str) = match run.strip_prefix('-') {
              Some(r) if !r.is_empty() => (true, r),
              _ => (false, run),
          };
          let bytes = rest.as_bytes();
          let mut pos = 0;
          while pos < bytes.len() {
              pos = simd_skip_non_token(bytes, pos);
              if pos >= bytes.len() { break; }
              let start = pos;
              let (end, had_upper, had_non_ascii) = simd_scan_token_run(bytes, pos);
              pos = end;
              if had_non_ascii || start == pos { continue; }
              if !had_upper {
                  f(negative, Cow::Borrowed(&rest[start..end])); // zero-copy, &'a str
              } else {
                  buf.clear();
                  buf.reserve(end - start);
                  for &b in &bytes[start..end] { buf.push(b.to_ascii_lowercase()); }
                  // SAFETY: ASCII alnum (lowercased) → valid UTF-8
                  let s = unsafe { std::str::from_utf8_unchecked(&buf) };
                  f(negative, Cow::Owned(s.to_owned()));
              }
          }
      }
  }
  ```

src/superfile/fts/mod.rs
```diff
@@ -8,6 +8,7 @@ pub mod bm25;
 pub mod builder;
 pub mod dict;
 pub(crate) mod fst_value;
+pub(crate) mod parser;
```
vinaykakade: Let us change so that a separate parser is not needed. See the comment in parser.rs regarding how this can be folded in AsciiLowercaseTokenizer.

src/superfile/fts/reader.rs
```diff
@@ -637,39 +637,84 @@ impl FtsReader {
...
+        // of this.
+        if negatives.is_empty() {
+            self.search_with_filters(column_id, positives, k, mode, None)
+                .await
+        } else {
+            let mut filter =
+                ExcludeFilter::new(self.build_term_cursors(column_id, negatives).await?);
+            self.search_with_filters(column_id, positives, k, mode, Some(&mut filter))
+                .await
+        }
```
vinaykakade: You can refactor ~10 lines above to:

```
  let mut filter = match negatives {
      [] => None,
      _ => Some(ExcludeFilter::new(
          self.build_term_cursors(column_id, negatives).await?,
      )),
  };
  self.search_with_filters(column_id, positives, k, mode, filter.as_mut())
      .await
```

src/superfile/fts/reader.rs
```diff
@@ -637,39 +637,84 @@ impl FtsReader {
...
+
+    /// BM25 search with negated (`-term`) terms excluded.
+    ///
+    /// `positives` are scored under `mode` as in [`Self::search`];
+    /// `negatives` filter out any doc containing one of them, regardless
+    /// of score. Both lists are already tokenized.
+    ///
+    /// No positives → [`FtsError::NegationOnly`] (nothing to rank).
+    /// Empty positives *and* negatives → empty result.
+    pub async fn search_excluding(
```
vinaykakade: This can be pub(crate) - as it does not have callers outside the crate

src/supertable/query/fts.rs
```diff
@@ -113,26 +113,28 @@ impl SupertableReader {
...
-        // eliminating the (N+1)·T redundant tokenizations a
-        // per-segment bm25_search would incur for N superfiles +
-        // a T-token query.
-        let term_strings: Vec<String> = AsciiLowerTokenizer.tokenize(query).collect();
-        let term_refs: Vec<&str> = term_strings.iter().map(|s| s.as_str()).collect();
+        // Parse + tokenize ONCE at the orchestrator (the `-` sigil
+        // splits negated terms out). The split lists are reused for the
+        // bloom-skip masks and for every per-segment search, instead of
+        // re-parsing per segment.
+        let parsed = crate::superfile::fts::parser::parse(query, &AsciiLowerTokenizer);
```
vinaykakade: This will change as you move the negation parsing inside AsciiLowerTokenizer.

vinaykakade (approved):
LGTM after making the suggested changes

src/superfile/fts/parser.rs
naman-modi: As discussed offline, we should keep the parser, and tokeniser clubbed for now. Can revisit it later.

src/superfile/fts/mod.rs
naman-modi: Done.

src/supertable/query/fts.rs
naman-modi: Done.

PR #84, opened 2026-06-10
feat: compaction policy v1

savannahar68:
What

  Adds supertable::compaction::select the policy function that decides which superfiles to compact.

  Pure function, no I/O. Takes a slice of SegmentStats and CompactionSettings, returns Vec<CompactionJob>. Each job is one merge: a set of input superfile IDs that should be merged into one output.

 Why

  Compaction needs a selection step that is testable in isolation, with no storage I/O or side effects. Keeping it pure
  means it can be called, tested, and reasoned about without a running supertable. The orchestration layer
  (supertable::compact) will gather SegmentStats from the manifest and call this.

### DIFF
--- a/src/supertable/compaction/mod.rs
+++ b/src/supertable/compaction/mod.rs
@@ -246,7 +246,10 @@ mod tests {
         }
         let jobs = select(&segs, &default_cfg());
         assert_eq!(jobs.len(), 2);
-        let a = jobs.iter().find(|j| j.partition_key == vec![0xA]).unwrap();
+        let a = jobs
+            .iter()
+            .find(|j| j.partition_key == vec![0xA])
+            .expect("partition A job");
         assert!(a.inputs.iter().all(|id| id.as_u128() < 5));
     }
 }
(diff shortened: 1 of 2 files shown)


savannahar68:
Rough sketch of what supertable::compact() looks like calling this:

```

  pub async fn compact(&self) -> Result<(), CompactionError> {
      let reader = self.reader();
      let manifest = reader.manifest();

      // Gather stats from manifest + tombstone sidecars
      let mut stats = Vec::new();
      for entry in &manifest.superfile_list.superfiles {
          let tombstone_state = self.tombstones.sidecar_state(entry.superfile_id).await?;
          stats.push(SegmentStats {
              superfile_id: entry.superfile_id,
              partition_key: entry.partition_key.clone(),
              size_bytes: entry.size_bytes,
              n_docs: entry.n_docs,
              tombstoned_docs: tombstone_state.tombstoned_count,
              sealed_by_other: tombstone_state.sealed_by != Some(my_compaction_id),
          });
      }

      let jobs = select(&stats, &self.config.compaction);

      for job in jobs {
          self.run_compaction_job(job).await?;
      }

      Ok(())
  }
```

Pranav2612000 (approved):

PR #85, opened 2026-06-10
improv superfile builder: return superfile stats when building from readers

Pranav2612000:
### What does this PR do?
- Returns superfile stats ( n_docs, min_id, max_id, scalar_stats ( min & max of each col ) ) when merging multiple superfiles

- These are required to generate a SuperfileEntry for the merged superfile.

### How do we do this change?
- Every append from a reader returns the statistics for the append ( in the future, this can be done for all appends - instead of doing this calculation at supertable level )
- These superfile stats for all appends are then merged together to get a cummulative SuperfileStats for the merged file, which is passed to the caller

### DIFF
--- a/src/superfile/builder.rs
+++ b/src/superfile/builder.rs
@@ -83,6 +83,7 @@ use crate::superfile::format::footer::{encode_parquet_body, splice_index_blobs};
 use crate::superfile::format::{self, kv};
 use crate::superfile::fts::builder::FtsBuilder;
 use crate::superfile::fts::tokenize::{AsciiLowerTokenizer, Tokenizer};
+use crate::superfile::stats::SuperfileStats;
 use crate::superfile::vector::builder::VectorBuilder;
 use crate::superfile::vector::reader::ColumnReader;
 use crate::superfile::{BuildError, SuperfileReader};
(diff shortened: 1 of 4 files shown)


src/superfile/stats.rs
```diff
@@ -0,0 +1,305 @@
...
+pub struct SuperfileStats {
+    pub n_docs: u64,
+    pub id_min: i128,
+    pub id_max: i128,
+    pub scalar_stats: ScalarStatsTable,
+    // TODO: Vector & FTS related stats could also be added here
+}
+
+impl SuperfileStats {
+    pub fn try_compute_from_record_batch(batch: &RecordBatch) -> Result<Self, BuildError> {
```
savannahar68, line 18: lets also pass FTS summary and vector summary so that this can be a single stats data structure at superfile level and we can get rid of other things

in separate PR

src/supertable/manifest/mod.rs
```diff
@@ -520,6 +520,177 @@ impl ScalarStatsTable {
         }
         Self { cols }
     }
+
```
savannahar68, line 523: TODO in future

maybe we can remove logic from from_batches (plural) and use from_batch and merge
just trying to avoid duplicate logic from maintanance pov

can be taken up later

src/superfile/builder.rs
savannahar68, line 638: lets remove fp32 restriction for compaction in separate PR

savannahar68 (approved):
lgtm
we will create couple of more PRs for enhancements

PR #86, opened 2026-06-11
Fix a few issues discovered during testing

ekechinwokah:
## Summary

This PR fixes some kernel-level issues the new bench instrumentation surfaced, and makes the bench harness measure superfiles and supertables consistently.

## Engine

- **Bare projection (`None`) returns `_id` + `score` only.**
  Previously `None` decoded every scalar column of every hit row. The query phase now follows the inverted-index contract: ids + scores, no row data. Materializing columns is an explicit opt-in by name. 

- **Hit → `_id` translation is manifest arithmetic, not a Parquet read.**
  Ids are minted in contiguous spans and segment rows are stored in id order, so when a segment's stats satisfy `id_max - id_min + 1 == n_docs`, `_id = id_min + local_doc_id`. Non-contiguous segments fall back to the id-page read.

  Bare-call latency at 256 segments / 1M docs: **456 µs → 32 µs**.

- **FTS: block-max fixed-point now rounds UP.**
  The builder truncated `max_bm25_x1000`, making stored block maxima a hair *below* the true block max, so floor-based pruning could skip blocks containing real top-k hits. This produced nondeterministic results under threshold sharing.

- **FTS: global top-k floor shared across the segment fan-out.**
  Every segment prunes against the table-wide threshold instead of only its local one.

- **Vector: folded SQ8 centroid scoring.**
  Cluster ranking runs in the quantized domain through the existing AVX-512 / AVX2 `sq8_dot` kernels with precomputed per-cluster L2 moments. The per-query dequantization pass is gone.

  `u8_sum_sumsq` accumulates in chunked `u32` so it autovectorizes. The L2 path is 2× faster than dequantize.

## Benches

- **Query/fetch phase split.**
  The superfile battery measured the raw hit kernel while the supertable battery measured the public search with full-row materialization — different operations, with 10 × ~1.5 KB of document decode billed to one tier only.

   Both tiers now report:

  - `warm` — id + score
  - `warm +fetch` — id + score + searched column

- **Per-cell process isolation.**
  RSS is per-process. A cell running after others inherited their residue: supertable FTS read ~9 GiB after the three superfile cells vs ~1.1 GiB isolated.

  Multi-cell runs now re-exec the bench binary per cell.

- **Cold split fixes.**
  Cold open and cold search are timed separately. The vector cold guard now opens its reader inside the timed open leg; it was constructing an empty struct there and hiding the S3 RTT in the search column.

- Corpus generation is mmap-streamed outside the RSS windows.
- The allocator is purged before each window.
- `smaps_rollup` anonymous/file-backed breakdowns log at phase boundaries.

## Tests

- `id_resolve` — arithmetic `_id`s == id-page-read `_id`s, rank order, multi-segment table.
- `fanout_floor` — warm fan-out decomposition probe: in-memory vs mmap-promoted readers, per-thread fault counters, superfile control column.
- Determinism canary extended for threshold sharing.

## Results

1M docs, real S3, isolated cells:

- Supertable warm FTS: **55 µs – 8 ms** by shape, down from **2.4 – 43 ms**.
- Fetch phase priced separately at ~**10–17 ms**.
- Supertable ingest RSS: **1.1–2.6 GiB** per cell, down from ~**9 GiB** of cross-cell contamination.
- Vector recall@10 = **0.995** (`nprobe=64`, `rerank=256`) at 1M.
- Correctness oracles green:
  - BMW == brute-force
  - df=1 exactness
  - SQL counts
- ASAN green on the FTS surface.


src/superfile/fts/builder.rs
```diff
@@ -2612,7 +2612,14 @@ fn encode_and_emit_term<W: Write>(
                 min_dl_per_block[i],
                 avgdl,
             );
+            // ceil(): the stored fixed-point value must stay a true
+            // UPPER bound after quantization — truncation would round
+            // it below the real block max and let BMW / floor skips
+            // drop blocks that still hold qualifying docs. (The reader
+            // additionally adds one step on decode to cover files
+            // written before this rounding fix.)
```
vinaykakade, line 2620: nit: this statement in the comment can be deleted (the reader doesn't work that way - see fts/reader.rs ~line 2670), and we don't need to cover superfiles before this fix):
"(The reader additionally adds one step on decode to cover files written before this rounding fix.)"

vinaykakade (approved):

src/superfile/fts/builder.rs
ekechinwokah: sounds good. will handle in next PR i am about to submit.

Issue #49, opened 2026-06-07
optimize cold vector queries


ekechinwokah:
on average, our vector latencies are about 1s on [cold queries](https://github.com/infino-ai/infino/tree/main/benches). can we drive this down by an order of magnitude?

Recall target | (p/seg, r) | hot | cold | Peak RSS | Median RSS | P90 RSS | Peak RSS Δ
-- | -- | -- | -- | -- | -- | -- | --
0.90 | (p=4, r=4) | 10.02 ms | 1.03 s | 10.35 GiB | 10.33 GiB | 10.35 GiB | —
0.95 | (p=8, r=4) | 10.73 ms | 1.07 s | 10.35 GiB | 10.33 GiB | 10.35 GiB | —
0.99 | (p=32, r=4) | 23.03 ms | 1.42 s | 10.35 GiB | 10.33 GiB | 10.35 GiB | —

muralikpbhat:
First trying to get a baseline in an Ec2(c7i.4xl) with latest code. Looks like given the test harness changes, there is probably some blocker for 10M. First ran it for 1M with superfile_vector.

INFINO_BENCH_STORE=s3 INFINO_REAL_S3_BUCKET="$BUCKET" cargo bench --bench superfile_vector
    Finished `bench` profile [optimized + debuginfo] target(s) in 0.18s
     Running benches/vector/main.rs (target/release/deps/superfile_vector-cfe323e3194a3074)
[superfile_vec] generating 1M×384 vector corpus...
[superfile_vec] correctness: using measured 1-writer artifact...
[superfile_vec] correctness OK: recall@10 = 1.000 (≥ 0.80)
[superfile_vec] committing measured 1-writer artifact to object storage...
[tiers] s3 prefix=infino-superfile-bench/43532-1781017794128726875
[tiers] superfile committed: s3 path=data/seg-26717ed4-cca1-459d-8f2d-794af0f8e3c0.sf.parquet (792 MiB)

══ Superfile vector — ingest, single-segment / in-memory (1M docs × dim=384) ══
Host: Intel(R) Xeon(R) Platinum 8488C · 8C/16T · 31 GiB RAM · linux/x86_64
| Build      | Time        | Throughput     | Bandwidth       | Peak RSS      | Median RSS    | P90 RSS       |
| ---------- | ----------- | -------------- | --------------- | ------------- | ------------- | ------------- |
[...]

PR #88, opened 2026-06-11
bench: faithful cold vector measurement at supertable scale

muralikpbhat:
Part of #49 (optimize cold vector queries) — lands the bench tooling to measure cold vector latency at supertable scale (256 segments), and corrects the cold-open measurement to reflect what production actually does.

## What

- `INFINO_BENCH_SKIP_CALIBRATION=1` — measure only the fixed `(nprobe, rerank)` `default` row; skip the correctness gate, the recall-target calibration sweep, and brute-force ground-truth generation.
- `INFINO_BENCH_VECTOR_NPROBE` / `INFINO_BENCH_VECTOR_RERANK` — override the `default` row's probe count / rerank multiplier.
- `INFINO_BENCH_PREFETCH_CONCURRENCY` — disk-cache prefetch fan-out for the cold-fill / promotion path.
- **Parallel cold open** — `open_all_segments` now `tokio::spawn`s each per-segment open instead of `try_join_all`, so the 256 per-segment parses distribute across the runtime workers, matching the production fan-out.

## Why — measuring cold at scale

At supertable scale (256 segments, 10M docs) the vector bench could not produce a cold number — two paths block it:

**1. The calibration sweep stalls.** A cold-only run enters the per-target probe/refine sweep (probes up to 800, refines up to 1024, 100 queries each) against the cold 256-segment table and effectively hangs on the first target:

```
[supertable_vector] calibrating recall@0.90: grid over probes/refines (100 queries)...
```

**2. Warm promotion times out.** The warm phase promotes every segment into the mmap cache, which does not converge within the 600s budget — even at `INFINO_BENCH_PREFETCH_CONCURRENCY=64`:

```
thread 'main' panicked at benches/utils/supertable.rs:
supertable warm promotion: SuperfileOpen("segment SuperfileUri(c629b8c0-...) not mmap-promoted within 600s")
```

`INFINO_BENCH_SKIP_CALIBRATION=1` with a cold-only run skips both and lands a fixed-config number directly:

```
[supertable_vector] skip-calibration: measuring only fixed (p=8, r=4)
```

## Why — faithful cold open

The cold measurement times open separately from search, but `open_all_segments` used `try_join_all`: it overlaps the 256 segment GETs yet runs all 256 `SuperfileReader` parses on the single poller thread, inflating reported cold-open by ~1s of serialized CPU. Production's fan-out (`query/dispatch.rs`) `tokio::spawn`s per segment, distributing the parses. Switching the bench to match:

| cold open | before (`try_join_all`) | after (`tokio::spawn`) |
|---|---|---|
| 1M  | 1.82 s | 0.82 s |
| 10M | 3.01 s | 1.98 s |

~1s of serialized parse recovered at both scales — the bench was overstating cold-open; production (which already spawns) never had it.

## Scope / safety

- **Bench-harness only** (`benches/utils/`). No changes to `src/`, the on-disk format, the manifest/commit path, or the public API.
- **Recall unaffected.** Skip-calibration only elides the measurement grid and its ground truth; the parallel-open change only affects how the bench opens segments, not the search/scoring math.

### DIFF
--- a/benches/utils/executors.rs
+++ b/benches/utils/executors.rs
@@ -830,62 +830,74 @@ pub mod vector {
         include_warm: bool,
         include_cold: bool,
         cold_iters: usize,
+        skip_calibration: bool,
         log_prefix: &str,
         anchor: &str,
         title: String,
         note: &str,
     ) {
-        eprintln!(
-            "[{log_prefix}] correctness: recall@{k} on {} queries (nprobe={CORRECTNESS_NPROBE}, rerank={CORRECTNESS_RERANK_MULT})...",
-            q_correct.len(),
-        );
-        let recall = mean_recall(
-            warm_reader,
-            column,
-            q_correct,
-            gt_correct,
-            k,
-            CORRECTNESS_NPROBE,
-            CORRECTNESS_RERANK_MULT,
-        );
-        assert!(
-            recall >= CORRECTNESS_RECALL_FLOOR,
-            "{log_prefix} vector recall@{k} {recall:.3} < floor {CORRECTNESS_RECALL_FLOOR:.2}"
-        );
-        eprintln!("[{log_prefix}] correctness OK: recall@{k} = {recall:.3}");
-
-        let cal: Vec<Option<Calibrated>> = RECALL_TARGETS
-            .iter()
-            .map(|&target| {
-                eprintln!(
-                    "[{log_prefix}] calibrating recall@{target:.2}: grid over probes/refines ({} queries)...",
-                    q_cal.len(),
-                );
-                calibrate(warm_reader, column, q_cal, gt_cal, target, k, log_prefix)
-            })
-            .collect();
-
         let q0 = &q_cal[0];
         let mut rows: Vec<RecallRow> = Vec::new();
-        for (i, &target) in RECALL_TARGETS.iter().enumerate() {
-            match cal[i] {
-                Some(c) => rows.push(RecallRow {
-                    target: format!("{target:.2}"),
-                    params: format!("p={}, r={}", c.probe, c.refine),
-                    recall: format!("{:.3}", c.recall),
-                    warm: include_warm
-                        .then(|| measure_warm(warm_reader, column, q0, k, c.probe, c.refine)),
-                    cold: include_cold.then(|| {
-                        measure_cold(&open_cold, column, q0, k, c.probe, c.refine, cold_iters)
+        if skip_calibration {
+            // Skip-calibration mode (INFINO_BENCH_SKIP_CALIBRATION): no
+            // correctness gate, no recall-target grid — just the fixed
+            // `(default_nprobe, default_rerank)` row below. Needs no ground
+            // truth and no warmed reader, so a cold-only run is fast and
+            // prod-shaped.
+            eprintln!(
+                "[{log_prefix}] skip-calibration: measuring only fixed (p={default_nprobe}, r={default_rerank})",
+            );
+        } else {
+            eprintln!(
+                "[{log_prefix}] correctness: recall@{k} on {} queries (nprobe={CORRECTNESS_NPROBE}, rerank={CORRECTNESS_RERANK_MULT})...",
+                q_correct.len(),
+            );
+            let recall = mean_recall(
+                warm_reader,
+                column,
+                q_correct,
+                gt_correct,
+                k,
+                CORRECTNESS_NPROBE,
+                CORRECTNESS_RERANK_MULT,
+            );
+            assert!(
+                recall >= CORRECTNESS_RECALL_FLOOR,
+                "{log_prefix} vector recall@{k} {recall:.3} < floor {CORRECTNESS_RECALL_FLOOR:.2}"
+            );
+            eprintln!("[{log_prefix}] correctness OK: recall@{k} = {recall:.3}");
+
+            let cal: Vec<Option<Calibrated>> = RECALL_TARGETS
+                .iter()
+                .map(|&target| {
+                    eprintln!(
+                        "[{log_prefix}] calibrating recall@{target:.2}: grid over probes/refines ({} queries)...",
+                        q_cal.len(),
+                    );
+                    calibrate(warm_reader, column, q_cal, gt_cal, target, k, log_prefix)
+                })
+                .collect();
+
+            for (i, &target) in RECALL_TARGETS.iter().enumerate() {
+                match cal[i] {
+                    Some(c) => rows.push(RecallRow {
+                        target: format!("{target:.2}"),
+                        params: format!("p={}, r={}", c.probe, c.refine),
+                        recall: format!("{:.3}", c.recall),
+                        warm: include_warm
+                            .then(|| measure_warm(warm_reader, column, q0, k, c.probe, c.refine)),
+                        cold: include_cold.then(|| {
+                            measure_cold(&open_cold, column, q0, k, c.probe, c.refine, cold_iters)
+                        }),
                     }),
-                }),
-                None => rows.push(RecallRow {
-                    target: format!("{target:.2}"),
-                    params: "—".into(),
-                    recall: "—".into(),
-                    warm: None,
-                    cold: None,
-                }),
+                    None => rows.push(RecallRow {
+                        target: format!("{target:.2}"),
+                        params: "—".into(),
+                        recall: "—".into(),
+                        warm: None,
+                        cold: None,
+                    }),
+                }
             }
         }
         rows.push(RecallRow {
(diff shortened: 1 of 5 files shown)


benches/utils/executors.rs
```diff
@@ -44,21 +44,43 @@ pub struct ColdTiming {
...
+        .map(|e| (e.uri, e.subsection_offsets.clone()))
+        .collect();
+    crate::tiers::block_on(async move {
+        let handles: Vec<_> = segments
+            .into_iter()
+            .map(|(uri, offsets)| {
+                let store = store.clone();
+                let disk_cache = disk_cache.clone();
+                let storage = storage.clone();
+                tokio::spawn(async move {
```
ekechinwokah, line 67: serial-parse diagnosis seems right, but this spawns all 256 opens with no bound — on the cold path that's 256 concurrent foreground fetch storms against S3, while the background fill is carefully bounded by prefetch_concurrency. the biggest impediment to performance atm is fanout. can we a) test on supertable and b) bound the fan-out (semaphore or buffer_unordered(N), ideally sized to the same prefetch_concurrency knob)?

benches/utils/executors.rs
```diff
@@ -830,62 +852,74 @@ pub mod vector {
...
-            match cal[i] {
-                Some(c) => rows.push(RecallRow {
-                    target: format!("{target:.2}"),
-                    params: format!("p={}, r={}", c.probe, c.refine),
-                    recall: format!("{:.3}", c.recall),
-                    warm: include_warm
-                        .then(|| measure_warm(warm_reader, column, q0, k, c.probe, c.refine)),
-                    cold: include_cold.then(|| {
-                        measure_cold(&open_cold, column, q0, k, c.probe, c.refine, cold_iters)
+        if skip_calibration {
```
ekechinwokah, line 863: i thought about this, but i'm not sure we should skip calibration at scale. i have a different approach i'm testing in #86 at 50m docs. will update here once done.

benches/utils/executors.rs
ekechinwokah, line 926: run-to-run deltas match rows by this label. with INFINO_BENCH_VECTOR_NPROBE/RERANK changing what "default" means, a knobbed run silently diffs against a baseline measured at a different (p, r). either bake the actual config into the label here, or perhaps we reject INFINO_BENCH_UPDATE_README=1 when the knobs are non-default. my vote would be for the simpler approach though.

benches/utils/tiers.rs
```diff
@@ -40,6 +40,13 @@ const INDEX_CACHE_HEADROOM_DIVISOR: u64 = 10;
 const SUPERFILE_CACHE_GIB: u64 = 4;
 /// Parallel cold-fetch streams used by the bench disk cache.
 const BENCH_COLD_FETCH_STREAMS: usize = 8;
+/// Default concurrent background-fill permits (mirrors the library
+/// default). Override with `INFINO_BENCH_PREFETCH_CONCURRENCY`: warm
+/// promotion runs at most this many segment fills at once, so a
+/// many-segment supertable promotes in `ceil(n_segments / this)` waves —
+/// at the default 8, a 256-segment table needs 32 waves and can miss the
+/// `wait_until_warm` timeout; 64 brings it to 4 waves.
+const BENCH_DEFAULT_PREFETCH_CONCURRENCY: usize = 8;
```
ekechinwokah: this duplicates the library default (DEFAULT_PREFETCH_CONCURRENCY in src/supertable/reader_cache/config.rs:134, currently private) as a literal — if the library default changes, the bench diverges silently. perhaps we make the library const reachable and reuse it?

benches/utils/tiers.rs
```diff
@@ -481,6 +488,20 @@ fn supertable_search_cache_gib() -> Option<u64> {
         .filter(|&v| v > 0)
 }
 
+/// Concurrent background-fill permits for the bench disk cache. Raising
+/// `INFINO_BENCH_PREFETCH_CONCURRENCY` lets a many-segment supertable
+/// finish `wait_until_warm` within the timeout (256 segments promote in
+/// `ceil(256 / concurrency)` waves). Background memory scales as
+/// `concurrency × cold_fetch_streams × cold_fetch_chunk_bytes`, so e.g.
+/// 64 × 8 × 8 MiB ≈ 4 GiB of in-flight fill buffers.
```
ekechinwokah, line 513: worth adding one sentence with the recommended ceiling for a standard 32 GB laptop?

ekechinwokah:
as i mentioned in the comment above, i'm also running into this vector-testing-at-scale problem so just a heads-up that #86  restructures `run_search` ( calibration above 1M docs, single streamed ground-truth pass, new n_docs param) — if it works, the motivation for `INFINO_BENCH_SKIP_CALIBRATION` mostly disappears.

benches/utils/executors.rs
muralikpbhat: ok, will wait

benches/utils/executors.rs
muralikpbhat: Good point, however the unbounded spawn isn't new to the bench — it mirrors the production fan-out in query/dispatch.rs, which already tokio::spawns per probed superfile unbounded. The bench was previously serializing the parse (try_join_all on one poller), so this PR makes it faithful, not stormier. 

However, I agree we need a bound in production not to exhaust the resources. We need to debate whether we should really match prefetch_concurrency, that doesn't seem right to me. Few points:

1. foreground is RTT bound (small pulls) where as background is bandwidth bound(large pulls). So, I see we needing 2 different knobs. Given the prefetch_concurrency is just 8, it would increase 'open' wave by 32X. So I will argue for a higher number for foreground.
2. FWIW, 256 might be a right foreground limit, but I agree we need to test and come up with a bound when the actual number of segments go even higher.
3. Foreground pulls might already have an implicit bound via the S3 connection pool

I would take up this bounding work separately (256 seems reasonable for now given the S3_POOL_MAX_IDLE_PER_HOST=1024), as we need to do that first in production path and then in bench?

>a) test on supertable

All the testing was done on supertable only with real s3 and ec2

> b) bound the fan-out (semaphore or buffer_unordered(N),

Will take this protection work outside of this issue-49.

benches/utils/executors.rs
muralikpbhat: This makes sense. I would prefer rejecting INFINO_BENCH_UPDATE_README, when NPROBE and RERANK are set, but I feel there may be other knobs as well which might influence the outcome and hard do to this comprehensively for all existing and future knobs. We could leave the README updates approval to the code reviewer (not a strong gate though).

The way it prints now is 

══ Supertable vector — search, multi-segment / object-store (1M docs × dim=384) ══
Host: Intel(R) Xeon(R) Platinum 8488C · 8C/16T · 31 GiB RAM · linux/x86_64
| Recall target | (p, r)   | recall | cold open                | cold search            |
| ------------- | -------- | ------ | ------------------------ | ---------------------- |
| default       | p=8, r=4 | —      | 823.08 ms  -54.3% better | 587.42 ms   |

the 'default' is in the recall target column and the passed values are displayed in (p, r) column. So your point on 'either bake the actual config into the label here' is the case already since the second column has p and r. The label is mainly for the recall rate and I am not comfortable baking in the p and r to it.  My suggestion is we leave this to the judgement of the person passing these arguments on how to interpret the results and when to update README.

Ideally, I would leave updating the numbers in README to a nightly build rather than a manual check in?

benches/utils/tiers.rs
muralikpbhat: Thanks for catching this. reused it via DiskCacheConfig::default().prefetch_concurrency instead of exposing the const, to avoid widening the public API surface — same single source of truth.

benches/utils/tiers.rs
muralikpbhat: I won't suggest running real S3 tests from laptop at all :). It is not just the memory concern, bandwidth especially from home is the blocker. No one should run S3 scale test from laptop, laptop runs are just for smoke testing with real-S3. This concurrency knobs are added for ec2 bench as I wanted to speed up the background fill.

benches/utils/executors.rs
ekechinwokah: sounds good - can you please quickly check with vinay on README expectations?

benches/utils/executors.rs
ekechinwokah: ok my fix works at 50m docs. however, it does take a while so i think we should keep this option if someone doesn't want to pay this cost.

ekechinwokah (approved):

benches/utils/executors.rs
vinaykakade: Yes. README update should be a nightly build - and let us not update README numbers in the PR. You can compare the numbers using a CI job and/or paste the numbers in a PR comment, but the README update should happen only via nighly build. Filed https://github.com/infino-ai/infino/issues/108

PR #90, opened 2026-06-11
feat: node bindings

ashishmishra26:
## Summary

First-class **Node.js / TypeScript bindings** for infino, under `infino-node/`.
Wraps the curated `Connection` + `Supertable` catalog API in a Node-idiomatic
surface: pass arrays of plain objects (or apache-arrow `Table`s) in, get plain
records out; opt into apache-arrow `Table`s with `{ arrow: true }`. Mirrors the
Python bindings' surface.

Additive only — the core crate is untouched. `infino-node/` is a separate,
workspace-excluded crate, so a core build never needs a Node toolchain.

## What's in it

- `infino-node/` — a napi-rs addon (`src/lib.rs`) plus a hand-written TypeScript
  wrapper (`infino/index.ts`, compiled by `tsc`).
- **API** — `connect(uri, options?)` → `Connection` (`createTable` / `openTable`
  / `dropTable` / `listTables` / `querySql`); `Table` (`append`, `bm25Search`,
  `vectorSearch`, `tokenMatch`, `exactMatch`, `schema`); `IndexSpec`.
- **CI** — a `node` job (build + smoke tests on ubuntu + macos-14) and a
  `node-addons` job (per-platform `*.node` artifacts), mirroring the Python jobs;
  `make node-test` / `make node-build` targets.
- Node quickstart added to the top-level `README.md` and `infino-node/README.md`.

## Design notes

- **Synchronous** API (matches the Rust + Python surfaces). A call blocks the
  event loop, so a long-running server should run calls in a `worker_thread`.
- **Objects in, records out** — the TS wrapper handles object ↔ Arrow
  conversion; bulk data crosses the JS↔Rust boundary as Arrow IPC, query vectors
  as `Float32Array`. `createTable` also accepts a plain `{ column: type }`
  descriptor (no apache-arrow needed).
- **`_id` is a `bigint`** — the 128-bit `_id` exceeds JS's safe-integer range.
- **Layout** — Rust in `src/`, TypeScript in `infino/` compiled by `tsc`; the
  standard napi-rs + TypeScript package layout.

## Testing

`make node-test` runs 8 smoke tests (create + append via descriptor and
apache-arrow schema, BM25, vector kNN, token/exact match, SQL, local-fs
persistence across reconnect) — green on `darwin-arm64`; CI covers ubuntu +
macos.

## Out of scope (follow-ups)

`update` / `delete`, disk-cache connect options, and npm registry publishing.


.github/workflows/ci.yml
```diff
@@ -139,6 +139,77 @@ jobs:
           name: wheel-${{ matrix.os }}
           path: infino-python/dist
 
+  node:
```
vinaykakade, line 142: nit: tagging here as there isn't otherwise a suitable place to tag. Please add a description to this PR. Should cover an overview of the changes - makes it easier to review, as well as easier to search later in the codebase.

ashishmishra26: thanks, added

infino-node/__test__/smoke.test.mjs
```diff
@@ -0,0 +1,127 @@
...
+  v[i] = 1.0;
+  return v;
+};
+
+test("memory roundtrip: create, append, search, drop", () => {
+  const db = connect("memory://");
+  const docs = db.createTable("docs", titleSchema(), new IndexSpec().fts("title"));
+
+  // append takes plain objects — the wrapper builds Arrow internally.
+  docs.append([{ title: "the quick brown fox" }, { title: "a lazy dog" }]);
```
vinaykakade, line 40: Do you want to add a test example that appends arrow RecordBatch? (in addition to the plain object - so that anyone looking at the code knows the RecordBatch is supported)

infino-node/__test__/smoke.test.mjs
```diff
@@ -0,0 +1,127 @@
...
+  assert.equal(ex.length, 1);
+});
+
+test("unknown table throws", () => {
+  const db = connect("memory://");
+  assert.throws(() => db.openTable("nope"));
+});
+
+test("localfs persists across reconnect", () => {
+  const dir = `${process.env.TMPDIR ?? "/tmp"}/infino-node-smoke-${process.pid}`;
```
vinaykakade: nit: rather than creating the dir name by hand, use something like fs.mkdtempSync(path.join(os.tmpdir(), "infino-node-smoke-"))

infino-node/examples/demo.mjs
```diff
@@ -0,0 +1,53 @@
+// A walkthrough of how someone uses the infino Node binding, end to end.
```
vinaykakade, line 4: License header is missing. Please check other files too

infino-node/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "types": "infino/index.d.ts",
+  "napi": {
+    "name": "infino",
+    "triples": {
+      "defaults": true,
+      "additional": [
+        "aarch64-apple-darwin",
+        "x86_64-unknown-linux-gnu",
+        "aarch64-unknown-linux-gnu",
+        "x86_64-pc-windows-msvc"
```
vinaykakade: Are you able to test on windows? We aren't supporting it yet in python / rust -- may want to take the same position here to reduce the test surface and add windows support later

vinaykakade (approved):
LGTM with comments

infino-node/package.json
ashishmishra26: Remove for now

infino-node/examples/demo.mjs
ashishmishra26: Updated

infino-node/__test__/smoke.test.mjs
ashishmishra26: Updated

infino-node/__test__/smoke.test.mjs
ashishmishra26: Updated

PR #100, opened 2026-06-11
[1/n] refactor supertable manifest: move manifest rebalancing to a separate function

Pranav2612000:
### What does this PR do?
- Moves the logic for rebalancing the manifest during `try_commit` to a separate function

- Makes the logic easier to understand
- Reduces complexity of the `try_commit` fn
- Adds tests for verifying the manifest behaviour


savannahar68 (approved):

savannahar68:
In a separate PR need to fix these issues 
https://github.com/infino-ai/infino/issues/104

PR #102, opened 2026-06-11
[2/n] feat manifest: support removing superfile entries from manifest

Pranav2612000:
### What does this PR do?
- Adds support for removing certain SuperfileEntries from the manifest during the `rebalance_for_commit`

- During compaction, multiple SuperfileEntries are merged into a single merged Superfile. These older entries need to be removed from the Manifest. This PR introduces this support

### How do we do this change?
- The `rebalance_for_commit` fn already returns the latest complete `ManifestListEntries` and the newly created `ManifestParts` ( non-modified, already existing `ManifestParts` are not returned )
- In this PR, before returning these values, we do a pass over these values to remove the `SuperfileEntries` that are to be deleted. Changes to the `ManifestListEntries` and `SuperfileEntries` due to this removal are taken care of.

### How is this tested?
- New unit tests added to test these newer scenarios

### Gotcha
- Multiple ManifestParts belonging to the same partition ( because having them in the same partition would breach threshold ) could be merged together to a single `ManifestPart` after removing superfiles. This PR does not yet take care of this edge case.

### DIFF
--- a/src/supertable/manifest/commit.rs
+++ b/src/supertable/manifest/commit.rs
@@ -482,6 +482,7 @@ pub async fn rebalance_for_commit(
     opts: &Arc<SupertableOptions>,
     old: &Arc<crate::supertable::Manifest>,
     new_entries: &[Arc<SuperfileEntry>],
+    entries_to_remove: &[Arc<SuperfileEntry>],
 ) -> Result<(Vec<ManifestListEntry>, Vec<ManifestPart>), CommitError> {
     // 1. Resolve the effective partition strategy. Locked at
     //    first commit: read from the existing manifest list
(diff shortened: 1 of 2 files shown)


src/supertable/manifest/commit.rs
```diff
@@ -482,6 +482,7 @@ pub async fn rebalance_for_commit(
     opts: &Arc<SupertableOptions>,
     old: &Arc<crate::supertable::Manifest>,
     new_entries: &[Arc<SuperfileEntry>],
+    entries_to_remove: &[Arc<SuperfileEntry>],
```
savannahar68, line 485: as discussed we need to refactor this entire function to return Manifest as a whole 

try_commit_attempt should ideally 

```
write new superfiles
manifest = rebalance
write new manifest
```

in this PR there will be old parts which are less than threshold and has space to consume more Partentries, given we are not merging. A separate PR which will take care of refactoring this function should take care of recreating Manifest.parts entirely so that we can merge multiple partions if its lesser than set target_superfiles_per_partition

savannahar68 (approved):
lgtm
refactoring and improvements to be taken up separately in next PR

PR #103, opened 2026-06-12
[3/n] feat supertable compaction: compact superfiles and update manifest for each job

Pranav2612000:


src/supertable/compaction/mod.rs
```diff
@@ -335,9 +282,106 @@ impl Supertable {
...
+
+        let merged_segment = self
+            .merge_superfiles(&inputs)
+            .await
+            .map_err(CompactionError::Build)?;
+
+        let new_entries = vec![merged_segment.entry];
+        let mut pending_storage_writes = if let Some(bytes) = merged_segment.bytes_for_storage {
+            vec![bytes]
+        } else {
```
savannahar68: error out instead of else?

savannahar68 (approved):
lgtm
one minor comment to be addressed

PR #112, opened 2026-06-13
Fix more testing issues: reduce aggregation latencies and clean up benchmark reporting

ekechinwokah:
## Summary

This branch lands SQL aggregate statistics, swaps the vector rotation for a structured FWHT, and adds a first-cut cost model to the bench suite — plus an Azure storage fix and a batch of bench-harness improvements. It also evaluated moving the hot vector read paths from rayon to tokio and **reverted that move** after measurement (details below). See `benches/README.md` or the CI run for perf comps.

### SQL aggregate statistics (the main change)
- Covered/residual rewrite for aggregates: when an aggregate is fully covered by build-time statistics, it's answered from the manifest instead of scanning; the residual (uncovered) portion falls back to a normal scan-and-fold.
- Build-time stat folding in the manifest so COUNT / SUM / MIN / MAX / AVG-style shapes resolve with minimal I/O.
- New coverage in `tests/supertable/query/covered_agg.rs` and `stats_fold.rs`.

### Vector rotation: structured FWHT
- Replaced the dense rotation with a structured Fast Walsh–Hadamard Transform —
  O(dim·log dim) instead of O(dim²) — in the superfile vector path. This is the
  real vector-search speedup at large dimensions (measured at dim=1024).

### Read-path concurrency: evaluated tokio, kept rayon
- The decode wave (`resolve_columns`) and the vector scoring/rerank waves were moved from rayon to tokio `spawn_blocking`, then **reverted to rayon** after A/B testing showed it didn't pay for itself:
  - **Decode wave:** tokio regressed warm FTS at 1M docs (CI, same machine) — reverted to the rayon `reader_pool` + `par_iter` + oneshot bridge.
  - **Vector scoring:** with FWHT held constant (1M docs, dim=1024, real S3, recall@10 = 0.995 on both), tokio was a **wash on warm** and **slower on cold search** than rayon. The earlier apparent tokio win was the FWHT rotation, not the runtime. Reverted to rayon (oneshot-bridged), keeping the async search signature and FWHT intact.
- Net: CPU waves stay on rayon (the standing "rayon for CPU, tokio for I/O" contract); the FWHT rotation is retained.

### Bench cost model + harness
- New cost model over measured bench rows: a resource meter (vCPU·seconds, PUT counts, stored GB·months, wall times), metered cold S3 HEAD/GET counts and bytes, and a per-cell rate card. Supertable cold "open" is treated as the one-time table open (amortized), priced separately from per-query search.
- Ingest tables now show corpus size, stored footprint, upload bandwidth, and writer count; ingest commits are capped by doc count; vector calibration staircases past 1M docs.

### Storage fix
- `AzureStorageProvider::tail()` now does HEAD + a bounded range request instead of a suffix range (Azure rejects suffix ranges); regression test in `tests/supertable/storage/smoke_azure.rs`.

### CI
- Bench job runs the superfile + supertable tiers and drops the SQL-only leg.

## Performance & recall
- **Vector recall@10 = 0.995** at dim=1024 (≥ 0.99 bar), unchanged across the FWHT swap and the rayon/tokio reverts.
- **rayon vs tokio A/B** (1M docs, dim=1024, real S3, FWHT on both sides):

  | Recall target | warm rayon | warm tokio | cold search rayon | cold search tokio |
  |---|---|---|---|---|
  | 0.90 (p=5,r=1) | 4.92 ms | 5.20 ms | 525 ms | 989 ms |
  | 0.95 (p=5,r=1) | 4.98 ms | 4.31 ms | 508 ms | 1.04 s |
  | 0.99 (p=5,r=1) | 4.85 ms | 4.92 ms | 486 ms | 691 ms |
  | default (p=8,r=20) | 8.21 ms | 8.11 ms | 939 ms | 851 ms |

  → warm is a wash; rayon is faster on cold search. Hence the revert.
- Attach the `cargo bench -- supertable` diff vs `main` (warm/cold search + the
  SQL resolve diagnostic) before merge to confirm no latency/cost regression.


src/supertable/query/exec/common.rs
```diff
@@ -17,22 +17,23 @@
...
+    ArrayRef, Decimal128Array, Float32Array, RecordBatch, RecordBatchOptions, UInt32Array,
+};
 use arrow_schema::{DataType, Field, Schema, SchemaRef};
 use datafusion::error::{DataFusionError, Result as DfResult};
 use datafusion::logical_expr::Expr;
 use datafusion::scalar::ScalarValue;
 use futures::TryStreamExt;
 use parquet::arrow::ProjectionMask;
 use parquet::arrow::async_reader::{ParquetObjectReader, ParquetRecordBatchStreamBuilder};
-use rayon::prelude::*;
```
vinaykakade, line 29: Please add before/after perf numbers in the PR. rayon->tokio is a subtle change, will be good to understand perf implications in tier1/tier2/tier3 (memory/disk/object storage) scenarios.

src/supertable/query/exec/common.rs
```diff
@@ -345,27 +342,29 @@ async fn resolve_columns(
...
-                DataFusionError::Execution("resolve decode: reader pool dropped result".into())
-            })?
-            .map_err(|e| DataFusionError::Execution(e.to_string()))
+        let mut handles = Vec::with_capacity(warm_inputs.len());
+        for (i, sf, locals) in warm_inputs {
+            let names = owned_names.clone();
+            handles.push(tokio::task::spawn_blocking(move || {
+                let name_refs: Vec<&str> = names.iter().map(String::as_str).collect();
+                sf.take_by_local_doc_ids(&locals, &name_refs)
+                    .map(|batch| (i, batch))
```
vinaykakade: For the hot (memory) path, when the decodes moved to tokio and the number of doc ids was large, there was a regression last time. Would be good to double check the current behavior.

src/supertable/query/covered_agg.rs
```diff
@@ -0,0 +1,651 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Infino Authors
+
+//! Covered/residual evaluation for filter-aligned aggregates —
+//! the manifest-statistics "aggregation tree" (claude-plans 022,
+//! phase 5).
```
vinaykakade: The comments need to be cleaned up to remove plan references.

src/superfile/vector/reader.rs
```diff
@@ -24,7 +24,6 @@ use crate::superfile::vector::rerank_codec::RerankCodec;
 use crate::superfile::vector::rotation::RandomRotation;
 use crate::superfile::{ReadError, error::VectorError};
 use bytes::Bytes;
-use rayon::prelude::*;
```
vinaykakade, line 27: Significant cpu-intensive work has moved to tokio from rayon. Assuming perf tests are already done - but let us attach the numbers to the pr.

vinaykakade (approved):

src/superfile/vector/reader.rs
ekechinwokah: decided to revert to rayon.

src/supertable/query/covered_agg.rs
ekechinwokah: done.

src/supertable/query/exec/common.rs
ekechinwokah: good catch. CI results also support. let me take another look.

src/supertable/query/exec/common.rs
ekechinwokah: decided to revert to rayon.

src/supertable/query/exec/common.rs
ekechinwokah: confirmed regression on tokio and confirmed that reverting to rayon fixes it.

PR #115, opened 2026-06-14
Fix memory leak on supertable/query: break SessionContext cycle that leaked consumers on cold reopen

ekechinwokah:
The per-snapshot SessionContext is cached on `SupertableInner`, and its search TVFs (bm25/vector/match/hybrid) held a strong `Arc<SupertableReader>`, which in turn holds `Arc<SupertableInner>`. That formed a reference cycle (`SupertableInner` -> cached `SessionContext` -> `TVF` -> `Arc<SupertableReader>` -> `SupertableInner`), so the consumer never dropped. A long-lived (warm) consumer hides it as a single un-freed cycle; reopening a fresh consumer per query (cold reads) leaks one consumer each time. At scale the cold SQL battery OOM-killed the process (anon-rss ~64.5 GiB on a 62 GiB host).

Fix: the TVFs now hold a WeakReader (`Weak<SupertableInner>` + the pinned `Arc<Manifest>`; a manifest never references the inner, so no cycle) and rebuild the strong pinned reader at call() time. The consumer is always alive while a query executes, so the upgrade succeeds; when the external handle drops, the inner now drops too. The warm `SessionContext` cache is unchanged, so warm latencies are unaffected.

Validated on Azure Blob: 1M SQL cold RssAnon flat ~0.6 GiB (was climbing 2.1 -> 4.6 GiB and still rising); 100M SQL cold completes at ~16.6 GiB peak (was OOM-killed at 64.5 GiB). FTS/vector unchanged.


vinaykakade (approved):
The change itself is good, but curious to know how did you catch it - and whether we need to incorporate anything in CI to catch such issues in the future.

ekechinwokah:
> The change itself is good, but curious to know how did you catch it - and whether we need to incorporate anything in CI to catch such issues in the future.

ran into OOM during SQL cold searches, but not warm, at 100m docs. then was able to repro same symptom at 1M docs - FTS and vector tests finished fine, but SQL RAM was growing unbounded and confirmed there was a leak so debugged from there. 

added a regression test to catch this particular bug in the future.

PR #117, opened 2026-06-14
parallelize corpus generation in benchmarks

ekechinwokah:
Corpus generation is a material part of benchmark testing times, and at 1B+ scale this can be several hours for each corpus. Moving corpus generation to parallel builds to speed this up.


benches/utils/corpus.rs
```diff
@@ -273,29 +311,66 @@ impl MmapTextCorpus {
...
+            let mut rng = StdRng::seed_from_u64(chunk_seed(seed, c));
+            let zipf = ZipfDistribution::new(VOCAB_SIZE);
+            let mut buf: Vec<u8> = Vec::with_capacity(cap);
+            for doc_id in start..end {
+                write!(buf, "doc{doc_id:07}").expect("fmt doc token");
+                for _ in 0..TOKENS_PER_DOC {
+                    write!(buf, " term{:05}", zipf.sample(&mut rng)).expect("fmt term");
+                }
+            }
+            debug_assert_eq!(buf.len(), cap, "deterministic doc-length mismatch");
```
vinaykakade: Benches run in release mode - so this is never reached. This should be assert, instead of debug_assert.

benches/utils/corpus.rs
```diff
@@ -172,8 +212,6 @@ const N_CENT_SMALL: usize = 64;
 /// Average bytes-per-token estimate used to pre-size a doc's `String`
 /// (`(TOKENS_PER_DOC + 1) * AVG_BYTES_PER_TOKEN`).
 const AVG_BYTES_PER_TOKEN: usize = 8;
-/// `BufWriter` capacity (8 MiB) for streaming a corpus to a mmap file.
-const CORPUS_WRITER_BUF_CAPACITY: usize = 8 << 20;
```
vinaykakade, line 176: The memory requirements are much higher with this change, so will OOM at larger corpus sizes. We probably still want buffer capacity - but could be capacity per individual writer, and the output is in memory only till this memory capacity is reached.

vinaykakade (approved):
LGTM with comments

benches/utils/corpus.rs
ekechinwokah: agreed. fixed.

benches/utils/corpus.rs
ekechinwokah: agreed. reverting.

PR #127, opened 2026-06-15
docs: restructure README for onboarding — TOC, Install, agent quickstart, hybrid retrieval

vinaykakade:
Reworks the README so a first-time reader (we're opening the repo to early users) sees what infino is, why to use it, and how to start — fast.

## Changes

- **Opening** leads with the value proposition (one engine for SQL + full-text + vector over a single copy of data on object storage) and a **"Why infino"** list led by best performance-per-dollar. Adds CI + license badges and a **Contents** TOC.
- **Install** section: `pip install infino` / `npm install infino` / `cargo add infino`.
- **Quickstart** — an agent retrieving over a knowledge base (`docs` table), showing all three paths: `bm25_search` (keyword), `vector_search` (vector), and `query_sql` (SQL filter), in Python, Node, and Rust. Copy-pastable: embeddings use a small, clearly labeled stand-in so it runs as-is.
- **Architecture** section folds in the old Links (overview / superfile / supertable), links preserved.
- **SQL joins across tables** now shows the end-to-end hybrid retrieval an agent actually does: BM25 + vector candidates fused with reciprocal-rank fusion (`FULL OUTER JOIN` on `_id`) and joined to a `sources` provenance table — with notes on swapping in a real embedding model. Replaces the old authors/posts example.

## Notes

- The README is the crate doc (`#![doc = include_str!("../README.md")]`), so the Rust quickstart and the hybrid-join example are **runnable doctests** — `make doctest` passes (16/16), so both examples are verified to execute against the engine.
- Docs-only; no code or public-API changes. Merged latest `main`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ infino = "0.1"
 
 ## Quickstart
 
-A memory store for your agent: persist what it learns on object storage, then recall the most relevant pieces to ground the next model call. One table holds the text and its embedding, indexed for both keyword and vector retrieval, with SQL over the same rows — no second system to sync. The backend is chosen by the connection URI (`memory://`, a local path, `s3://…`, `az://…`).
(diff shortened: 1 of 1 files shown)


pratyushlokhande (approved):
Tested on python sdk. It's working.

PR #131, opened 2026-06-15
chore ci: fail ci if test coverage drops below 90

Pranav2612000:


savannahar68 (approved):

Makefile
```diff
@@ -29,7 +29,7 @@ test:
 
 # Coverage (cargo-llvm-cov; install: cargo install cargo-llvm-cov)
 coverage:                      # CI gate: ≥90% overall + lcov.info for codecov upload
-	cargo llvm-cov --summary-only --features test-helpers
+	cargo llvm-cov --summary-only --features test-helpers --fail-under-lines 90
```
vinaykakade: Lines by default also count test lines as well. It can be turned off by annotating test functions with coverage(off) -- but it is only in nightly.

Let us fail under 90% lines as well as under 90% function coverage. We'll need to get the current coverage above these numbers before this PR can be checked-in.

Issue #48, opened 2026-06-07
optimize long multi-term OR queries


ekechinwokah:
our [performance degrades](https://github.com/infino-ai/infino/tree/main/benches) on long FTS `OR` queries:

Query | hot | cold | Peak RSS | Median RSS | P90 RSS | Peak RSS Δ
-- | -- | -- | -- | -- | -- | --
single_rare | 2.78 ms | 320.45 ms | 10.88 GiB | 10.81 GiB | 10.82 GiB | —
single_common | 2.79 ms | 316.83 ms | 10.88 GiB | 10.81 GiB | 10.82 GiB | —
two_term_or | 3.20 ms | 434.43 ms | 10.88 GiB | 10.81 GiB | 10.82 GiB | —
three_wide_or | 5.65 ms | 523.44 ms | 10.88 GiB | 10.81 GiB | 10.82 GiB | —
three_similar_or | 12.06 ms | 495.48 ms | 10.88 GiB | 10.81 GiB | 10.82 GiB | —
five_term_or | 24.29 ms | 523.85 ms | 10.88 GiB | 10.81 GiB | 10.82 GiB | —
ten_term_or | 65.30 ms | 441.77 ms | 10.88 GiB | 10.81 GiB | 10.82 GiB | —
prefix | 51.61 ms | 484.51 ms | 10.88 GiB | 10.81 GiB | 10.82 GiB | —

`AND` queries seem ok for now.

Query | hot | cold | Peak RSS | Median RSS | P90 RSS | Peak RSS Δ
-- | -- | -- | -- | -- | -- | --
two_term_and | 199.03 µs | 139.63 ms | 5.56 GiB | 3.44 GiB | 3.51 GiB | -5.7% improved
three_wide_and | 3.39 ms | 163.54 ms | 5.56 GiB | 3.44 GiB | 3.51 GiB | -5.7% improved
three_similar_and | 5.64 ms | 158.71 ms | 5.56 GiB | 3.44 GiB | 3.51 GiB | -5.7% improved
five_term_and | 6.94 ms | 159.05 ms | 5.56 GiB | 3.44 GiB | 3.51 GiB | -5.7% improved
ten_term_and | 7.99 ms | 182.98 ms | 5.56 GiB | 3.44 GiB | 3.51 GiB | -5.7% improved

vinaykakade:
## Baseline — current `main`, 10M docs

[...]

PR #160, opened 2026-06-16
feat(fts): add Supertable::count — count-only matcher with O(1) df fast path

vinaykakade:
## What

Adds a public `Supertable::count(column, query, mode) -> u64`: count the documents matching an FTS query, with **no scoring and no row materialization**.

- **single token, no deletes** → resolves from the term dictionary's stored document frequency (`term_df`, O(1) per superfile, no posting decode).
- **multi-token / deleted rows** → falls back to the candidate doc-id set minus tombstones.

It's the count-collector analog of `bm25_search`.

## Why

Counting matches previously had no dedicated path: the only way to get a match count was `bm25_search(k = usize::MAX)`, which scores every hit and builds a top-k heap just to discard the scores. For a common term that's tens of milliseconds (score 1M docs) to produce a number the dictionary already stores. This adds the missing count path: O(1) from the stored document frequency for the common single-term case, and a scoreless candidate-set walk otherwise.

## Benchmarks

Count battery added to the superfile + supertable FTS benches: `count()` vs the old `bm25_search(k = usize::MAX)` path, both asserted to return the same count. Superfile tier, 1M docs:

**OR / single-term**

| Query | matches | count() | bm25 (k=MAX) | speedup |
|---|---|---|---|---|
| single_common | 1M | 250 ns | 44.12 ms | **176,495×** |
| single_rare | 2.0K | 209 ns | 5.12 µs | 24.5× |
| two_term_or | 1M | 5.67 ms | 50.28 ms | 8.9× |
| ten_term_or | 977.5K | 34.01 ms | 118.87 ms | 3.5× |

**AND** — the second commit unifies the unranked AND match path with the ranked scorer's block flat-merge, so a count is never slower than scoring-everything (it was, for tiny high-frequency intersections):

| Query | matches | count() before → after | speedup before → after |
|---|---|---|---|
| two_term_and | 335.9K | 3.63 ms → 3.09 ms | 2.9× → **3.4×** |
| five_term_and | 3.5K | 5.69 ms → 4.09 ms | 0.7× → **1.0×** |
| ten_term_and | 4 | 9.06 ms → 5.06 ms | 0.5× → **1.0×** |

Ranked `bm25_search` is unchanged across the battery (within run-to-run noise) — the shared traversal is monomorphized per sink, so the scorer keeps its exact codegen.

## Scope

- Public surface: **+1 method** (`public-api.txt` regenerated).
- `superfile/fts`: the multi-term AND intersection is now one traversal shared by the ranked scorer (`ScoreSink`) and the unranked matcher (`CollectSink`) — removes the duplicate per-leader-doc leapfrog and guarantees the two agree on which docs match.
- Benches: `count` section per tier, sharing the measurement via two new `FtsRead` methods.
- Tests: df fast path across superfiles, `And`/`Or` modes, agreement with `token_match`, empties, the tombstone-subtraction path, and the brute-force BM25 oracle (unchanged, still green).
- `make ci` green (check + doctest + coverage; coverage stays >90% on all three metrics).

1. **SQL `COUNT(*)` pushdown** — make `COUNT(*) … WHERE <fts predicate>` resolve via the same core instead of a DataFusion Parquet scan. Larger change: the `covered_agg` rewrite is plan-level and decoupled from the reader, so it needs either a reader-aware fold or exact `statistics()` on the `token_match` TVF. Separate PR.

### DIFF
--- a/src/superfile/fts/reader.rs
+++ b/src/superfile/fts/reader.rs
@@ -766,10 +766,11 @@ impl FtsReader {
     /// ascending doc-id order.
     ///
     /// Reuses the same [`build_term_cursors`](Self::build_term_cursors)
-    /// the scored path uses, then walks the cursors with
-    /// [`walk_cursors_unranked`] — no BM25 scoring and no top-k heap, so
-    /// nothing is ranked. Cursors traverse blocks in doc-id order, so
-    /// the result is already ascending (no re-sort).
+    /// the scored path uses, then walks the cursors —
+    /// [`collect_and_intersect`](Self::collect_and_intersect) for `And`,
+    /// [`or_merge_unranked`] for `Or` — with no BM25 scoring and no
+    /// top-k heap, so nothing is ranked. Cursors traverse blocks in
+    /// doc-id order, so the result is already ascending (no re-sort).
     pub async fn token_match(
         &self,
         column: &str,
@@ -792,6 +793,34 @@ impl FtsReader {
                 self.collect_and_intersect(column_id, cursors)
             }
             BoolMode::Or => or_merge_unranked(cursors),
+        })
+    }
+
+    /// Unranked token-match **count** — the cardinality
+    /// [`token_match`](Self::token_match) would return, without
+    /// materializing the doc-id `Vec`. The AND path tallies through a
+    /// [`CountSink`], the OR path counts the union walk; both skip the
+    /// `Vec<u32>` so a high-cardinality count doesn't allocate one id
+    /// per match.
+    pub async fn token_match_count(
+        &self,
+        column: &str,
+        tokens: &[&str],
+        mode: BoolMode,
+    ) -> Result<u64, FtsError> {
+        let column_id = self.resolve_column_id(column)?;
+        if tokens.is_empty() {
+            return Ok(0);
+        }
+        let cursors = self.build_term_cursors(column_id, tokens).await?;
+        Ok(match mode {
+            BoolMode::And => {
+                if cursors.len() != tokens.len() {
+                    return Ok(0);
+                }
+                self.count_and_intersect(column_id, cursors)
+            }
+            BoolMode::Or => or_count_unranked(cursors),
         })
     }
 
@@ -1474,6 +1503,22 @@ impl FtsReader {
         let mut sink = CollectSink { out: Vec::new() };
         self.and_flat_merge(&mut cursors, dl_norm_k1, &mut sink);
         sink.out
+    }
+
+    /// Unranked multi-term AND **count**: the size of the intersection
+    /// via the same flat-merge as [`collect_and_intersect`](Self::collect_and_intersect),
+    /// but through a [`CountSink`] that tallies hits instead of
+    /// collecting them — no `Vec<u32>` materialized.
+    fn count_and_intersect(&self, column_id: u32, mut cursors: Vec<TermCursor>) -> u64 {
+        if cursors.is_empty() {
+            return 0;
+        }
+        let col_meta = &self.columns[column_id as usize];
+        let dl_norm_k1 = col_meta.dl_norm_k1.as_slice();
+        cursors.sort_by_key(|c| c.block_count());
+        let mut sink = CountSink { n: 0 };
+        self.and_flat_merge(&mut cursors, dl_norm_k1, &mut sink);
+        sink.n
     }
 
     /// Dispatch to the 2-term specialization or the general `n >= 3`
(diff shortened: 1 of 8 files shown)


src/superfile/fts/reader.rs
```diff
@@ -1440,30 +1447,54 @@ impl FtsReader {
...
+    /// over high-frequency terms costs the same posting-list work as the
+    /// ranked search minus the scoring.
+    fn collect_and_intersect(&self, column_id: u32, mut cursors: Vec<TermCursor>) -> Vec<u32> {
+        if cursors.is_empty() {
+            return Vec::new();
+        }
+        let col_meta = &self.columns[column_id as usize];
+        let dl_norm_k1 = col_meta.dl_norm_k1.as_slice();
+        cursors.sort_by_key(|c| c.block_count());
+        let mut sink = CollectSink { out: Vec::new() };
```
savannahar68, line 1503: updates will be rare so we won't have tombstones most of the times
So does it make sense to also have CountSink along with CollectSink? This saves vec allocation

For a query demanding high document count we can save couple of mbs of data

src/superfile/fts/reader.rs
```diff
@@ -2534,48 +2644,25 @@ fn read_u64_le(b: &[u8]) -> u64 {
 /// [`FtsReader::run_and_intersect`] (And) / the OR merge. Cursors
 /// traverse blocks in doc-id order, so the result is ascending with no
 /// extra sort. Used by [`FtsReader::token_match`].
-fn walk_cursors_unranked(mut cursors: Vec<TermCursor>, mode: BoolMode) -> Vec<u32> {
+/// Unranked multi-term OR: the union of the cursors' doc ids in
+/// ascending order. A k-way merge — each step emits the minimum
+/// current doc id across the live cursors and advances every cursor
+/// sitting on it. No scoring; the caller wants membership, not rank.
+fn or_merge_unranked(mut cursors: Vec<TermCursor>) -> Vec<u32> {
```
savannahar68: Here we are scanning all cursors twice per emitted doc. For a 10-term OR with 1M docs that will be worst case 20M scans. A BinaryHeap<(Reverse<u32>, cursor_idx)> would cut this to O(M log k) from O(M*K)

pop min, advance that cursor, push its next value back. Makes a real difference for the OR heavy COUNT queries

src/supertable/query/fts.rs
```diff
@@ -473,6 +473,120 @@ impl SupertableReader {
...
+        let handles: Vec<_> = kept
+            .into_iter()
+            .map(|entry| {
+                let store = Arc::clone(&store);
+                let disk_cache = disk_cache.clone();
+                let storage = storage.clone();
+                let tombstone_cache = tombstone_cache.clone();
+                let column_arc = Arc::clone(&column_arc);
+                let term_arc = Arc::clone(&term_arc);
+                tokio::spawn(async move {
```
savannahar68: if we error on any single superfile we should return early 
similar to other places in code we should use `futures::future::try_join_all` which will help with returning from this path early

src/supertable/query/fts.rs
```diff
@@ -1475,4 +1632,119 @@ mod tests {
         }
         assert_eq!(cursor, 200_000, "sub-ranges tile the whole superfile");
     }
+
+    #[test]
+    fn count_single_term_sums_df_across_superfiles() {
```
savannahar68, line 1619: can we also add a test for multi term across superfiles so that other branch of code is also tested?

savannahar68 (commented):
some perf related changes
pls check

src/supertable/query/fts.rs
vinaykakade: Done

src/supertable/query/fts.rs
vinaykakade: Good call. Done - and also refactored the dispatch::fanout helper so that other queries get this benefit

src/superfile/fts/reader.rs
vinaykakade: The heap did not help for the current dataset. see benchmarks below, so it's better to keep the flat scan. Here's the data (superfile FTS count battery, 1M docs — the OR rows route through or_merge_unranked):
     
  Query (multi-term OR)   matches   scan (current)   heap      delta
  two_term_or             1M        5.61 ms          8.02 ms   +42.8% worse
  three_wide_or           1M        7.82 ms          12.11 ms  +54.8% worse
  three_similar_or        700K      8.88 ms          9.44 ms   +6.3%  worse
  five_term_or            861K      15.90 ms         22.22 ms  +39.7% worse
  ten_term_or             977K      33.63 ms         52.01 ms  +54.6% worse

Two reasons it loses (I think):
  
1. Small K favors the flat scan. With 2–10 terms, the per-doc min-search is over a tiny contiguous array of plain u32s — branch-predictable and cache-resident. The heap trades that for branchy sift-up/down and Reverse tuple comparisons per posting, whose constant factor swamps the log K ≈ 3.3 advantage at K=10.
2. The real cost is O(P·log K), not O(M·log K). P is the total number of postings (sum of term dfs), which is pushed and popped once each. For common OR queries P is much larger than the union size M, so the heap does more work than the O(M·K) scan rather than less.

That said, we might see an advantage of the heap approach at very large K. Let us take it up as part of https://github.com/infino-ai/infino/issues/48

savannahar68: yeah the comment was with intention that k is large and not just couple of terms because that's where we are poor in search bench game compared with turbopuffer

src/superfile/fts/reader.rs
vinaykakade: Done

savannahar68 (approved):
lgtm
we need to also update search benchmark game infino engine to use the count api added in this PR

PR #173, opened 2026-06-17
Add Filtered Vector Search via Predicate Pushdown

ekechinwokah:
### **What**

Adds filtered kNN vector search: the k-nearest rows **among those matching a text predicate**, by pushdown. A SQL `WHERE` clause on an FTS column is lowered into the vector kernel so the allow-set is applied *inside* the coarse 1-bit shortlist. The result is the true k-nearest among matching rows, with no over-fetch and no post-filter underflow.

Three layers:

1. **Superfile** (`superfile/vector/reader.rs`, `superfile/reader.rs`) — `vector_hits_filtered_async` and `vector_search_clusters_filtered` accept an optional `RoaringBitmap` allow-set. When present, the IVF shortlist + rerank score only allowed doc-ids. `None` is the unfiltered fast path (zero overhead — the old code path is preserved as a delegation).
2. **Supertable** (`supertable/query/vector.rs`) — `vector_hits_filtered_async` resolves the text predicate per superfile into a `SuperfileUri → RoaringBitmap` allow map via `candidate_bitmaps` (concurrent `token_match` fan-out), then feeds it into the existing global cluster-selection + fan-out, which now skips superfiles whose predicate matched nothing (zero vector GETs) and carries the bitmap into each kernel. Also adds `vector_search_filtered` (the public row-returning form) and `vector_hits_filtered` (the hit form with a BM25 text predicate). The fan-out is now wave-capped by reader-pool width to bound transient memory from concurrent cold kernels.
3. **SQL pushdown** (`supertable/query/exec/vector_exec.rs`) — `VectorExec` recognizes a `WHERE <fts_col> = <literal>` or `WHERE <fts_col> IN (...)` predicate on the DataFusion plan, extracts it, and pushes it down as a `CandidatePlan` filter into the vector kernel. Queries like `SELECT _id, score FROM vector_search('docs', 'emb', ARRAY[...], 10) WHERE body = 'rust'` get true filtered kNN, not scan-then-discard.

Also adjusted vector search parameter defaults (p, r) for both filtered and unfiltered search based on the below. Tested sweeps across a matrix; biased towards test corpora but gives reasonable starting point for users and trades a some latency and CPU cycle for high recall.

# Supertable Vector Search — 1M Docs × 1024-dim, Real S3

| Filter | Nominal `(p, r)` | Effective `(p, r)` | Recall@10 | p50 |
|---|---|---|---:|---:|
| unfiltered | `p=6, r=256` | `p=6, r=256` | 1.000 | 3.18 ms |
| filtered (~10%) | `p=8, r=256` | `p=80, r=2560` | 1.000 | 4.23 ms |

| Recall | `(p, r)` | Main warm | Branch warm | Δ warm | Main cold search | Branch cold search | Δ cold |
|---:|---|---:|---:|---:|---:|---:|---:|
| 0.90 | `p=5, r=1` | 14.27 ms | 4.95 ms | −65% | 4.40 s | 534 ms | −88% |
| 0.95 | `p=5, r=1` | 11.10 ms | 4.85 ms | −56% | 3.73 s | 549 ms | −85% |
| 0.99 | `p=10, r=1` | 17.80 ms | 5.42 ms | −70% | 3.41 s | 596 ms | −83% |
| default | `p=8, r=20` | 20.51 ms | 7.85 ms | −62% | 3.86 s | 941 ms | — |
|


src/supertable/query/vector.rs
```diff
@@ -207,33 +282,266 @@ impl SupertableReader {
...
+                // `evaluate` is postings-only (no column decode), the same
+                // retrieval the FTS scan pushdown uses. `None` (unbounded)
+                // cannot occur for a bounded plan; treat it as empty. The
+                // fan-out hit shape is `(local_doc_id, score)`; the score is
+                // an unused `0.0` placeholder (the predicate is unranked).
+                let docs = plan
+                    .evaluate(r.as_ref())
+                    .await
+                    .map_err(|e| QueryError::Parquet(e.to_string()))?
+                    .unwrap_or_default();
```
naman-modi: The `unwrap_or_default()` will make `None` fail silently, which means the plan was unbounded. It is safer to rather throw an error here, so ensuring correctness is paramount.

Something like this?

```
let docs = plan
    .evaluate(r.as_ref())
    .await
    .map_err(|e| QueryError::Parquet(e.to_string()))?
    .ok_or_else(|| QueryError::Execute(
      "bounded CandidatePlan evaluated to Unbounded — planner bug: an Unbounded \
       subtree leaked past from_filters' and/or_combine into the kNN pushdown path".into(),
  ))?;
```

src/supertable/query/vector.rs
```diff
@@ -207,33 +282,266 @@ impl SupertableReader {
...
+                Ok(docs.into_iter().map(|d| (d, 0.0f32)).collect::<Vec<_>>())
+            }
+        };
+        let per_superfile = crate::supertable::query::dispatch::fanout(self, units, kernel).await?;
```
naman-modi: Can we import the crate at top, rather than using inline?
```suggestion
        let per_superfile = dispatch::fanout(self, units, kernel).await?;
```

src/supertable/query/vector.rs
```diff
@@ -207,33 +282,266 @@ impl SupertableReader {
...
+                // retrieval the FTS scan pushdown uses. `None` (unbounded)
+                // cannot occur for a bounded plan; treat it as empty. The
+                // fan-out hit shape is `(local_doc_id, score)`; the score is
+                // an unused `0.0` placeholder (the predicate is unranked).
+                let docs = plan
+                    .evaluate(r.as_ref())
+                    .await
+                    .map_err(|e| QueryError::Parquet(e.to_string()))?
+                    .unwrap_or_default();
+                Ok(docs.into_iter().map(|d| (d, 0.0f32)).collect::<Vec<_>>())
```
naman-modi: [perf, non-blocking] From what I understand, we are essentially inflating the doc-id of each so it can be used in `dispatch::fanout`, then to `tag_hits`, and then use the `group_doc_ids_to_allow` to remove all the empty bitmaps. (doc_ids -> 24B -> doc_ids)

So, we can think of creating a doc-id based list and a native fan-out path basis this doc-ids (returns Vec<u32> / builds the per-superfile RoaringBitmap in-task, tombstone-filtered via roaring set-difference). This would help in gaining performance, and can be taken as a separate PR.

src/supertable/query/vector.rs
```diff
@@ -207,33 +282,266 @@ impl SupertableReader {
...
+        Ok(group_doc_ids_to_allow(per_superfile))
+    }
+}
+
+/// Group per-superfile candidate hits into a `superfile → allow-set` map
+/// of matching `local_doc_id`s, dropping superfiles whose set is empty (so
+/// the caller skips them entirely). Shared by the text-predicate
+/// ([`SupertableReader::candidate_bitmaps`]) and SQL-plan
+/// ([`SupertableReader::candidate_bitmaps_from_plan`]) allow-set builders.
+fn group_doc_ids_to_allow(
```
naman-modi: [non-block, perf] this function is the "undo" half as I mentioned above: it reads each 24B struct, drops the id and score, and keeps just the doc-id to build the bitmap. If the fan-out handed back doc-ids in the first place, this whole function goes away.

src/superfile/vector/reader.rs
```diff
@@ -1393,6 +1395,10 @@ impl VectorReader {
         k: usize,
         nprobe: usize,
         rerank_mult: usize,
+        // Filtered search allow-set (per-superfile matching doc-ids).
+        // `None` = unfiltered; threaded to the coarse shortlist so the
+        // top-k is the true k-nearest among matching rows.
+        allow: Option<std::sync::Arc<roaring::RoaringBitmap>>,
```
naman-modi: Can we please move imports at the top rather than inline?

src/superfile/vector/reader.rs
```diff
@@ -1446,6 +1461,10 @@ impl VectorReader {
         k: usize,
         clusters: &[u32],
         rerank_mult: usize,
+        // Filtered search allow-set (per-superfile matching doc-ids).
+        // `None` = unfiltered; threaded to the coarse shortlist so the
+        // top-k is the true k-nearest among matching rows.
+        allow: Option<std::sync::Arc<roaring::RoaringBitmap>>,
```
naman-modi: Similar, please move imports at the top than inline.

src/superfile/vector/reader.rs
```diff
@@ -1481,6 +1510,9 @@ impl VectorReader {
         chosen: &[usize],
         k: usize,
         rerank_mult: usize,
+        // Filtered search allow-set (per-superfile matching doc-ids),
+        // threaded down to the coarse shortlist. `None` = unfiltered.
+        allow: Option<std::sync::Arc<roaring::RoaringBitmap>>,
```
naman-modi: Similar, please move imports at the top than inline.

src/superfile/vector/reader.rs
```diff
@@ -1765,6 +1798,10 @@ async fn build_shortlist(
     survivor_only_rerank_fetch: bool,
     k: usize,
     rerank_mult: usize,
+    // Filtered search: when `Some`, only doc-ids in this per-superfile
+    // bitmap may enter the coarse shortlist (the predicate's allow-set).
+    // `Arc` so the parallel branch can move a clone into the rayon task.
+    allow: Option<std::sync::Arc<roaring::RoaringBitmap>>,
```
naman-modi: Similar, please move imports at the top than inline.

src/superfile/reader.rs
```diff
@@ -1102,13 +1102,31 @@ impl SuperfileReader {
...
+    /// shortlist, so the returned top-k is the true k-nearest among
+    /// matching rows — pushdown, not post-filter, with no underflow.
+    /// `allow == None` is identical to [`Self::vector_hits_async`].
+    pub async fn vector_hits_filtered_async(
+        &self,
+        column: &str,
+        query: &[f32],
+        k: usize,
+        options: VectorSearchOptions,
+        allow: Option<std::sync::Arc<roaring::RoaringBitmap>>,
```
naman-modi: Similar, please move imports at the top than inline.

src/superfile/reader.rs
```diff
@@ -1125,13 +1143,30 @@ impl SuperfileReader {
...
+    /// predicate allow-set), applied inside the coarse shortlist.
+    /// `allow == None` is identical to [`Self::vector_search_clusters`].
+    pub async fn vector_search_clusters_filtered(
+        &self,
+        column: &str,
+        query: &[f32],
+        k: usize,
+        clusters: &[u32],
+        options: VectorSearchOptions,
+        allow: Option<std::sync::Arc<roaring::RoaringBitmap>>,
```
naman-modi: Similar, please move imports at the top than inline.

src/superfile/vector/reader.rs
```diff
@@ -1826,7 +1874,15 @@ async fn build_shortlist(
...
-                            &codes, &doc_ids, cnt, off, c as u32, &quant, &q_rot_v, &mut heap,
+                            &codes,
+                            &doc_ids,
+                            cnt,
+                            off,
+                            c as u32,
+                            &quant,
+                            &q_rot_v,
+                            allow_owned.as_deref(),
+                            &mut heap,
```
naman-modi, line 1896: [suggestion, non-blocking] We are passing multiple variable right from search_async → probe_clusters_async → build_shortlist → score_cluster_codes_into_heap like the newly added `allow`, and other like `k`, `q_rot`. To make the code cleaner, and more readable we can create a struct like ProbeCtx which can have these vars, and used.

benches/utils/bench/supertable/query.rs
```diff
@@ -16,7 +16,8 @@ pub fn vector_topk_global(
     options: VectorSearchOptions,
 ) -> Vec<u32> {
     let hits: Vec<SuperfileHit> = st
-        .vector_search(VEC_COLUMN, query, k, options)
+        .reader()
```
naman-modi, line 19: We are only measuring the correctness of none-filtered vector queries over here. Can we also have a similar correctness/recall measurement for filtered vector queries?

src/supertable/query/vector.rs
```diff
@@ -207,33 +282,266 @@ impl SupertableReader {
...
+    /// [`Self::vector_hits_filtered`].
+    pub(crate) async fn vector_hits_filtered_async(
+        &self,
+        column: &str,
+        query: &[f32],
+        k: usize,
+        options: VectorSearchOptions,
+        filter_col: &str,
+        filter_query: &str,
+        mode: BoolMode,
```
naman-modi:
```suggestion
    pub(crate) async fn vector_hits_filtered_async(
        &self,
        column: &str,
        query: &[f32],
        k: usize,
        options: VectorSearchOptions,
        filter: VectorFilter<'_>,
```

Small thought: is there a reason we pass the filter as loose args here rather than the VectorFilter struct? Since the public vector_search already groups them into VectorFilter, it might be a touch cleaner to thread the struct straight through instead of unpacking it. People/Agents could by mistake swap the column (which is for vector search) and f.column (which is for filtering column)....

src/supertable/query/vector.rs
ekechinwokah: done. `ok_or_else` with planner-bug message.

src/supertable/query/vector.rs
ekechinwokah: done. moved.

src/supertable/query/vector.rs
ekechinwokah: done. `fanout_with` builds bitmap directly in-task, tombstone-subtracted.

src/supertable/query/vector.rs
ekechinwokah: done. `group_doc_ids_to_allow` removed.

src/superfile/vector/reader.rs
ekechinwokah: done. moved to top.

src/superfile/vector/reader.rs
ekechinwokah: done. moved.

src/superfile/vector/reader.rs
ekechinwokah: done. moved.

src/superfile/vector/reader.rs
ekechinwokah: done. moved.

src/superfile/reader.rs
ekechinwokah: done. moved.

src/superfile/reader.rs
ekechinwokah: done. moved.

src/superfile/vector/reader.rs
ekechinwokah: done — `probe_clusters_async` 8→5 params, `build_shortlist` 9→6 params.

benches/utils/bench/supertable/query.rs
ekechinwokah: fair. added brute-force filtered ground truth + recall gate.

src/supertable/query/vector.rs
ekechinwokah: done. `vector_hits_filtered_async` takes `VectorFilter<'_>`.

src/supertable/query/vector.rs
```diff
@@ -184,15 +242,33 @@ impl SupertableReader {
...
-            }
+                Probe::Nprobe
+            } else {
+                continue;
+            };
+            let bitmap = allow.as_ref().map(|m| {
+                Arc::clone(
+                    m.get(&entry.uri)
+                        .expect("selected superfile is present in allow map"),
+                )
```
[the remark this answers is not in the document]
ekechinwokah: replaced with continue (skip the superfile instead of panicking).

src/supertable/query/exec/vector_exec.rs
```diff
@@ -312,16 +343,43 @@ impl ExecutionPlan for VectorSearchExec {
...
+            // distance only among matching rows); `Unbounded` (no
+            // FTS-resolvable predicate, or none pushed) runs the plain kNN
+            // and lets the `FilterExec` above apply the predicate.
+            let manifest = reader.manifest();
+            let fts_cols: HashSet<String> = manifest
+                .options
+                .fts_columns
+                .iter()
+                .map(|c| c.column.clone())
+                .collect();
```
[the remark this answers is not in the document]
ekechinwokah: done. changed to reference.

src/supertable/query/vector.rs
```diff
@@ -184,15 +242,33 @@ impl SupertableReader {
...
+            });
```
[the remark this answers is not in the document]
ekechinwokah: done.  replaced with `continue`.

src/supertable/query/exec/vector_exec.rs
```diff
@@ -312,16 +343,43 @@ impl ExecutionPlan for VectorSearchExec {
...
+            let plan = CandidatePlan::from_filters(
+                &filters,
+                &fts_cols,
+                manifest.options.tokenizer.as_ref(),
+            );
```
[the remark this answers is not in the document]
ekechinwokah: `from_filters` and all callees now take `HashSet<&str>`, zero cloning.

src/supertable/query/exec/vector_exec.rs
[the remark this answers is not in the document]
ekechinwokah: done.

src/supertable/query/vector.rs
```diff
@@ -164,12 +214,44 @@ impl SupertableReader {
...
+                let selectivity = allowed_docs as f64 / total_docs as f64;
+                let boosted = (options.nprobe as f64 / selectivity).ceil() as usize;
+                let total_clusters = scored.len() + fallback.len();
+                boosted.min(total_clusters).max(options.nprobe)
+            } else {
+                options.nprobe
+            }
+        } else {
+            options.nprobe
+        };
```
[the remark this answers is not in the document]
ekechinwokah: `MAX_FILTER_NPROBE_MULTIPLIER` = 64, uses `saturating_mul`, capped before the cast.

src/supertable/query/exec/vector_exec.rs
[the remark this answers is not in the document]
ekechinwokah: done.

src/supertable/query/vector.rs
```diff
@@ -164,12 +214,40 @@ impl SupertableReader {
...
```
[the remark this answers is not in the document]
ekechinwokah: `total_docs` scoped to only superfiles present in the allow map.

src/supertable/query/vector.rs
```diff
@@ -207,33 +289,334 @@ impl SupertableReader {
...
+        plan: &CandidatePlan,
+    ) -> Result<HashMap<SuperfileUri, Arc<RoaringBitmap>>, QueryError> {
+        let units: Vec<(Arc<SuperfileEntry>, ())> =
+            superfiles.iter().map(|e| (Arc::clone(e), ())).collect();
+        let plan_arc = Arc::new(plan.clone());
+        let body =
+            move |r: Arc<SuperfileReader>,
+                  entry: Arc<SuperfileEntry>,
+                  tombstone_cache: Option<Arc<crate::supertable::tombstones::SidecarCache>>,
+                  now: std::time::Instant,
```
naman-modi: Can we please import at top rather than inline?

src/supertable/query/vector.rs
```diff
@@ -207,33 +289,334 @@ impl SupertableReader {
...
+                    Ok((entry.uri, bm))
+                }
+            };
+        let pairs: Vec<(SuperfileUri, RoaringBitmap)> =
+            dispatch::fanout_with(self, units, body).await?;
+        Ok(pairs
+            .into_iter()
+            .filter(|(_, bm)| !bm.is_empty())
+            .map(|(uri, bm)| (uri, Arc::new(bm)))
+            .collect())
```
naman-modi: There is a lot of repeat code in `candidate_bitmaps` and here `candidate_bitmaps_from_plan`. The only difference is getting doc_ids from token_match there vs plan.evaluate here. The tombstone removal, dropping emptying etc is the same. 

Can we extract the common pieces out? Maybe have a new wrapper like  `pub(crate) async fn fanout_bitmaps<L, Fut>` on top of `dispatch::fanout_with`, or like just extracting the tombstone to a separate method?

naman-modi (approved):
LGTM after resolving these comments.

src/superfile/vector/reader.rs
```diff
@@ -1924,6 +1980,26 @@ async fn build_shortlist(
...
+    let n = n_docs as u64;
+    let allowed = bm.len();
+    if n == 0 || allowed == 0 {
+        return 1;
+    }
+    let selectivity = allowed as f64 / n as f64;
+    (1.0 / selectivity)
+        .ceil()
+        .min(MAX_FILTER_NPROBE_MULT as f64) as usize
+}
```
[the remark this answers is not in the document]
ekechinwokah: done. empty allow-set now returns 0 from `filter_selectivity_mult`; search_async / search_clusters_async return early and skip probe work.

src/supertable/query/exec/vector_exec.rs
```diff
@@ -312,16 +343,48 @@ impl ExecutionPlan for VectorSearchExec {
...
+                    reader
+                        .vector_search_async(&column, &query, k, options)
+                        .await
+                }
+                bounded => {
+                    let pushed_k = k.saturating_mul(INEXACT_OVERFETCH_MULT);
+                    reader
+                        .vector_hits_filtered_by_plan(&column, &query, pushed_k, options, &bounded)
+                        .await
+                }
```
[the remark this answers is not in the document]
ekechinwokah: disagree. bounded TVF pushdown already ranks at true k inside the allow-set; we’re not adding k×N over-fetch + `FilterExec` trim.

src/superfile/vector/reader.rs
```diff
@@ -1924,6 +1983,37 @@ async fn build_shortlist(
...
 /// in-memory subsection slice or the just-fetched centroids
 /// region bytes from [`Source::get_range`] — both reach this
+/// Compute the inverse-selectivity multiplier for filtered search.
+/// Returns 1 (no boost) when `allow` is `None` (unfiltered) or the
+/// bitmap covers the full column. Capped at 64× so very sparse
+/// predicates don't turn every query into a full cluster scan.
+/// Maximum multiplier applied to filtered-search probe breadth and
+/// rerank width. Caps the inverse-selectivity boost so very sparse
+/// predicates don't turn every query into a full cluster scan.
+const MAX_FILTER_NPROBE_MULT: usize = 64;
```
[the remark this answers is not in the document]
ekechinwokah: renamed to `MAX_FILTER_SELECTIVITY_MULT` with named empty/unfiltered constants.

src/supertable/query/vector.rs
```diff
@@ -207,33 +284,334 @@ impl SupertableReader {
...
+    /// fixture.
+    #[cfg(feature = "test-helpers")]
+    pub async fn vector_hits_global_allow_async(
+        &self,
+        column: &str,
+        query: &[f32],
+        k: usize,
+        options: VectorSearchOptions,
+        allow_global: Arc<RoaringBitmap>,
+    ) -> Result<Vec<SuperfileHit>, QueryError> {
```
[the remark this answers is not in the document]
ekechinwokah: left as pub + #[cfg(feature = "test-helpers")] — infino-bench-utils is a separate crate, so pub(crate) doesn’t compile for the bench hook.

src/supertable/query/exec/vector_exec.rs
```diff
@@ -312,16 +343,48 @@ impl ExecutionPlan for VectorSearchExec {
...
+            }
```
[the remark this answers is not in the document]
ekechinwokah: same as earlier — no over-fetch multiplier on the bounded path.

src/superfile/vector/reader.rs
```diff
@@ -1428,7 +1458,14 @@ impl VectorReader {
...
         let _ = sub_start;
         let chosen: Vec<usize> = centroid_scores.iter().map(|&(c, _)| c).collect();
-        self.probe_clusters_async(col, query, &q_rot, &cluster_idx, &chosen, k, rerank_mult)
+        let rerank_mult_eff = rerank_mult.saturating_mul(filter_mult);
+        let ctx = ProbeCtx {
+            q_rot: &q_rot,
+            k,
+            rerank_mult: rerank_mult_eff,
+            allow,
+        };
```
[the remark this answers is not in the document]
ekechinwokah: fair. added `effective_filtered_rerank_mult()` capped at `MAX_EFFECTIVE_FILTERED_RERANK_MULT` (16_384) in both filtered search paths.

benches/utils/supertable.rs
```diff
@@ -1051,6 +1084,94 @@ pub mod vector {
...
+                            offsets
+                                .get(&h.superfile)
+                                .map(|base| (base.saturating_add(h.local_doc_id), h.score))
+                        })
+                        .collect();
+                    recalls.push(corpus::recall_at_k(&global_hits, gt));
+                }
+                let mean_recall: f32 = recalls.iter().sum::<f32>() / recalls.len() as f32;
+                latencies.sort_unstable();
+                let p50_ns = latencies[latencies.len() / 2].as_secs_f64() * 1e9;
```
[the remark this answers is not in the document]
ekechinwokah: done. supertable filtered bench skips stats/report when recalls or latencies is empty.

src/supertable/query/vector.rs
```diff
@@ -251,14 +634,23 @@ impl SupertableReader {
...
+                Some(f) => {
+                    self.vector_hits_filtered_async(
+                        column,
+                        query,
+                        k,
+                        options.for_filtered_path(),
+                        f,
+                    )
+                    .await?
+                }
```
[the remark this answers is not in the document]
ekechinwokah: agree. removed stacked `for_filtered_path()` calls; one options.resolve(filtered) at engine entry.

benches/utils/supertable.rs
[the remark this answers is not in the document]
ekechinwokah: dup. same fix as earlier.

benches/utils/supertable.rs
```diff
@@ -1051,6 +1084,94 @@ pub mod vector {
...
+                    .expect("filtered recall query");
+                    latencies.push(t0.elapsed());
+                    let global_hits: Vec<(u32, f32)> = hits
+                        .iter()
+                        .filter_map(|h| {
+                            offsets
+                                .get(&h.superfile)
+                                .map(|base| (base.saturating_add(h.local_doc_id), h.score))
+                        })
+                        .collect();
```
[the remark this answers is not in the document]
ekechinwokah: done. missing manifest offset for a hit superfile now panics instead of silently dropping via filter_map.

src/superfile/reader.rs
```diff
@@ -1203,6 +1260,16 @@ impl VectorSearchOptions {
...
+
+    /// When stock unfiltered defaults are passed into a filtered kNN path,
+    /// upgrade to [`Self::new_filtered()`]. Explicit overrides are kept.
+    pub(crate) fn for_filtered_path(self) -> Self {
+        if self.nprobe == Self::DEFAULT_NPROBE && self.rerank_mult() == Self::RERANK_MULT {
+            Self::new_filtered()
+        } else {
+            self
+        }
+    }
```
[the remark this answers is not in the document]
ekechinwokah: done. replaced magic rewrite with Option overrides + resolve(filtered).

src/supertable/query/vector.rs
ekechinwokah: done. extracted `fanout_candidate_bitmaps` + `subtract_tombstones`; token-match and plan paths share the tombstone/drop-empty logic.

src/supertable/query/vector.rs
ekechinwokah: done. `SidecarCache` imported at the top of vector.rs; inline path removed from the fan-out closure.

PR #183, opened 2026-06-17
benches: stream the dataset-prepare corpus instead of staging it

pratyushlokhande:
## Problem

Dataset prepare filled the VM disk and panicked before writing anything to object storage — surfacing as `No space left on device` (errno 28) at the corpus `pwrite`, not OOM. `prepare_corpus` generated the **whole** corpus to a local temp file and mmapped it before ingesting. At large doc counts that file exceeds any VM disk (a billion-doc text corpus is ~2 TB; vectors ~4 TB), so a `bench-1b` prepare ran for an hour and died with an empty container.

## Fix

Prepare reads each commit chunk once, in order, so there is no reason to stage the whole corpus first.

- Add `StreamingTextCorpus` / `StreamingVectorCorpus` that generate one append chunk at a time and hold only that chunk in memory.
- Route **prepare-only** runs (build, no read phases) through them via a `ChunkSource` seam; the resident bench path is unchanged.
- Streamed bytes are **identical** to the mmap corpus — shared per-doc generation (`write_text_doc` / `fill_vector_row` / `make_centers`) and the same per-chunk seeds — so a benched dataset that regenerates the corpus for its recall ground truth is unaffected. Commit count, `n_cent` sharding, and the options hash are untouched.
- Lower the prepare workflow's default `docs` from 1,000,000,000 to 10,000,000 (the size the rest of the suite uses).

Prepare now scales to any doc count regardless of local disk; RAM (one commit chunk + engine working set) becomes the ceiling for very large vector runs, noted in the workflow input.

## Validation

- New byte-identity unit tests: streamed text/vector == mmap corpus; uneven `next_*` splits are stable.
- End-to-end on real Azure at 1k docs: all 3 modalities prepared (sidecars + blobs written, no disk-full), then benched — vector **recall@10 = 1.000**, calibration hit 0.90 / 0.95 / 0.99.
- `cargo fmt --check` and `cargo clippy --all-targets` clean.

## Notes

- The `ChunkSource` refactor keeps the column layout in a single `assemble_batch`; both sources are mirror images over it.
- Pre-existing (unrelated): `SequentialSyntheticCorpus` in the grading path does not match `MmapVectorCorpus` (continuous RNG vs per-chunk seed) — 3 tests already red on `main`. Out of scope here; worth a follow-up.


.github/workflows/dataset-prepare.yml
```diff
@@ -23,9 +23,9 @@ on:
         default: "fts vector sql"
         type: string
       docs:
-        description: "Documents to generate (plain integer)."
+        description: "Documents to generate (plain integer). Larger runs scale fine (the corpus streams), but size the VM's RAM to the per-commit working set."
         required: true
-        default: "1000000000"
+        default: "10000000"
```
naman-modi, line 28: Do we want to keep the default as 100M?

benches/utils/corpus.rs
```diff
@@ -1218,6 +1439,67 @@ mod tests {
...
+    /// streaming and disk both reseed per chunk with `chunk_seed` and share
+    /// `write_text_doc` / `fill_vector_row`.
+    const STREAM_TEST_DOCS: usize = 4096;
+    /// Seed shared by the streaming and disk corpora under test.
+    const STREAM_TEST_SEED: u64 = 7;
+    /// Cluster count for the streaming vector-equivalence test.
+    const STREAM_TEST_N_CENT: usize = 16;
+
+    #[test]
+    fn streaming_text_matches_disk_corpus() {
```
naman-modi, line 1568: Hey Pratyush, I don't think any of the tests is testing the `gen_next_chunk`? Can we have a test for it?

naman-modi (approved):

naman-modi:
Please make sure the `make ci` passes.

PR #185, opened 2026-06-17
feat: add gc support

savannahar68:
### what it does

GC supprot added to clear manifests superfiles and parts

### Questions

Should we have a default safety gap (buffer) which GC should consider while deleting or take it as a user input. This PR is taking user input. So this is TBD and might chagne in next PR

### DIFF
--- a/src/supertable/gc/mod.rs
+++ b/src/supertable/gc/mod.rs
@@ -1,0 +1,136 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Infino Authors
+
+use std::collections::HashSet;
+use std::time::{Duration, SystemTime};
+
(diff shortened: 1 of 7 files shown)


src/supertable/gc/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+        live.insert(sf.uri.storage_path());
+    }
+    live
+}
+
+impl Supertable {
+    pub fn gc(&self, safety_gap: Duration) -> Result<GcReport, GcError> {
+        crate::runtime_bridge::bridge_on_runtime(
+            self.gc_async(safety_gap),
+            &self.inner().query_runtime(),
```
Pranav2612000, line 39: Nit: should we reuse the query runtime here?

Pranav2612000 (approved):

PR #231, opened 2026-06-22
feat manifest: add cache for manifest parts

Pranav2612000:
### What does this PR do?
- Adds a cache for storing manifest parts

- The manifest data is the first thing a query touches, to prune superfiles. Without a cache, the manifest parts need to be fetched from S3 ( across restarts ). This PR reduces this round trip and network calls to S3
- This speeds up the query, and reduces S3 costs

### How do we do the change?
- We allow the user to set the maximum size of the cache for manifest parts ( default 2GB )
- Manifest Loader adds a manifest_disk_cache which takes care of all the cache responsibilities

### How has this been tested?
- Tests to verify that number of network calls have decreased.
- Current benchmark tests to ensure that warm queries ( across restarts ) are faster.

### DIFF
--- a/src/supertable/manifest/disk_cache.rs
+++ b/src/supertable/manifest/disk_cache.rs
@@ -1,0 +1,515 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Infino Authors
+
+//! [`ManifestDiskCache`] — a content-addressed on-disk cache for the
+//! compressed (Avro+zstd) bytes of manifest parts.
+//!
+//! Manifest parts are immutable and addressed by the blake3 hash of
+//! their compressed bytes (see [`crate::supertable::manifest::part`]).
+//! That makes them an ideal disk-cache target: the content hash is the
+//! key, so a cached file can never be stale — if the logical content
+//! changes, the hash changes, and the new content lands under a new
+//! filename. No invalidation logic is needed; old files simply become
+//! eviction candidates.
+//!
+//! ## What this buys
+//!
+//! [`ManifestPartLoader`](super::ManifestPartLoader) consults this
+//! cache before issuing a `StorageProvider::get`. On a hit the part
+//! bytes come off local disk instead of object storage, eliminating
+//! the network round-trip (the dominant cost on a cold part load over
+//! S3). The cache also survives process restarts: the on-disk files
+//! persist and the in-memory accounting index is rebuilt by a
+//! directory scan at construction.
+//!
+//! ## What this does NOT cache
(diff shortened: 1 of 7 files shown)


src/supertable/manifest/disk_cache.rs
```diff
@@ -0,0 +1,516 @@
...
+            return;
+        }
+        if !self.reserve(size) {
+            return;
+        }
+
+        let final_path = self.cache_path(&hash);
+        let tmp_path = self.tmp_path(&hash);
+        let write_result = async {
+            tokio::fs::write(&tmp_path, bytes).await?;
```
savannahar68, line 219: what is the idea being write and rename?

src/supertable/manifest/disk_cache.rs
```diff
@@ -0,0 +1,516 @@
...
+            // Roll back the reservation; leave no partial tmp behind.
+            self.current_bytes.fetch_sub(size, Ordering::Release);
+            let _ = tokio::fs::remove_file(&tmp_path).await;
+            return;
+        }
+
+        // Commit into the accounting index. If a concurrent put raced
+        // us to the same hash, release our reservation — the file on
+        // disk is byte-identical either way (content addressing).
+        use dashmap::mapref::entry::Entry;
```
savannahar68: import at top

src/supertable/manifest/mod.rs
```diff
@@ -955,32 +969,64 @@ fn rebuild_part_and_entry(
...
+/// parts are content-addressed, a cache hit can never be stale.
 pub struct ManifestPartLoader {
     storage: Arc<dyn StorageProvider>,
     /// Maps `PartId → (expected content_hash, uri)`. Built from
     /// the manifest list at construction; immutable per-`Manifest`.
     parts_index: HashMap<PartId, (ContentHash, String)>,
+    /// On-disk cache for compressed part bytes. `None` disables the
+    /// cache (in-process-only supertables, tests, or storage attached
+    /// without a `disk_cache_root` configured).
+    manifest_cache: Option<Arc<ManifestDiskCache>>,
```
savannahar68: rename this to manifest_disk_cache ? 
so that its clear this is not in memory. Same rename changes in other files as well

savannahar68 (approved):
minor comments
lgtm

src/supertable/manifest/disk_cache.rs
Pranav2612000: Prevent other readers from reading incomplete/partial data. Provides some semblance of atomicity for the write and reduce CRC mismatch errors

PR #233, opened 2026-06-22
docs: rename compact to optimize in Python/Node READMEs

pratyushlokhande:
## What
Python and Node READMEs still documented the table-maintenance call as `compact` / `CompactOptions`. Renamed to match the shipped binding API:

- Python: `docs.optimize(...)`, `infino.OptimizeOptions(...)`
- Node: `docs.optimize(...)`, `OptimizeOptions`
- Section heading `Compaction` → `Optimization`; API-reference entries updated.

## Why
The bindings expose `optimize` (a wrapper over the internal compaction path) and the options type is `OptimizeOptions` in both bindings. The READMEs were the only stale surface — binding source, Rust `public-api.txt`, `docs/architecture/`, and the examples already use the current names.

## Scope
Docs only; no code change. Verified no other user-facing `compact(` / `CompactOptions` references remain in READMEs, `docs/`, or examples.

### DIFF
--- a/infino-node/README.md
+++ b/infino-node/README.md
@@ -211,14 +211,14 @@ console.log(stats.matched, stats.nTombstoned, stats.nNotFound);
 number you supply, otherwise it throws. Both methods return `{ matched,
 nTombstoned, nNotFound }`.
 
-## Compaction
+## Optimization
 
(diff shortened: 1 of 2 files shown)


savannahar68 (approved):
lgtm

Issue #111, opened 2026-06-12
Supertable: TimeRange-only partitioning — implement the partition-aware write path, drop Hash/ColumnRange


vinaykakade:
## Context

The hierarchical (two-tier) manifest format is in place: `ManifestList` + content-addressed `ManifestPart`s, lazy part loading, list-level aggregate pruning (`manifest/list_prune.rs`, `query/hierarchical_iter.rs`), and a partition-aware rebalance at commit (`manifest/commit.rs::rebalance_for_commit` — rewrite-latest-part per touched partition, part split, cold-partition parts, untouched parts carried over without re-encode).

What's missing is the writer-side partitioning logic. Today every production commit emits superfiles with `partition_hint: None` / `partition_key: []` (`supertable/writer.rs` `PreparedSuperfile` construction — its comment says assignment is "populated by the per-shard `PartitionStrategy` wiring elsewhere", but that wiring doesn't exist yet). Only the default single-bucket sentinel strategy — observationally "no partitioning", one rewritten part per commit — works end-to-end.

## Decision

**TimeRange is the only partition strategy.** It becomes the default and the only variant we implement; `Hash` and `ColumnRange` are removed rather than finished:

- `Hash` was never usable for `n_buckets > 1` (requires writer pre-sharding that doesn't exist; pinned as an expected-error test) and for `n_buckets == 1` it's just a "no partitioning" sentinel.
- `ColumnRange` assignment was an unconditional `SuperfileSpansPartition` error ("lands in a follow-up") — that follow-up is now: don't.

We're pre-1.0 with no persisted-format compatibility burden, so the variants come out of the format, options, and `options_hash` cleanly.

## Work items

- [ ] **Remove `PartitionStrategy::Hash` and `PartitionStrategy::ColumnRange`** from `manifest/list.rs`, `manifest/partition.rs` (assignment + key encode/decode), options, and the Avro/list encoding. A table with no configured strategy stays unpartitioned (single partition), replacing today's single-bucket Hash sentinel.
- [ ] **Remove `SuperfileEntry.partition_hint`** — it existed solely for multi-bucket hash pre-sharding and has no other consumer.
- [ ] **Remove `part_size_threshold_bytes`** from `SupertableOptions` — documented as a part-split trigger but never consulted; `target_superfiles_per_partition` (which `rebalance_for_commit` does check) remains the split knob.
[...]

PR #245, opened 2026-06-22
feat manifest partition: use ingestion time with daily granularity as default partition

Pranav2612000:


savannahar68 (approved):

muralikpbhat:
What was the motivation for this change? Is there any GitHub issue explaining the problem?

Pranav2612000:
> What was the motivation for this change? Is there any GitHub issue explaining the problem?

https://github.com/infino-ai/infino/issues/111 Although I'm still not convinced that time based sharding is the best strategy. We decided to keep it this way to have a sensible default for now.

PR #248, opened 2026-06-22
ci: trigger Node and Python publishes on a v* release tag

ashishmishra26:
A `v<version>` tag now publishes all three artifacts as one coordinated release. The crate workflow already triggers on the tag; this wires the Node and Python publish workflows to the same `v[0-9]*` pattern.

**Node** (`node-publish.yml`)
- Triggers on a `v*` tag and publishes for real. The version comes from `infino-node/package.json`; a guard asserts it matches the tag so a release can't ship a mismatched version.
- Manual `workflow_dispatch` runs are unchanged — still default to a dry run.

**Python** (`publish-python.yml`)
- Triggers on a `v*` tag, derives the version from the tag (`v0.1.0` → `0.1.0`), and publishes to **real PyPI**.
- Manual `workflow_dispatch` runs keep their `version` input and the TestPyPI (default) / PyPI target choice.

**Semantics this establishes**
- A `v` tag = coordinated release of crate + Node + Python at that version.
- Single-package patches (Node-only, Python-only, Rust-only) still go through a manual workflow run, so they stay independent of the coordinated release — consistent with `docs/versioning.md` (major.minor locked, patch independent).

Merge after the initial `v0.1.0` has been published manually.

### DIFF
--- a/.github/workflows/node-publish.yml
+++ b/.github/workflows/node-publish.yml
@@ -4,12 +4,16 @@ name: Node publish
 # thin `infino` package plus its per-platform binary packages (`infino-<triple>`)
 # to npm.
 #
-# Triggered manually. The default run is a dry run: it packs and validates the
-# main package and every platform package without uploading anything. Re-run
-# with dry_run unchecked to publish for real. A real run needs the NPM_TOKEN
+# Triggered two ways: (1) a `v<version>` tag pushes a real publish as part of a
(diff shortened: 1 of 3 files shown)


ashishmishra26:
Superseded by a fresh PR rebased on current main — the scoped `@infino-ai/infino` rename and `--access public` landed since this was drafted, and this branch carried a stale package-lock. Reopening the same change cleanly.

PR #251, opened 2026-06-22
feat SQL: prune manifest-parts/superfiles for SQL IN-list filters

### LINKED ISSUE #180: SQL: prune segments for IN-list scalar predicates
Sub-issue of #179.

## Problem

`WHERE col IN (a, b, c)` doesn't prune — `collect_conjuncts` (`src/supertable/query/provider.rs`) ignores `Expr::InList`, so every segment opens even when its min/max excludes all listed values.

## Approach

Lower a non-negated `InList` to a disjunction of equality `ScalarPredicate`s; keep a segment if any listed value is in `[min, max]` (mirrors `candidate.rs::in_list_leaf`). Add a `BETWEEN` regression test while here (it desugars to `>= AND <=` and should keep pruning).

## Soundness

Min/max is exact; an over-wide keep-set is slow, never wrong (`FilterExec` re-verifies). `NOT IN` is out of scope.

## Acceptance

- `IN` prunes segments whose min/max excludes every value; result identical to the scan path (oracle test).
- No regression on the SQL bench battery.

naman-modi:
Closes #180 

### What does this PR do?
Extends `WHERE`-clause superfile pruning to `column IN (...)`. So an `IN` query opens only the superfiles whose stats say they might hold a listed value ; the same "ask the manifest before touching bytes" pruning.

Today an `IN` filter reaches `SupertableProvider::scan` but isn't lowered to any prune leaf, so it contributes nothing and **every** superfile is opened. Equality (`col = x`) and prefix already prune; `IN` was the gap.

### How do we do this change?
Each `column IN (...)` filter becomes a manifest check that skips superfiles before opening them:
- a **min/max** check (every column); skip if the value range can't hold any listed value;
- on a **full-text column**, also a **bloom** check; skip if the text holds none of the values' words.

Run in two tiers (manifest parts, then superfiles); only survivors are opened. This PR deals only with  4+ value lists (DataFusion rewrites smaller `IN`s to `=`/`OR`); `NOT IN`, non-literals, and function/unknown columns produce no leaf and just scan.

### Performance
Cold tier, 1M docs / ~16 superfiles, `key IN (4 keys)` on the FTS-indexed, unsorted `key` column (min/max can't prune it; only the bloom can):

| | cold search p50 | superfiles opened |
| --- | --- | --- |
| before | 2.86 s | all ~16 |
| after | 0.51 s | only the ~4 holding |

≈ 5.6× improv; the bloom skips the cold opens (object-store GET + footer + decode) of the superfiles holding none of the keys.

### Notes
- Multi-word FTS values (`'Orange Juice'` → `orange, juice`) flatten into one `Or` bloom over all tokens - a conservative superset (it over-keeps a superfile holding only `orange`); `FilterExec` verifies. A tighter per-value `(orange AND juice) OR pineapple` is possible but needs a disjunctive leaf.

### DIFF
--- a/src/supertable/query/provider.rs
+++ b/src/supertable/query/provider.rs
@@ -305,7 +305,9 @@ impl SupertableProvider {
     // survive the two-tier prune — per-part aggregates (ManifestPartEntry)
     // first, then per-superfile stats (SuperfileEntry).
     //
-    // These superfiles are then opened and scanned by DataFusion.
+    // The survivors are returned to `scan`, which opens each through the
+    // superfile reader for its byte source + footer, then lets
+    // DataFusion's Parquet engine decode the rows over that byte source.
     async fn select_survivors(&self, filters: &[Expr]) -> DfResult<Vec<Arc<SuperfileEntry>>> {
         let predicates = exprs_to_scalar_predicates(filters, &self.schema);
         let mut leaves = self.predicates_to_prune_leaves(predicates);
@@ -1133,8 +1135,10 @@ pub(crate) fn exprs_to_scalar_predicates(
 /// A `NOT IN`, a non-literal item, a function-wrapped or unknown column
 /// yields no leaf — that filter just isn't pruned (the scan stays correct).
 ///
-/// Only sees `Expr::InList` (4+ values): DataFusion rewrites a 1-value
-/// `IN` to `=` and a 2–3 value `IN` to `OR` before the provider.
+/// Handles any `Expr::InList`, whatever its length. In practice SQL
+/// planning rewrites a short `IN` first — a 1-value to `=`, a 2–3 value
+/// to `OR` — so only 4+ value lists usually reach here, but that's the
+/// optimizer's behavior, not a precondition this helper relies on.
 pub(crate) fn exprs_to_in_list_leaves(
     filters: &[Expr],
     schema: &SchemaRef,
@@ -1166,6 +1170,10 @@ fn collect_in_list_leaves(
     out: &mut Vec<PruneLeaf>,
 ) {
     match expr {
+        // Filters reach us alias-free (Filter::try_new runs unalias_nested),
+        // but an alias is a pure rename; descend it so pruning is unaffected
+        // if one ever survives (e.g. a metadata-carrying alias).
+        Expr::Alias(a) => collect_in_list_leaves(&a.expr, schema, fts_cols, tokenizer, out),
         // Descend AND; an IN can sit on either side.
         Expr::BinaryExpr(be) if be.op == Operator::And => {
             collect_in_list_leaves(&be.left, schema, fts_cols, tokenizer, out);
@@ -1518,6 +1526,18 @@ mod tests {
             .gt(lit(0_i64))
             .and(col("y").in_list(vec![lit(7_i64)], false));
         assert_eq!(in_list_leaves(&[expr], &s), vec![("y".to_string(), 1)]);
+    }
+
+    #[test]
+    fn in_list_under_alias_is_found() {
+        // Filters reach us unaliased, but the descent must still find an
+        // IN wrapped in an alias if one ever survives (a pure rename
+        // doesn't change the column the leaf prunes on).
+        let s = schema_xy();
+        let expr = col("x")
+            .in_list(vec![lit(1_i64), lit(2_i64)], false)
+            .alias("k");
+        assert_eq!(in_list_leaves(&[expr], &s), vec![("x".to_string(), 2)]);
     }
 
     #[test]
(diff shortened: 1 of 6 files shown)


src/supertable/query/provider.rs
```diff
@@ -295,20 +301,32 @@ impl SupertableProvider {
         leaves
     }
 
-    /// Lower `filters` to prune leaves and return the superfiles that
-    /// survive the two-tier prune — exactly the inputs the scan hands to
-    /// DataFusion.
+    // Lower `filters` to prune leaves and select the superfiles that
```
ekechinwokah, line 304: datafusion does not read superfiles. please fix the comment.

src/supertable/query/provider.rs
```diff
@@ -1107,6 +1125,104 @@ pub(crate) fn exprs_to_scalar_predicates(
...
+/// only `orange` is kept though no value matches. A tighter prune would be
+/// per-value `(orange AND juice) OR pineapple`; FilterExec verifies either way.
+fn collect_in_list_leaves(
+    expr: &Expr,
+    schema: &SchemaRef,
+    fts_cols: &HashSet<&str>,
+    tokenizer: Option<&dyn Tokenizer>,
+    out: &mut Vec<PruneLeaf>,
+) {
+    match expr {
```
ekechinwokah, line 1171: we only match bare `Expr::Column`, so an `IN` under an alias silently won't prune. we should probably descend `Expr::Alias` here too. and should we leave `Cast/TryCast` out? those cross a coercion boundary and may need their own soundness argument before we treat them as a bare column for `min/max`.

src/supertable/query/provider.rs
```diff
@@ -1107,6 +1125,104 @@ pub(crate) fn exprs_to_scalar_predicates(
...
 
+/// Build prune leaves for the `column IN (...)` filters in `filters`. Each yields:
+///  i)  a `ScalarInList` (min/max) leaf — on every column;
+///  ii) plus, on an FTS-indexed column, a `TermPresence{Or}` leaf over the
+///      values' tokens, which prunes on which superfiles hold the term.
+///
+/// A `NOT IN`, a non-literal item, a function-wrapped or unknown column
+/// yields no leaf — that filter just isn't pruned (the scan stays correct).
+///
+/// Only sees `Expr::InList` (4+ values): DataFusion rewrites a 1-value
```
ekechinwokah: note: don't write this as an invariant — the new tests build 1- and 3-value `InList` directly. it's just what the optimizer usually gives; the helper takes any `InList`. 

also consider rewording so nobody builds on false precondition.

ekechinwokah:
should we add a BETWEEN regression test?

src/supertable/query/provider.rs
```diff
@@ -295,20 +301,32 @@ impl SupertableProvider {
...
 
-    /// Lower `filters` to prune leaves and return the superfiles that
-    /// survive the two-tier prune — exactly the inputs the scan hands to
-    /// DataFusion.
+    // Lower `filters` to prune leaves and select the superfiles that
+    // survive the two-tier prune — per-part aggregates (ManifestPartEntry)
+    // first, then per-superfile stats (SuperfileEntry).
+    //
+    // These superfiles are then opened and scanned by DataFusion.
     async fn select_survivors(&self, filters: &[Expr]) -> DfResult<Vec<Arc<SuperfileEntry>>> {
```
ekechinwokah, line 310: should we gate this on actually having a non-negated `InList`: build fts_cols_set() and call `exprs_to_in_list_leaves` only then? so the common non-IN scan pays zero....

ekechinwokah (approved):

src/supertable/query/provider.rs
```diff
@@ -1107,6 +1122,106 @@ pub(crate) fn exprs_to_scalar_predicates(
...
+            }
+
+            // Every item must be a literal to bound min/max; else skip.
+            let mut values = Vec::with_capacity(il.list.len());
+            for item in &il.list {
+                let Expr::Literal(v, _) = item else {
+                    return;
+                };
+                values.push(v.clone());
+            }
```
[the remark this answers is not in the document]
naman-modi: keep-on-incomparable is the existing invariant; exact-type guard reduces pruning. Non-issue.

src/supertable/query/provider.rs
naman-modi: corrected my wordings, thanks.

src/supertable/query/provider.rs
naman-modi: By the time, the filters reach here, the aliases are generally resolved, but there could be some cases of alias with metadata, which needs to be handed - added an arm for that, thanks. Agreed, leaving the CAST for now, can pick up later to have proper support.

src/supertable/query/provider.rs
naman-modi: makes sense, updated.

src/supertable/query/provider.rs
naman-modi: to gate itself, we will have to pay a cost, i.e. filter walk etc, so doesn't come at no cost, so I think it does the work efficiently the way it is, as it pays almost nothing.

PR #255, opened 2026-06-23
bench: simplify the PR comment and split supertable CI by modality

pratyushlokhande:
Two related cleanups to the Azure benchmark CI/reporting.

## 1. Simplify the PR comment

The comment carried a per-row "Main (10c)" annotation and Mermaid charts tracking the last 10 main commits, plus a glossary footnote. It now reports only the clean **PR-vs-latest-main** delta.

- Remove `load_history` / `main_trend` and the `Main (10c)` column from `bench_summary.py`; drop the "Render trend charts" and "Collect main history" workflow steps.
- Per-commit history blobs are still published for archival; only their display is gone.
- Rename user-facing `regression`/`improvement` to neutral **Worse / Better**.

## 2. Split supertable CI by modality

The `supertable` leg ran all three modalities on one VM and posted one combined comment, so a noisy modality muddied the rest with no per-modality control.

- Fan supertable into three CI legs — **fts**, **vector**, **sql** — each on its own VM with its own baseline and sticky PR comment.
- Add `supertable_{fts,vector,sql}` tokens to the reusable workflow; collapse the duplicate `supertable`/`supertable_all` arm and drop the now-dead `fts`/`vector`/`combined`/`sql` tokens (unreferenced).
- `supertable`/`supertable_all` stay for manual all-in-one dispatch.

## Validation

- `bench_summary.py` validated end-to-end against the real Azure container (deterministic table, live Foundry verdict, and the model-unavailable fallback).
- Both workflow YAMLs parse; unique-resource keys (VM name, NIC, NSG, comment marker, concurrency group) confirmed distinct per leg; per-modality baseline blobs already exist so deltas render immediately.

### DIFF
--- a/.github/scripts/bench_summary.py
+++ b/.github/scripts/bench_summary.py
@@ -1,17 +1,16 @@
 #!/usr/bin/env python3
 """Summarize benchmark deltas for a PR comment.
 
-Diffs this run against the `main` baseline, keeps only changes past the noise
-threshold, annotates each with main's own trend, and writes a lean markdown
-summary. Foundry narrates; the percentages are computed here, never by the
-model. On any model failure or missing creds the deterministic tables are the
-summary on their own. Stdlib only — the runner needs no pip install.
+Diffs this run against the latest `main` baseline, keeps only changes past the
+noise threshold, and writes a lean markdown summary. Foundry narrates; the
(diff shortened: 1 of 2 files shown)


naman-modi (approved):
Lets move on to CI for main-vs-PR comparison too.

PR #257, opened 2026-06-23
Fix vector search under-returning after deletes

pratyushlokhande:
## Problem

Unfiltered `vector_search` excludes tombstoned rows in `dispatch::fanout` —
*after* each superfile's per-kernel top-k has already been truncated to `k`.
A deleted row sitting inside that top-k is dropped with no backfill, so the
result shrinks below `k`. A small-`k` search whose nearest rows were all
deleted returns fewer than `k` live rows — empty at `k=1`.

Reproduces wherever the deleted rows exhaust the `k · rerank_mult` candidate
pool (more visible on native x86, but architecture-independent). The filtered
path is unaffected: its allow-set has tombstones subtracted before ranking, so
it never selects them.

## Fix

The unfiltered fan-out now over-fetches each superfile's kNN by that
superfile's tombstone count, so `k` live rows remain once the post-rank
tombstone retain runs. Bounded by the actual tombstone count; in-memory and
no-delete paths are unchanged (over-fetch is 0). Filtered search is untouched.

## Test

Adds `vector_search_backfills_past_deleted_rows`: deletes the nearest rows from
a storage-backed table and asserts a `k=1` search still returns the nearest
live row. Full lib suite passes (1560 tests); fmt + clippy clean.


src/supertable/query/vector.rs
```diff
@@ -1025,6 +1053,97 @@ mod tests {
...
+    /// Regression: after deletes, vector search must still return `k` *live*
+    /// rows when at least `k` remain. The tombstone filter runs on each
+    /// superfile's already-truncated top-k, so deleted rows in the top-k were
+    /// dropped with no backfill — a small-`k` search whose nearest rows are
+    /// all deleted returned fewer than `k` (empty at `k=1`).
+    #[test]
+    fn vector_search_backfills_past_deleted_rows() {
+        use datafusion::prelude::{Expr, col, lit};
+
+        use crate::{storage::LocalFsStorageProvider, supertable::storage::StorageProvider};
```
Pranav2612000, line 1065: Nit: move imports to the top

src/supertable/query/vector.rs
```diff
@@ -320,11 +348,11 @@ impl SupertableReader {
             while !units.is_empty() {
                 let n = fanout_width.min(units.len());
                 let wave: Vec<_> = units.drain(..n).collect();
-                collected.extend(dispatch::fanout(self, wave, kernel.clone()).await?);
+                collected.extend(dispatch::fanout_with(self, wave, body.clone()).await?);
             }
             collected
         } else {
-            dispatch::fanout(self, units, kernel).await?
```
Pranav2612000, line 327: Are these changes required for the fix?

Pranav2612000 (changes requested):

PR #266, opened 2026-06-24
fts: exclude negated terms from count and token_match

### LINKED ISSUE #261: incorrect values for count queries when query contains negated terms
Check the difference in doc between other engines and infino
The result returned by count doesn't factor in documents from negated terms

Image

vinaykakade:
## Problem

Closes #261. Count queries (and the unranked `token_match`) returned incorrect values when the query contained a negated term. For a query like `alpha -beta`, the count included the documents that matched `beta` instead of excluding them — diverging from what a scored search returns and from other engines.

## Root cause

`token_match_count_async` and `token_match_async` parsed the query with the *indexing* tokenizer (`tokenize()`), which knows nothing about query syntax. `alpha -beta` was split into `["alpha", "beta"]`, silently turning the negated term into a positive one. The scored `bm25_search` path already parsed correctly (`parse()` → positives + negatives) and excluded negatives; the two unranked paths did not.

## Fix

- New shared `parse_and_prune` helper: parses positives/negatives and prunes superfiles on **positives only** (a negated term must never drop a superfile — same rule the scored path follows).
- `token_match_count_async`: the `term_df` (O(1)) and counting-sink fast paths now run only when there are no negatives and no deletes. When negatives are present, positive matches are materialized and any doc that also matches a negated term (union) or carries a tombstone is dropped in one filter pass.
- `token_match_async`: same negation handling, preserving the documented `count == token_match.len()` invariant.

## Tests

Added regression coverage that fails before the fix and passes after:
- `count_excludes_negated_terms` — `alpha -beta` → 2, not 3, across multiple superfiles.
- `count_with_negation_agrees_with_token_match` — the count↔token_match invariant under negation, across OR/AND and single/multi-positive shapes.
- `count_excludes_negated_terms_and_tombstones` — negation and deletes compose.

`cargo fmt --check`, `cargo clippy --lib --tests`, and the full `fts` query test module are green.

## Performance

Touches the FTS query fan-out but not BM25/vector scoring math, distance/SIMD kernels, codecs, or the commit path. The negation-free, delete-free common case keeps its existing fast paths unchanged; negation now forces a materialized walk (correctness over the prior — wrong — fast count). No bench-relevant hot path changed.

### DIFF
--- a/src/supertable/query/fts.rs
+++ b/src/supertable/query/fts.rs
@@ -80,6 +80,7 @@ use std::{
 };
 
 use arrow::record_batch::RecordBatch;
+use roaring::RoaringBitmap;
 use uuid::Uuid;
 
 pub use crate::superfile::fts::reader::BoolMode;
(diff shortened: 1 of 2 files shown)


src/supertable/query/fts.rs
```diff
@@ -446,41 +447,84 @@ impl SupertableReader {
...
+                        .await
+                        .map_err(|e| QueryError::Parquet(e.to_string()))?
+                        .into_iter()
+                        .collect();
+                    docs.into_iter()
+                        .filter(|d| !excluded.contains(*d))
+                        .collect::<Vec<_>>()
+                } else {
+                    docs
+                };
```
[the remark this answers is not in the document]
vinaykakade: not worth the code for a rare query

src/supertable/query/fts.rs
```diff
@@ -446,41 +447,84 @@ impl SupertableReader {
...
+        &self,
+        column: &str,
+        query: &str,
+        mode: BoolMode,
+    ) -> Result<(Vec<String>, Vec<String>, Vec<Arc<SuperfileEntry>>), QueryError> {
+        let parsed = AsciiLowerTokenizer.parse(query);
+        let positives: Vec<String> = parsed.positives.into_iter().map(Cow::into_owned).collect();
+        let negatives: Vec<String> = parsed.negatives.into_iter().map(Cow::into_owned).collect();
+        if positives.is_empty() {
+            return Ok((positives, negatives, Vec::new()));
```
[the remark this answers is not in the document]
vinaykakade: not worth the code for a rare query / not the worth the tax for common queries

src/supertable/query/fts.rs
```diff
@@ -446,41 +447,84 @@ impl SupertableReader {
     ///
     /// `pub(crate)` async kernel; the public surface is the sync
     /// [`SupertableReader::token_match`].
```
naman-modi: Hey Vinay, this comment explaining the `token_match` should be above the method on L485.

src/supertable/query/fts.rs
```diff
@@ -446,41 +447,84 @@ impl SupertableReader {
...
+        column: &str,
+        query: &str,
+        mode: BoolMode,
+    ) -> Result<(Vec<String>, Vec<String>, Vec<Arc<SuperfileEntry>>), QueryError> {
+        let parsed = AsciiLowerTokenizer.parse(query);
+        let positives: Vec<String> = parsed.positives.into_iter().map(Cow::into_owned).collect();
+        let negatives: Vec<String> = parsed.negatives.into_iter().map(Cow::into_owned).collect();
+        if positives.is_empty() {
+            return Ok((positives, negatives, Vec::new()));
+        }
```
naman-modi, line 487: Negation-only here returns empty kept → count yields 0 and token_match yields []. 

But the scored path errors on the same input: bm25_search("-foo") returns `FtsError::NegationOnly`. Let us do the same here?

naman-modi (approved):
LGTM with some minor comments.

src/supertable/query/fts.rs
vinaykakade: Done

src/supertable/query/fts.rs
vinaykakade: Done

savannahar68:
confirmed that the fix worked - https://infino-ai.github.io/search-benchmark-game/full/

PR #278, opened 2026-06-24
fix vector query: exclude tombstones before kNN ranking

### LINKED ISSUE #259: Vector search returns fewer than k results after deletes (no backfill past tombstones)
## Summary

Unfiltered `vector_search` can return **fewer than `k`** rows (empty at `k=1`)
after deletes, even when ≥`k` live rows remain. Tombstones are dropped *after*
each superfile's top-k is chosen, with no backfill — so a deleted row sitting
in the top-k just shrinks the result.

## Repro

1. Create a vector table, insert N rows.
2. Delete the row(s) nearest to query `q`.
3. `vector_search(q, k=1)` → `[]` (expected: the nearest *live* row).

Generalizes: delete the `D` nearest rows and the result drops below `k` whenever
the tombstoned rows occupy the candidate pool. Surfaced via LangChain's standard
`VectorStore` delete compliance tests; reproduces engine-side on any arch (e.g.
delete the ~10 nearest of 60, `k=1`).

## Root cause

The unfiltered fan-out filters tombstones *post-rank*: each per-superfile
kernel truncates to `k`, then `dispatch::fanout` drops tombstoned hits via
`apply_tombstone_filter` (`hits.retain(...)`, `src/supertable/query/dispatch.rs`),
then `top_k_ascending(k)`. There is no backfill, so any tombstoned row in the
top-k reduces the count.

The **filtered** path is immune: its allow-set has tombstones subtracted
(`subtract_tombstones`) and is applied *inside* the coarse shortlist
(`score_cluster_codes_into_heap`, `src/superfile/vector/reader.rs`), so it ranks
only among live rows. The unfiltered path has no equivalent.

Note on the arch-dependence (red herring): it first showed only on native x86
CI. That's just the trigger — SIMD FP rounding decides whether the deleted row
wins the `k=1` slot. The bug itself is arch-independent.

## Impact

Correctness: delete-then-search silently under-returns / returns empty at small
`k`. Affects any delete + kNN workload. Workaround: query with a larger `k`.

## Suggested direction

Exclude tombstones *before* ranking, mirroring the filtered allow-set — pass a
per-superfile `deny` set into the coarse shortlist so deleted rows never enter
the candidate heap (top-k is then the true k-nearest among live rows, no
over-fetch, rerank cost bounded by `k`). A simpler over-fetch-by-tombstone-count
variant also works but reranks extra candidates proportional to delete volume.

naman-modi:
### What does this PR do?
Fixes #259. Unfiltered vector_search no longer under-returns after deletes. For e.g. k=1 returns the nearest live row instead of an empty result, and any k returns the true k-nearest among live rows.

A delete only writes a tombstone; the row stays in the superfile until compaction. The unfiltered fan-out removed tombstones "after" each superfile's kernel ranked all rows (tombstoned included) and truncated to its top-k, so the final result has fewer than k responses, i.e,  giving incomplete results.
  
However, the filtered path was already immune: its allow-set has tombstones subtracted (subtract_tombstones) and is applied inside the coarse shortlist. The unfiltered path had no equivalent. Surfaced via LangChain's VectorStore delete-compliance tests.

### How do we do this change?
By excluding tombstones before ranking, not after.

For each superfile we resolve its tombstone set once and hand it to the scorer as a **deny set**. A deleted row is skipped as candidates are scanned; before it can take a slot in the candidate heap. So, the top-k is filled purely from live rows. No over-fetch, no post-rank trimming, no backfill to get wrong.

## Performance

Two infino strategies measured head-to-head across `k` × delete-ratio (6K docs, 2 superfiles, dim 1024). `deny` is what this PR ships; `over-fetch` is the alternative (see Notes).

  | k | delete ratio | deny (this PR) | over-fetch |
  |---|---|---|---|
  | 1 | 0% | `15.8 ms` | `15.9 ms` |
  | 1 | 5% | `15.9 ms` | `21.4 ms` |
  | 1 | 50% | `13.2 ms` | `20.6 ms` |
  | 10 | 50% | `13.1 ms` | `20.4 ms` |
  | 100 | 50% | `13.4 ms` | `20.7 ms` |
 
  - No regression with no deletes: all paths tie (deny = None short-circuits the hot path).
  - Faster under deletes:  excluding tombstones from the coarse heap shrinks the rerank survivor pool, so there's less full-precision work; deny gets faster as deletes grow, ~1.3–1.6x ahead of over-fetch at 50%.
  - Cold object-store cost is strategy-neutral: nprobe is fixed, so all paths issue the same range GETs (123–124 GETs, ~46 MiB, within ~1%); the rerank vectors ride in those same per-cluster block fetches.

### Notes
- Over-fetch (the alternative): fetch k + tombstone_count then drop post-rank. Correct, but rerank width grows with delete volume per superfile, so it carries the warm-latency / RSS regression seen above and on bench, with no compensating cold-cost saving. Deny is exact (no margin to tune) and cheaper, so this PR supersedes that approach.
- Recall is unchanged by construction: deny only removes tombstoned rows from the candidate set; live-row ranking is identical.

### DIFF
--- a/benches/utils/bin/profile_superfile.rs
+++ b/benches/utils/bin/profile_superfile.rs
@@ -156,6 +156,7 @@ fn mean_recall_filtered(
             TOP_K,
             opts(nprobe, rerank_mult),
             Some(Arc::clone(allow)),
+            None,
         ))
         .expect("vector_hits_filtered");
         sum += corpus::recall_at_k(&hits, truth);
--- a/benches/utils/superfile.rs
+++ b/benches/utils/superfile.rs
@@ -1089,6 +1089,7 @@ pub mod vector {
                 TOP_K,
                 opts,
                 Some(Arc::clone(allow)),
+                None,
             ))
             .expect("filtered sweep query");
             sum += corpus::recall_at_k(&hits, gt);
@@ -1228,6 +1229,7 @@ pub mod vector {
                     TOP_K,
                     opts,
                     Some(Arc::clone(allow)),
+                    None,
                 ))
                 .expect("filtered latency query");
                 std::hint::black_box(hits);
@@ -1408,6 +1410,7 @@ pub mod vector {
                                 FILTERED_DEFAULT_RERANK_MULT,
                             ),
                             Some(Arc::clone(&allow)),
+                            None,
                         ))
                         .expect("filtered recall query");
                         recalls.push(corpus::recall_at_k(&hits, gt));
@@ -1495,6 +1498,7 @@ pub mod vector {
                                 TOP_K,
                                 exec_vec::search_opts(nominal_nprobe, nominal_rerank),
                                 set.clone(),
+                                None,
                             ))
                             .expect("filtered vector search");
                             samples.push(t0.elapsed());
--- a/src/superfile/reader.rs
+++ b/src/superfile/reader.rs
@@ -1120,7 +1120,7 @@ impl SuperfileReader {
         k: usize,
         options: VectorSearchOptions,
     ) -> Result<Vec<(u32, f32)>, ReadError> {
-        self.vector_hits_filtered_async(column, query, k, options, None)
+        self.vector_hits_filtered_async(column, query, k, options, None, None)
             .await
     }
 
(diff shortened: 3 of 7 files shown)


pratyushlokhande (approved):

muralikpbhat (commented):
LGTM

muralikpbhat (approved):

PR #290, opened 2026-06-26
docs: separate the Rust crate readme from the project README

ashishmishra26:
## What

The Rust crate's `readme` (Cargo.toml) and the crate-level
`#![doc = include_str!("../README.md")]` both pointed at the multi-language
project README. As a result, **docs.rs** and **crates.io** rendered Python and
Node Quickstart snippets on the Rust API reference — wrong audience for those
surfaces.

## Change

- Add `crate-docs.md` — a Rust-focused crate page: Rust quickstart + a short
  API/module overview + a pointer to the Python (`pip install infino`) and Node
  (`@infino-ai/infino`) bindings.
- Point `Cargo.toml` `readme` and `src/lib.rs` `#![doc]` at `crate-docs.md`.
- The project `README.md` is unchanged — it stays the multi-language GitHub front page.

## Notes

- The Rust quickstart in `crate-docs.md` is the existing example, kept as a live
  doctest. Verified locally with `cargo test --doc` (16 passed, 0 failed; the
  `crate-docs.md` example runs green).
- `crate-docs.md` is not in the package `exclude` list, so it ships in the crate
  tarball and is used as the crates.io readme.

### DIFF
--- a/.github/workflows/node-publish.yml
+++ b/.github/workflows/node-publish.yml
@@ -4,12 +4,17 @@ name: Node publish
 # thin `infino` package plus its per-platform binary packages (`infino-<triple>`)
 # to npm.
 #
-# Triggered two ways: (1) a `v<version>` tag pushes a real publish as part of a
-# coordinated release (crate + Node + Python) — the version comes from
-# package.json and a guard asserts it matches the tag; (2) a manual run
-# (workflow_dispatch) defaults to a dry run — pack + validate, upload nothing —
-# re-run with dry_run unchecked to publish. A real run needs the NPM_TOKEN
-# secret, with publish rights to the packages under the infino-ai org.
+# Triggered two ways: (1) a `v<version>` tag — but ONLY for a coordinated
+# minor/major release (patch == 0, e.g. `v0.2.0`); a crate-only patch tag (e.g.
+# `v0.1.1`) skips the bindings, which patch independently. The version comes from
+# package.json and a guard asserts it matches the tag. (2) a manual run
+# (workflow_dispatch) — the path for an independent Node patch — defaults to a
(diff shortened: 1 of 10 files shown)


ashishmishra26:
Added the patch version bump (`0.1.0` → `0.1.1` in `Cargo.toml` + `Cargo.lock`) so this PR is release-ready. Per `docs/versioning.md`, after merge a `v0.1.1` tag triggers the *Publish crate* workflow → crates.io (and the docs.rs rebuild). It's a Rust-only patch, so the Node/Python bindings don't move.

ashishmishra26:
Added the release-model fix on top of the readme change (per the discussion), so this PR now also:

**Gates binding publishes to coordinated minor/major tags.** After #252, `node-publish` and `publish-python` fired on every `v*` tag — which broke the independent-patch model: a crate-only patch tag (e.g. `v0.1.1`) would red-fail Node's version guard and queue an unintended Python PyPI deploy. Both binding workflows now publish on a tag push **only when patch == 0** (coordinated minor/major, tag ends in `.0`); a crate patch tag skips them cleanly. Independent binding patches still go via manual `workflow_dispatch`.

**Why patch is independent:** tying the crate's patch number to the bindings collides on immutable registries (if a binding sprinted ahead, the crate's patch can't republish that number). So `major.minor` is synced; `patch` is per-package. `docs/versioning.md` is updated to spell out the tag-driven model.

Net for the pending `v0.1.1`: it's a crate patch (`patch > 0`), so tagging `v0.1.1` publishes the **crate only** — Node/Python skip. Full PR now: (1) Rust crate readme split, (2) `0.1.1` bump, (3) coordinated-release tag gating.

pratyushlokhande (approved):

Issue #286, opened 2026-06-25
cpu: consolidate tokio runtimes and route all rayon work through controlled pools


savannahar68:
### Why

Infino has a clear concurrency contract: tokio owns I/O waves, rayon owns CPU waves. The implementation has grown gaps that violate it. On a machine with N logical cores and K open supertables, the actual thread count today is:

* reader_pool: N threads     (all cores, one pool shared by all supertables)
* writer_pool:  N/2 threads   (half cores, shared)
* tokio runtimes: N × (1 + K) threads  
   - 1 per Connection  (catalog-query)  catalog/mod.rs:491
   - 1 per Supertable  (supertable-query)  handle.rs:187
 each scaled to available_parallelism()

With two supertables on a 16-core host: 16 + 8 + 16×3 = 72 threads fighting for 16 cores. That's before considering the global rayon pool also leaking in

This results in build-time CPU competes with query-time isolation. A commit and a concurrent warm vectorquery share the global pool with no priority or bound.

savannahar68:
There are other issues alongside which ill keep attaching to this issue, create subtasks and tackle one by one

savannahar68:
CPU thrash confirmed. Here's what the profile shows:

59 threads competing for 10 cores. All threads are equally saturated (29-30k samples each) — nobody is blocked or waiting. That's the oversubscription signature.

No lock/IO contention. Zero futex, mutex, park, or semaphore frames in the top 50 hottest functions. If it were IO or lock contention, you'd see threads parked in futex_wait or pthread_mutex_lock.

High context-switch overhead. libsystem_pthread shows up heavily — the OS is constantly switching between 59 runnable threads on 10 cores.

Where the ~59 threads come from (1 table):
```
┌────────────────────────────────────────────────────────────────┬─────────┐
│                              Pool                              │ Threads │
├────────────────────────────────────────────────────────────────┼─────────┤
│ sim runtime workers (10 base + ~9 block_in_place replacements) │ ~19     │
├────────────────────────────────────────────────────────────────┼─────────┤
│ Supertable internal query_runtime                              │ 10      │
├────────────────────────────────────────────────────────────────┼─────────┤
│ reader_pool rayon                                              │ 10      │
├────────────────────────────────────────────────────────────────┼─────────┤
│ writer_pool rayon                                              │ 5       │
├────────────────────────────────────────────────────────────────┼─────────┤
│ Global rayon pool                                              │ 10      │
[...]

PR #294, opened 2026-06-26
chore: concurrent bench to find cpu contention

savannahar68:
### What is the change about

The change adds a harness for figuring out cpu contention in the library. This will help us have a better setup for cpu pooling.
This is a setup as part of #286 

This PR doesn't adds vector search in the mix, the next PR will. Vector search will help us profile how rayon and tokio are working with each other as for a given vector query we spawn rayon tasks equal to number of cpu.

### What it does

* It ingest 20k docs
* Collect stats
   - 8 readers no writer
   - 8 reader 1 writer (ingesting some data but no commit)
* Compares the numbers across both and against previous runs.

### How it helps?

In next set of PRs we would do profiling using tokio-concole and take task dump to figure out issues around cpu contention. This will also help us identify if sync-async bridge build is working as intended and interoperatability between tokio and rayon as they would be figuring for CPU.

### How to run

```
cargo bench -- concurrent
   Compiling infino v0.1.1 (/Users/savan/code/infino-ai/infino)
   Compiling infino-bench-utils v0.1.0 (/Users/savan/code/infino-ai/infino/benches/utils)
    Building [=======================> ] 498/499: bench(bench)
    Finished `bench` profile [optimized + debuginfo] target(s) in 2m 21s
     Running benches/bench/main.rs (target/release/deps/bench-0851075c4a1b4deb)
[bench] === concurrent ===
[concurrent] 1 table(s), 200000 docs/8 superfiles, 8 reader tasks, 15s window (3s warmup discarded)
[concurrent] table 0: baseline (8 readers, no writer)...
[concurrent] table 0: contention (8r+1w)...
[concurrent] table 0: baseline 19578 q/s | contention 5210 q/s (-73.4%) | writer 20.5 commits/s | p99 +336.5%

══ Concurrent ingest+query — 1 table(s), 200000 docs, 8 readers, 15s window ══
Host: unknown CPU · 10C/10T · macos/aarch64
| Table        | Condition | p50                      | p95                      | p99                      | q/s                               | n / commits                       |
| ------------ | --------- | ------------------------ | ------------------------ | ------------------------ | --------------------------------- | --------------------------------- |
| single table | baseline  | 389.54 µs  -72.1% better | 649.08 µs  -74.3% better | 903.67 µs  -77.6% better |         19578 q/s  +274.0% better |                            234941 |
| single table |    8r+1w  |   1.42 ms  +2.0% ~       |   2.48 ms  -1.7% ~       |   3.94 ms  -2.2% ~       | 5210 q/s (-73.4%)  -0.5% ~        | 62524 / 246 commits (p99 +336.5%) |
```

This stores json in target/infino-bench/concurrent.json which becomes a baseline and comparision number for next set of runs which helps us do comparision.

### DIFF
--- a/benches/utils/concurrent.rs
+++ b/benches/utils/concurrent.rs
@@ -1,0 +1,391 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Infino Authors
+
+//! Concurrent ingest + query contention harness.
+//!
+//! Measures sustained reader latency and throughput under two conditions:
+//! - **baseline**: N readers firing queries in a tight loop, no writer.
+//! - **contention**: same N readers + 1 writer committing continuously.
+//!
+//! **Duration-based** (not iteration-based): each condition runs for a fixed
+//! wall-clock window (default 15 s; 3 s warmup discarded). Every query that
+//! completes during the measurement window is recorded. This guarantees
+//! sustained overlap between readers and the writer — the failure mode of
+//! iteration-based benches is that readers finish in milliseconds before the
+//! writer commits even once.
+//!
+//! Runs on a `multi_thread` tokio runtime so `bridge_sync_to_async` takes
+//! the `block_in_place` path — the same code path as axum/SaaS production.
+//!
+//! Knobs (env vars):
+//!   INFINO_BENCH_CONCURRENT_DOCS      corpus size (default 200_000)
+//!   INFINO_BENCH_CONCURRENT_READERS   concurrent reader tasks (default 8)
+//!   INFINO_BENCH_CONCURRENT_TENANTS   tables open simultaneously (default 1)
+//!   INFINO_BENCH_CONCURRENT_DURATION  measurement window in seconds (default 15)
+//!   INFINO_BENCH_CONCURRENT_WARMUP    warmup seconds to discard (default 3)
+//!
+//! Invoked as `cargo bench -- concurrent`.
+
+use std::{
+    hint::black_box,
+    sync::{
+        Arc,
+        atomic::{AtomicBool, Ordering},
+    },
+    time::{Duration, Instant},
+};
+
+use futures::future::join_all;
+use infino::{
+    storage::{LocalFsStorageProvider, StorageProvider},
+    superfile::fts::reader::BoolMode,
+    supertable::Supertable,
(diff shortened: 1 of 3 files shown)


benches/utils/concurrent.rs
```diff
@@ -0,0 +1,391 @@
...
+    while !stop.load(Ordering::Relaxed) {
+        if let Ok(mut w) = st.writer() {
+            let start = batch_start;
+            let titles_owned: Vec<String> = (start..start + WRITER_BATCH)
+                .map(|i| format!("alpha extra{i:08}"))
+                .collect();
+            let titles: Vec<&str> = titles_owned.iter().map(|s| s.as_str()).collect();
+            let batch = build_title_batch(&titles);
+            let _ = w.append(&batch);
+            let _ = w.commit();
```
Pranav2612000, line 214: Thoughts on using supertable api ( instead of writer )? The same api that users will use

Pranav2612000: If this is a suprtable, then do we need to call commit? `append` already calls commit

savannahar68: this is superTableWriter append which doesn't do commit so we have to do commit
Commit helps us conflict rayon's (reader and writer pool) so we see contention here

savannahar68:
| Condition              | p50      | p99      | q/s    | delta       |
|------------------------|----------|----------|--------|-------------|
| 8 readers, no writer   | 339 µs   | 667 µs   | 22,899 | baseline    |
| 8 readers + 1 writer   | 391 µs   | 4,030 µs | 5,235  | -77%, +510% |
| 8 readers + append only (no commit) | 391 µs | 797 µs | 19,629 | -14%, +20% |

### What this tells us

* append() alone barely affects readers it's just buffering rows in memory.
* commit() is the problem. It builds a Parquet file and an FTS index using a rayon thread pool sized to N cores (writer_pool). At the same time, readers are scoring BM25 queries using a separate rayon pool also sized to N cores (reader_pool). Both pools think they own all the CPU — so under concurrent ingest + query we have roughly 2× as many rayon threads as cores, causing thrash.


This is with a single table. In a multi-tenant process with M tables open, each table gets its own pools the thread count multiplies by M.

### What comes next
Instrument thread counts and add a flamegraph run to confirm CPU vs I/O causing CPU thrash

Pranav2612000 (approved):

PR #296, opened 2026-06-27
fts: faster multi-term OR count via block-at-a-time disjunction cardinality

vinaykakade:
## What

Speeds up the matching-doc **count** for multi-term OR queries, with supporting bench coverage and a small serving-path win.

### Multi-term OR count — block-at-a-time disjunction cardinality

Multi-term OR count used a per-doc k-way merge: for every doc in the union it rescanned all cursors to find the minimum and again to advance. Cost scaled with `union_size × term_count`, degrading super-linearly on long common-term unions.

This replaces it with a block-at-a-time disjunction count — walk the cursors one fixed doc-id window at a time, mark each matching doc in a small presence bitset, and accumulate the per-window popcount. Windows partition the doc-id space disjointly, so a doc matching several terms is counted once and nothing is double-counted across windows; the tally equals the distinct-doc union size. No scoring, no top-k heap, no doc-id list materialized. The per-doc k-way merge stays for the path that genuinely needs ordered ids.

**Numbers (1M-doc single superfile, warm, OR-count p50):**

| terms | before | after | speedup |
|---|---|---|---|
| 2 | 5.50 ms | 2.46 ms | 2.2× |
| 5 | 15.82 ms | 2.96 ms | 5.3× |
| 10 | 33.52 ms | 5.71 ms | 5.9× |
| 20 | 65.40 ms | 10.65 ms | 6.1× |
| 40 | 118.39 ms | 18.76 ms | 6.3× |

The scaling flattens from linear-in-terms to decode-bound. The win carries to the multi-superfile object-store path (10-term OR count −85% at 1M docs on the Azure bench). Single-term count (df fast path) and AND count are unchanged.

### Single-superfile fan-out runs inline

The per-superfile query fan-out spawned a `tokio` task even for a single work unit. With one unit there's no concurrency to gain — the spawn + join is pure overhead — so it now runs the body inline. Helps every query (count and search) on a compacted, single-superfile table.

### Bench coverage

- Added 20- and 40-term OR shapes to the count battery (the large-union shapes that degraded most), on both the superfile and supertable tiers.
- Added a top-1000 query-phase gate (representative shapes) on both tiers, so large-`k` collection cost is tracked going forward.
- The superfile count battery now measures the dedicated count primitive — it had been measuring the ordered-merge path, so it never exercised the count fast path; now consistent with the supertable tier.
- Trimmed the superfile large-`k`/cold cost to keep the bench within its time budget.

## Tests

- Extended the count-vs-merge equivalence test and added a multi-window count test (corpus spans multiple windows; checks agreement with the merge and an exact known cardinality).
- Full FTS reader unit suite, the supertable integration suite (198 tests, over LocalFs), `cargo fmt`, and `cargo clippy` all green. All four Azure bench legs pass.

### DIFF
--- a/benches/utils/executors.rs
+++ b/benches/utils/executors.rs
@@ -103,6 +103,78 @@ pub mod fts {
     /// Nanoseconds per second, for time-cell formatting.
     const NS_PER_SEC: f64 = 1e9;
 
+    /// Twenty mid-rank common terms (`term00050`..`term00069`) — a dense
+    /// disjunction whose match set covers a large fraction of the corpus.
+    /// Exercises the large-union count path, where a naive per-doc k-way
+    /// merge degrades super-linearly in the term count.
+    const TWENTY_COMMON_TERMS: &[&str] = &[
+        "term00050",
+        "term00051",
+        "term00052",
+        "term00053",
+        "term00054",
+        "term00055",
+        "term00056",
+        "term00057",
+        "term00058",
+        "term00059",
+        "term00060",
+        "term00061",
+        "term00062",
+        "term00063",
+        "term00064",
+        "term00065",
+        "term00066",
+        "term00067",
+        "term00068",
+        "term00069",
+    ];
+
+    /// Forty mid-rank common terms (`term00050`..`term00089`) — the extreme
+    /// large-union shape; the count path's worst case at this scale.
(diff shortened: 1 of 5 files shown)


naman-modi (approved):
LGTM.

PR #303, opened 2026-06-29
examples: retry transient HuggingFace Hub errors when loading datasets

pratyushlokhande:
## What

Wrap `datasets.load_dataset` in the shared example loaders with tenacity retry/backoff.

## Why

CI ([run #28363283248](https://github.com/infino-ai/infino/actions/runs/28363283248/job/84022862125)) failed loading `microsoft/ms_marco` with `HfHubHTTPError: 504 Gateway Time-out` while the Hub resolved the dataset file tree. Transient Hub-side fault, not our code. All six loaders route through `load_dataset`, so one wrapper covers them.

## Details

- `load_dataset` now retries `HfHubHTTPError` / `requests` `RequestException`, 3 attempts, exponential backoff capped at 32s, `reraise=True` so the original error surfaces on final failure.
- Added `tenacity>=8` to `examples/requirements.txt` (examples-only dep, not the shipped crate).

### DIFF
--- a/infino-python/examples/_shared/loaders.py
+++ b/infino-python/examples/_shared/loaders.py
@@ -10,7 +10,18 @@ import os
 os.environ.setdefault("HF_HUB_DISABLE_PROGRESS_BARS", "1")
 os.environ.setdefault("HF_HUB_VERBOSITY", "error")
 
-from datasets import load_dataset
+from datasets import load_dataset as _hf_load_dataset
+from huggingface_hub.utils import HfHubHTTPError
+from requests.exceptions import RequestException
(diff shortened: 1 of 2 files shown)


naman-modi (commented):
Looks good. Can you please recheck other examples too, and make sure we don't face similar issues?

pratyushlokhande:
> Looks good. Can you please recheck other examples too, and make sure we don't face similar issues?

Yes, All the examples use the same `load_dataset` method. So this will make sure such transient errors are caught and retried thrice.

naman-modi (approved):

PR #308, opened 2026-06-29
storage: cloud-agnostic config-driven credentials + fail-fast connect

pratyushlokhande:
## What

Make object-store credentials cloud-agnostic and config-driven across the core crate and both language bindings.

- **`storage_options` map** — credentials/tuning are passed as a flat `HashMap<String, String>` keyed by `object_store`'s config strings (`aws_*` / `azure_*`), the de-facto standard shared by delta-rs / Polars / LanceDB. One generic helper (`storage::options::apply`) folds the map onto any backend's builder; a new cloud backend reuses it with no new public API.
- **No environment reads** — the S3 and Azure providers no longer call `from_env()`. Credentials come from config; ambient cloud identity (IAM role / managed identity) still resolves inside `object_store` at request time when none are supplied.
- **Opt-in connect probe** — `ConnectOptions::with_validate(true)` (default `false`) probes the object store at `connect` (reusing the catalog read), so bad credentials or an unreachable bucket fail at `connect` instead of on the first table operation. Off by default, so `connect` does no I/O unless asked.
- **One credential path** — `storage_options` is the only way to pass credentials. The earlier S3-endpoint shorthand (`with_s3_endpoint` / the bindings' `endpoint`/`region`/`access_key`/`secret_key` kwargs) is removed; a custom endpoint is just `aws_endpoint` (+ `aws_allow_http`).

## Why

Azure credentials previously had no programmatic path (env-only), and each backend was a separate special-case in `ConnectOptions`. This unifies credential passing, removes the env coupling, and removes the redundant S3-only shim.

## API surface

- **Added:** `ConnectOptions::with_storage_option`, `ConnectOptions::with_validate`.
- **Removed:** `ConnectOptions::with_s3_endpoint` (superseded by `with_storage_option`).
- **Bindings:** Python `connect(storage_options=..., validate=...)`; Node `connect(uri, { storageOptions, validate })`. The S3-endpoint kwargs are dropped (bindings are pre-1.0).
- `public-api.txt` updated accordingly.

## Keys

Standard `object_store` strings (the names the AWS/Azure SDKs use):
- S3: `aws_access_key_id`, `aws_secret_access_key`, `aws_region`, `aws_session_token`, `aws_endpoint`
- Azure: `azure_storage_account_name`, `azure_storage_account_key`, `azure_storage_sas_key`, `azure_storage_client_id`, `azure_storage_client_secret`, `azure_storage_tenant_id`

Custom S3 endpoint is detected by config-key *variant* (`Endpoint`/`S3Endpoint`), robust to every alias — not a literal string. An unknown or cross-backend key is rejected at connect.

## Docs

Documented consistently in every place a user looks: README "Cloud storage" section (URI table + key table + examples), crate docs, architecture overview, the shipped `config.yaml` (`storage.storage_options`), and both binding READMEs.

## Testing

- `make ci`: fmt, clippy, tests, doctests, public-api — green. Bindings: fmt + clippy + `mypy --strict` (Python) + `tsc` (Node).
- Rust: unit tests for option apply/validation, endpoint-alias detection, and the default-off probe.
- Python: smoke (`storage_options` apply, unknown-key reject, default-off) + **gated real-Azure e2e (6/6)**.
- Node: smoke (built) + **gated real-Azure e2e (6/6)**, mirroring Python.
- **CI:** a non-required `object-store-e2e` job runs the bindings' e2e against the real Azure container (self-skips on fork PRs where secrets are absent), with a run-scoped prefix + best-effort blob cleanup.
- Benches reviewed + compile clean; they don't use the `connect` path, so no impact.

## Performance

Off the hot path — provider construction + an optional connect-time read; no change to scoring / distance / quantization / commit / reader cache. No regression expected; the full `make bench` perf gate can run against `INFINO_BENCH_STORE=azure`.


benches/utils/tiers.rs
```diff
@@ -211,6 +212,38 @@ fn azure_prefix_root(default: &str) -> String {
...
+    ])
+}
+
+/// Standard Azure credential options from the environment.
+pub fn azure_storage_options_from_env() -> HashMap<String, String> {
+    storage_options_from_env(&[
+        ("AZURE_STORAGE_ACCOUNT_NAME", "azure_storage_account_name"),
+        ("AZURE_STORAGE_ACCOUNT_KEY", "azure_storage_account_key"),
+    ])
+}
```
Pranav2612000: Nit: Is `tiers` the best file for defining these functions

infino-node/README.md
```diff
@@ -226,24 +226,43 @@ docs.optimize({ targetSuperfileSizeMb: 256, minFillPercent: 50 });
...
+    azure_storage_account_key: "…",
+  },
 });
 ```
 
+`connect` probes the backend before returning, so wrong credentials or an
+unreachable bucket throw immediately instead of failing on the first query. An
+unknown key is rejected at `connect`. For an S3-compatible endpoint (MinIO / R2
+/ Ceph), `endpoint` / `region` / `accessKey` / `secretKey` remain as a shorthand
+for the matching `aws_*` options.
```
Pranav2612000: Like in the python readme, can we also include a list of all the possible keys that can be passed

src/catalog/mod.rs
```diff
@@ -92,6 +95,8 @@ pub fn connect_with(
         _ => {
             let root = backend_to_provider(&backend, &options)?
                 .expect("non-memory backend yields a storage provider");
+            // Probe now so bad credentials fail at connect, not first use.
+            bridge_sync_to_async(read_catalog(root.as_ref()))?;
```
Pranav2612000: Let's discuss if we want to perform these checks during connect.

benches/utils/tiers.rs
pratyushlokhande: Good call — moved them to a dedicated `benches/utils/storage_options.rs` module (`tiers` was the wrong home for credential gathering). 8c64680

infino-node/README.md
pratyushlokhande: Done — added the per-backend key table, matching the Python README. 57bb8b8

src/catalog/mod.rs
pratyushlokhande: Made it opt-in: the probe now defaults **off**, enabled via `ConnectOptions::with_validate(true)` (and `validate` in the Python/Node bindings). So `connect` does no I/O by default; callers opt into fail-fast. 978a3b9 + binding commits.

Pranav2612000 (approved):

ashishmishra26 (approved):

PR #315, opened 2026-06-30
storage: rotate static credentials on a live connection (no reconnect)

pratyushlokhande:
## What

Add `Connection::rotate_credentials(storage_options)` (Python `rotate_credentials`, Node `rotateCredentials`) to swap a rotated **static** object-store key into a **live** connection — and every table opened from it — without dropping the connection or restarting the worker. The connection pool stays warm.

## Why

Static keys (Azure account key, AWS access/secret/session) get rotated on a schedule. Today a long-lived worker holds a `Connection` whose credentials are baked in at `connect` time; picking up a new key means tearing down and reconnecting (cold pools, re-plumbing every handle). This lets the worker hand infino the new key and keep serving.

## How

`object_store` queries a `CredentialProvider` per request. We inject a custom provider backed by `ArcSwap` instead of baking a static key via `with_config`:

- `RotatingCredentialProvider<T>` (storage/credentials.rs) — `get_credential` reads the current value lock-free; `store` publishes a new one.
- `BackendCredentials` — a connection-scoped cell built **once** at `connect`. The **same `Arc`** is threaded (via `with_credentials`) into the root provider and every per-table provider, so one `rotate_credentials` swap propagates to all current and future tables on their next request.
- The credential keys are kept out of `with_config` (so a static key can't shadow the rotating provider); region/endpoint still apply normally.

`rotate_credentials` is a pure in-memory swap — no I/O, no reconnect. It parses the new credential from the passed options (other keys ignored), validates it (e.g. Azure base64), and atomically swaps; a bad key leaves the old one live.

## Scope

- **Rotatable:** static keys — AWS `access_key_id`/`secret_access_key`/`session_token`, Azure `account_key`.
- **Not rotated (object_store self-refreshes):** IAM role / STS / web-identity, Azure service principal / managed identity. `rotate_credentials` on an ambient-identity connection errors (nothing static to swap).
- **Not rotatable:** endpoint / region / bucket — changing those still needs a reconnect (documented).
- SAS / bearer-token rotation deferred (no caller yet); account-key + AWS static cover the ask.

## API surface

Additive. `public-api.txt`: `+Connection::rotate_credentials`. The credential types are `pub` only behind the `test-helpers` feature (off the shipped surface), matching the existing storage-provider pattern.

## Testing

- Rust unit: ArcSwap swap, credential extraction/validation, partial-cred error.
- Rust catalog: `rotate_credentials` swaps on an S3 connection; errors on `memory://`.
- **Gated real-Azure e2e (Python + Node, 7/7 each):** rotate a *wrong* key into a live connection → the next write **fails** (proving the swap reached the open table, no reconnect) → rotate the good key back → it works. Runs in the `object-store-e2e` CI job.
- Bindings smoke (ungated): rotate without credentials → error.
- `make ci` green (fmt, clippy `-D warnings`, tests, doctests, public-api); bindings clippy + `mypy --strict` + `tsc` clean.

## Performance

Off the hot path — credential plumbing only; no change to scoring / distance / quantization / commit / reader cache. `rotate_credentials` is an `ArcSwap` store.


Pranav2612000 (commented):
Instead of these changes, where we also add credentials to a `Connection`. Can we expose and api to update just the `storage_options`

pratyushlokhande:
@Pranav2612000 good push — I dug into whether a generic `update_storage_options` is possible, and it hinges on one `object_store` constraint:

**A store bakes its config at build time — only credentials are live-swappable.** Once a store is built with `with_config(region / endpoint / bucket / …)`, those are fixed; they can't change without rebuilding the store. Credentials are the sole exception, because `with_credentials` takes a *provider* that's polled on every request (that's exactly what the `ArcSwap`-backed provider here exploits).

So a generic `update_storage_options(map)` is feasible, but only the **credential keys** in that map would take effect on a live connection — `region` / `endpoint` / backend changes would silently no-op until reconnect. I want to avoid an API that looks like it updates everything but quietly only updates creds.

Given that, three ways forward — which do you prefer?

1. **Reshape to `update_storage_options(map)`, honestly scoped.** Drop the typed `BackendCredentials` from `Connection`; store a shared `ArcSwap<storage_options>` instead, and have the credential provider read creds out of it per request. Keeps live rotation, matches your mental model, less type surface. Documented clearly: only credential keys apply live; other keys need reconnect.

2. **Keep `rotate_credentials`, just narrower/renamed.** Same as today but framed explicitly as credentials-only, so the API never implies it can change region/endpoint.

3. **Drop the feature entirely, document reconnect.** `connect()` again with the new `storage_options` and re-open tables — no new API, smallest surface. Trade-off: table handles opened before the swap keep the old key until re-opened (in-place rotation is what lets already-open handles survive transparently).

My lean is (1) if we want open handles to survive a rotation transparently, or (3) if callers always go through `open_table` and re-opening is cheap for them. Happy to go whichever way you think fits the usage better.

pratyushlokhande:
@Pranav2612000 went with **option 1** — pushed. `Connection` no longer carries a typed credentials concept; it holds one `SharedStorageOptions` cell (`ArcSwap<StorageOptions>`) shared into every provider, and the API is now `update_storage_options(map)`.

What changed vs the first version:
- **No `BackendCredentials` on `Connection`.** Just the options cell. Credential providers re-read it per request (one generic `OptionsCredentialProvider<T>` over an extractor, replacing the two-variant enum).
- **`rotate_credentials` → `update_storage_options(map)`**, merge semantics: the given `aws_*` / `azure_*` keys overlay the existing options.
- **`new_with_prefix`** now takes `Option<AwsCredentialProvider>` / `Option<AzureCredentialProvider>` (object_store's own types) instead of a crate-internal type, so the `credentials` module is back to `pub(crate)`.

Behavior, stated honestly given the object_store constraint (only credentials are live-swappable; `with_config` bakes region/endpoint at build):
- **Credential keys** take effect immediately on the connection and its already-open tables — no reconnect.
- **Other keys** (region, endpoint) apply to tables opened *after* the update; changing them on open tables still needs a reconnect. Documented on the method.
- Errors on non-object-store backends (`memory://`, local fs) and on a malformed merged credential.

Gates green: `make check`, `make public-api` (one-line diff — the rename), rust storage/catalog tests, node smoke 22/22, both binding crates fmt-clean. Public API surface change is just `Connection::update_storage_options`.

Pranav2612000 (approved):

PR #316, opened 2026-06-30
[1/n] feat memory: per-connection memory budget primitive (counter + reservation)

naman-modi:
### What does this PR do?
Adds the per-connection memory budget: a counter (`ConnectionMemoryBudget`), a reservation guard (`Reservation`), and an `OverBudget` error. It's just the building block.

A query's heap grows with the data and the query: result batches, the vector shortlist, ingest buffers. With no limit, one connection can eat memory until the process is OOM-killed, taking every other connection down with it. We'd rather refuse that one query and keep running. Nothing today lets a connection say "no, this is too much."

### How do we do this change?
- A caller asks the budget for `n` bytes before it allocates them. If they fit it gets a guard, and the guard hands the bytes back when it's dropped.
- Two modes: **measured** (default, just counts usage, never refuses) and **bounded** (refuses once usage would cross the limit).
- A bounded budget enforces at 90% of the configured value; the spare 10% covers the small allocations we don't bother tracking.
- One budget per connection, shared, so what's capped is the total it's holding at once, not any single allocation.

### Performance
Nothing calls it yet and the default only measures, so no behaviour change. The path allocates nothing and takes no locks. Concurrency is tested: 32 threads racing a bounded budget land exactly on the limit and no more, and heavy reserve/release churn balances back to zero.

### Notes
- Default is measure-only, so zero change until a limit is set and the call sites land.
- `allow(dead_code)` is temporary, removed as the call sites are wired in.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -126,6 +126,7 @@ pub use roaring;
 // public items are re-exported at the crate root below.
 mod catalog;
 mod error;
+mod memory;
 mod runtime_bridge;
 
 // ---- Curated public surface ----
--- a/src/memory/mod.rs
+++ b/src/memory/mod.rs
@@ -1,0 +1,464 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Infino Authors
+
+//! Per-connection memory budget.
+//!
+//! Infino keeps data on object storage. To answer a query it pulls the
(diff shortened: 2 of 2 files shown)


vinaykakade (approved):
LGTM. When you get to the implementation, pl make sure that there are no per-query reservation calls in the hot paths (query or ingest).

savannahar68 (commented):
lgtm

PR #333, opened 2026-07-01
catalog: preserve schema for zero-row query_sql results

### LINKED ISSUE #326: query_sql() on a zero-row result raises 'Must pass schema, or at least one RecordBatch'
**Context:** building an app on the 0.1.2 Python wheel. A faceted search took down the **entire** facet response because one facet field (an all-empty column) produced a zero-row `GROUP BY`.

**Bug:** `query_sql()` on a result with **zero rows** returns an object whose `.to_pylist()` *and* `.num_rows` both raise, so an empty result can't be inspected or converted:

```python
db.query_sql("SELECT idx FROM jobs WHERE idx = -1").to_pylist()
# ValueError: Must pass schema, or at least one RecordBatch
db.query_sql("SELECT idx FROM jobs WHERE idx = -1").num_rows
# ValueError: Must pass schema, or at least one RecordBatch
```

Non-empty queries are fine. It looks like the empty result is materialized as a pyarrow Table from zero batches **without carrying the projected schema**, so pyarrow can't build it.

**Impact:** any query that legitimately matches nothing crashes the caller — a very common case (selective filter, absent term, empty facet column). Callers must wrap every `query_sql` in try/except on the message string, which is brittle.

**Expected:** a zero-row result should be an empty Arrow table **with the projected schema**, so `.num_rows == 0` and `.to_pylist() == []`.

adityasd314:
## What

`query_sql()` raises `ValueError: Must pass schema, or at least one RecordBatch` when the query matches zero rows. Calling `.num_rows` or `.to_pylist()` on the result both fail, making any legitimately-empty result (selective filter, absent term, empty facet column) crash the caller.

## Why 

`df.collect()` returns an empty `Vec<RecordBatch>` when no rows match. The schema is not carried on the `Vec` itself, so the Python binding's `pyarrow.Table.from_batches([])` call has no schema to build from and raises.

## How

Capture the output schema from the `DataFrame` before consuming it with `collect()`. If the collected batches are empty, return a single `RecordBatch::new_empty(schema)` so callers always receive a well-formed result with the projected schema intact.

## Testing

Added a Rust unit test (`query_sql_zero_row_result_carries_projected_schema`) and a Python integration test (`test_query_sql_zero_row_result_returns_empty_table_with_schema`).

## Performance
Off the hot path. Schema cloning is a simple O(1) clone, and `RecordBatch::new_empty` is only called on the empty query results path.

---
Fixes #326

### DIFF
--- a/infino-python/tests/test_smoke.py
+++ b/infino-python/tests/test_smoke.py
@@ -119,6 +119,18 @@ def test_query_sql_returns_pyarrow_table():
     assert out.column("n")[0].as_py() == 3
 
 
+def test_query_sql_zero_row_result_returns_empty_table_with_schema():
+    # A zero-row result must be an empty table with the projected schema.
+    db = infino.connect("memory://")
+    table = db.create_table("docs", _title_schema(), infino.IndexSpec().fts("title"))
+    table.append(_title_batch(["alpha", "beta"]))
+
+    out = db.query_sql("SELECT title FROM docs WHERE title = 'no_match'")
+    assert out.num_rows == 0
+    assert out.to_pylist() == []
+    assert out.schema.get_field_index("title") != -1
+
+
 def test_query_sql_bm25_tvf():
     db = infino.connect("memory://")
     table = db.create_table("docs", _title_schema(), infino.IndexSpec().fts("title"))
--- a/infino-python/tests/test_smoke.py
+++ b/infino-python/tests/test_smoke.py
@@ -119,16 +119,20 @@ def test_query_sql_returns_pyarrow_table():
     assert out.column("n")[0].as_py() == 3
 
 
-def test_query_sql_zero_row_result_returns_empty_table_with_schema():
-    # A zero-row result must be an empty table with the projected schema.
+def test_query_sql_zero_row_filter_preserves_projected_schema():
     db = infino.connect("memory://")
     table = db.create_table("docs", _title_schema(), infino.IndexSpec().fts("title"))
     table.append(_title_batch(["alpha", "beta"]))
 
+    # Ground-truth schema from a query that returns rows.
+    with_rows = db.query_sql("SELECT title FROM docs")
+    expected_schema = with_rows.schema
+
+    # Zero-row result must carry the identical schema.
     out = db.query_sql("SELECT title FROM docs WHERE title = 'no_match'")
     assert out.num_rows == 0
     assert out.to_pylist() == []
-    assert out.schema.get_field_index("title") != -1
+    assert out.schema == expected_schema
 
 
 def test_query_sql_bm25_tvf():
--- a/src/catalog/mod.rs
+++ b/src/catalog/mod.rs
@@ -502,7 +502,15 @@ impl Connection {
                 .sql(&sql)
                 .await
                 .map_err(|e| InfinoError::Query(e.to_string()))?;
-            df.collect().await.map_err(sql_exec_error)
+            // Preserve the output schema so a zero-row result is a well-formed
+            // empty batch rather than a schema-less Vec.
+            let output_schema: SchemaRef = df.schema().inner().clone();
+            let batches = df.collect().await.map_err(sql_exec_error)?;
+            if batches.is_empty() {
+                Ok(vec![RecordBatch::new_empty(output_schema)])
+            } else {
+                Ok(batches)
+            }
         };
         // A query that names a `FROM` catalog table drives on that table's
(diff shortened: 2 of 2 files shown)


adityasd314:
@naman-modi Can you please review it ?

src/catalog/mod.rs
```diff
@@ -476,9 +476,18 @@ impl Connection {
                 .sql(&sql)
                 .await
                 .map_err(|e| InfinoError::Query(e.to_string()))?;
-            df.collect()
+            // Preserve the output schema so a zero-row result is a well-formed
+            // empty batch rather than a schema-less Vec.
```
naman-modi: A comment like this might be more helpful to understand:

```
// A RecordBatch carries the schema; an empty Vec does not. Capture the output schema before collect()
// consumes the DataFrame so a zero-row result returns one empty batch with the projected schema, 
// instead of a schema-less Vec - which the Python binding's Table.from_batches([]) can't build.
```

src/catalog/mod.rs
```diff
@@ -859,6 +868,35 @@ mod tests {
...
+        let total_rows: usize = batches.iter().map(|b| b.num_rows()).sum();
+        assert_eq!(total_rows, 0, "no rows should match");
+        assert!(
+            batches[0]
+                .schema()
+                .fields()
+                .iter()
+                .any(|f| f.name() == "title"),
+            "projected schema must include 'title'"
+        );
```
naman-modi: To test the multi-column projection, can you select more columns and assert their schema too?

src/catalog/mod.rs
```diff
@@ -859,6 +868,35 @@ mod tests {
...
+        let conn = connect("memory://").expect("connect");
+        let docs = conn
+            .create_table("docs", schema_id_title(), IndexSpec::new().fts("title"))
+            .expect("create docs");
+        docs.append(&build_title_batch(&["alpha", "beta"]))
+            .expect("append");
+
+        let batches = conn
+            .query_sql("SELECT title FROM docs WHERE title = 'no_match'")
+            .expect("zero-row query must not error");
```
naman-modi, line 1155: Can we also add a few more tests with other operators like `GROUP_BY` etc?

infino-python/tests/test_smoke.py
```diff
@@ -118,6 +118,18 @@ def test_query_sql_returns_pyarrow_table():
...
+def test_query_sql_zero_row_result_returns_empty_table_with_schema():
+    # A zero-row result must be an empty table with the projected schema.
+    db = infino.connect("memory://")
+    table = db.create_table("docs", _title_schema(), infino.IndexSpec().fts("title"))
+    table.append(_title_batch(["alpha", "beta"]))
+
+    out = db.query_sql("SELECT title FROM docs WHERE title = 'no_match'")
+    assert out.num_rows == 0
+    assert out.to_pylist() == []
+    assert out.schema.get_field_index("title") != -1
```
naman-modi: We want to ideally assert the entire schema, so a better way would be to also run a query which returns results, and then assert the entire `schema` with this query which returns 0 results.

naman-modi (changes requested):

naman-modi:
Please rebase, and solve merge conflicts too.

src/catalog/mod.rs
adityasd314: understood!

src/catalog/mod.rs
adityasd314: yes, I'll add checks for multi-column projects

src/catalog/mod.rs
adityasd314: yes

infino-python/tests/test_smoke.py
adityasd314: understood!

src/catalog/mod.rs
adityasd314: Added `query_sql_zero_row_group_by_preserves_projected_schema` to verify this exact DataFusion operator path.

src/catalog/mod.rs
adityasd314: Done! Updated the query to `SELECT _id, title` and asserted strict schema equality against a query returning non-empty result.

adityasd314 (commented):
We're now strictly asserting schema equality across filters, aggregations (GROUP BY). Let me know if anything else is needed!

naman-modi (approved):

naman-modi:
Thanks @adityasd314!

Issue #325, opened 2026-07-01
memory: per-connection memory budget (config + public API + bindings)


naman-modi:
## Problem

A query allocates heap that grows with the data and the query: result batches, the vector shortlist, ingest buffers. Nothing today lets a caller bound it, so one connection can grow the process heap until the OS OOM-kills the process, taking every other connection in it down too. We want a per-connection ceiling so an oversized query fails cleanly instead of killing the process.

## Approach

The core primitive (`ConnectionMemoryBudget`: a per-connection byte counter, measure-only by default, bounded when a limit is set) lands separately. This issue exposes and wires it:

- `ConnectOptions::with_connection_memory_budget_bytes(u64)`, a per-connection override.
- A `memory.budget_bytes` default in `config.yaml`, read at `connect`. Precedence: an explicit `ConnectOptions` value wins (including `0` for measure-only), otherwise the config default, otherwise measure-only. Minted once per `Connection` and shared (cloned `Arc`) across all its tables.
- Expose the knob on the Python (`infino-python`) and Node (`infino-node`) `connect` APIs, plus their config surface.

Default is `0` (measure-only): usage is tracked but never refused, so there is no behavior change until a budget is set. Enforcement at the allocation sites (SQL, vector, ingest) is separate follow-up work; this issue covers the knob, the config surface, and the bindings.

## Public API

New public item on the Rust surface:

    pub fn infino::ConnectOptions::with_connection_memory_budget_bytes(self, u64) -> Self

`public-api.txt` is regenerated to match. The `Config` `memory` section stays internal (not on the public snapshot).

Note: `connect` now reads the config layering to resolve the default, which it did not before. With the default `0` this is behavior-preserving.

## Bindings

- `infino-python`: a budget argument on `connect` (bytes) mapping to `with_connection_memory_budget_bytes`; expose the config surface.
- `infino-node`: the same on the Node `connect`.

## Acceptance

- `ConnectOptions::with_connection_memory_budget_bytes` + `memory.budget_bytes`, resolved at connect with the precedence above; unit-tested.
- Python and Node `connect` accept a budget; a smoke test in each sets one and confirms it reaches the connection.
- `public-api.txt` updated; unset / `0` is measure-only with no behavior change.

naman-modi:
#396 #397 solves this.

PR #334, opened 2026-07-01
[2/n] feat memory: wire the connection budget through connect + config

naman-modi:
### What does this PR do?
Makes the per-connection memory budget from #316 actually usable: each connection gets one budget, shared by all its tables, and you can set it either at connect or in the config file. The default stays measure-only, so nothing changes until a limit is set.

The budget existed but nothing could reach it. It belongs to the whole connection, not a single table, so it has to be built once when you connect and shared by every table underneath. And it needs a way to be set, in the two places infino already takes config: at connect, and from the config file.

### How do we do this change?
- At connect, the budget is built once and shared (a cloned pointer) into every table the connection opens.
- It's set one of two ways: an option at connect, or a `memory` key in the config file. The connect path never reads the config file, the same split we use for storage credentials.
- Both ways agree on one rule, kept in a single place so they can't drift: `0` means measure-only, anything positive is a limit.

### Performance
No behaviour change. The default only measures and nothing reserves against it yet. Connect reads no extra files, and each table just clones a pointer, so there's nothing on the query path.

### Notes
- `public-api.txt` gains one line for the new connect option. It's additive, so the existing python / node bindings keep working unchanged.

A part for #325

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -85,7 +85,43 @@ pub struct Config {
     /// Compaction settings.
     #[serde(default)]
     pub compaction: CompactionSettings,
+    /// Per-connection memory budget.
+    #[serde(default)]
+    pub memory: MemorySettings,
 }
+
+/// Memory subsection of [`Config`]. All memory-related settings for Infino here.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(default)]
+pub struct MemorySettings {
+    /// Per-connection memory (heap) budget in bytes. `0` (the default) is
+    /// measure-only: usage is tracked but never refused. A positive value
+    /// enforces a ceiling so one connection can't exhaust process memory.
+    ///
+    /// Applies to connections built from a config file (`apply_config`). Code
+    /// that opens a connection programmatically sets the budget on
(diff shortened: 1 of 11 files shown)


src/config/mod.rs
```diff
@@ -85,8 +85,44 @@ pub struct Config {
...
+}
+
+/// Engine default connection budget when none is configured; used by both
+/// [`MemorySettings`] and the connect path. `0` is the deliberate measure-only
+/// (no-ceiling) sentinel that `from_budget_bytes` maps to a measured budget.
+///
+/// A future non-trivial default (e.g. a fraction of system RAM) changes here.
+/// `from_budget_bytes` stays a pure value mapper; the only added work then is
+/// letting the config field distinguish "unset" from an explicit `0`.
+pub(crate) const DEFAULT_CONNECTION_BUDGET_BYTES: u64 = 0;
```
savannahar68, line 124: can be picked in next PR - if possible lets declare all the const after imports

savannahar68 (approved):
lgtm

src/config/mod.rs
naman-modi: noted.

PR #338, opened 2026-07-01
fix table manifest: allow opening tables without any data appended

### LINKED ISSUE #336: bug catalog: open_table fails with PointerNotFound for a table created but never appended
Creating a table and ending the program without appending any data leaves the table in an inconsistent state: the catalog durably records the table, but no manifest pointer is written to storage. A subsequent `open_table` then fails with `ManifestLoadError::PointerNotFound`, even though `list_tables` reports the table as existing.

Steps to reproduce**

1. `connect(...)` to a storage-backed catalog.
2. `create_table("posts", schema, spec)`.
3. Do **not** call `append` / `commit`. End the program.
4. In a new run: `open_table("posts")`.

**Expected**

Opening a registered-but-empty table succeeds and yields an empty table (zero rows). The catalog is the source of truth for table existence; a table that was successfully created should be openable.

**Actual**

`open_table` returns `InfinoError` from `OpenError::ManifestLoadError(ManifestLoadError::PointerNotFound)`. The table is registered in the catalog (so `list_tables` shows it) but cannot be opened — a half-created state with no user-facing recovery path.

Pranav2612000:
### What does this PR do?
- Fixes #336

### DIFF
--- a/src/catalog/mod.rs
+++ b/src/catalog/mod.rs
@@ -771,6 +771,43 @@ mod tests {
     }
 
     #[test]
+    fn create_without_append_reopens_as_empty_table() {
+        // A table created but never appended to is still durably
+        // registered in the catalog. Reopening it in a fresh connection
+        // (i.e. after a program restart) must succeed and yield an empty
+        // table — not fail because no manifest pointer was ever written.
+        // The pointer is only written on the first commit, so `create`
+        // alone leaves the physical table with a catalog entry but no
+        // `_supertable/current`; open must tolerate that the same way
+        // `create` does when it probes and finds no pointer.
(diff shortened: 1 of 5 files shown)


src/catalog/mod.rs
```diff
@@ -316,6 +316,14 @@ impl Connection {
                 if let Some(cache) = disk_cache {
                     opts = opts.with_disk_cache(cache);
                 }
+                // A table can be registered in the catalog yet have no
```
savannahar68: do we need the comment?
I think this is a expected behaviour so can be skipped

savannahar68 (approved):
lgtm

PR #341, opened 2026-07-02
supertable: persist initial manifest on create; keep open strict (supersedes #338)

vinaykakade:
## Problem

Creating a table on durable storage (`file://`, S3, …) returned an in-memory handle but did not persist a manifest pointer until the first append's commit. So `open_table` on a freshly created table — from another process, after a restart, or before the first append — failed with a manifest-load error (`PointerNotFound`). This only worked on `memory://`, so the documented `create_table` → `open_table` sequence was effectively broken on durable backends.

## Fix

`Supertable::create` now materializes and persists the initial empty manifest — its list plus the pointer at `manifest_id 0` — so a freshly created durable table is openable immediately.

- **No id shift:** persisted at `manifest_id 0`, so the first append still commits `manifest_id 1`.
- **Race-tolerant:** if a concurrent creator wins the initial-pointer CAS, `create` adopts their committed manifest.
- **No orphan:** the stale id-0 manifest is reclaimed by GC after compaction.

## Relationship to #338 (supersedes)

#338 fixed the same bug (#336) the other way — it made `open` **tolerant**: a missing pointer returns an empty table. This PR takes the **eager + strict** approach instead, and reverts #338's tolerant-open, because:

> With tolerant-open, *any* absent current-pointer reads as an empty table — including a pointer **lost/deleted after data was committed** (bad GC, partial delete, storage issue). That silently surfaces committed data as "0 rows" instead of erroring.

Since `create` now guarantees a persisted pointer for every registered table, `open` can stay **strict**: a genuinely missing pointer is `PointerNotFound` — the open-or-create trigger for a never-created table, or a lost pointer on a created one — surfaced as an error, never masked as an empty table. (A *corrupt* pointer remains a distinct error too.)

## Tests

- `create_without_append_reopens_as_empty_table` (catalog) and `durable_open_before_first_append_yields_empty_usable_table` — a created durable table reopens as empty+usable via the persisted pointer.
- `open_on_fresh_tempdir_returns_pointer_not_found` — a never-created dir still errors (strict open-or-create trigger).
- Crash-safety contract (`supertable_commit_crash_localfs`) updated for the new reality that `create` publishes an id-0 pointer, incl. `crash_post_list_no_prior_commit_yields_pointer_unreadable` (strict). All 5 pass.
- Full `make ci` green: fmt, clippy, doctests, **public-api (no drift)**, commit / catalog / supertable suites.

### DIFF
--- a/src/catalog/mod.rs
+++ b/src/catalog/mod.rs
@@ -727,6 +727,46 @@ mod tests {
         ));
     }
 
+    /// Regression: on durable storage, `open_table` on a table that was
+    /// created but never appended to must succeed and yield an empty,
+    /// usable table. `create` leaves no pointer file until the first commit,
+    /// so a fresh `open` — any reconnect (another process, a restart) before
+    /// the first append — must treat the missing pointer as an empty table
+    /// rather than failing. Previously it surfaced a "manifest load error",
+    /// and the create handle only worked because it never went through `open`.
+    #[test]
+    fn durable_open_before_first_append_yields_empty_usable_table() {
+        let dir = tempfile::tempdir().expect("tempdir");
+        let uri = dir.path().to_str().expect("utf8 path").to_string();
+        let conn = connect(&uri).expect("connect");
+
+        // Create, but do NOT append through the returned handle.
+        let _created = conn
+            .create_table("docs", schema_id_title(), IndexSpec::new().fts("title"))
+            .expect("create_table");
+
+        // Open fresh — the reconnect path. This must not error.
+        let opened = conn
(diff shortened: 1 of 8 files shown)


savannahar68:
just going through the description of PR - is this duplicate of https://github.com/infino-ai/infino/pull/338

Pranav2612000 (approved):

vinaykakade:
Compared to #338, we make open_table stricter (treat a table that doesn't exist yet as error, as opposed to assuming it is empty).

PR #352, opened 2026-07-03
[3/n] feat memory: gate SQL heap on the connection budget

naman-modi:
### What does this PR do?
Makes the connection memory budget actually bound SQL. DataFusion runs every SQL query and allocates the heavy parts itself (sort, GROUP BY, join buffers), so we hand it a memory pool backed by the connection budget. When the budget tightens it spills to disk; if it still can't fit, the query fails with an over-budget error. The default is measure-only, so nothing changes until a limit is set.

The budget was plumbed into every connection but nothing enforced it yet, so setting a limit did nothing. SQL is the right first place: it's where the biggest, data-proportional allocations happen, and DataFusion already knows how to spill when memory runs low. We just point it at our budget.

### How do we do this change?
- A small adapter (`ConnectionBudgetPool`) implements DataFusion's memory-pool interface over the budget: measure-only never refuses, a bounded budget refuses past the 90% gate.
- Both SQL context sites (catalog `query_sql` and the reader's cached session) build with this pool.
- A refusal surfaces as `InfinoError::OverBudget`, not a generic query error.
- One helper builds the context, and every pool on a connection shares the same counter, so SQL (and later vector / ingest) all compete for one ceiling.

### Performance
No behaviour change by default. The pool is atomic-only: no locks, nothing allocated on the query path. Streaming queries (scans, filters) reserve nothing and run untouched even at a tiny budget; only the buffering operators are bounded. Tested that a sort larger than the budget spills and still returns correct results, and that a query with no room fails cleanly as OverBudget.

### DIFF
--- a/src/catalog/mod.rs
+++ b/src/catalog/mod.rs
@@ -30,7 +30,7 @@ use std::{
 
 use arrow::record_batch::RecordBatch;
 use arrow_schema::SchemaRef;
-use datafusion::{config::Dialect, execution::context::SessionContext};
+use datafusion::{config::Dialect, error::DataFusionError};
 use futures::future::try_join_all;
 pub use index_spec::IndexSpec;
 use manifest::{
@@ -44,8 +44,8 @@ use uri::{Backend, parse_uri};
 use crate::{
     InfinoError,
     config::DEFAULT_CONNECTION_BUDGET_BYTES,
-    memory::ConnectionMemoryBudget,
-    runtime_bridge::{bridge_on_runtime, bridge_sync_to_async, shared_io_runtime},
+    memory::{ConnectionMemoryBudget, budgeted_session_context},
+    runtime_bridge::{bridge_on_runtime, bridge_sync_to_async, build_query_runtime},
     storage::{StorageError, StorageProvider},
     superfile::{
         builder::FtsConfig,
@@ -454,7 +454,11 @@ impl Connection {
     /// ```
     pub fn query_sql(&self, sql: &str) -> Result<Vec<RecordBatch>, InfinoError> {
         debug!(sql, "running sql query");
-        let ctx = SessionContext::new();
+
+        // Gate SQL heap on the connection budget: DataFusion allocates the
+        // working set (sort / aggregate / join), so its pool is the gate.
+        let ctx = budgeted_session_context(&self.inner.connection_memory_budget)
+            .map_err(|e| InfinoError::Query(e.to_string()))?;
 
         // Resolve the relations the query names and register each that is a
         // catalog table. Unknown names (CTEs, search TVFs, aliases) are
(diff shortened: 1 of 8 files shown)


savannahar68:
when does it spills to disk and when does it throw back OverBudget?

src/catalog/mod.rs
```diff
@@ -454,7 +454,11 @@ impl Connection {
     /// ```
     pub fn query_sql(&self, sql: &str) -> Result<Vec<RecordBatch>, InfinoError> {
```
savannahar68, line 455: unrelated to this PR but imo this method shouldn't be part of this file

savannahar68: nvm cross table joins can happen so fair to have it here

savannahar68 (approved):
lgtm

naman-modi:
> when does it spills to disk and when does it throw back OverBudget?

We leverage Datafusion for this in the SQL query path. 

- It spills whenever a spillable operator ([sort](https://github.com/apache/datafusion/blob/53.1.0/datafusion/physical-plan/src/sorts/sort.rs#L548), [grouped aggregate](https://github.com/apache/datafusion/blob/53.1.0/datafusion/physical-plan/src/aggregates/row_hash.rs#L1026), sort-merge join) can free memory by writing buffered data to disk and keep going. 
- It returns OverBudget only when spilling can't rescue it like: 
   - the operator can't spill at all: 
        -  it's non-spillable (hash join build side, nested-loop join, window), or 
        -  it's a streaming operator (scan/filter/projection) that buffers nothing, so a single allocation already exceeds the budget and there's nothing to write out; 
   - the operator is spillable but can't even reserve the minimum it needs to run the spill/merge (e.g. sort's sort_spill_reservation_bytes for the merge).
   - No `DiskManager`, so nowhere to spill; but this case is not applicable to infino, as the DiskManager exists.

savannahar68:
> > when does it spills to disk and when does it throw back OverBudget?
> 
> We leverage Datafusion for this in the SQL query path.
> 
> * It spills whenever a spillable operator ([sort](https://github.com/apache/datafusion/blob/53.1.0/datafusion/physical-plan/src/sorts/sort.rs#L548), [grouped aggregate](https://github.com/apache/datafusion/blob/53.1.0/datafusion/physical-plan/src/aggregates/row_hash.rs#L1026), sort-merge join) can free memory by writing buffered data to disk and keep going.
> * It returns OverBudget only when spilling can't rescue it like:
>   
>   * the operator can't spill at all:
>     
>     * it's non-spillable (hash join build side, nested-loop join, window), or
>     * it's a streaming operator (scan/filter/projection) that buffers nothing, so a single allocation already exceeds the budget and there's nothing to write out;
>   * the operator is spillable but can't even reserve the minimum it needs to run the spill/merge (e.g. sort's sort_spill_reservation_bytes for the merge).
>   * No `DiskManager`, so nowhere to spill; but this case is not applicable to infino, as the DiskManager exists.

nice, thanks that is helpful

PR #358, opened 2026-07-05
catalog: re-export arrow_array and arrow_schema at crate root

### LINKED ISSUE #357: Re-export arrow crates from public API to prevent version-mismatch errors for users
### Problem

`infino` exposes `arrow` types in its public API (e.g. `Catalog::create_table` takes `arrow_schema::Schema`, table append takes `arrow_array::RecordBatch`), but does **not** re-export the `arrow` crates it depends on.

This makes `arrow` a **public (leaked) dependency**. When a user adds their own `arrow-schema` / `arrow-array` to `Cargo.toml` at a different major version than infino uses internally, Cargo keeps both copies in the dependency graph. The types then have identical names but are considered incompatible, producing confusing errors:

```
error[E0308]: mismatched types
  expected `arrow_schema::schema::Schema`, found `arrow_schema::Schema`
  note: there are multiple different versions of crate `arrow_schema` in the dependency graph
        arrow-schema-58.3.0  (used by infino)
        arrow-schema-59.0.0  (used by user)
```

Arrow bumps major versions frequently, so this will hit users repeatedly and is hard to diagnose.

### Reproduction

`Cargo.toml`:

```toml
[dependencies]
infino = "0.1.2"          # pins arrow 58.3 internally
arrow-schema = "59.0.0"   # user picks newer major
arrow-array = "59.0.0"
```

Call `create_table(..., schema.clone(), ...)` and `append(&RecordBatch::try_new(...)?)?` → both fail with `E0308` mismatched types across the two arrow versions.

### Proposed fix

Re-export the arrow crates that appear in infino's public API, so users import them through infino and can never pick a mismatched version:

```rust
// src/lib.rs
pub use arrow_schema;
pub use arrow_array;
```

Users then write `use infino::arrow_schema::Schema;` and do **not** add their own `arrow-*` dependency. This is the standard approach in the arrow ecosystem (`datafusion`, `lance`, `deltalake`, `polars` all re-export their arrow).

### Follow-ups

- [x] Add re-exports for every arrow crate appearing in the public API (audit `Schema`, `Field`, `DataType`, `RecordBatch`, arrays, etc.)
- [x] Document the required import path in README + docs.rs; add a runnable example
- [ ] Treat future `arrow` major bumps as **breaking** infino releases (arrow version is now part of our public API contract — semver coupling)
- [ ] (optional) Run `cargo check -Z public-dependency` on nightly in CI to catch any remaining leaked deps in the public API

Prashantsinghchouhan:
Implemented the core fix: pub use arrow_schema and pub use arrow_array at the crate root, updated public doctests + public-api.txt, and added tests/arrow_reexport.rs as a smoke test.

Audited public-api.txt — only those two Arrow crates appear in the public contract; types like Schema, Field, DataType, RecordBatch are covered via those re-exports.

Left for follow-up unless you want them in this PR: README / crate-docs.md / examples still use direct arrow_* imports; semver policy for Arrow major bumps; optional public-dependency CI check.

cargo test --doc and make public-api pass on my branch.

Closes #357

### DIFF
--- a/src/catalog/mod.rs
+++ b/src/catalog/mod.rs
@@ -159,7 +159,7 @@ impl Connection {
     ///
     /// ```
     /// # use std::sync::Arc;
-    /// # use arrow_schema::{DataType, Field, Schema};
+    /// # use infino::arrow_schema::{DataType, Field, Schema};
     /// use infino::{connect, IndexSpec};
     ///
     /// let db = connect("memory://")?;
@@ -272,7 +272,7 @@ impl Connection {
     ///
     /// ```
     /// # use std::sync::Arc;
-    /// # use arrow_schema::{DataType, Field, Schema};
+    /// # use infino::arrow_schema::{DataType, Field, Schema};
     /// # use infino::{connect, IndexSpec};
     /// # let db = connect("memory://")?;
     /// # let schema = Arc::new(Schema::new(vec![Field::new("body", DataType::LargeUtf8, false)]));
@@ -355,7 +355,7 @@ impl Connection {
     ///
     /// ```
     /// # use std::sync::Arc;
-    /// # use arrow_schema::{DataType, Field, Schema};
+    /// # use infino::arrow_schema::{DataType, Field, Schema};
     /// # use infino::{connect, IndexSpec};
     /// # let db = connect("memory://")?;
     /// # let schema = Arc::new(Schema::new(vec![Field::new("body", DataType::LargeUtf8, false)]));
@@ -441,8 +441,8 @@ impl Connection {
     ///
     /// ```
     /// # use std::sync::Arc;
-    /// # use arrow_array::{LargeStringArray, RecordBatch};
-    /// # use arrow_schema::{DataType, Field, Schema};
+    /// # use infino::arrow_array::{LargeStringArray, RecordBatch};
+    /// # use infino::arrow_schema::{DataType, Field, Schema};
     /// # use infino::{connect, IndexSpec};
     /// # let db = connect("memory://")?;
     /// # let schema = Arc::new(Schema::new(vec![Field::new("body", DataType::LargeUtf8, false)]));
--- a/tests/arrow_reexport.rs
+++ b/tests/arrow_reexport.rs
@@ -8,7 +8,10 @@
 use std::sync::Arc;
 
 use infino::{
-    IndexSpec, arrow_array::{LargeStringArray, RecordBatch}, arrow_schema::{DataType, Field, Schema}, connect,
+    IndexSpec,
+    arrow_array::{LargeStringArray, RecordBatch},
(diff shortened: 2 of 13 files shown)


Prashantsinghchouhan:
Hi @naman-modi, could you please review this PR when you have a chance? Thanks!

tests/arrow_reexport.rs
```diff
@@ -0,0 +1,25 @@
...
+
+use std::sync::Arc;
+
+use infino::{
+    arrow_array::{LargeStringArray, RecordBatch},
+    arrow_schema::{DataType, Field, Schema},
+};
+
+#[test]
+fn arrow_types_reexported_at_crate_root() {
```
naman-modi, line 18: This tests looks good, but can we have a more end-to-end test for the issue fix. Something like:

```
let db = infino::connect("memory://").expect("connect");
let schema = Arc::new(Schema::new(vec![Field::new("body", DataType::LargeUtf8, false)]));

let t = db.create_table("docs", schema.clone(), infino::IndexSpec::new().fts("body")).expect("create");

let batch = RecordBatch::try_new(schema, vec![Arc::new(LargeStringArray::from(vec!["ok"]))]).expect("batch");
t.append(&batch).expect("append accepts the re-exported RecordBatch");   // <- this is the regression guard
```

src/catalog/mod.rs
naman-modi, line 1: I think we should also update the `README.md`, and the `examples/` to use the `infino::arrow*` deps in this PR itself, as folks will generally refer to these as a starting point, what do you think? 

We can defer the other follow-ups.

naman-modi (changes requested):

tests/arrow_reexport.rs
Prashantsinghchouhan: makes sense.  will update/ create a new test to cover the full create_table + append flow using the re-exported Arrow types.

src/catalog/mod.rs
Prashantsinghchouhan: cool, will make the changes in this PR only.

Prashantsinghchouhan:
Hi @naman-modi, made the requested changes. could you please review? 
Thanks!

naman-modi:
Thanks @Prashantsinghchouhan. `make ci` [ref](https://github.com/infino-ai/infino/actions/runs/28775805678/job/85324989414?pr=358#step:7:1) is failing, can you please check once?

Prashantsinghchouhan:
@naman-modi 
Fixed the CI fmt failure — our rustfmt config merges infino::arrow_* into one infino::{ ... } import block. make check passes locally now

naman-modi (approved):
LGTM.

naman-modi:
Thanks for the contribution, @Prashantsinghchouhan! merging.

PR #364, opened 2026-07-06
docs: add infino's on-disk layout architecture doc

naman-modi:
### What
Adds `docs/architecture/on-disk-layout.md`, a physical view of what infino writes under a database root. Linked from `llms.txt` and cross-referenced from `supertable.md`.

### Why
`superfile.md` covers the single-file byte format and `supertable.md` covers the manifest/commit model, but nothing showed the actual directory tree a user sees under a database root, or how a read resolves from the catalog down to a superfile. Someone inspecting a `data/` directory had no doc to map files to meaning.


ashishmishra26 (approved):

Pranav2612000 (commented):
LGTM

PR #365, opened 2026-07-06
improv tracing spans: add spans for tracking duration of important functions

Pranav2612000:
### What does this PR do?
- Adds a mechanism to track the time taken by some important functions ( e.g `commit`, `append`, `bm25_query`, `sql_query` etc

- Having the exact duration of the functions helps get insights on how long each function takes, which aids performance improvements

### How do we do this change?
- Introduce a `detailed-tracing` feature. When the engine is built with this feature, it logs the spans of some important functions. Having a separate feature prevents the production performance from being affected

### How has this been tested?
- Manual test
- Example added in examples/

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,6 +86,15 @@ test-helpers = []
 # work or release-profile recall gates, but should not run as part of a
 # plain `cargo bench`.
 bench-diagnostics = []
+# Compiles the per-function timing spans (`#[tracing::instrument]`) on
+# the append/commit, SQL, search, superfile-open, and object-store
+# fetch paths. OFF by default: with the feature absent the
+# `cfg_attr`-gated attributes expand to nothing, so there is zero
+# runtime cost in a production build. Turn it on to profile where a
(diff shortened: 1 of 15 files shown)


Cargo.toml
naman-modi, line 1: Should we also add a check in the ci.yml to check if the compilation with `detailed-tracing` passes? Something like: `cargo check --features detailed-tracing --all-targets`. No strong opinion.

src/supertable/writer.rs
naman-modi, line 1682: Should we enable the detailed-tracing feature for `persist_commit` too?

naman-modi (approved):
Overall LGTM, thanks. Some minor comments.

src/supertable/writer.rs
Pranav2612000: Yes. Will add separately.

Cargo.toml
Pranav2612000: Let's add it later if we see the need for this. Adding it now will slow down CI builds, and most of the changes will not need this check.

PR #368, opened 2026-07-06
catalog: reject range_only + cache_dir at connect with InfinoError::C…

### LINKED ISSUE #340: range_only + cache_dir surfaces a leaky internal error instead of validating at connect()
**Context:** measuring cold reads from S3.

`connect(uri, cache_dir=…, cold_fetch_mode="range_only")` then a search raises a leaky internal error:
```
ValueError: superfile store error during query: failed to open superfile bytes: io error during read:
disk cache fetch: superfile reader failed to open mmap'd bytes:
ColdFetchMode::RangeOnly bypasses the disk cache; construct StorageRangeSource + open_lazy directly
```
So `range_only` + a `cache_dir` is an invalid combo, but it fails deep in the read path with an internal instruction rather than up front.

**Suggestion:** validate at `connect()` time — either make `range_only` ignore `cache_dir` cleanly, or reject the combo with a clear message (“range_only does not use a disk cache; omit cache_dir”). Right now it looks like a crash to a Python user.

adityasd314:
## What

`connect_with(uri, cache_dir=…, cold_fetch_mode="range_only")` crashes deep in the read path with a leaky internal message instead of a clear error at connect time. See #340.

Invalid configurations combining `cold_fetch_mode = "range_only"` and a configured disk cache (`cache_dir` or `disk_cache_root` in config) are now validated and rejected upfront at `DiskCacheStore` construction time.

This surfaces a clean `InfinoError::Config` instead of a leaky internal error deep in the query path (see #340).

## Why

`ColdFetchMode::RangeOnly` issues range requests directly into the query with no cache fill, so pairing it with a `cache_dir` is a contradiction. The engine accepted it silently and only surfaced a raw internal error string at query time.

## How

- Added a validation guard at the top of `DiskCacheStore::new` (the common entry point for both connections and configuration files) that rejects `RangeOnly` mode with `DiskCacheError::Config`.
- Mapped `DiskCacheError::Config` to the public `InfinoError::Config` variant in the catalog's `build_disk_cache`.

## API surface

Adds `InfinoError::Config(String)`. `InfinoError` is `#[non_exhaustive]` so downstream match arms are unaffected. `public-api.txt` updated with `make public-api-update`.

## Testing

Updated existing test and added new ones:
- Edited `supertable::reader_cache::disk::tests::reader_range_only_mode_is_rejected` to verify constructor validation.
- Added `catalog::tests::create_table_range_only_with_cache_dir_is_rejected` (verifies rejection on `connect_with`).
- Added `catalog::tests::create_table_cache_dir_with_non_range_only_mode_is_accepted` (verifies positive cache validation).
- Added `supertable::options::tests::apply_config_disk_cache_root_with_range_only_is_rejected` (verifies rejection on `apply_config`/`config.yaml`).

## Performance

Off the hot path.

Fixes #340

### DIFF
--- a/src/catalog/mod.rs
+++ b/src/catalog/mod.rs
@@ -56,7 +56,7 @@ use crate::{
     supertable::{
         Supertable,
         options::{Consistency, SupertableOptions},
-        reader_cache::{DiskCacheConfig, DiskCacheStore},
+        reader_cache::{DiskCacheConfig, DiskCacheError, DiskCacheStore},
     },
 };
 
@@ -94,12 +94,6 @@ pub fn connect_with(
     uri: impl AsRef<str>,
     options: ConnectOptions,
 ) -> Result<Connection, InfinoError> {
-    if options.cold_fetch_mode == ColdFetchMode::RangeOnly && options.cache_dir.is_some() {
-        return Err(InfinoError::Config(
-            "range_only does not use a disk cache; omit cache_dir".into(),
-        ));
-    }
-
     let backend = parse_uri(uri.as_ref())?;
     let store = match &backend {
         Backend::Memory => CatalogStore::Memory(Mutex::new(HashMap::new())),
@@ -754,8 +748,13 @@ fn build_disk_cache(
     if let Some(budget) = options.cache_budget_bytes {
         cfg.disk_budget_bytes = budget;
     }
-    let cache = DiskCacheStore::new_unpinned(Arc::clone(storage), cfg)
-        .map_err(|e| InfinoError::Io(e.to_string()))?;
+    let cache = DiskCacheStore::new_unpinned(Arc::clone(storage), cfg).map_err(|e| {
+        if let DiskCacheError::Config(msg) = e {
+            InfinoError::Config(msg)
+        } else {
+            InfinoError::Io(e.to_string())
+        }
+    })?;
     Ok(Some(cache))
 }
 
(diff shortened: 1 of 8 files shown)


adityasd314:
@naman-modi Can you please review it ?

src/catalog/mod.rs
```diff
@@ -93,6 +93,12 @@ pub fn connect_with(
     uri: impl AsRef<str>,
     options: ConnectOptions,
 ) -> Result<Connection, InfinoError> {
+    if options.cold_fetch_mode == ColdFetchMode::RangeOnly && options.cache_dir.is_some() {
+        return Err(InfinoError::Config(
+            "range_only does not use a disk cache; omit cache_dir".into(),
+        ));
+    }
```
naman-modi: Hey @adityasd314, a connection in infino is made via 2 ways: 
- one here, using connect_with() & ConnectOptions
- secondly via config.yaml where `disk_cache_root` and `cold_fetch_mode` is defined.

Here, you are only handling the first option, so better to have this check where there is a common path, like in the `DiskCacheStore` contructor.

src/catalog/mod.rs
```diff
@@ -93,6 +93,12 @@ pub fn connect_with(
     uri: impl AsRef<str>,
     options: ConnectOptions,
 ) -> Result<Connection, InfinoError> {
+    if options.cold_fetch_mode == ColdFetchMode::RangeOnly && options.cache_dir.is_some() {
+        return Err(InfinoError::Config(
+            "range_only does not use a disk cache; omit cache_dir".into(),
```
naman-modi: Please consider reframing from a permanent contradiction to a current limitation. 

We may want to support later, at which point this guard relaxes (reject → accept, a non-breaking widening).

```suggestion
            "range_only + disk cache is not supported yet; omit cache_dir".into(),
```

src/catalog/mod.rs
```diff
@@ -738,6 +744,17 @@ mod tests {
         ));
     }
 
+    #[test]
+    fn connect_range_only_with_cache_dir_is_rejected() {
```
naman-modi: Please check if other tests already check these,  and we want to cover all these cases:

1. range_only without cache_dir → Ok (the legit cacheless use isn't accidentally rejected).

2. cache_dir with a non-RangeOnly mode (e.g. HybridWithPrefetch) → Ok (guard isn't too broad).

3. The config-file path: apply_config with disk_cache_root + cold_fetch_mode: range_only must also be rejected — that's the entry point the connect_with guard misses.

src/error.rs
naman-modi, line 1: Please also check the comments in the other places like [here](https://github.com/infino-ai/infino/blob/main/src/catalog/options.rs#L14), where the comment should be updated w.r.t this change. eg: `the cache-servicing modes need a cache; RangeOnly is the cacheless path (for now).`

naman-modi (changes requested):

src/catalog/mod.rs
adityasd314: understood, moved the validation to `DiskCacheStore::new` so it blocks invalid configurations coming from both `ConnectOptions` and `config.yaml`

src/catalog/mod.rs
adityasd314: Added test cases to cover all connection and configuration paths.

src/catalog/mod.rs
adityasd314: Updated the error message and doc comments to frame this as a current limitation

src/error.rs
adityasd314: Updated the doc comment on ColdFetchMode to clarify that RangeOnly is the cacheless path and does not currently support caching.

naman-modi:
Thanks @adityasd314, please check and fix the `make ci` error.

adityasd314:
> Thanks @adityasd314, please check and fix the `make ci` error.

On it !

adityasd314 (commented):
Fixed the failing ci test to include constructor validation

src/catalog/options.rs
naman-modi, line 100: Would be helpful to add this comment here: 
```
/// `RangeOnly` is the cacheless path and is rejected if a cache_dir is set.`
```

naman-modi (approved):
Looks good now, minor comments. Thanks @adityasd314

adityasd314 (commented):
Added the Comment!

naman-modi:
Thanks, please rebase with main, and have CI green.

PR #371, opened 2026-07-07
[5/n] feat memory: gate vector search's cold fetch on the connection memory budget

naman-modi:
### What does this PR do?
Makes the connection memory budget bound vector search. A vector query over object storage fetches its IVF cluster blocks onto the heap (the cold GET); we now reserve those bytes against the budget before fetching. Over budget, the search fails with `InfinoError::OverBudget` before it allocates; a warm query whose blocks are already resident reserves nothing. Measure-only by default, so nothing changes until a limit is set.

#352 gated SQL, but a vector search allocates outside DataFusion: the cold cluster-block fetch, which grows with `nprobe` and the number of superfiles probed. Until now a limit did nothing for it, so one vector query could still push memory arbitrarily high. This puts the same ceiling on the query's largest allocation.

### How do we do this change?
- The vector reader reserves the cold cluster-block bytes right before the fetch (`reserve_cold_fetch`) and holds the guard for the probe; drop frees them.
- Only non-resident ranges are reserved, so a warm (mmap / in-memory) query slices its blocks zero-copy and reserves nothing: the hot path is untouched.
- A refusal is a new `VectorError::OverBudget`, routed through `ReadError::Vector` to the public `InfinoError::OverBudget`.
- The budget is read once in the vector fan-out, so every path (public `vector_search`, the SQL `vector_search` / `hybrid_search` functions, and `WHERE`-pushdown) is gated at the same point.

### Performance
No behaviour change by default. Warm queries reserve nothing, so there's no cost on the hot path even at a tiny budget; the cold fetch pays one atomic reservation, off the object-store GET's critical path. Recall@10 is unchanged (the reservation gates fetching, not scoring): 0.994 L2Sq and 0.998 Cosine at nprobe 8 and 32 on the scale bench, both above the 0.99 bar.

### DIFF
--- a/src/memory/mod.rs
+++ b/src/memory/mod.rs
@@ -109,8 +109,12 @@ const ENFORCED_BUDGET_DENOMINATOR: u128 = 10;
 /// limit, not a different policy — so there is no trait. The one trait we do
 /// implement is the query engine's memory-pool interface, in a thin adapter
 /// that forwards to an `Arc<ConnectionMemoryBudget>`.
+///
+/// Must be `pub`: the `pub` vector-search functions take this type as an
+/// argument. It is still not public API, because `lib.rs` makes the whole `memory`
+/// module `pub` only under `test-helpers``, so a shipped build keeps this type crate-internal.
 #[derive(Debug)]
-pub(crate) struct ConnectionMemoryBudget {
+pub struct ConnectionMemoryBudget {
     // Enforced ceiling in bytes, already reduced to the headroom gate.
     // `None` is measure-only: count usage, never refuse.
     limit: Option<usize>,
@@ -179,7 +183,7 @@ impl ConnectionMemoryBudget {
     /// Reserve `n` bytes, returning a guard that frees them on drop. Fails with
     /// [`OverBudget`] if the reservation would cross the ceiling; a measured
     /// budget always succeeds.
-    pub fn try_reserve(self: &Arc<Self>, n: usize) -> Result<Reservation, OverBudget> {
+    pub(crate) fn try_reserve(self: &Arc<Self>, n: usize) -> Result<Reservation, OverBudget> {
         self.try_grow(n)?;
 
         Ok(Reservation {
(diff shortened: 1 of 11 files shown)


savannahar68 (approved):
lgtm

src/test_helpers/mod.rs
```diff
@@ -78,6 +78,33 @@ pub fn default_disk_cache(
...
 }
 
+/// A `DiskCacheStore` in `LazyForegroundWithBackgroundFill`: the foreground
+/// query reads through an `open_lazy` `StorageRangeSource`, so the superfile's
+/// bytes stay non-resident and every cold read is an object-store GET. That is
+/// the path the connection budget gates. `default_disk_cache`
+/// (`HybridWithPrefetch`) instead collects the cold responses into a resident
+/// in-memory reader, which reads warm and reserves nothing. Same budget,
+/// timers, and eviction otherwise.
+pub fn lazy_foreground_disk_cache(
```
muralikpbhat, line 88: this new fn not used?

muralikpbhat (commented):
left a minor comment. 

A larger question is on the motivation for budget at connection level. Is this for really protecting clients who might use multiple connection in the same process? It is hard to actually define a budget at the connection level without knowing the workload of each connection upfront. How is the budget managed per connection and how do we ensure fairness (many connection case) and avoid starvation(say limited connection).  Is the budget configurable in the client connection and server decides to accept/reject?

src/memory/mod.rs
```diff
@@ -179,7 +183,7 @@ impl ConnectionMemoryBudget {
     /// Reserve `n` bytes, returning a guard that frees them on drop. Fails with
     /// [`OverBudget`] if the reservation would cross the ceiling; a measured
     /// budget always succeeds.
-    pub fn try_reserve(self: &Arc<Self>, n: usize) -> Result<Reservation, OverBudget> {
+    pub(crate) fn try_reserve(self: &Arc<Self>, n: usize) -> Result<Reservation, OverBudget> {
```
muralikpbhat, line 186: reservations are the precise control; do we also want a coarse process-RSS admission backstop for uninstrumented allocations? We might miss some places where allocations might happen without looking at budget

src/test_helpers/mod.rs
naman-modi: We are using it in the tests, [here](https://github.com/infino-ai/infino/pull/371/changes#diff-29118d4fb5f9bb1f2cb2249e0ca2eea3c374c766968557166b26da9373ab9087L71)...

naman-modi:
> left a minor comment.
> 
> A larger question is on the motivation for budget at connection level. Is this for really protecting clients who might use multiple connection in the same process? It is hard to actually define a budget at the connection level without knowing the workload of each connection upfront. How is the budget managed per connection and how do we ensure fairness (many connection case) and avoid starvation(say limited connection). Is the budget configurable in the client connection and server decides to accept/reject?

Thanks Murali, the motivation is the multi-connection case you mentioned: when the _infino engine is used as a library inside an end-to-end system_, we don't want one connection's query (a wide vector fetch, a big SQL sort) eating all the process memory and starving the other connections. The budget is the lever to isolate them.

A few things on how it's meant to be managed:

- the budget is configurable [per connection](https://github.com/infino-ai/infino/blob/main/public-api.txt#L165), so on a single process you can run a few connections with a higher budget and a few with a lower one, based on the workload. it's not one global number.
- the [default is measure-only](https://github.com/infino-ai/infino/blob/main/src/memory/mod.rs#L44): it tracks usage and never refuses, so nothing changes until someone sets a limit. the intended workflow is to run the workload first and then set an informed limit rather than guessing upfront. one honest gap today: the usage/peak is tracked internally but not yet exposed on the public API, so that "observe then size it" loop isn't fully wired for an integrator. that's the natural next step and i'd like your view on it.
- infino is an embedded library, not a server, so there's no server-side accept/reject step, imo. the application integrating infino owns the connections and sets the budgets directly. we deliberately just give the levers (per-connection limit, measured vs bounded); how to split a machine's memory across N connections is the integrator's policy, since they know their deployment.

On fairness and starvation across many connections: you're right that this PR doesn't solve cross-connection scheduling. Today each connection gets its own independent ceiling and there's no arbiter rebalancing between them. I've been treating that as an application-level concern rather than an engine one, but let me know what you thikn about this.

src/memory/mod.rs
naman-modi: good point. right now the precise reservations cover the big, data-proportional allocations we instrument, and for the smaller ones we reserve the headroom (10% of the mentioned budget). we dont want to gate too much also, as it would affect the latencies...

PR #375, opened 2026-07-08
catalog: memoize table handles so query_sql reuses the warm cache

### LINKED ISSUE #373: SQL: reuse the table handle across query_sql instead of rebuilding it every query
## Problem

Every `Connection::query_sql` on a storage-backed table rebuilds the whole `Supertable` from scratch. `query_sql` calls `open_table` for each relation it names, on every query (`src/catalog/mod.rs`), and for `CatalogStore::Storage` `open_table` builds a fresh `Supertable` plus a fresh empty `DiskCacheStore` each time. So the in-memory reader cache and the disk cache's `cached` map are empty on every query, and every query re-fetches all superfiles from storage. Nothing warms across queries.

`CatalogStore::Memory` already reuses a stored handle (`Mutex<HashMap<String, Supertable>>`), so its caches persist. The storage backends do not.

This is the normal usage: open one `Connection`, run many `query_sql`. Search (`bm25_search` / `vector_search`) avoids this because it is a method on `Supertable`, so the caller holds the handle and its cache stays warm. SQL sits on `Connection` (it joins across tables and hosts the search TVFs), so there is no warm handle to hold, and every query pays a cold rebuild. `SELECT COUNT(*)` hot stays ~0.10s (it whole-fetches every superfile) rather than a few ms.

## Possible Approach

Memoize the table handle per name on the `Connection` for storage backends, the same way `Memory` already does. `open_table` returns the cached handle if present, else builds it once and inserts it; `create_table` seeds it; `drop_table` removes it. Snapshot isolation stays intact via the manifest refresh that already runs per query (`ensure_fresh` / `read_consistency`), which leaves the disk cache untouched.

## Acceptance

- A second `query_sql` on the same table hits the warm cache (no whole-superfile re-fetch), asserted via `DiskCacheStore` stats.
- A commit between two `query_sql` calls is visible to the second.
- `memory://` unchanged; no regression on the SQL bench battery.

naman-modi:
## What

When you run SQL against a table that lives on object storage, the engine now keeps the table open and reuses it across queries, instead of opening it again from scratch on every query. The first query still loads what it needs; every query after that reads from the cache that is already warm.

## Why

Today every query reopens the table and starts with an empty cache. So each query downloads the whole table from object storage all over again, and the cache never gets a chance to warm up. Every query ends up as slow as the very first one, no matter how many times you have already read the same data. Keeping the table open between queries fixes this: the data is fetched once, and later queries reuse it.

## How

- **Two maps on `CatalogStore::Storage`.** `handles: DashMap<name, Supertable>` is the warm cache, sharded so parallel queries on one connection don't serialize. `building: DashMap<name, Arc<Mutex<()>>>` is a per-name build/evict lock.
- **`open_table` single-flights the build.** Lock-free `handles` lookup; on a miss, take the name's gate, re-check, build once, insert. Same-name openers wait on the gate so exactly one store is built (two would race their cold-fetch finalizers on the same cache file); different names build in parallel.
- **`create_table` and `drop_table` take the same gate.** `create_table` seeds the memo under it; `drop_table` evicts under it, across both the memo removal and the catalog commit. So build, create, and drop of one name never overlap, and a racing `open_table` cannot re-insert a handle after a drop removes it.
- **Memoized handles open `Strong`.** They re-read the manifest pointer per query, so they stay as fresh as the old rebuild-per-query, in-process and across connections.

## Performance

Every query after the first drops from a whole-object fetch of all superfiles to a warm range-read of the projected columns. The warm path is a lock-free sharded lookup with no serialization across concurrent queries; same-name first-opens single-flight so one cold fetch serves them all. 

## Testing

- Warm-cache reuse: the second `query_sql` hits the cache, no extra cold fetch.
- Freshness: a commit is visible on the next query both in-process and from a second `Connection` on the same storage (guards the `Strong` choice).
- Concurrency: same-name first-opens build one store (one cold fetch across 8 threads); `drop_table` racing `open_table` leaves no stale handle; `create_table` racing `open_table` stays consistent. Stressed 50x, no flake.

Closes #373

### DIFF
--- a/src/catalog/mod.rs
+++ b/src/catalog/mod.rs
@@ -30,6 +30,7 @@ use std::{
 
 use arrow::record_batch::RecordBatch;
 use arrow_schema::SchemaRef;
+use dashmap::DashMap;
 use datafusion::{config::Dialect, error::DataFusionError};
 use futures::future::try_join_all;
 pub use index_spec::IndexSpec;
@@ -54,7 +55,7 @@ use crate::{
     },
     supertable::{
         Supertable,
-        options::SupertableOptions,
+        options::{Consistency, SupertableOptions},
         reader_cache::{DiskCacheConfig, DiskCacheStore},
     },
 };
@@ -103,7 +104,11 @@ pub fn connect_with(
             if options.validate {
                 bridge_sync_to_async(read_catalog(root.as_ref()))?;
             }
-            CatalogStore::Storage(root)
+            CatalogStore::Storage {
+                root,
+                handles: DashMap::new(),
+                building: DashMap::new(),
+            }
         }
     };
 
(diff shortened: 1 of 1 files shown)


Pranav2612000 (approved):

Pranav2612000:
Looks good for now. The decision of not removing tables from 'building' may need to be changed in the future...

PR #380, opened 2026-07-09
node-publish: restore workspace ownership after the container build

ashishmishra26:
Second follow-up to #376: the glibc-baseline container runs as root, so the `npm install` / `napi build` outputs on the mounted workspace come back root-owned. The host-side `make node-verify` step then fails with `EACCES: permission denied, rename .../node_modules/apache-arrow` when npm (running as the `runner` user) tries to rework that directory.

Fix: a `sudo chown -R` of the workspace back to the runner user between the container build and the verify step. The build itself was succeeding — only the handoff to the host verify was broken.

### DIFF
--- a/.github/workflows/node-publish.yml
+++ b/.github/workflows/node-publish.yml
@@ -102,6 +102,12 @@ jobs:
                      npm install --ignore-scripts && \
                      npx napi build --platform --release --strip --js native.js --dts native.d.ts infino && \
                      npx tsc"
+      # The container runs as root, so everything it wrote on the mounted
+      # workspace (node_modules, target/, the addon) is root-owned — the
+      # host-side verify step then fails with EACCES. Hand it all back.
+      - name: Restore workspace ownership after container build
+        if: ${{ matrix.glibc_image }}
+        run: sudo chown -R "$(id -u):$(id -g)" .
       - name: Check binary glibc floor
         if: ${{ matrix.glibc_image }}
         run: |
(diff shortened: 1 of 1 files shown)


ashishmishra26:
Pushed a second commit: the glibc floor check now runs **after** `node-verify` instead of before it. Reason: `verify-pack.sh` rebuilds the addon on the host when any generated file is missing (the failed x64 run shows it took that path — "building addon + wrapper"), which would replace the container-built 2.31 binary with a 2.39 one after an early floor check had already passed. Checking last makes that failure mode break the build instead of shipping. Also added an `ls` of the container build outputs so the next run shows exactly which generated file, if any, the rebuild guard is missing.

pratyushlokhande (approved):

ashishmishra26:
Third commit — root cause of the rebuild found via the new inventory step: all five generated files existed, so it was the *other* guard condition. `verify-pack.sh` greps `native.js` for `${PKG_NAME}-${TRIPLE}` where `PKG_NAME` is the npm package name (`@infino-ai/infino`), but the generated loader references per-platform packages named after `napi.package.name` (`infx-<triple>`). The grep could never match, so the reuse guard has rebuilt the addon on **every** run since the `infx` rename — previously just ~10 wasted minutes per leg, but with the container builds it replaced the glibc-2.31 binary with a host-linked 2.39 one (exactly what the post-verify floor check flagged in the last run). Fixed to grep the `napi.package.name` prefix; verified locally that `infx-linux-x64-gnu` matches the current `native.js`. With this, the expected log line is `==> [1/5] reusing existing infino.<triple>.node` and the floor check should pass.

Issue #378, opened 2026-07-09
optimize(): min_fill_percent floor bug, permanently stranded seals, and a measured latency tradeoff


savannahar68:
Follow-up investigation on #372. Splitting this out because #372 was about one reporter's numbers, and this is about what we found digging into the compaction code itself, which is broader than their specific run.

## What we ruled out

Two things that looked like plausible explanations for "optimize() did not reduce segment count" turned out not to be it, at least not directly:

- `target_superfile_size_mb` is a packing ceiling, not a per-job minimum. The packer in `pack_partition` greedily accumulates candidates up to that size before emitting a job, so a large target should merge lots of small files into a few big jobs, not block merging.
- The default partition strategy is `IngestionTime` with 1 day buckets (`options.rs`, `effective_partition_strategy`). This is not an age gate on compaction eligibility, it is just how superfiles get grouped for skip pruning. Files ingested in the same session land in the same bucket and merge fine together. The doc comment above `effective_partition_strategy` actually claims the default is `Hash { n_buckets: 1 }`, which is wrong, the code returns `IngestionTime`. Worth fixing as a separate doc nit.

## Two real bugs

**1. `min_fill_percent` clamp turns "no floor" into a hidden 1% floor, and that floor can silently kill legitimate merges.**

`min_fill_percent.clamp(1, 100)` in the config means `min_fill_percent: 0` does not mean "no minimum," it means 1%. `min_output_bytes = target_bytes * clamped_percent / 100`. If a partition's total live data is smaller than that floor (which can easily happen if `target_superfile_size_mb` is set generously and the actual data is modest), `pack_partition` will still pack every candidate into one job, then throw the entire job away for missing the floor, even when merging all of them into one small file would obviously help.

I reproduced this directly, no synthetic setup needed: 40 tiny superfiles, ~560KB combined. With `target_superfile_size_mb: 1024, min_fill_percent: 1`, the floor works out to about 10.7MB. All 40 files pack into one candidate job, then get discarded because 560KB < 10.7MB. Zero merges, even though merging all 40 into one ~560KB file is a completely reasonable thing to do. Test: `compact_does_not_regress_warm_bm25_latency` (see below, same test surfaced this before I fixed the config to make the merge actually happen).

This would not fully explain the original report at 847k docs / ~5GB, since 5GB is well above a 40MB floor, but it is a real, independently reproducible way for `optimize()` to silently do nothing, and worth fixing regardless.

[...]

PR #385, opened 2026-07-10
chore: compaction unseal tombstones if errored

savannahar68:
As part of https://github.com/infino-ai/infino/issues/378

### What is this change required?

Today in case of error or crash we don't have a mechanism to unseal the sealed tombstones during compaction process. This will lead to certain superfiles not getting compacting and always sealed.

### How do we fix it?

We introduce a default_stale_timeout which helps the next running compaction job to include sealed tombstones if they have surpassed the stale timeout. This ensures that the orphan superfile will be picked up eventually. Right now the stale timeout is set to 2mins. Through process is for 1gb of compacted merged file size it takes 12seconds, so with current default CompactionSettings and with some buffer 2mins is enough for compaction jobs to complete successfully. Even after that if the file is sealed which means the process crashed.

In case of errors during compaction we unseal the files explicitly so that the next optimize/compaction run can pick it.

### DIFF
--- a/src/supertable/compaction/mod.rs
+++ b/src/supertable/compaction/mod.rs
@@ -232,6 +232,7 @@ impl Supertable {
             };
 
         // Build SuperfileStats for every superfile in the snapshot.
+        let now = Utc::now();
         let stats: Vec<SuperfileStats> = manifest
             .get_all_superfiles()
             .iter()
@@ -241,7 +242,9 @@ impl Supertable {
                     .cloned()
                     .unwrap_or_else(|| (Arc::new(RoaringBitmap::new()), None));
                 let tombstoned_docs = bitmap.len();
-                let sealed_by_other = seal.is_some();
+                let sealed_by_other = seal
+                    .as_ref()
+                    .is_some_and(|s| !tombstones_admin::is_seal_stale(s.sealed_at, now));
                 SuperfileStats {
                     superfile_id: entry.superfile_id,
                     partition_key: entry.partition_key.clone(),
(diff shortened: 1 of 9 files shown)


src/supertable/wal/tombstones_admin.rs
```diff
@@ -59,11 +61,27 @@ pub enum TombstonesAdminError {
     WalStore(#[from] WalStoreError),
 }
 
+/// How old a seal has to be before we treat its owner as dead and
+/// take over, instead of backing off. A crashed compactor never
+/// gets to unseal its own inputs, so this is what actually
+/// recovers them.
+pub const DEFAULT_STALE_SEAL_TIMEOUT: Duration = Duration::from_secs(2 * 60);
```
Pranav2612000: Is this low? Can compaction take more than this to complete? Especially with CAS errors ?

src/supertable/compaction/mod.rs
```diff
@@ -479,21 +493,35 @@ impl Supertable {
...
     }
 }
 
+/// Best-effort of clearing every seal this attempt placed
+async fn unseal_all(wal_store: &WalStore, compaction_id: Uuid, sealed_ids: &[Uuid]) {
+    for id in sealed_ids {
+        if let Err(e) = tombstones_admin::unseal(wal_store, *id, compaction_id).await {
+            warn!(superfile_id = %id, error = %e, "compact: failed to unseal after aborting");
+        }
+    }
```
Pranav2612000: Thoughts on unsealing parallely, instead of one after the other?

src/supertable/wal/tombstones_admin.rs
```diff
@@ -129,6 +146,37 @@ pub async fn seal(
...
+/// Clear the seal on a sidecar, but only if it's still sealed by
+/// same compaction_id
+pub async fn unseal(
+    wal_store: &WalStore,
+    superfile_id: Uuid,
+    compaction_id: Uuid,
+) -> Result<(), TombstonesAdminError> {
+    let Some((existing, etag)) = wal_store.get_tombstones(superfile_id).await? else {
+        return Ok(());
+    };
```
Pranav2612000: Do we need the extra GET?

src/supertable/wal/tombstones_admin.rs
savannahar68: discussed offline. And have pushed this control to user level via CompactionSettings

src/supertable/compaction/mod.rs
savannahar68: done

src/supertable/wal/tombstones_admin.rs
savannahar68: Removed it entirely

savannahar68:
addressed all the review comments @Pranav2612000

Pranav2612000 (approved):

PR #387, opened 2026-07-11
[6/n] feat memory: gate the ingest - used while building superfiles - on connection memory budget

naman-modi:
### What does this PR do?
Bounds heap memory used during ingestion, against the connection memory budget.

#352 gated SQL and #371 gated vector search, but the write path was still open. Building a superfile (the FTS blob, the vector index, and the Parquet columnar data) from the buffered rows is the largest allocation on the write path: it materializes the whole index and the serialized file in RAM before the object-store PUT. This happens at commit, or mid-`append` once `commit_threshold_size_mb` is breached, so one big append could exhaust memory even with a budget set.

### How do we do this change?
- Before draining the buffer, the writer reserves the estimated build heap against the budget via `reserve_build_scratch`, held across the concurrent shard build and the publish, freed on drop. Reserving before the drain is what leaves the buffer untouched on a refusal.
- Per-column-type estimate: The peak-to-input ratio differs a lot by column kind (measured ~2-3x scalar-only, ~3-4x with FTS, ~5-8x with vectors). So,  each input class is weighted on its own and summed: `2.5*scalar + 6.5*vector + 1.5*fts_text`; the FTS term is additive on top of the scalar hold since the same text drives the index build.

These are central estimates, not worst-case ceilings; a fully dynamic per-schema factor can be future work.

### Performance
No behaviour change by default.  The only added work is one atomic reservation per commit, off the per-row path. 

### Notes
**The update path isn't gated yet.** `Supertable::update` builds its new-row superfile through a separate WAL pipeline. Append is the common, high-volume write path, so it's gated first; the update build is worth designing when an update-heavy workload makes it matter, so it lands as its own slice. The design there is a placement plus refusal-semantics choice, not a WAL change: the new rows are durable before the build, so a refusal either rejects before durability (clean, retryable) or defers to the recovery sweep (which isn't budget-aware and could get stuck).

### DIFF
--- a/src/supertable/writer.rs
+++ b/src/supertable/writer.rs
@@ -124,17 +124,26 @@ use crate::{
     },
 };
 
-// An approximate multiple for the memory the build will use, reserved up front rather than accounted for exactly.
-// Building the superfile here holds the FTS and vector blobs plus the serialized file in memory at once, so the
-// real peak is a few times the raw ingested bytes.
+// Approximate multiples for the memory the build will use, reserved up front rather than
+// accounted for exactly. Building the superfile holds the FTS and vector blobs plus the
+// serialized file in memory at once, so the real peak is a few times the raw ingested bytes.
 //
-// Reserving "5 x ingested_raw_bytes" is a cheap estimate. It covers the measured scalar-only (~2-3x) and
-// FTS (~3-4x) cases, but vector-heavy builds reach ~5-8x (worse for small batches and across concurrent
-// shards), so 5x can still under-reserve those.
+// Each kind of blob has a separate factor, so the estimate tracks the schema & ingestion data
+// closely: memory for building vector blobs >> memory for the FTS blob >> memory for plain
+// scalar columns.
 //
-// A better way later is a dynamic factor, since it depends on the schema a lot:
-// memory for building vector blobs >> memory for the FTS blob >> memory for plain scalar columns
-const BUILD_SCRATCH_FACTOR: usize = 5;
+// Stored as numerator over DENOM so the estimate stays integer-only (halves).
+const BUILD_SCRATCH_DENOM: usize = 2;
+
+// Scalar columns, held then serialized into the Parquet body: ~2.5x.
+const BUILD_SCALAR_NUM: usize = 5;
+
+// f32 vector payload, rebuilt as quantized + rerank codecs alongside the raw input: ~6.5x.
+const BUILD_VECTOR_NUM: usize = 13;
+
+// FTS text, ~1.5x for the FST + postings structures. Added on top of the scalar factor, not
+// instead of it: the same text bytes are held as a column and drive the index build at once.
+const BUILD_FTS_NUM: usize = 3;
 
 /// Single-writer append + commit handle.
 ///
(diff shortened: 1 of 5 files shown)


src/supertable/writer.rs
```diff
@@ -1068,6 +1094,19 @@ impl ShardOutput {
...
 }
 
+/// Reserve the build's estimated heap (`raw_bytes` times [`BUILD_SCRATCH_FACTOR`]);
+/// returns `OverBudget` when a bounded budget can't fit it.
+fn reserve_build_scratch(
+    budget: &Arc<ConnectionMemoryBudget>,
+    raw_bytes: usize,
+) -> Result<Reservation, BuildError> {
+    budget
+        .try_reserve(raw_bytes.saturating_mul(BUILD_SCRATCH_FACTOR))
```
savannahar68: instead of multiplying with factor
can we come up with simpler but a bit sophesticated formula to compute estimate bytes?

eg: during append we count number of docs, so based on number of docs 
and if number of vectors present 

come up with a simple formula and add some multiple to it

This will give a better estimate compared to 5x

in cases where there are no vectors we will still reserve 5x that will affect queries

this looks good but at high throughput both ingest and queries this will bite us

src/supertable/writer.rs
```diff
@@ -974,6 +987,16 @@ impl SupertableWriter {
         if self.buffer.is_empty() {
             return Ok(());
         }
+
+        // Try reserving the transient heap from the ConnectionMemoryBudget before draining the buffer.
+        // For now, if memory reservation is refused, the buffer is left untouched, but this behaviour can be changed.
+        //
+        // Held until this function returns, i.e. past `publish_superfiles` below.
+        let _build_guard = reserve_build_scratch(
```
savannahar68, line 1027: trying to think here:

instead of throwing err here can we instead retry couple of times maybe in smaller chunks which will create smaller superfiles but its fine given we have compaction

throwing entirely will push clients to retry or in our saas we will have to have some queuing or wal mechanism

wdyt?

beacuse as per current logic we already have raw data in memory

savannahar68:
should we also reserve raw bytes?

wdyt? because with vectors even those will have significant memory consumption

ideally the idea is to have a memory allocator on ingestion side to build exactly number of superfiles based on how much memory we have, and reject incoming append in case we are under pressure

there should be some mechanisms to reject appends all together instead of throwing errs at commit

src/supertable/writer.rs
naman-modi: Agreed, I was thinking of picking it later, but makes sense to do it now. Updated the logic.

src/supertable/writer.rs
naman-modi: As we discussed offline, the change to "create smaller batches and ingest" would require more thoughts around how we handle this at scale aka have a durable ingest queue/ingest-side WAL we push to, and for the time being its a conscious choice to push the error to the process using the engine, and handle on their end. We can always come back to update this behaviour.

naman-modi:
> should we also reserve raw bytes?
> 
> wdyt? because with vectors even those will have significant memory consumption
> 
> ideally the idea is to have a memory allocator on ingestion side to build exactly number of superfiles based on how much memory we have, and reject incoming append in case we are under pressure
> 
> there should be some mechanisms to reject appends all together instead of throwing errs at commit

Good point, two separate things here though, imo:

on reserving raw bytes: we don't reserve them incrementally as they buffer, but the commit-time reserve is a multiple of exactly those raw bytes (scalar + vector), so holding them through the build is already covered.

Regarding the allocator idea (size superfiles to available memory, reject appends under pressure instead of erroring at commit) is the bigger separate stream, that's more of admission control / backpressure, and I don't see that as part of this PR. But I agree, and we can pick it up soon.

savannahar68 (approved):
lgtm
if not already added lets add a test with multiple supertables readers and writers if things are working fine and budget is being honored (can be in different PR)

PR #389, opened 2026-07-12
Fix/profile superfile budget arg

Prashantsinghchouhan:
## Summary
- Follow-up to #371: `profile_superfile` still called `vector_hits_filtered_async` with 7 args after `budget` was added.
- Pass `None` for the connection memory budget (same as other unbudgeted bench paths).

### DIFF
--- a/benches/utils/bin/profile_superfile.rs
+++ b/benches/utils/bin/profile_superfile.rs
@@ -158,6 +158,7 @@ fn mean_recall_filtered(
             Some(Arc::clone(allow)),
             None,
             None,
+            None
         ))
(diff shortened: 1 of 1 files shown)


Prashantsinghchouhan:
Hi, @naman-modi 
small miss from PR (#371): `profile_superfile` still uses the pre-budget `vector_hits_filtered_async` signature (7 args). `cargo check -p infino-bench-utils` fails on current `main`. 

Fix: added a trailing `None` for `budget` #389

naman-modi (approved):
Thanks for catching this, @Prashantsinghchouhan.

PR #396, opened 2026-07-14
[9/n] feat python: expose the connection memory budget in connect()

naman-modi:
### What does this PR do?
Exposes the connection memory budget in the Python binding: `connect()` gains a `connection_memory_budget_bytes` keyword, and an over-budget refusal raises a typed `ConnectionMemoryBudgetError`.

The engine already caps a connection's heap, but Python callers had no way to set it, so the budget was unreachable from Python. And an over-budget error mapped to a generic `RuntimeError`, indistinguishable from any other failure.

### How do we do this change?
- **`connect()` keyword.** `connection_memory_budget_bytes` forwards to `ConnectOptions::with_connection_memory_budget_bytes`; `0` or omitting it measures without enforcing. Sits beside `cache_budget_bytes` so heap and disk budgets read distinctly.
- **Typed error.** Adds an `InfinoError` base exception with `ConnectionMemoryBudgetError` under it, so a caller can catch this specifically (or the family) and back off. Other errors still map to builtins for now; moving them under `InfinoError` is a follow-up.


infino-python/examples/connection_memory_budget.py
pratyushlokhande, line 1: @naman-modi do we really need this example? The existing examples are real e2e usecases. Not sure if this aligns.

naman-modi: Fair point, Pratyush. Removed it.

pratyushlokhande: thanks

pratyushlokhande (approved):

PR #398, opened 2026-07-14
feat: add code-context support

Pranav2612000:
### What does this PR do?
- Adds support for using code-context(https://github.com/infino-ai/code-context) with this repo


ashishmishra26:
@Pranav2612000 Not needed, this was a side effect, removing that from package itself.

PR #412, opened 2026-07-17
fts: lucene-style query clauses — +must, should, -must-not

vinaykakade:
## What

`bm25_search` / `bm25_hits` / `token_match` / `count` queries now support lucene `BooleanQuery`-style clause sigils, parsed natively:

- `+term` — **must**: every hit contains it; the musts' intersection defines the match set.
- `-term` — **must-not**: hard exclusion (unchanged from the existing negation support).
- bare term — polarity from `BoolMode`, which becomes the **default operator**: `And` requires bare terms like musts; `Or` makes them scoring-only **shoulds** once any must exists, and a plain union when none does.

`"+climate policy"` under `Or` matches exactly the docs containing `climate` and ranks those also mentioning `policy` higher. Counts follow the match set: `count("+climate policy")` is the number of docs containing `climate`, and a lone must keeps the O(1) df fast path.

## Why

Mixed must/should queries are a standard lucene query shape that previously mis-executed: `+` was stripped as punctuation, so `+climate policy` silently ran as the union `climate OR policy` — returning wrong match sets and wrong counts. Every sigil-less query behaves bit-for-bit as before; the only behavior change is that `+` now means must, which is what anyone writing it intended.

## How

- Parser (`Tokenizer::parse`): a `+`-prefixed run is a must clause; `ParsedQuery::into_clauses(mode)` resolves bare-term polarity once at parse time.
- Kernel: the must intersection drives the existing AND flat-merge; should cursors ride in the emit sink, `skip_to`-streamed to each intersection doc (each should list is traversed at most once per query). Block-max pruning stays sound by lowering the admission bar by the shoulds' summed term-max upper bound. Single-atom, pure-OR, and pure-AND queries route to the exact kernels they used before — zero added cost to existing shapes. The clause path also threads the cross-segment score floor that the negation path previously lacked.
- Supertable: musts prune superfiles with AND-presence semantics; unranked `token_match`/`count` use the must set (there are no scores for a should to raise).

## Tests

Brute-force clause oracle (`top_k_clauses`) in `test_helpers`; planted-corpus tests at the superfile layer (12, incl. multi-block corpora crossing PFOR block boundaries and truncated-top-k pruning) and supertable layer (4, cross-superfile prune + count); parser unit tests pinning every legacy shape parses identically. `make ci` green (fmt, clippy, api-parity, doc-check, doctests, coverage).

## Performance

Validated on the public search-benchmark-game harness (5M-doc English Wikipedia corpus, full 962-query set): infino-ai/search-benchmark-game#1 — mixed must/should counts now agree exactly with lucene 10.4 and tantivy 0.26 on every command (previously all 40 mixed queries diverged), mixed-shape COUNT dropped from 224µs to 50µs mean (−78%, must-driven count replaces the full-union walk), and no query shape regressed. Three mixed-clause rows were added to the internal fts bench battery (`must_common_should_common`, `must_rare_should_common`, `must_two_should_two`).

### DIFF
--- a/src/supertable/query/fts.rs
+++ b/src/supertable/query/fts.rs
@@ -228,41 +228,49 @@ impl SupertableReader {
         let pool_threads = manifest.options.reader_pool.current_num_threads();
         let column_owned = column.to_owned();
 
-        // Parse the query once here, not per superfile. The fan-out
-        // closures below need owned ('static) data for tokio::spawn,
-        // so this is the one place the tokens are copied — the prune
-        // and every per-superfile search reuse them.
-        let parsed = AsciiLowerTokenizer.parse(query);
-        let positives: Vec<String> = parsed.positives.into_iter().map(Cow::into_owned).collect();
-        let negatives: Vec<String> = parsed.negatives.into_iter().map(Cow::into_owned).collect();
+        // Parse the query once here, not per superfile, resolving the
+        // bare tokens' polarity from the default operator (`And` ⇒
+        // must, `Or` ⇒ should). The fan-out closures below need owned
+        // ('static) data for tokio::spawn, so this is the one place
+        // the tokens are copied — the prune and every per-superfile
+        // search reuse them.
+        let clauses = AsciiLowerTokenizer.parse(query).into_clauses(mode);
+        let musts: Vec<String> = clauses.musts.into_iter().map(Cow::into_owned).collect();
+        let shoulds: Vec<String> = clauses.shoulds.into_iter().map(Cow::into_owned).collect();
+        let negatives: Vec<String> = clauses.negatives.into_iter().map(Cow::into_owned).collect();
 
-        // Negation-only (e.g. `-foo`): no positive anchor to rank. Reject
-        // up front so the per-superfile excluding kernel never has to, and
-        // so the unranked count / token_match path surfaces the identical
-        // error (see `parse_and_prune`).
-        if positives.is_empty() && !negatives.is_empty() {
+        if musts.is_empty() && shoulds.is_empty() {
+            // No scorable clause at all. Empty / punctuation-only
+            // queries match nothing (not an error); negation-only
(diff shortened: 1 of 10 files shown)


src/supertable/query/fts.rs
```diff
@@ -471,33 +475,51 @@ impl SupertableReader {
...
+            (musts, BoolMode::And)
+        };
         let prune_leaf = PruneLeaf::TermPresence {
             column: column.to_owned(),
-            terms: positives.clone(),
-            mode,
+            terms: match_terms.clone(),
+            mode: match_mode,
         };
         let kept =
```
savannahar68, line 520: one small optimisation can be done here for must-intersection COUNT query 

Something like 

```
if musts.len() == 1 && shoulds.is_empty() {
    // O(1) path: single must = dictionary df lookup per superfile
    return term_df(musts[0]) * superfile_count;  // (pseudo-code)
}
```

This would make count("+rare_term") O(1) globally

wdyt ?

savannahar68 (approved):
lgtm

src/supertable/query/fts.rs
vinaykakade: Good instinct — this is already wired up, and it actually covers a wider case than the single-clause query. `parse_and_prune` returns `match_terms` = the musts alone whenever any must exists, so both `count("+rare_term")` and `count("+rare_term plus any shoulds")` arrive at the count kernel with `match_terms.len() == 1` → the `single_term` branch → per-superfile `term_df` (the O(1) dictionary-header read; see the `single_term` binding just below this hunk and the `term_df` branch in the fanout body). Per-superfile dfs are **summed** across the manifest (not multiplied — `count_single_term_sums_df_across_superfiles` pins that), and the term-bloom prune drops superfiles lacking the must before any fanout. The bench battery's `must_common_should_common` count row lands at ~375ns for exactly this reason. Two deliberate exceptions where we still materialize: negated terms and tombstoned superfiles, since a bare df can't subtract exclusions/deletes.

PR #416, opened 2026-07-17
sql dep: upgrade datafusion from  53 to 54

naman-modi:
### What does this PR do?
Upgrades the query engine dependency from DataFusion 53 to 54.

### List of changes
- **Bump the dependency** across the library, benches, and the Python and Node bindings. Arrow stays at 58 (54 accepts it), so nothing else in the read/write path moves.
- **Move the S3 test dependency** `s3s` / `s3s-fs` 0.13 to 0.14: DataFusion 54's crypto functions require the final `md-5` 0.11, while the old `s3s` pinned the release candidate.
- **Adapt to the API changes:**
  - `MemoryPool` now requires `name()` and `Display`.
  - `TableProvider` / `ExecutionPlan` / `TableSource` dropped `as_any` for an `Any` supertrait; downcasts use DataFusion's provided `downcast_ref`.
  - `PartitionedFile` extensions are keyed by concrete type, so the access plan is attached with `with_extension` and read back via `extensions.get::<ParquetAccessPlan>()`. This one is behavioral: the old slot would no longer be found, silently disabling row-group pruning.
  - `TableFunctionImpl::call` is replaced by `call_with_args`, migrated across the search table functions.

### Testing
- Full test suite green (library plus every integration binary), no failures.
- `clippy --all-targets` clean, fmt clean, public-API snapshot unchanged.
- Python and Node bindings compile against 54.

### DIFF
--- a/src/supertable/query/exec/vector_exec.rs
+++ b/src/supertable/query/exec/vector_exec.rs
@@ -34,14 +34,14 @@
 //! score` ascending lists nearest neighbours first. See
 //! [`SuperfileHit::score`].
 
-use std::{any::Any, collections::HashSet, fmt, sync::Arc};
+use std::{collections::HashSet, fmt, sync::Arc};
 
 use arrow::compute::cast;
 use arrow_array::{Array, ArrayRef, Float32Array, ListArray};
 use arrow_schema::{DataType, SchemaRef};
 use async_trait::async_trait;
 use datafusion::{
-    catalog::{Session, TableFunctionImpl, TableProvider},
+    catalog::{Session, TableFunctionArgs, TableFunctionImpl, TableProvider},
     error::{DataFusionError, Result as DfResult},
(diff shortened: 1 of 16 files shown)


src/supertable/query/exec/vector_exec.rs
```diff
@@ -1017,15 +1010,15 @@ mod tests {
         let st = supertable_one_superfile(dim, 8);
         let reader = Arc::new(st.reader());
         let scalar_schema = reader.options().scalar_schema();
+        use crate::supertable::query::exec::common::test_support::call_tvf;
```
Pranav2612000, line 1013: Nit: imports at top

Pranav2612000: here and at some other places

Pranav2612000 (approved):

PR #420, opened 2026-07-19
fts: token positions and exact phrase queries

vinaykakade:
Adds opt-in token positions to the full-text index and makes double-quoted exact phrases first-class query atoms in `bm25_search`, `token_match`, and `count`.

## What

- **Positions are a per-column opt-in** (`FtsConfig { positions: true }`, default off). A positional column stores, per (term, doc), the token ordinals as delta-encoded varint runs in a dedicated, CRC-guarded region addressed from the term dictionary — a query fetches only the runs it touches. Positions roughly double a column's FTS footprint, which is why they're opt-in.
- **Quoted phrases are query atoms** under the existing flat clause model: any polarity works (`"new york" hotel`, `+"new york" +hotel`, `-"memory unsafe"`). A phrase scores as one BM25 atom — `tf` = verified occurrences, `idf` = Σ member idf. A single-word phrase degrades to that term.
- **No silent fallback**: a phrase against a positionless column is a typed error (`PositionsUnavailable`), never a bag-of-words approximation. No public signatures change — the feature rides in the query string and the build config.
- **Counts are exact**: the count path verifies adjacency for every candidate.

## Format

The FTS blob header grows 48 → 56 bytes (version 2) and records the positions-region offset; the region is empty for positionless columns. New code always writes v2 and reads v1; term metadata for positional columns carries the run offset/length (32 bytes vs 20), skip entries address the per-block first run, and a df=1 tf=1 term packs its single position into the existing inline-value fast path. Cold-open GET count is unchanged.

## Query path

Phrase-bearing queries route to heterogeneous atom walks (term + phrase cursors); term-only queries keep the existing optimized kernels byte-for-byte. A phrase cursor leapfrogs its member postings and verifies adjacency via per-block run-offset caches with binary-search probes. Ranked walks hand each atom a pruning bar (kth-best minus the other atoms' score bounds): since a phrase's tf can't exceed any member's tf, candidates whose min-member-tf bound falls below the bar skip position work entirely. That pruning is what makes dense phrases ("secretary of state") competitive: verification only runs for candidates that can still change the top-k.

## Performance

External: the full 962-query standard suite (5M-doc Wikipedia corpus, c7i.2xlarge) on [search-benchmark-game](https://github.com/infino-ai/search-benchmark-game). All 962 queries are now answered — including the 300 phrase queries previously reported UNSUPPORTED — and every phrase count was verified exact, query by query.

Phrase-tag latency (mean / median of per-query best, ms):

| mode | latency |
|---|---|
| TOP_10 | 1.41 / 0.54 |
| COUNT | 1.78 / 0.43 |
| TOP_100 | 1.91 / 0.84 |
| TOP_1000 | 2.44 / 1.02 |
| TOP_100_COUNT | 3.74 / 1.26 (†) |

(†) That mode issues a full top-k search *plus* a separate exhaustive count (3.74 ≈ 1.91 + 1.78); a count-while-ranking walk is a natural follow-up.

Non-phrase tags are unregressed vs current main — term/intersection/union deltas within cross-run noise in every mode (e.g. TOP_10 global mean 0.52 vs 0.51 ms; COUNT intersection 0.29 vs 0.27 ms).

### Internal A/B (branch vs `main`, same host)

Azure A/B runs (Standard_D8s_v7, 1M docs, both arms built and run on one VM) plus a local three-arm A/B at the same scale:

- **Warm/count queries, supertable tier**: every non-phrase row within the ±5% noise threshold (best: `twenty_term_or` p90 −5%). Phrase rows land as new cells (e.g. `phrase_two_mixed` warm min 1.55 ms across 112 segments).
- **Warm queries, superfile tier**: an early A/B caught a real regression — multi-term OR shapes +6–9%, scaling with term count — traced to the positional fields having grown the two structures the OR kernels iterate hottest (`BlockMeta` 24→32 bytes, `TermCursor` +3 fields). Fixed by moving all positional state onto the phrase members, which re-derive it from the term's own bytes; `BlockMeta`/`TermCursor` are back to their pre-positions layout.
- **Re-verification after the fix** (same A/B rig): every non-phrase FTS query row is back under the ±5% gate — `three_similar_or` +2.3%, `five_term_or` +2.1%, `ten_term_or` +0.6%, `twenty_term_or` +3.6%, `forty_term_or` +3.9% (from +6–9% before the fix), with AND rows within noise. The small remaining drift tracks a codegen/layout shift, not a code path: the term-only kernels are source-identical to main, yet the BMM kernel's machine code moved +128 bytes on a 6.6 KB symbol under `codegen-units = 1`. The real-corpus check agrees: non-phrase tags on the external suite are flat vs main in all five modes.
- **Positional build cost (opt-in, per column)**: enabling positions costs roughly +50–70% index build time and +39% stored bytes — positions add about as many encoded bytes as the postings themselves. Positionless columns are unaffected: the default build produces byte-identical output (606.55 MiB at 1M docs) in the same time as main. The superfile bench now reports both configs as separate rows — the historical `1 writer` / `N writers` rows keep measuring the default build (and compare clean vs main on the A/B rig), while new `(positions)` rows track the opt-in cost explicitly.
- **Final A/B state**: no FTS row is flagged by the ±5% gate. The single remaining flag in the last run (`token_match (all rows)` p90 +9%) is run-to-run variance — the library code was identical to the preceding run, where that row was unflagged.

## Tests

- Phrase semantics suite against a brute-force BM25 phrase oracle (set + score agreement), including boundary, wrong-order, non-adjacent, overlapping-occurrence, and multi-block corpora.
- Truncated top-k oracle agreement for every pruning site of the ranked atom walk (bare-phrase union, phrase musts, should drag, multi-phrase union, negated phrase), and ids/count parity of the unranked walk against the oracle's full match set in both combination modes. These caught a pre-existing top-k bug on main — on score ties at the kth-best boundary the heap evicted the smaller doc id, opposite to the documented tie-break — fixed here with the tests as its regression guard.
- Positional build/read round-trips for both the in-RAM and spilled builder paths; a synthesized v1 blob read by new code; positions-region CRC corruption rejected; lazy (range-read) open of a positional blob.
- Parser coverage for quoted runs with all sigil combinations, unbalanced quotes, and single-token degradation.
- End-to-end supertable tests: phrase search/count across segments, and the typed error on a positionless column.

### DIFF
--- a/src/superfile/fts/builder.rs
+++ b/src/superfile/fts/builder.rs
@@ -100,6 +100,7 @@ use crate::superfile::{
         bm25,
         dict::{DictBuilder, StreamingDictBuilder},
         fst_value::FstValue,
+        positions::encode_run,
         posting::{BLOCK_LEN, Block, EncodedBlock, encode_block},
         tokenize::{AsciiLowerTokenizer, Tokenizer},
     },
@@ -151,6 +152,14 @@ type TermIdMap = HbHashMap<&'static str, u32, FxBuildHasher>;
 /// and cleared before the arena is reset, so only its bucket allocation
 /// persists across calls.
 type DocTfMap = HbHashMap<&'static str, u32, FxBuildHasher>;
+
+/// Per-doc positional chain-head map — same key lifetime rules as
+/// [`DocTfMap`] (bump-arena keys, cleared before every arena reset).
+type DocPosHeadMap = HbHashMap<&'static str, u32, FxBuildHasher>;
+
+/// Chain terminator for the per-doc position chains
+/// (`FtsBuilder::doc_pos_chain`): no previous occurrence.
+const CHAIN_END: u32 = u32::MAX;
 
 #[derive(Default)]
 struct FinishProfile {
@@ -303,8 +312,16 @@ enum ColumnPostings {
     /// In-RAM term → posting list map. Small builds stay here forever.
     InRam {
         terms: FxHashMap<Box<str>, Vec<(u32, u32)>>,
-        /// Estimated bytes held by `terms` — used to drive the spill
-        /// threshold check. Approximate (see `ACCUM_*_BYTES`).
+        /// Per-term position runs for a **positional column**: the
+        /// concatenated [`positions`](crate::superfile::fts::positions)
+        /// varint runs, one run per `(doc, tf)` pair in `terms`, in the
+        /// same doc order — pair `i`'s run is the `i`-th run in the
+        /// buffer, delimited by its `tf`. Never populated for
+        /// positionless columns (zero cost off).
+        pos_runs: FxHashMap<Box<str>, Vec<u8>>,
+        /// Estimated bytes held by `terms` (+ `pos_runs`) — used to
+        /// drive the spill threshold check. Approximate (see
+        /// `ACCUM_*_BYTES`).
         bytes: usize,
     },
     /// Hash-partitioned spill files plus the per-column term
@@ -368,6 +385,7 @@ impl ColumnPostings {
     fn new() -> Self {
         Self::InRam {
             terms: FxHashMap::default(),
+            pos_runs: FxHashMap::default(),
             bytes: 0,
         }
     }
@@ -1106,6 +1124,20 @@ pub struct FtsBuilder {
     /// keys borrow from that arena under the lifetime invariant described
     /// on [`DocTfMap`].
     doc_tf: DocTfMap,
+    /// Per-doc chain heads for the positional capture path: term →
+    /// index of its LAST occurrence in `doc_pos_chain`. Shares its
+    /// `&'static str` keys with `doc_tf` (same bump arena, same
+    /// lifetime invariant), so it is likewise declared before `bump`
+    /// and cleared before the arena resets. Untouched for
+    /// positionless columns.
+    doc_pos_head: DocPosHeadMap,
+    /// Per-doc occurrence chain: `(position, prev_index)` per token,
+    /// `CHAIN_END` terminating each term's chain. Cleared per doc;
+    /// capacity persists.
+    doc_pos_chain: Vec<(u32, u32)>,
+    /// Scratch for one term's positions while draining a doc (chain
+    /// walk emits them descending; reversed here before encoding).
+    pos_scratch: Vec<u32>,
     /// Per-shard bump arena reused across every `add_doc` call.
     /// Holds the transient `&str` keys of the per-doc tf hashmap.
     /// Reset at the top of each `add_doc` so the leftover bytes are
@@ -1166,6 +1198,9 @@ impl FtsBuilder {
             max_partition_bytes: DEFAULT_MAX_PARTITION_BYTES,
             n_docs: 0,
             doc_tf: DocTfMap::with_hasher(FxBuildHasher),
+            doc_pos_head: DocPosHeadMap::with_hasher(FxBuildHasher),
+            doc_pos_chain: Vec::new(),
+            pos_scratch: Vec::new(),
             bump: Bump::new(),
         }
     }
@@ -1495,6 +1530,7 @@ impl FtsBuilder {
             .as_any()
             .downcast_ref::<AsciiLowerTokenizer>();
         let mut tokens_in_doc: u64 = 0;
+        let positional = self.columns[col_idx].positions;
 
         // Clear borrowed keys before resetting their backing arena.
         // Both containers retain their allocations for the next doc.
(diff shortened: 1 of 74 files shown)


vinaykakade:
## Benchmark: PR vs `main` (same-VM A/B)

PR arm `5499d0c` was built and run side by side with `main` on one Standard_D8s_v7 (1M docs), twice on matching hardware (base arms `8d45bdc` then `4fc663b`). Deltas are PR-vs-main on the same host, so they reflect the change, not cross-host variance.

### Non-phrase: existing (positionless) indexes unregressed

Positions are off by default, so an existing table stays positionless after this merges. That path is unchanged: a query with no quoted phrase routes to the same single-term/OR/AND/must-should kernels as main with the same fan-out, and every addition in this PR sits behind an `if positions` branch (inert when off) or a `has_phrases` branch (false for term queries). Term postings are byte-for-byte identical between the two layouts — positions live in a separate region — so nothing on the positionless term path moved.

The single-superfile **in-memory** tier confirms it (no object-store fan-out, so the lowest-noise read on the kernels): every non-phrase shape within ±3.5% of main in both runs.

| shape | warm p50 Δ |
|---|---|
| single_common / two_term_or / three_wide_or | −0.2% / +0.3% / −5.4% |
| five/ten/twenty/forty_term_or | −1.4% / −0.5% / +3.3% / +3.1% |
| two/three/five/ten_term_and | −1.3% / −0.5% / −0.5% / −0.3% |
| must_common_should / must_rare_should / must_two_should | −3.3% / +0.2% / +1.3% |

Default-config build time and stored bytes are unchanged — byte-identical output (606.55 MiB), build 1-writer +1.5% / 7-writer −0.4%, within noise.

On the multi-superfile / object-store tier, longer queries and all counts are flat-to-better and memory is 2–11% lower, but three high-frequency two-atom shapes (`two_term_or`, `two_term_and`, `must_common_should_common`) reproduce a ~30% warm-latency increase across both runs; other short shapes scatter run-to-run (noise). This tier benches a **positions-enabled** PR index against a **positionless** `main` index (main has no positions), so the two arms differ in the index, not just the code — an apples-to-oranges the in-memory tier and the static argument above both rule out as a kernel regression. There is no positional baseline to A/B against until this merges (`main` can't build one), so the positions-on object-store path is the one thing measured only against a positionless reference; on realistic (non-pathological-frequency) term queries it does not surface, and on positionless columns it does not exist.

### Phrase queries (new capability)

Exact phrase search is new, so these have no main baseline. Multi-superfile / object-store, warm latency and native count (counts are exact — every candidate is adjacency-verified):

| query | warm min / p50 / p90 | matches | count() |
|---|---|---|---|
| phrase_two_mixed | 1.52 / 1.58 / 1.73 ms | 4.2K | 4.75 ms |
| phrase_two_common | 5.15 / 9.06 / 13.06 ms | 648.9K | 34.99 ms |
| phrase_plus_must_term | 9.79 / 12.78 / 14.03 ms | 565.1K | 33.79 ms |
| phrase_three_common | 16.87 / 20.92 / 22.87 ms | 34.2K | 40.79 ms |

`phrase_two_mixed` (a common word paired with a rarer one) is the representative shape; the `_common` rows pair the two most frequent Zipfian terms, so their phrase matches ~65% of the corpus — a deliberate stress shape for the intersect-then-verify pipeline, not typical selectivity. On a real corpus (the external suite) phrase-tag latency is well under a millisecond at the median.

### Positional storage (opt-in, per column)

Positions are off by default. Enabling them roughly doubles the FTS footprint and adds build time — measured as its own bench rows so the cost is explicit:

| | default (positionless) | positions on |
|---|---|---|
| stored (single-superfile, 1M) | 606.55 MiB | 840.55 MiB (+39%) |
| build 1-writer / 7-writer | 13.81 s / 3.32 s | 23.22 s / 4.98 s |

This matches the cost of storing positions in any inverted index; columns that don't opt in pay nothing.

src/superfile/fts/builder.rs
```diff
@@ -1376,60 +1550,119 @@ impl FtsBuilder {
...
-            tokenizer.tokenize_each(text, &mut on_token);
+            // Positional twin of the scan: same interning + dense-tf
+            // bump, plus each occurrence links its position into the
+            // per-doc chain (the same chain the in-RAM path uses),
+            // headed per term in `dense_doc_poshead`.
+            self.doc_pos_chain.clear();
+            let doc_pos_chain = &mut self.doc_pos_chain;
+            let mut on_token = |tok: &str| {
+                let position = tokens_in_doc;
+                tokens_in_doc += 1;
```
naman-modi: We only give a position number to tokens the tokenizer keeps. if the tokenizer drops a token, it gets no number, and it doesn't leave an empty slot behind either. so the words around it become neighbours.

example: `new café york`. café is dropped (non-ASCII), so we store new=0, york=1. now they look adjacent, and "new york" matches a doc that actually says "new café york", and that wouls be a wrong match

Today it's narrow: the ASCII tokenizer only drops non-ASCII. but Tokenizer is public, so the first custom tokenizer that drops stopwords quietly breaks phrase matching. no error, no failing test.

Is this intended? either count every token the tokenizer sees (so a dropped one still leaves a gap), or document that phrases only match on kept tokens and a custom tokenizer must leave a gap for anything it drops. Either way, can we add a test: "new york" should NOT match "new café york"?

tests/superfile/fts/phrase.rs
```diff
@@ -0,0 +1,295 @@
...
+        r#"+"inverted index""#,
+    ] {
+        assert_matches_oracle(&r, &oracle, query, BoolMode::Or, K_ALL).await;
+    }
+}
+
+// ── multi-block corpus (positions cross skip boundaries) ─────────────
+
+#[tokio::test]
+async fn phrase_oracle_agreement_multi_block() {
```
naman-modi, line 216: Hey Vinay, nothing here exercises the block-crossing rejection path, we only ever check docs that match. wrong-order/non-adjacent rejection is only in the 8-doc single-block corpus.

Can we also have a  query "alpha gamma", and assert they don't match? That pins the block-crossing offset logic on a real rejection.

src/supertable/query/fts.rs
naman-modi: Hey Vinay, not related to this PR completely, but I don't think `Parquet` is the correct error for this. Would something like  FtsError::PositionsUnavailable { .. } => QueryError::InvalidQuery(e.to_string()) make more sense ti the user?

naman-modi (approved):
Overall LGTM, few comments we can think about...

src/supertable/query/fts.rs
vinaykakade: Done - here and other places in fts.rs

tests/superfile/fts/phrase.rs
vinaykakade: Thanks, done in a new test ` phrase_block_crossing_rejection`.

src/superfile/fts/builder.rs
vinaykakade: Good catch. The tokenizer now advances the position when a token is dropped to fix this issue. Added a test for "new york" as suggested.

PR #425, opened 2026-07-21
supertable: default fine probe to four, multipart large slow-state, + bench tooling

muralikpbhat:
## Summary

Three changes on top of the global vector index, coming out of large-scale (100M) validation of the adaptive cell split.

## 1. Multipart the slow vector-state blob above a size threshold (`95761b9`)

Part of the ongoing adaptive cell-split work at large scale. The per-superfile slow vector-state blob (fp32 centroids + membership) grows with cell size, and at 100M-scale drains it exceeds the object store's single-PUT ceiling (~5 GiB), failing the drain with a 413. This routes the write through the existing multipart upload path above a 100 MiB threshold — the same chunking superfiles already use — while blobs below the threshold keep the atomic single-PUT path. Unblocks large-cell drains.

## 2. Lower the default fine-cell probe to four (`cb82a8b`)

The per-cell fine IVF probe (`DEFAULT_CELL_FINE_NPROBE`) was 8. Sweeping it against reopened 1M and 10M tables shows recall is **flat from 4 upward**, so 8 only inflated the per-query cold read. Dropping the default to 4 is free on recall and roughly **halves the steady-state cold GET count at scale**.

Also adds an `INFINO_VECTOR_FINE_NPROBE` override (unfiltered queries only, off by default) to sweep the fine probe without disturbing filtered queries, which keep their own deeper fine floor.

Measured, same-build reopens (Azure Blob, cells 1..1, sq8 fixed-residual). Cold is split into the **first** cold query (pays a one-time metadata warmup) and the **steady** cold query (each subsequent cold query):

| scale | probe | recall unf / filt | warm p50 | warm $/1M | first-cold: lat · GET · MiB · $ | steady-cold: GET · MiB · $ |
|-------|-------|-------------------|----------|-----------|----------------------------------|-----------------------------|
| 1M | 8 | 0.995 / 0.894 | 2.01 ms | $0.07 | 161 ms · 2 · 11.50 · $1.20 | 1 · 11.00 · $0.40 |
| 1M | 4 | 0.995 / 0.894 | 2.06 ms | $0.07 | 198 ms · 2 · 11.50 · $1.20 | 1 · 9.00 · $0.40 |
| 10M | 8 | 0.996 / 0.945 | 3.87 ms | $0.10 | 462 ms · 6 · 72.85 · $2.80 | 4 · 28.72 · $1.60 |
| 10M | 4 | 0.996 / 0.945 | 3.86 ms | $0.09 | 425 ms · 5 · 58.85 · $2.40 | 2 · 10.00 · **$0.80** |

Cost basis: the `$` columns are object-store **request** cost (~$0.40 per GET per 1M queries; same-region reads, egress free). **MiB is a transfer/latency signal, not a cost driver** — cost only moves when the GET *count* changes. Only the first-cold *latency* is measured; the steady query is reported by GET/MiB/$.

- **Recall: no cost.** Unfiltered flat (0.995 / 0.996); filtered flat (0.894 / 0.945) and identical at 8 vs 4 — the override never touches the filtered path.
- **Steady cold is the win.** At 10M, GET 4 → 2, 28.72 → 10.00 MiB, **$1.60 → $0.80 (−50%)**. At 1M steady is GET-floored at 1, so cost holds at $0.40 (bytes nudge 11 → 9, latency only).
- **First cold barely moves — the centroid tax.** The first cold query fetches the fp32 centroid section covering *all* cells (~5 MiB at 1M) plus a data prime, a fixed cost that does not scale with the fine probe. So first-cold is flat at 1M ($1.20, 11.50 MiB) and only −14% at 10M ($2.80 → $2.40). This is why the headline cold latency barely improves (462 → 425 ms). Fetching only the probed cell's centroid slice (the format already stores per-cell slices) would remove this tax and recover small-scale scaling — a separate follow-up, not part of this change.
- **1M vs 10M.** At 1M the probed cell holds only ~4 fine clusters, so probing 8 vs 4 reads the same set; the dial only bites once cells are large enough to hold more (10M+). The 1M 161-vs-198 ms first-cold latency is single-shot noise.

## 3. Bench: local-fs store backend + force-compact knob (`bbf32bb`)

Test tooling. `INFINO_BENCH_STORE=local` runs the supertable benches against a local-filesystem provider on fast disk, isolating index behavior from object-store write-path limits while keeping the full OCC + multipart commit path. `INFINO_BENCH_FORCE_COMPACT` forces the compaction pass on a reopened, already-drained table, so the compaction/split path can be exercised when reusing an existing prefix rather than only in a full drain run.

## 4. Config-drive the fine-probe knobs; add opt-in proportional depth (`32f34c2`)

Addresses the review feedback on env-var litter. The `INFINO_VECTOR_FINE_NPROBE` override added in §2 is **replaced** by two config fields:

- `vector.fine_nprobe_floor` (default `4`) — the per-cell fine-probe floor. The user-table routing default takes this; the hidden index's per-table manifest stamp still overrides it.
- `vector.fine_nprobe_pct` (default `0.0`) — the proportional fine-probe fraction.

Both live in `config.yaml` (discoverable), and are retunable at runtime through the layered override files (`./infino.yaml` or `~/.config/infino/config.yaml`) with **no rebuild** — the embedded `config.yaml` is `include_str!`'d, but those files are merged over it at load. This removes the raw `std::env::var` read from the query hot path and drops the `DEFAULT_CELL_FINE_NPROBE*` constants from `list.rs`; the routing default now reads from config.

### Why `fine_nprobe_pct`, and why it defaults to `0.0`

The per-cell probe becomes `max(floor, floor(pct × the cell's fine-cluster count))`. A fixed absolute probe (§2's floor of 4) is a **shrinking fraction of a cell as the cell grows**: a cell's fine-cluster count scales with its row count (≈ one fine cluster per ~945 rows), so 4 clusters is ~10% of a 10M/256-cell but ~0.3% of a 100M/128-cell — and recall tracks that fraction, not the absolute count. Probing a constant *fraction* holds recall as cells grow.

`pct = 0.0` makes the proportional term inert (`max(floor, 0) = floor`), so the default path is **bit-identical** to the fixed-floor behavior in §2 — no production behavior change. The proportional depth is **opt-in** (`pct > 0`) pending large-scale validation that it is the right lever; shipping it off-by-default lands the mechanism, config-driven, without committing to a recall/cost trade-off that isn't yet measured at scale. A unit test covers the `pct > 0` gating.


src/supertable/query/vector.rs
```diff
@@ -1358,6 +1358,19 @@ impl SupertableReader {
             } else {
                 CellRoutingParams::default()
             };
+            // Diagnostic sweep knob: override the fine-first gate for
+            // UNFILTERED vector queries only. Off by default (unset ⇒ the
+            // manifest routing value). Filtered queries keep their own fine
+            // floor untouched, so the knob never widens or narrows a
+            // filtered sweep.
+            if !filtered
```
ekechinwokah: claude has been pretty loose with env var litter. do we need this or can it go info config yaml?

benches/utils/supertable.rs
```diff
@@ -2977,6 +2977,11 @@ pub mod vector {
                     == Some("1");
             let skip_vector_delta =
                 std::env::var(SKIP_VECTOR_DELTA_ENV).ok().as_deref() == Some("1");
+            // Force compaction (optimize) + post-compact on a reopened, already-drained
+            // table -- otherwise a drained existing prefix reports phase-derived only and
+            // never exercises the split.
+            let force_compact =
```
ekechinwokah, line 3131: same as my later comment re: env var litter.

src/supertable/query/vector.rs
muralikpbhat: Agreed, moved this one to config.

benches/utils/supertable.rs
muralikpbhat: keeping the bench env as such, but moved the production setting to config.

ekechinwokah (approved):

PR #427, opened 2026-07-21
fts: byte-quantize the per-doc length-norm table (4x smaller)

vinaykakade:
## What

Quantize the per-doc BM25 length normalizer (`dl_norm_k1`) from an `f32`
table (4 bytes/doc) to one byte per doc plus a 256-entry decode table.

## Why

The length-norm table is indexed once per scored doc. Held as `f32`, it is
`4 × n_docs` bytes — at multi-million-doc scale it no longer stays
cache-resident, so each scored doc pays a scattered load from a table that
overflows cache. Byte-quantizing the doc length cuts the resident table 4×
and keeps the decode table (1 KiB) L1-resident:

| docs | norm table (before) | after | reduction |
|---|---|---|---|
| 1M | 4.0 MB | 1.0 MB | −75% |
| 5M | 20.0 MB | 5.0 MB | −75% |
| 10M | 40.0 MB | 10.0 MB | −75% |

A scored doc now reads `lut[bytes[doc]]` — one load from the small per-doc
table plus one L1 lookup.

## How

- Each doc's length is quantized to one byte via an 8-bit float (3-bit
  mantissa, truncating so the map is monotonic and idempotent). Relative
  length error ≤ 12.5%, and BM25 feels only a fraction of that since the
  length enters the norm as `1 - b + b·dl/avgdl` (the `1 - b` term carries
  no error).
- A 256-entry table decodes each bucket to its `dl_norm_k1` value.
- Built at reader open from the existing on-disk doc-lengths array — **the
  on-disk format is unchanged.**

## Correctness

- Documents up to length 15 quantize exactly, so existing BM25 oracle tests
  are bit-identical.
- New codec unit tests: exactness below threshold, monotonicity, bounded
  relative error, idempotence.
- New long-doc ranking-preservation test: on varied, well-separated lengths
  the quantized top-k equals the exact-BM25 top-k element-for-element.
- New memory-footprint guard test: the resident table stays one byte/doc.
- Full FTS unit + integration suites pass; `fmt`/`clippy` clean.

Localized to the FTS read path; no public API or on-disk format change.

### DIFF
--- a/src/superfile/fts/reader.rs
+++ b/src/superfile/fts/reader.rs
@@ -225,6 +225,75 @@ fn two_term_has_rare_anchor(cursors: &[TermCursor]) -> bool {
     lo > 0 && hi >= lo.saturating_mul(WAND_BMW_2TERM_DF_RATIO)
 }
 
+/// Per-doc BM25 length normalizer, quantized to one byte per doc.
+///
+/// The scorer needs `dl_norm_k1[doc] = K1·(1 - B + B·dl/avgdl)` for
+/// every scored doc. Held as an `f32` per doc, that table is 4 bytes ×
+/// n_docs — at multi-million-doc scale too large to stay cache-resident,
+/// so each scored doc pays a scattered load from a table that overflows
+/// cache. Instead the doc length is quantized to one byte
+/// ([`bm25::quantize_len`]) and a 256-entry table decodes each bucket to
+/// its norm value: the per-doc table is 4× smaller (one byte), and the
+/// decode table is 1 KiB (L1-resident). A scored doc reads
+/// `lut[bytes[doc]]` — one load from the small per-doc table plus one L1
+/// lookup — instead of one load from a 4×-larger table.
+#[derive(Debug, Clone)]
+pub struct NormTable {
+    /// Per-doc quantized length bucket. Empty for a column with no docs.
+    bytes: Vec<u8>,
+    /// Bucket → `K1·(1 - B + B·dequantize_len(bucket)/avgdl)`. 256
+    /// entries; empty when the column is empty.
+    lut: Vec<f32>,
+}
+
+impl NormTable {
+    /// Build from a column's per-doc lengths and average length. An
+    /// `avgdl` of `0.0` (empty column) yields an empty table; it is
+    /// never indexed because `search` short-circuits on empty columns.
+    fn new(doc_lengths: impl Iterator<Item = u32>, n_docs: usize, avgdl: f32) -> Self {
+        if avgdl <= 0.0 {
+            return Self {
+                bytes: Vec::new(),
+                lut: Vec::new(),
+            };
+        }
+        let inv_avgdl = 1.0_f32 / avgdl;
+        let lut: Vec<f32> = (0..=u8::MAX)
+            .map(|b| {
+                let dl = bm25::dequantize_len(b) as f32;
+                bm25::K1 * (1.0 - bm25::B + bm25::B * dl * inv_avgdl)
+            })
+            .collect();
+        let mut bytes = Vec::with_capacity(n_docs);
+        for dl in doc_lengths {
+            bytes.push(bm25::quantize_len(dl));
+        }
+        Self { bytes, lut }
+    }
+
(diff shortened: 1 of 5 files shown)


vinaykakade:
## No FTS performance regression (same-host A/B vs `main`)

The Azure benchmark ran a same-host A/B — this branch vs `main` on one VM — for the FTS path. The `supertable_fts` and `supertable_sql` arms are green, and FTS warm p50 latency is unchanged within run-to-run noise.

**FTS search — warm p50 (this branch vs `main`):**

| shape | main | this PR | Δ |
|---|---|---|---|
| single_rare | 242.7 µs | 245.3 µs | +1.1% |
| single_common | 725.9 µs | 756.3 µs | +4.2% |
| two_term_or | 262.3 µs | 242.2 µs | −7.7% |
| two_term_and | 346.7 µs | 271.9 µs | −21.6% |
| three_wide_or | 526.7 µs | 500.1 µs | −5.1% |
| five_term_or | 1.07 ms | 1.11 ms | +3.7% |
| ten_term_or | 1.78 ms | 1.83 ms | +2.8% |
| twenty_term_or | 2.92 ms | 3.09 ms | +5.8% |
| forty_term_or | 5.01 ms | 5.23 ms | +4.4% |
| must_common_should_common | 362 µs | 267 µs | −26.3% |
| phrase_two_common | 9.60 ms | 9.84 ms | +2.5% |

**FTS top-1000 — warm p50:** single_common −1.9%, ten_term_or −5.9%, two_term_and +1.2%, two_term_or +2.6%.

Every warm-p50 change is within **±6%** (worst +5.8%), and the deltas are **mixed** — several shapes are faster, some by a lot — rather than a systematic slowdown. That's run-to-run A/B noise, not a regression. This is expected: the change only shrinks the resident length-norm table 4× (4 → 1 byte/doc) and is latency-neutral, which the numbers confirm.

**On the summary's "regressions" count:** that tally flags every metric (p90/p99, RSS, cost, cold-open) across all shapes at a low threshold, so it over-reports noise. The warm-p50 latency that matters shows no regression.

src/superfile/fts/reader.rs
```diff
@@ -225,6 +225,77 @@ fn two_term_has_rare_anchor(cursors: &[TermCursor]) -> bool {
...
+/// decode table is 1 KiB (L1-resident). A scored doc reads
+/// `lut[bytes[doc]]` — one load from the small per-doc table plus one L1
+/// lookup — instead of one load from a 4×-larger table.
+#[derive(Debug, Clone)]
+pub struct NormTable {
+    /// Per-doc quantized length bucket. Empty for a column with no docs.
+    bytes: Vec<u8>,
+    /// Bucket → `K1·(1 - B + B·dequantize_len(bucket)/avgdl)`. 256
+    /// entries; empty when the column is empty.
+    lut: Vec<f32>,
```
naman-modi: We can make this [f32; 256] instead of Vec<f32>. `get()` does `self.lut[self.bytes[doc] as usize]`, and the index is a u8 so it's always 0..=255. 

With a fixed array the compiler can prove that and drop the bounds check, and the table is inline instead of a heap pointer chase. `empty()` becomes [0.0; 256].

naman-modi (approved):
LGTM, minor possible optimisation in the the `NormTable::lut`.

src/superfile/fts/reader.rs
vinaykakade: Good call — done in afba2f8a, with one tweak: Box<[f32; 256]> rather than a bare inline [f32; 256]. The fixed-length array still lets the compiler drop the bounds check on the u8 bucket index, but boxing keeps ColumnMeta pointer-sized instead of embedding 1 KiB per column — the non-scoring paths scan columns (name lookup, listing), so I didn't want to bloat that struct.

Benched superfile fts warm (1M docs) before/after: no change beyond run-to-run noise — aggregate p50 −0.6%, heavy OR/AND queries +0.4–1.2%, everything inside jitter. Expected, since the LUT base is loop-invariant either way and the 1 KiB table is L1-resident regardless. So checking it in as a codegen/clarity cleanup (bounds-check elision + the 256-bucket invariant in the type), not as a speedup.

PR #430, opened 2026-07-21
ci(bench): don't check out fork code on the trusted runner

vinaykakade:
## Problem

The Azure bench started failing on **fork PRs** (e.g. any external / fork-based PR). Root cause is a recent GitHub-side change: **`actions/checkout` now refuses to check out fork-PR code under `pull_request_target`** (hardening against "pwn requests"). The runner-side `Checkout (runner-side tooling)` step fails, every later step is skipped, and it surfaces confusingly downstream as `python3: can't open .github/scripts/bench_summary.py` (exit 2).

This is not caused by any product change — the same `@v4` tag changed behavior underneath us, which is why the bench "worked hours ago" and then broke. (Same-repo PRs are unaffected; their vector-arm failures are a separate recall-floor issue.)

## Why not just `allow-unsafe-pr-checkout: true`

The runner is the **trusted** context (base `GITHUB_TOKEN` + secrets) and only runs our own `bench_summary.py` over metrics `scp`'d back from the VM — it never needs the code under test. Checking out fork code there and running a helper script from that checkout is exactly the pwn-request surface the guardrail blocks, and an **environment approval does not make it safe** (a reviewer won't reliably catch a malicious helper-script edit). So re-enabling the unsafe checkout would reopen a real hole.

## Fix

- **Runner-side checkout uses base tooling for fork PRs** (`ref: is_fork_pr ? default_branch : ref`); same-repo PRs keep using the ref under test. No fork code on the trusted runner.
- The fork's **code-under-test still runs** — on the isolated, approval-gated **VM** (the correct boundary for untrusted code), unchanged.
- **Instrument the VM clone** to report whether the ref resolves from the base repo (directly, or via the base repo's PR-head object) and **fail loudly** if not — so we can tell whether the VM clone must target the fork repo directly (thread `head_repo`) instead of silently benchmarking base code.
- Add an `is_fork_pr` `workflow_dispatch` input so the fork-safe path can be smoke-tested by manual dispatch before merge.

## Note on verifying

`pull_request_target` always runs the **base** repo's copy of the workflow, so this change can't take effect on PR bench runs until it's on `main`. After merge, re-running a fork PR's bench will confirm the runner checkout passes and the `[vm-clone]` diagnostic line will tell us whether a `head_repo` follow-up is needed.

### DIFF
--- a/.github/workflows/supertable-bench-azure.yml
+++ b/.github/workflows/supertable-bench-azure.yml
@@ -45,6 +45,11 @@ on:
         required: false
         default: "main"
         type: string
+      is_fork_pr:
+        description: "Simulate a fork PR (runner checks out base tooling, not the ref; routes to the approval-gated env). For verifying fork-safe checkout."
+        required: false
+        default: false
+        type: boolean
       keep_table:
         description: "Keep the bench blob prefix after the run (debug)."
         required: false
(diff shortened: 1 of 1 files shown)


naman-modi (approved):
LGTM.

PR #445, opened 2026-07-23
fts: share single-term top-k floor + deterministic stable-id tie-break

vinaykakade:
**For review.** Two related fixes to the cross-superfile scored top-k path.

## 1. Deterministic, compaction-stable tie-break (correctness)

The supertable top-k merge selected purely by score — no tie-break — so among **score-tied** hits, which ones survived depended on segment completion order (and stayed unstable across runs). A single-segment engine avoids this by breaking ties on a stable, total docid order.

Fix: break ties on the **stable `_id`** — `(score desc, _id asc)`. `_id` is invariant across compaction (unlike physical superfile-uuid/offset keys), so the top-k is now deterministic *and* stable across reorganizations. `_id` resolution stays top-k-sized (partition at the k-th score, keep only boundary ties, resolve those) — never a per-candidate decode.

Also makes the per-block BM25 max decode a guaranteed upper bound (`+1` fixed-point step): `ceil()` on encode plus f32 division could leave the decoded bound a hair below a tied block's true max, letting a rising floor skip that block and drop tied hits. Costs ~1/scale of pruning tightness.

## 2. Restore the single-term cross-superfile floor (perf)

Single-term scored queries were opting out of the shared kth-best floor, so each superfile returned a full local top-k and the fan-out merged ~(superfiles × k) candidates. Sharing the floor for single-term too lets each superfile prune against the running global kth-best. This was originally disabled because of the tie-nondeterminism above — now fixed at the root by (1), so pruning can be restored safely.

## Validation

- **All 214 supertable tests pass**, including the two determinism gates (`repeated_bm25_is_deterministic_under_threshold_sharing`, `fanout_under_concurrency_is_live_and_deterministic`) that fail on a naive floor restore; 71 FTS oracle tests pass.
- `fts-diag` (1M, k=1000): single common-term kernel **2.28ms → ~0.92ms**; small/large AND recovered; no regression on the AND paths.

## Scope note

This resolves the determinism hole and the single-term floor cost. It is **not** the full fix for the broader high-k scored-path slowdown in the same area — that is a separate, still-unpinned issue and is being investigated with real-corpus (not synthetic) measurements.


ekechinwokah (approved):

naman-modi (approved):
LGTM Vinay!

src/supertable/query/fts.rs
```diff
@@ -405,16 +405,18 @@ impl SupertableReader {
...
+                // scan against the running top-k instead of returning a full
+                // local top-k for the merge to re-sort. Without this the
+                // fan-out churns ~(superfiles × k) candidates through the
+                // merge heap at large k, which dominates high-k latency.
+                // Ties stay correct: the floor prunes only scores strictly
+                // below the published kth-best (kernels compare via
+                // `floor.next_down()`), so the merged top-k — score ties
+                // included — matches an uncoordinated run; only the amount
+                // of skipped work depends on segment completion order.
+                let floor = shared.floor();
```
muralikpbhat, line 419: do we need a test asserting single-term always shares the floor or an higher level unit test catching the extra fanout that this resulted in?

PR #446, opened 2026-07-23
supertable/query: cache-backed _id resolution for scored top-k

vinaykakade:
## Problem

The scored full-text fan-out stamps each result's stable `_id` after
top-k selection. That stamping went through a per-superfile
`take_by_local_doc_ids`, which re-runs a scattered Parquet page read
(roughly one page fetch + decompress per hit) on **every** query. On a
real corpus with many pages this dominates large-k scored latency; it is
near-free only on a tiny table with few pages, which is why the in-repo
synthetic benches never surfaced it.

## Fix

Route the stamp through the shared, cache-backed resolution already used
for an id-only projection:

- inline id / contiguous-span arithmetic where they apply (unchanged, cheap);
- otherwise a `_id`-column read via the decoded scalar cache, so a warm
  reader decodes each superfile's `_id` column at most once and later
  queries hit the cache.

New helper `stamp_stable_ids` in `exec/common.rs`;
`attach_stable_ids_to_hits` in `dispatch.rs` now calls it instead of the
per-superfile page-read loop. The per-superfile `attach_stable_ids` is
retained for the vector path.

## Results

5M-document full-text corpus, scored top-k, mean latency per query:

| command | before | after | Δ |
|---|---:|---:|---:|
| TOP_10 | 1,323µs | 722µs | −45% |
| TOP_100 | 4,992µs | 939µs | −81% |
| TOP_1000 | 27,792µs | 1,534µs | −94% |
| COUNT | 980µs | 870µs | −11% |

Median-of-min is sharper still (TOP_1000 37,470µs → 816µs, −98%), so the
shift is distribution-wide, not driven by a few outliers. The gain scales
with `k` because the old path paid the per-hit page read for every one of
the top-k results.

## Correctness

All 298 tests pass, including the brute-force BM25 top-k oracle and both
fan-out determinism gates. `cargo fmt --check` + clippy clean, no new
warnings. No format or public-API change.

### DIFF
--- a/src/supertable/query/dispatch.rs
+++ b/src/supertable/query/dispatch.rs
@@ -53,7 +53,7 @@ use crate::{
         handle::SupertableReader,
         manifest::SuperfileEntry,
         query::{
-            exec::common::{take_rows_byte_source, take_rows_object_store},
+            exec::common::{stamp_stable_ids, take_rows_byte_source, take_rows_object_store},
             superfile_reader::superfile_reader,
             vector::row_id_from_manifest_entry,
         },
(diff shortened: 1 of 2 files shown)


ekechinwokah (approved):

ekechinwokah:
note that `decoded_scalar_cache` will hold the merged text shard's full decoded _id column. that's ~16 MB at 1M rows but ~160 MB _per shard_ at 10M docs. how are you thinking of scaling?

also in pending PR our compaction now sorts merge inputs by id_min, which restores id-ordering in merged user files — if you ever want to swap the full-column decode for a binary search we now have ordering guarantees.

vinaykakade:
> note that `decoded_scalar_cache` will hold the merged text shard's full decoded _id column. that's ~16 MB at 1M rows but ~160 MB _per shard_ at 10M docs. how are you thinking of scaling?

Unless I am missing something, the cache doesn't hold the full _id column; it holds only the projected top-k rows for a given hit set. Concretely: resolve_columns groups hits by superfile (common.rs:428-439) and decodes each superfile's rows with take_by_local_doc_ids(&locals, &["_id"]) (common.rs:526), where locals is just that superfile's hit doc-ids. 
 
> also in pending PR our compaction now sorts merge inputs by id_min, which restores id-ordering in merged user files — if you ever want to swap the full-column decode for a binary search we now have ordering guarantees.

Awesome, will be great to get that merged and will optimize further.

PR #447, opened 2026-07-23
supertable: extend hidden table to support FTS

ekechinwokah:
### Problem

Cold and warm FTS/SQL queries pay an O(N-superfiles) fanout tax: every surviving
user superfile costs its own fetch waves and kernel invocation. At 1M docs /
128 superfiles a cold ten-term OR pays one posting GET per file (128 GETs) and
the warm path pays the same tax in kernel/merge overhead. OPANN already solved
this shape for vectors with a hidden, drain-built acceleration table; this PR
extends that same hidden supertable to FTS.

### Architecture

ONE hidden supertable, one manifest, one drain. Alongside the existing
cell-ordered vector shards, the drain now also merges the user superfiles'
FTS blobs into **term-organized text shards**: a merged FST + concatenated
PFOR postings + positions over the whole drained corpus, with global idf.
Postings keep u32 shard-local doc ids; the shard's parquet stub carries the
`_id` column, resolved through the existing id ladder for the top-k only.
Queries run **two waves** under one shared top-k floor: the text shards, plus
the undrained user tail; deletes are identity-filtered via the hidden deleted
set. A slow-CAS FTS routing blob keeps 1-byte-quantized per-block BM25 bounds
resident (df ≥ 1024), so single- and multi-term kernels admit posting blocks
best-first and fetch only what can beat the floor. Drain-emitted **bigram
terms** (adjacent pairs where both members ≥ 200‰ df) make exact phrases
first-class: `build_atom_cursors` rewrites a k-word phrase onto k−1 pair
cursors with a Σ-unigram-idf override; scores are identical to the unigram
walk (equivalence-tested). The drain merge is a tiered LSM cascade
(fan-in 16) — O(N log N) merged bytes instead of the quadratic chained fold.
Everything is MVCC append + atomic manifest swap; no committed bytes are
mutated.

### Performance at 1M docs (Azure, 4C/8T; standard config)

Warm p50 (bm25_search, k=10):

| Shape | main | branch pre-drain | branch post-drain | post-compact |
|---|---|---|---|---|
| single_common | 0.82 ms | 0.93 ms | ~0.9 ms | ~0.9 ms |
| ten_term_or | 2.51 ms | 2.17 ms | 2.1–3.7 ms | 3 ms |
| twenty_term_or | 3.84 ms | 3.75 ms | 3.6 ms | 4 ms |
| forty_term_or | 6.06 ms | 6.32 ms | 9.5 ms | 7 ms |
| phrase_two_common | n/a* | 9.25 ms | **2.61 ms** | **4.4 ms** |

Note that the 2x increase in some post-compact latencies will be addressed as update to this PR before merging.

Cold, per query on a fresh consumer (data GETs; measured):

| | main, 1M (recorded baseline) | main @ e1f1139, 10M | branch pre-drain, 1M (= main's coalesced path) | branch post-drain, 1M |
|---|---|---|---|---|
| first cold query | **1280 GET** / 3.1 MiB per cold open+search | 134 GET / 33.3 MiB (ten_term_or 526 ms) | 130 GET / 4.1 MiB | **22 GET / 10.3 MiB** |
| steady cold query | — | 128 GET | 128 GET / 3.1 MiB | **12 GET / 2.0 MiB** |
| request cost | — | $85.53/1M queries | $51.20/1M | **$8.80/1M** first, **$4.80/1M** steady |

Main's recorded 1M baseline paid ~10 small GETs per file per cold search
(1280 total); with the coalesced open that becomes one posting GET per file
(128), and the hidden text shards then collapse the per-file fan entirely:
**1280 → 12 GETs steady-cold**. A fresh current-main 1M baseline re-runs
with the 10M A/B and this table will be updated with it.

The bench now gates the branch exactly: steady-cold 128/12/20/12 data GETs
(pre-drain/post-drain/post-delta/post-compact); any increase fails the run.

**Post-compact latencies and the next lever.** The post-compact column runs
above post-drain on broad ORs (forty 9.5 → 11.5–13 ms) because consolidation
concentrates each term's postings into fewer, longer contiguous runs: the
cold/fetch path aggregates them into fewer, larger GETs served by a single
reader stream, and the walk covers longer per-term lists per cursor. Both
are addressable by ranging: split large coalesced reads into parallel
sub-range GETs (the fetch analog of the existing 8-way ranged kernel
slicing), so latency scales with the largest sub-range instead of the total
aggregated bytes. Identified as the first fast-follow after this PR; the
GET-count gates above are intentionally tight so that change must justify
its request-count trade-off explicitly.

One-time maintenance (1M): drain text pass 64.7 s (was 278.7 s before the
tiered merge), optimize re-fold 37.3 s, final 8-run merge 3.7 s.

### Cost trade-off

The hidden text index is a second on-storage copy of the inverted index:
at 1M, user 912 MiB + hidden shards 593 MiB + slow-CAS state 4.4 MiB =
1.47 GiB live (priced total 1.61 GiB vs 1.03 GiB user-only before). Bigrams
are ~70 MB of that. The cost model now prices the live LIST, so this is
visible in every report. In exchange, steady cold queries drop ~10× in
GETs/cost and phrases drop 2–6× warm.

### Interactions with recent main changes

- #445 (stable-id tie-break): adopted; additionally the kth-score tie band is
  now capped per file before id resolution (locals are id-monotone per file,
  so results are identical) — without the cap a heavily tied band read most
  of a shard's `_id` pages per cold query.
- #446 (cache-backed id resolution): the scored-top-k attach is routed back
  through the single `attach_stable_ids` ladder with the decoded-scalar cache
  fronting it. The generic column-resolve path issued concurrent duplicate
  block fetches on 100MB+ shard stubs (op-timeline: same 512 KiB block
  fetched 6–8×; ~15 MiB per cold attach vs ~3 MiB via the ladder). Repeat-
  query caching is preserved; `stamp_stable_ids` is removed.
- Compaction sorts merge inputs by id_min, restoring the merged file's global
  id/time order (this had been silently broken by deleted-first packing) —
  required by the placement bisection and the tie cap's monotonicity
  invariant.

### Also in this PR

- Bench harness: routing-state observability for FTS and SQL lifecycles
  (per-state warm GET/query, cold split with user/hidden/manifest
  attribution, expected-tier assertions), per-shape cold GET/hit logging
  with a >0-hits assertion, speed-per-dollar fixed to 1/(s×$), stored bytes
  priced from the live post-optimize LIST, scale-independent df=1 battery
  term.

### Verification

- 2037 lib + integration tests green; `make public-api` parity (no surface
  change); clippy clean.
- Oracles: BM25 brute-force equivalence on planted corpora; drained-shard
  stats textbook-exact; bigram/unigram score equivalence end-to-end.
- Regression tests added for: gapped-file placement + bare projection,
  bigram phrase results, tie-band determinism (600 tied docs → k smallest
  ids), cold-battery hit assertions, background-fill rollback, settle-waiter
  fill.
- Crash-resume, watermark exactly-once, and delete filtering covered by the
  existing drain test batteries.
- PENDING (running now, will update the PR): 10M A/B vs current main
  (FTS + SQL + vector suites) including the recall@10 ≥ 0.99 gate on the
  standard vector bench.


muralikpbhat:
> **Automated first-pass review** — I'm posting this ahead of my own manual review so you get faster feedback. Treat it as a triage signal, not a final verdict; I'll follow up with a manual pass. I have few concerns at high scale about posting list skewness and the hidden table litter at many columns.

## Review: hidden FTS index

Independent high-effort review of the diff (`e92542c…758ecb9`) across correctness, removed-behavior, cross-file, efficiency, reuse, simplification, altitude, and public-hygiene angles, with the flagged candidates verified against the code. Public-hygiene is clean, and the merge/reader paths correctly reuse the existing PFOR/FST/range-fetch helpers (no hand-rolled codecs). The architecture is sound; the items below are fixable without rework.

**Headline:** three correctness issues worth resolving before merge — one silent index-corruption in the drain (**F1**) and two tie-determinism holes (**F2**, **F3**) — plus a confirmed single-term routing regression (**F4**) and a set of efficiency/cleanup items. None of F1–F3 are covered by the current test suite.

| ID | Severity | Verdict | Location | Finding | Why it matters | Suggested fix |
|----|----------|---------|----------|---------|----------------|---------------|
| F1 | **High** | Confirmed | `src/supertable/writer.rs:3654,3663` | Text drain collects readers via `buffer_unordered` (completion order), then `batch.iter().zip(&readers)` pairs them by index | On any batch with a non-resident superfile / out-of-order completion, entry A's tombstone bitmap + stable-ids are applied to superfile B's postings → the merged shard maps docs to the **wrong `_id`s** and deletes the wrong docs, silently. The `len` guard only catches unequal doc counts | Return `(reader, entry-derived data)` from the future — as the vector drain at `:4557` already does — or use order-preserving `buffered` |
| F2 | **High** | Confirmed | `src/supertable/query/fts.rs:2013` | Per-file tie-band cap assumes local-doc order == `_id` order, with no `is_sorted` guard | Legacy pre-PR compaction packed "most-deleted first" (non-monotone) and the new `id_min` sort isn't retroactive; merged shards likewise assume commit order == `_id` order. A tie-heavy query then drops a smaller-`_id` doc from the true top-k — silent wrong result | Guard the cap on known monotonicity (mirror the vector `bisect`'s `is_sorted` fallback), else resolve the full tie band. Confirm `_id`s can't be user-supplied |
| F3 | Med-High | Confirmed | `src/superfile/fts/reader.rs:1156` (+ OR/AND kernels) | Block-selected per-unit k-heap replaces only on `score > min_score`, discarding local-kth ties before the global merge | A unit seeing >k docs at its kth score drops tied (smaller-`_id`) docs that `select_top_k_stable` never receives → breaks the tie-determinism guarantee *within a unit*, independent of F2 | Carry the full kth-score tie band out of each unit, or tie-break units on `_id` as well |
| F4 | Med | Confirmed | `src/supertable/query/fts.rs:425` | Single-bare-term partition admits at `spread ≥ 1/255`, but the shared `row_can_prune` requires `≥ 16/255` — the two gates the comment says "MUST agree" disagree | A low-spread single-term query is routed to the block-selected path where bounds prune nothing, degrading to a serial whole-shard walk (the 1-vs-8-thread cliff). Latency regression, not wrong results | One source of truth for the admission verdict; use the same spread threshold in both places |
| F5 | Med | Confirmed | `src/superfile/fts/reader.rs:1468,1897` | Per-block `.clone()` of both decoded `u32` posting vecs in the survivor scan, which are then only read | Hundreds of needless allocations per query on the ranking hot path | Immutable borrow (`&decoded[slot].0`, `&decoded[slot].1`) — zero-copy |
| F6 | Med | Plausible | `src/supertable/query/vector.rs:1114` | `user_placement_for_scalar_resolve` awaits a manifest lookup per hit | Up to k serialized manifest round-trips per scalar-projection query, before the already-batched placement step | Collect keys up front and resolve in one batched/concurrent pass |
| F7 | Low-Med | Plausible | `src/supertable/query/fts.rs:996` | `hidden_epoch_has_text()` does an O(shards) `.any()` scan twice per query, no memoization | Repeated full scan on the query hot path for a value fixed within a pinned snapshot | Compute once per query / cache on the pinned reader |
| F8 | Low (latent) | Plausible | `src/supertable/reader_cache/disk.rs:131` | `reader_blocks_background_fill` keys off a store-**global** `n_promotion_waiters`, not per-URI | Contradicts the per-URI quiescence doc-comment; if a waiter ever overlaps live queries, unrelated-URI fills stop yielding bandwidth. Latent today (only the post-drain settle waits) | Make the politeness override per-URI |
| F9 | Med | Confirmed | `src/supertable/writer.rs` (~10 sites from `:3418`) | Raw `eprintln!("[supertable drain] …")` in the production drain path | A library unconditionally spamming stderr on every drain, with no way for the host to silence it | Route through `tracing::debug!`/`info!`, or delete bring-up prints |
| F10 | Med | Confirmed | `src/supertable/slow_fts_state.rs:280`; `writer.rs:4135,3512` | `slow_fts_state` forks `slow_vector_state`'s content-addressed CAS-blob machinery; the FTS drain commit is forked across two sites; the resident-or-fetch reader block is copy-pasted 4× (error strings already drifted) | The publish/verify/GC protocol now lives in copies that can silently diverge; the commit-ordering invariant is defined twice | Extract a generic content-addressed routing-blob helper, a shared `resident_or_fetch_reader`, and one `publish_text_shards_and_stamp` stage |
| F11 | Low | Confirmed | `src/supertable/slow_fts_state.rs:233` | `Vec::with_capacity` on an untrusted `u32` length prefix, before consuming the bytes | A corrupt length aborts the process instead of returning `Err`, violating the module's documented no-panic contract. **Mitigated** by content-hash verify on fetch | Bound the capacity by remaining bytes |
| F12 | Low-Med | Plausible | `reader.rs:4011,2074`; `writer.rs:4145` | Dispatch preconditions enforced by `debug_assert` (stripped in release); an `expect()` encoding a caller invariant ~70 lines away; bigram k>2 phrase correctness held by one equivalence test rather than a shared typed contract | The recurring "invariant by assert, not type" pattern — a future dispatch/anchor change misbehaves silently in release | Encode the shapes in types (non-empty mixed clause, `Phrase` with ≥2 members, a shared anchor-position constant) |

### Cleared / refutable (not bugs)
- **Avro backward-compat** (`manifest/part.rs:404`): the appended `row_group_stats` field is decoded positionally with the new schema, which would desync a pre-PR manifest part — but on-disk backward-compat is a non-concern for a new product. Noted for the record; refutable.
- **Vector-codec guard early-return** (`dispatch.rs:248`): intentional and correct — text shards legitimately carry no vector index.

### Test coverage
F1–F3 are not exercised by the current suite. Worth adding: (a) a concurrent, mixed-residency drain fixture (catches F1), and (b) a non-`_id`-monotone file with a tie-heavy query at the kth-score boundary (catches F2/F3).

ekechinwokah:
overall: agree with most of the automated review. f1/f3/f4/f5/f6/f11/f12 confirmed and fixed, f2 pinned with a test, f8 was already fixed after your diff range, f7 declined on numbers, f9 spun off, f10 half done half deferred with reason. 

f1: real, and the worst of the set. fixed with buffered — keeps the read overlap, restores order. carrying (entry, reader) pairs like the vector drain wasn't enough here: the text merge also needs batch ORDER, shard-local doc ids are assigned in commit order. regression test drives a delayed out-of-order open through a 3-file batch with a tombstone on the slow file — fails on buffer_unordered, passes now. why no test caught it: the options default is 1 file/batch; the test pins 3.

f2: local order == _id order isn't an assumption, it's the table's ordering contract — appends are snowflake-ordered, compaction sorts inputs by id_min over disjoint ranges (deleted-first is job SELECTION order only, the merge re-sorts), drain shards emit locals in commit order. _ids are engine-generated, never user-supplied. pre-PR merged files are the same non-concern as your avro note. an is_sorted guard on the kept ties can't actually protect the cap (dropped smaller ids are invisible post-hoc), so the honest fix is contract + tripwire: the post-compact test now deletes one doc per commit before optimize and asserts the merged file's _id column strictly ascends.

f3: confirmed for the single-term kernel only — OR/AND already push through and_heap_push (ties displace on ascending doc id) and next_down their bars. single-term had both holes: strict-> eviction and bars that pruned tie-band blocks. now uses the same shared push + next_down bars. new planted test: tie pair split across unequal-bound blocks, high-bound block visited first — fails before, passes after. also strengthened the existing equivalence test from scores-only to full (doc, score).

f4: confirmed. both single-term gates now use row_can_prune — one predicate, same 16/255 spread bar as the multi-term path. dominance stays multi-term-only, it's meaningless for one term.

f5: fixed, borrows.

f6: real but smaller than it reads — lookup_superfile_entry is an in-memory scan of the resident manifest, not a round trip. the misuse was per-hit awaits and one-element batches. now memoized per unique superfile + one batched hidden_hits_user_ids call.

f7: declined. get_all_superfiles returns a borrowed slice; the probe is a zero-alloc .any() over the resident entry list — nanoseconds at current shard counts. if the scalar-index family grows hidden entries 100x we memoize per pinned epoch then, on that branch, where the growth actually happens.

f8: already fixed — your diff ends at 758ecb9, 8ba98b0 landed after it and made promotion waiters per-uri. the store-wide n_settle_waiters that remains is deliberate: a settle waiter waits for the whole store to quiesce, so every fill must be allowed to finish.

f9: agree it's wrong for a library, but it's the pre-existing drain idiom on main (9 eprintln sites in writer.rs before this pr — the text drain matched it). tracing is a new dependency and a repo-wide change; spun off as its own issue/pr. happy to fold it here if you'd rather.

f10: partially done: resident-or-fetch is one helper now (the pasted copies had already drifted their error strings — you were right), and the drained-watermark arithmetic is one fn instead of two definitions. deferring the generic cas-blob extraction: the genuinely shared surface is ~50 lines against two different wire formats and error enums, and the sql scalar-index branch adds a third user of exactly this protocol — i'd rather extract once with three concrete users than force a generic over two. the two drain commit sites share shape, not content (fts-only vs combined vector+text commit).

f11: fixed — capacities clamped by remaining bytes; a corrupt count now errors at the first truncated element instead of aborting on allocation.

f12: the two asserts whose release violation returns silently wrong results (dropped negation, empty-side mixed kernel) are typed errors now (FtsError::DispatchInvariant), and the drain expect is a typed error. the rest stay debug_asserts — they fail loud and local in any build that exercises them; encoding them in types buys little here for real complexity cost.

cleared items: agree on both.

on posting skewness at scale: the block-selected kernels exist for exactly this — a mega-term is walked by resident block-max admission, so fetch volume scales with blocks-that-beat-the-bar, not list length; and doc-partitioning means no shard-level skew at all (every shard holds the same doc budget — that's why term-range sharding was abandoned). the 10m gate exercises the worst shapes (single_common, common-must ANDs).

on hidden-table litter at many columns: text shards are per-table, not per-column — all fts columns of a doc range live in one shard blob, so column count grows shard size, not entry count. the thing that does multiply entries is the coming scalar-index family (per-column value shards) on the sql branch; the shard-count cap (target = max(config, total/TEXT_SHARD_MAX_COUNT)) is queued for exactly that.

muralikpbhat:
Thanks for the response. Let us wait until we understand/address the following :

1. Unlike vectors where we genuinely rearrange data to reduce both gets and bytes, here in FTS, bytes that we need to read and amount of CPU work doesn't change much, it just optimizes the number of GETs (which is critical though). I was just hoping S3 gives a cheaper way to transfer range data out of multiple objects :)
2. Common term scale: how can we still make multiple machines work in parallel like before?
3. Is this worth for selective terms which must be pruning already?
4. what happens to multi term and intersection queries which are getting pushed down currently to each shard?
5. Since the gains must be mainly for common terms, should we do this only for such fields (configured or auto-learned)?
6. Is there a different approach for hybrid queries since we might be able to use the global vectors (for pruning) and use the superfiles found from there to prune the super files?
7. Then, the pure 'FTS' query, that too common terms is where we get benefits and is this really worth the complexity?
8. Vector and Hybrid is where the high scale is and cold cost concern is. Does the pure FTS earn this cold specific solution?

ekechinwokah:
needs more testing.

PR #449, opened 2026-07-23
Hosted (remote) transport for the SDK

ashishmishra26:
Adds an optional `remote` feature so a connection can target a hosted endpoint over HTTP instead of local object storage — same API, one argument changed:

```rust
connect("https://host/<db>", api_key)   // vs. connect("s3://bucket/db")
```

### Design
- `Supertable` becomes a thin wrapper over an internal `Table` trait; the local engine handle and a new remote client both implement it, so **one public type serves both targets**. The local path is unchanged — this is additive, and with the feature **off the build is byte-identical to before**.
- `parse_uri` gains an `http(s)://` arm (plain `http` is restricted to localhost so a bearer credential never travels in the clear); `connect` routes a remote target to a synchronous `ureq` client.
- Wire codec: JSON requests, Arrow-IPC responses decoded to `Vec<RecordBatch>`, schema / index-spec translation, and HTTP-status → `InfinoError` mapping.
- `api_key` on `ConnectOptions` (or the `INFINO_API_KEY` env var), sent as `Authorization: Bearer`.

### Feature gating
`remote` is **off by default for the crate** (so an embedder pulls in no networking deps) and **on for the Node and Python bindings** (which thread `api_key` through `connect`).

### Scope (vertical slice)
Wired end to end: `create_table`, `append` (Arrow), `bm25_search`, `query_sql`, `list_tables`, `drop_table`, `schema`. Other table ops return a clear "not yet supported over the remote transport" error. Full op parity, typed-error round-tripping, and the local→cloud docs/quickstart are follow-ups.

### Testing
- Hermetic HTTP tests (`tests/remote.rs`) + wire-codec unit tests, plus a new CI lane (`make check` lints the feature; `make ci` runs `test-remote`) so it can't bit-rot.
- A quickstart example (`examples/remote_quickstart.rs`) and skipped live smoke tests for the bindings.
- Manually validated the full round-trip from Rust, Node, and Python against a running hosted endpoint.


src/catalog/uri.rs
```diff
@@ -110,6 +121,39 @@ pub(crate) fn parse_uri(uri: &str) -> Result<Backend, InfinoError> {
...
+            )));
+        }
+        // A bearer credential must never travel in the clear, so plaintext
+        // `http://` is accepted only for a local endpoint; any other host
+        // must use `https://`.
+        if is_http && !is_localhost(host) {
+            return Err(InfinoError::Backend(format!(
+                "http:// is only allowed for localhost; use https:// for a remote host: {uri}"
+            )));
+        }
```
Pranav2612000, line 147: In what scenario will a user use localhost? If its just for test, can we gate it behind a #cfg(test) ?

Pranav2612000 (approved):

Issue #437, opened 2026-07-21
Cell split OOMs and rewrites the full corpus at 100M (batch-major layout vs cell-major split)


muralikpbhat:
## Summary

The adaptive cell split (`split_overflow_cells` / `split_overflow_cell`, `src/supertable/writer.rs`) is unusable at 100M. During `optimize`/compaction it OOMs (~64 GiB anonymous RSS → SIGKILL on a 64 GiB machine), and even setting the OOM aside it incurs **O(cells × corpus)** write amplification.

## Root cause: layout mismatch (batch-major superfiles vs cell-major split)

The split was designed for a **cell-major** layout (a cell owns its superfiles). The drain now produces a **batch-major** layout — each drain batch writes superfiles spanning ~all cells — so a single cell's rows are scattered across roughly **one superfile per drain batch** (~16–32 superfiles at 100M).

To split one cell, `split_overflow_cell`:
- gathers every superfile that contains the cell (`writer.rs` ~5220), then
- re-decodes and re-packs the **kept** cells too — every *other* cell that merely shares those superfiles (`writer.rs` ~5281–5320, via `load_materialized_rows_by_cell_from_ivf_superfile` + `build_merged_subsection_from_materialized`).

So one split materializes a large fraction of the corpus in RAM (→ OOM) and rewrites all those superfiles. With N over-cap cells each rewriting ~all superfiles, total I/O is O(N × corpus).

## Observed

- 100M, 128-cell grid, 64 GiB machine: OOM at `compacting (optimize: user + hidden)`, ~64 GiB anon RSS, SIGKILL.
- The split is already **sequential** (largest over-cap cell one at a time, atomic commit each), so this is not a cross-cell parallelism problem — the cost is within a single cell's split.

## Options (increasing scope)

1. **Splice kept cells through** — byte-copy the untouched cells' subsections instead of decode+repack. Removes the OOM; the per-split superfile rewrite (write-amp) remains.
2. **Append child + tombstone parent** — write the two child sub-cells as a new superfile, tombstone the parent's rows in place, and let merge repack. Avoids both the OOM and the per-split rewrite (log-structured). Recall layout is preserved (child cells are freshly built; tombstones skip the parent rows). Depends on compaction tolerating superseded inputs (#431).
3. **Cell-aligned drain layout** — lay out superfiles so a cell lives in a few contiguous superfiles; then any split strategy is cheap (root-cause fix).
[...]

PR #450, opened 2026-07-24
Split overflowing hidden-index cells in place (append child, supersede parent)

muralikpbhat:
Fixes the cell-split OOM and O(cells × corpus) write-amplification (#437) with an in-place, log-structured split, and makes that split **recall-safe at scale** by deriving each cell's sub-grid from its own data.

## What
When a hidden vector-index cell grows past `cell_split_doc_cap`, split it **in place** instead of rewriting whole superfiles:
- read **only the over-cap cell's rows** (RSS bounded to one cell),
- partition it with **capacitated k-means** (sizing/seeding described below),
- append the children as **one packed superfile** (child 0 reuses the cell id),
- mark the parent cell **superseded** via a manifest marker — no republish, no removal at split time.

Readers, per-cell counts, merges, and split selection all skip the superseded parent; a later compaction reclaims its dead blocks. Per-split cost: RSS ≈ one cell, writes = 1 child superfile + a few cell-id markers. Total over N splits = **O(corpus)**, not O(cells × corpus).

## Why the split must be data-derived (recall at 100M+)
The coarse grid is trained once, on the first commit. At 100M a fixed grid — even a much finer one — leaves recall on the table: a single coarse cell packs many natural clusters, and the initial centroids never move as data arrives. The split is where the grid re-learns from the data that actually landed. But a *naive* split makes recall worse, not better: if the children don't match the cell's real cluster structure, a query's true neighbors end up in a child the query never probes at nprobe=1.

Three changes make the split data-faithful:

- **Greedy k-means++ seeding (child count > 2).** Random seeding in high dimension drops several seeds into one dense blob and merges the other blobs into one oversized child. D²-weighted greedy seeding spreads the initial centers so each natural cluster gets its own seed — the difference between balanced children and one child swallowing several clusters.
- **Capacitated assignment.** Each row goes to its nearest sub-centroid under a per-child capacity; overflow spills to the next-nearest, strongest-preference-first. This balances child sizes *without* leaving centroid geometry, so split membership still equals query routing.
- **Self-tuned child count.** A fixed `cap` alone doesn't work — the neighbors of one query can straddle a cap-sized cut. So the child count starts at `⌈rows/cap⌉` and rises (more, smaller children) until the assignment reaches a **route-fidelity** target: the fraction of rows sitting in their nearest child, which is exactly what nprobe=1 recall tracks. The right cell size is a function of the incoming data and can't be guessed up front, so the split learns it per cell.

Nearest-centroid assignment == query routing → a split doc lands in the cell its query probes → **split recall at native parity**, now holding even when a coarse cell packs many clusters.

## Validation
Measured with a streaming ingest → drain → optimize → measure loop at a fixed **nprobe=1** (the minimal-GET routing this index is built around), recall@10 vs exact NN, on Azure Blob object store and on localfs (they agree to ±0.001):

- **Split quality at 5M / 10M:** recall holds at **0.977 at 5M** and **rises to 0.984 at 10M** as the split refines the initial 256-cell grid (256 → ~547 live cells) — the data-derived partitioning does what it should, and the small/mid regimes are not degraded.
- **10M steady RSS / OOM: confirmed** — split-time RSS bounded to ≈ one cell (no cliff vs the old ~64 GiB).
- **Beyond 10M, a *separate* coarse-routing issue surfaces** (not a split-quality regression): from 15M on, nprobe=1 recall dips (e.g. 15M → 0.807) while a breadth sweep recovers it to ~0.96 flat — i.e. the neighbors are present but in a cell the query doesn't probe first. The cause is upstream of this PR: the drain keeps assigning new data against the **initial** coarse grid while splits accumulate underneath, so the coarse routing goes stale. The **cap-aware merge** and **grid-refresh-from-new-data** follow-ups below address it; the split changes here are a prerequisite, not the fix for it.
- **100M end-to-end in progress** — the full 5M→100M streaming sweep is running (Blob); numbers and the routing follow-up land as it completes.
- A merge with all-dead inputs now **reclaims** (returns empty) instead of erroring — a latent panic the split made common.

Per-split tracing (starting child count, tuned child count, achieved route-fidelity, child-size spread) is emitted at debug for field diagnosis.

Design writeup: https://github.com/infino-ai/infino/issues/437#issuecomment-5065708897

## Tests
Coverage for every path — split (supersede parent, grid growth, doc conservation, split-then-merge reclaim), merge skip/reclaim of superseded cells, manifest field roundtrip, reader skip — plus the split-quality guards: **k-means++ balances many equal clusters** at child count > 2, and the **self-tuning raises the child count** on a cell packing more clusters than the starting count can hold cleanly. `make ci` green (fmt + clippy + all test targets incl. integration).

## Follow-ups (tracked, not in this PR)
- **Coarse routing goes stale as data grows** — the drain keeps assigning against the initial coarse grid; past ~10M this depresses nprobe=1 recall (breadth recovers it). Refresh the coarse grid from the accumulated data rather than the first commit.
- **Corpus realism** — the training cluster count is flat above ~5M; it should scale sub-linearly with corpus size so the coarse grid stays proportionate at 100M/1B.
- **Cap-aware merge** — post-split, a later merge can fold cells back over cap (re-coarsening); the merge should refuse a union that exceeds cap and/or re-split after merging.
- **Dead-fraction reclaim eligibility** — at 100M+ the superseded blocks live in large batch-major drain superfiles the raw-size compaction gate excludes; they need dead-fraction-triggered eligibility (+ a 1-input reclaim rewrite). At ≤10M, parents are under target and reclaim via normal merge.
- **Parallelize the split under an RSS budget** — currently serial per cell (RSS-cheap); for 100M/1B, bound concurrent splits by an RSS budget, compute-parallel / commit-serial through the manifest CAS.

## Why productionize the split (even though it's optional)

Measured on a streaming-ingest recall benchmark (synthetic multi-modal 1024-d cosine corpus, nprobe=1), at 10M documents:

| config | recall@10 | cells | trigger |
|---|---|---|---|
| modality split (R=4 + self-tune) | **0.991** | ~3548 | modality-driven |
| fixed grid (no split) | 0.983 | 256 | — |

The split refines the grid for ~1 recall point over a fixed grid. The cost is bounded and off the hot path:
- **Front-loaded and convergent** — the grid moves 256 → ~3541 on the first drain, then only +7 across the next 8M; the `≤ R`-modes stop keeps the cell count from running away under streaming.
- **Periodic maintenance only** — split runs on the optimize/compaction cadence, never on ingest.

It ships **off by default** (`cell_split_modality_d = 0.0`) because a fixed grid already holds ~0.98 well past 10M and the finer grid costs ~14× the cells; enabling it is the right call once the recall/cost trade is validated at 100M.


src/supertable/opann.rs
```diff
@@ -368,31 +373,90 @@ fn dequantize_row(row: &EncodedCellRow, dim: usize) -> Vec<f32> {
...
+            cents.extend_from_slice(&c);
+        }
+        return (cents, assign);
+    }
+
+    // k-means: assign each row to its NEAREST sub-centroid — identical to
+    // query-time routing, so a split doc lands in the very cell its query
+    // probes. Trains on a sample; assigns the full cell under `metric`.
+    if use_kmeans {
+        let sample_n = rows.len().min((k * 2048).max(4096));
```
ekechinwokah: ask claude to adhere to agents.md. these should be moved to constants at top of file. there are a few others like this.

src/supertable/opann.rs
```diff
@@ -368,31 +373,90 @@ fn dequantize_row(row: &EncodedCellRow, dim: usize) -> Vec<f32> {
...
+                    let d = distance(metric, &rv, &cents[c * dim..(c + 1) * dim]);
+                    if d < best_d {
+                        best_d = d;
+                        best = c;
+                    }
+                }
+                assign[i] = best as u32;
+                populated[best] = true;
+            }
+            // Accept any non-degenerate result (≥ 2 populated sub-cells). We do
```
ekechinwokah: hmm. accepting any ≥2 children even when over cap means the outer loop is the only bound. you have `MAX_SPLITS_PER_OPTIMIZE` which helps, but under pathological collapse→diameter-cut churn this might still thrash?

src/supertable/handle.rs
```diff
@@ -4149,6 +4175,164 @@ mod tests {
...
+            live_docs, N as u64,
+            "post-merge physical docs equal N — no loss, no resurrection"
+        );
+        assert!(
+            manifest.get_superseded_cells().is_none_or(|m| m.is_empty()),
+            "the reclaimed parent's superseded marker is dropped by the merge"
+        );
+        // Retrievable after the merge (routing/recall is a separate concern; a
+        // split needs nprobe >= 2 to cover both sub-cells near the boundary).
+        assert!(live_hit_count() > 0, "docs still resolve after the merge");
```
ekechinwokah: i think `split_then_merge` softens to `live_hit_count() > 0` after merge. might want to assert == N after split and after merge. "> 0" doesn't catch silent loss or half-recall.

src/supertable/opann.rs
muralikpbhat: Given I fallback to median-cut in pathological case, which guarantees an ~equal split, it can't thrash. However, will consider adding that explicit guard with MAX_SPLITS_PER_OPTIMIZE.  k-means doesn't balance, but median does, but the recall is bad with median compared to k-means.

src/supertable/opann.rs
muralikpbhat: Done — moved to named constants at the top of `opann.rs`:

| literal | → const |
|---|---|
| `2048` | `SPLIT_KMEANS_SAMPLE_PER_CLUSTER` |
| `4096` | `SPLIT_KMEANS_SAMPLE_MIN` |
| `10` | `SPLIT_KMEANS_ITERS` |
| `0x5157_5f4b_4d45_414e` | `SPLIT_KMEANS_SEED_XOR` |

src/supertable/opann.rs
muralikpbhat: Can't thrash: the fallback is a diameter cut into `K = ⌈rows/cap⌉` equal-population bins, so every child is `≈rows/K ≤ cap` in **one pass** — no re-split. The k-means path accepts only ≥2 populated (each child strictly smaller than the parent → monotonic shrink), and a still-over-cap child re-splits until near-identical rows collapse to <2 populated and hit that balanced cut. Termination is guaranteed by construction, so **`MAX_SPLITS_PER_OPTIMIZE` is redundant for correctness** — a pure safety backstop, never the actual bound.

Added a test, `plan_sq8_split_kway_median_balances_skewed_cell_under_cap_in_one_pass`, asserting the K-way median puts every child `≤ ⌈rows/K⌉` in one pass, even on a density-skewed cell.

src/supertable/handle.rs
muralikpbhat: Agreed, changed it — the retrieval check now probes exhaustively (every cell) and asserts `== N` after split **and** after merge (was `> 0`), so a silent drop or half-recall now fails the test. (Physical `live_docs == N` conservation was already asserted; this adds the retrievability side.)

src/supertable/opann.rs
muralikpbhat: This did not work as expected at higher scale, even with both kmeans and median cut,  and I had to change this completely, see the next update.

ekechinwokah:
Automated Review.

## Review Summary

I reviewed against PR head `53e9597` / merge-base `0612afd`, including the two large diffs (`opann.rs`, `writer.rs`) that GitHub didn't render.

### 1. Significant — superseded cells are excluded from routing counts but not from fine scoring or fetch

The supersede filter was added in exactly one place — `postings_by_cell_from_summaries` (`query/vector.rs:396`) — but the fine-scoring and fetch paths still iterate the full superfile list without filtering superseded cells:

- `score_fine_candidates` (`vector.rs:492`)
- `rescore_deferred_cells` (`vector.rs:1134`)

Both score every summary cell in `superfiles`, keyed by `cell.cell_id`.

Because child 0 reuses `split_cell`'s ID (`insert_split_centroids`, `opann.rs:451`), the cell remains live in the posting map via child 0's count, so it is still selected. The superseded parent (which remains in `superfiles` because it also contains the neighboring cell) then has its dead `split_cell` blocks:

- fine-scored,
- retained by `gate_fine_candidates_by_fragment` (which gates by `cell_id`, not by superfile), and
- fetched through the parent's fan-out.

Correctness is preserved only because of the downstream cross-superfile stable-ID dedup (`vector.rs:1571`, `k_fetch = k + replica_overhead`), so the final results remain correct.

However, there are two important consequences:

- The parent's dead blocks are re-fetched and re-scored on every query that hits the split cell until a merge eventually reclaims that parent. That window could be quite long if dead-fraction compaction never prioritizes it. This directly contradicts the PR's stated behavior ("never routed to, scored, or fetched", `vector.rs:398` and the `opann.rs` header) and violates the project's cost-first design goals.
- The posting-aware probe budget (`candidate_counts`, `vector.rs:1445`) includes postings from the parent's dead blocks, allowing the probe budget to be satisfied early and potentially reducing coverage of other cells.

This is also currently untested.

`split_then_merge_reclaims_superseded_without_resurrection` only asserts the result count (`.len()`), never the number of GETs, so the wasted fetches remain invisible.

**Suggested fix**

Thread the superseded map into both:

- `score_fine_candidates`
- `rescore_deferred_cells`

and skip superseded `(superfile, cell)` pairs using the same three-line guard already present in `postings_by_cell_from_summaries`.

Also add a GET-count assertion to the split end-to-end test.

I'm confident in this from static tracing. If useful, I can build the end-to-end test with GET instrumentation and provide before/after numbers to include in the review.

---

### 2. Confirm — recall@10 ≥ 0.99 with splits actually engaged

I have two questions regarding the `cell_split_doc_cap` change from **500K → 40K**.

First, the updated configuration comment describes the shipping default (256 cells at 10M vectors) as approximately **0.98 recall** (`config.yaml`, `config/mod.rs`).

That is below the repository's stated **0.99 recall** requirement.

Since a 40K cap still leaves the standard 10M benchmark split-free, this appears to be documenting an existing characterization rather than introducing a regression. However, the comment now effectively enshrines **0.98** as the expected behavior, so I'd like confirmation that the measured Recall@10 on the standard benchmark still satisfies the ≥0.99 requirement.

Second, because the standard benchmark never actually triggers a split at this cap, the recall gate never exercises the new split logic.

The entire PR is about the split path, so I'd expect to see a Recall@10 ≥0.99 measurement in a workload where splits are actually engaged (for example, >10M vectors or a deliberately skewed cell).

The existing route-fidelity ≥0.97 unit tests are useful, but they are not a substitute for an end-to-end recall measurement.

---

### 3. Confirm — split supersede stamping assumes single-flight maintenance

`parents` and `superseded_additions` are captured from a pre-commit snapshot (`writer.rs:5223`, `writer.rs:5341`) and replayed verbatim on each OCC retry.

If a concurrent commit were to add another superfile containing `split_cell` between the snapshot and the manifest swap, that new superfile would never be marked superseded and would remain live alongside the children.

This is perfectly safe if hidden-index maintenance (drain, split, merge, additional splits, etc.) is guaranteed to be single-flight for a given hidden table.

I'd just like confirmation that this invariant is guaranteed, since correctness of the supersede stamp depends on it.

---

## Minor

- `superseded_cells` in the DTO lacks `#[serde(default)]` (`manifest/list.rs:1079`). A manifest written before this field existed will fail to deserialize, while a rollback to an older binary silently ignores the field and can resurrect dead cells. This mirrors the existing `tombstone_seqs` pattern (which also lacks `#[serde(default)]`), so it's not new, but adding `#[serde(default)]` to both would make upgrades and rollbacks safer. It would also be helpful to confirm whether manifest versioning already prevents old binaries from reading newer manifests.

- `neighborhood_slice = [split_cell]` (`writer.rs:5215`) is now effectively a one-element array left over from the previous neighbor-rebalance implementation. A simple scalar comparison (`== split_cell`) would be clearer.

- Several tests use inline `std::collections::BTreeSet::from(...)` rather than importing `BTreeSet` at the top of the module. This is a minor deviation from the project's import conventions.

- The `stats.n_docs` override keys off `scalar_schema.fields().len() == 1` as a proxy for "ID-only hidden index" (`builder.rs:1035`). The logic is correct because `n_docs` is derived from `all_stable_ids`, but a short assertion or explanatory comment documenting that the hidden stub is always single-field would make the intent more explicit.

src/supertable/opann.rs
muralikpbhat: Followed up on the remaining literals in the new code: the per-branch recursion seed perturbations are now named consts (`MODALITY_RECURSE_SEED_LEFT`/`RIGHT`) alongside the `SPLIT_KMEANS_*` set. No bare literals left in the split planner.

src/supertable/opann.rs
muralikpbhat: Update: the accept-any-≥2 path + median-cut fallback this thread is about has been removed. The trigger is now modality-driven — a recursive two-means mode count splits a cell into ⌈K/R⌉ groups of ≤R=4 whole modes, and the k-way executor self-tunes k up to a route-fidelity target (capped). The original thrash concern no longer applies: there's no accept-any loop, and the "≤R modes" stop bounds the grid (children settle at ≤R modes and aren't re-split, so the cell count converges). Ships off by default (`cell_split_modality_d=0.0`) since a fixed grid holds ~0.98 past 10M; modality is opt-in pending wider-scale validation.

src/supertable/handle.rs
muralikpbhat: Confirmed on the current branch: the retrieval check asserts `== N` before split, after split, and after merge (no loss, no half-recall).

muralikpbhat:
## Changes since the last review

Five incremental commits on top of the originally-reviewed split (`7e419f1`):

**`365f777` — review response.** Addresses the three inline threads: split-planner magic numbers → named consts (`SPLIT_KMEANS_*`); a K-way termination test proving every child lands at ≤⌈rows/K⌉ in one pass (so the split loop can't thrash — `MAX_SPLITS_PER_OPTIMIZE` is only a backstop); and `split_then_merge` now probes exhaustively and asserts exactly `N` after split **and** after merge (was `> 0`).

**`53e9597` — derive each cell's sub-grid from its own data.** The in-place split had fixed OOM/write-amp but sized and seeded sub-cells naively, cratering nprobe=1 recall when a coarse cell packs many natural clusters (the 100M regime). Now: k-means++ seeding (spreads seeds so each cluster gets one, instead of merging blobs into one oversized child), capacitated nearest-centroid assignment (balances child sizes without leaving centroid geometry, so split membership == query routing), and a self-tuned child count that rises from ⌈rows/cap⌉ until the assignment reaches a route-fidelity target. 1M/10M unchanged.

**`1a749ef` — drain reads and extends the live hidden grid.** The drain re-seeded its assignment grid from the frozen user grid on every pass, discarding the split's growth — so new data routed against the initial coarse grid while refined cells went unused, and streaming recall decayed (0.984 @10M → 0.477 @100M). Now the hidden grid is the source of truth after the first drain: the split grows it, the drain reads and extends it, refinement accumulates. Decay is gone (~0.96–0.997 through 20M, verified at 100M scale); the global-aligned assign-skip is gated to the first drain only.

**`ff1b487` — modality-aware split trigger.** An optional Ashman-D trigger that splits a cell holding more than R=4 well-separated modes into ⌈K/R⌉ children of ~R modes each; the k-way split self-tunes to the route-fidelity target. **Off by default** (`cell_split_modality_d = 0.0`) — the plain doc-cap trigger governs and the cap is 500K.

**`544f2fd` — address the automated review.** (1) Superseded cells are now skipped in `score_fine_candidates` (both its admit-shortlist and scoring/defer loops), not only in routing — so a split parent's dead blocks are never scored, gated, deferred, or fetched, and no longer pollute the probe budget; `rescore_deferred_cells` and the fragment gate are transitively clean, and a new unit test asserts a superseded cell is in neither the scored nor the deferred set. (2) The `cell_split_doc_cap` / `cell_split_modality_d` comments are reframed so ~0.98 reads as the nprobe=1 stress-corpus floor (not a target), with the levers that lift it named. (3) Confirmed hidden-index maintenance is single-flight, so the supersede stamp can't miss a concurrently-added superfile.

Measured impact at 10M (streaming ingest, synthetic multi-modal 1024-d cosine, nprobe=1):

| config | recall@10 | cells |
|---|---|---|
| modality split (R=4 + self-tune) | **0.991** | ~3548 |
| fixed grid (no split) | 0.983 | 256 |

The refinement is front-loaded and convergent (256 → ~3541 cells on the first drain, then +7 over the next 8M as cells settle to ≤ R modes) and runs only as a periodic maintenance pass, off the ingest path.

All commits: `make ci` green (fmt + clippy + full test suite + coverage gate); the last commit's fmt/clippy + affected tests re-verified.

muralikpbhat:
Thanks for the thorough review — addressed in `544f2fd`.

**1 — superseded cells fine-scored / fetched (Significant).** Fixed. `score_fine_candidates` now takes the superseded map and skips superseded `(superfile, cell)` pairs in **both** its admit-shortlist loop and its scoring/defer loop — the same three-line guard already in `postings_by_cell_from_summaries`. Because a superseded cell is no longer *deferred*, `rescore_deferred_cells` (which only walks the `deferred` list, indexing back into `superfiles`) and `gate_fine_candidates_by_fragment` (which gates the candidate set) are **transitively clean** — they never see a superseded cell. So a split parent's dead blocks are no longer scored, gated, deferred, or fetched, and they no longer pollute the posting-aware probe budget. Added `score_fine_candidates_skips_superseded_cells`, which asserts a superseded cell appears in neither the scored candidates nor the deferred set (the two paths that would otherwise fetch it).

**2 — recall ≥0.99 with splits engaged / the config comment.** Confirmed with a splits-engaged measurement: streaming ingest to 10M with the modality split on (R=4 + self-tune, grid grown to ~3548 cells) gives **recall@10 = 0.991 at nprobe=1**, versus 0.983 for the fixed 256-cell grid (table now in the PR body and the "Changes since the last review" comment). On the comment itself — agreed it shouldn't enshrine 0.98; it's reworded so ~0.98 reads as the **nprobe=1 floor on the synthetic multi-modal stress corpus**, not a target, with the levers that lift it named (`fine_nprobe_pct` for within-cell depth, or the modality split → 0.991). Note the cap is now **500K** (not 40K), so the standard 10M benchmark stays split-free by design; the split path is exercised end-to-end by a deliberately split-triggering run (and by the split unit tests).

**3 — supersede stamping assumes single-flight maintenance.** Confirmed. Hidden-index maintenance (drain / split / merge / further splits) is single-flight per hidden table as of now, so no concurrent commit can add a superfile containing `split_cell` between the pre-commit snapshot and the manifest swap. The supersede stamp's correctness relies on that invariant, and it holds.

muralikpbhat:
> **Preliminary** — the 100M split-recall row and the split cold-GET figures are still validating (marked `*` below); this comment will be updated with the final numbers.

## Why the modality split ships on by default

`recall@10` at nprobe=1, **modality split (R=4 + self-tune)** vs the **fixed 256-cell grid**, across the streaming ingest/drain lifecycle.

### 1. Recall across scale

| docs | fixed grid (no split) | modality split | Δ | split cells |
|---|---|---|---|---|
| 100K | 0.940 | **0.998** | +0.058 | ~577 |
| 1M   | 0.962 | **0.997** | +0.035 | ~625 |
| 10M  | 0.983 | **0.991** | +0.008 | ~3548 |
| 40M  | 0.981 | **0.985** | +0.004 | ~3593 |
| 60M  | 0.982 | **0.987** | +0.005 | ~3593 |
| 80M  | 0.981 | **0.988** | +0.007 | ~3593 |
| 100M | ~0.979 | *(pending)* | * | ~3595 |

The split lifts recall above the fixed-grid floor at **every** measured scale, and never regresses it.

### 2. The win is crowding-conditional — biggest where the fixed grid is too coarse

| scale band | behaviour | why |
|---|---|---|
| sub-1M   | large win (**+0.035 – +0.058**) | 256 cells badly under-provisioned; split grows the grid the moment cells cross the modality floor (a 10K→100K stream ramps 0.94 → 0.998) |
| 2M – 10M | small win / near-parity (~+0.005) | grid is already healthy here; treat small deltas as noise, no gate added |
| ≥ 10M    | steady small win (+0.005 – +0.008) | split keeps cells resolved as the corpus grows |

### 3. It holds only under *periodic* optimize() (production cadence)

A **single end-of-build** `optimize()` over the whole corpus collapses at 100M; the periodic cadence does not:

| drain cadence | 100M recall@10 |
|---|---|
| one-shot (single end-of-build `optimize()`) | **0.094** |
| incremental (periodic `optimize()`)         | **~0.98** |

This is a **within-cell** effect, not routing: the drain doesn't refine the global grid (only split grows it), so both runs share the same grid + routing; the loss is in the per-cell fine index, which the incremental cadence builds from many small batches vs one ~390K-doc pass. Split lifts recall on top of that cadence.

### 4. Cold object-store cost — the finer grid *lowers* read cost

Steady cold query (nprobe=1), reopened 100M-scale table on an object-store backend:

| table | cells | cold GET / query | bytes / query | cold $ / 1M queries |
|---|---|---|---|---|
| fixed grid (no split) | 256 | 18 | 108.5 MiB | $7.20 |
| modality split | ~3595 | **9** * | **50.0 MiB** * | **$3.60** * |

`*` split row measured on the 90M reopen (100M resume pending); fixed-grid row is the 100M native table.

Smaller cells mean **smaller per-generation reads**, roughly halving cold GET/bytes/cost. The residual GET floor is the number of un-consolidated drain generations of the probed cell (one per checkpoint) — compaction is meant to fold those to ~1 but is currently blocked at scale by the merge target cap; unblocking it takes the 9 toward ~1 (tracked separately). So split already halves cold IO, with more available once merge consolidation lands.

*(Corpus: synthetic multi-modal 1024-d cosine stress stream; harness #466.)*

ekechinwokah (approved):
please enable benches and confirm

muralikpbhat:
Pre-merge status:

- **Bench**: all four `bench` CI lanes green (superfile, supertable_vector, supertable_sql, supertable_fts @ 1M).
- **Merge conflict with main resolved**: the only conflict was a split unit test against main's `reader() -> Result<...>` signature change — adapted the call sites, engine logic unchanged.
- **Smoke-tested again at 1M**: local `recall_while_ingest`, 1M docs / 100k checkpoints — recall@10 held **0.995–0.999** across all ten checkpoints, no regression from the merge.

PR #451, opened 2026-07-24
fts: cheaper windowed union + route non-dominant large-k ORs to it

vinaykakade:
## Problem

On the sub-range fan-out path, every multi-term OR is scored by MaxScore.
For a **non-dominant** union at large `k` (no single term's score upper
bound dominates), MaxScore can't prune — its essential set never shrinks —
so it scores the whole union with per-doc f-way merge overhead. The
windowed union scorer is the right structure there (O(f) per window), but
two things kept it from being usable as the default:

1. it allocated its per-window scratch (`scores`/`present`) **per call**,
   and the fan-out calls the scorer once per (superfile, sub-range) — many
   calls per query — so small/sparse unions paid allocation they couldn't
   amortize;
2. its drain scanned the **full** presence bitset every window regardless
   of how many docs the window held, so spread unions paid a fixed
   per-window cost.

## Change

- **Reuse the windowed scratch across calls** via a thread-local (each
  reader-pool worker its own; the drain leaves the buffers zeroed, so reuse
  needs no re-zero).
- **Drain only the presence words a window actually set** (tracked in a
  small reused list, sorted for ascending order) — O(touched) per window
  instead of O(window size).
- With the scorer no longer penalizing small/sparse unions, **route
  non-dominant unions to it at large `k`**. A `k`-gate keeps small-`k`
  queries on MaxScore, where its top-k threshold rises fast and block-skip
  pruning stays effective even without a dominant term; dominant unions
  stay on MaxScore at every `k`.

## Correctness

Results are exact — same top-k, same `(score desc, doc asc)` order —
verified against the brute-force BM25 oracle and the fan-out determinism
tests (both pass). The scratch changes are bit-identical; `fmt` + clippy
clean; no format or public-API change.

## Results (5M-document corpus, multi-term OR queries, warm, min-of-runs)

`k`-gate makes TOP_10/TOP_100 run the unchanged MaxScore path — a built-in
control that came out flat (±2%, the measurement floor), confirming small-`k`
and dominant queries are unaffected:

| command | before | after | median Δ |
|---|---:|---:|---:|
| TOP_10 (control) | 183µs | 186µs | +0.8% |
| TOP_100 (control) | 338µs | 343µs | +1.7% |
| TOP_1000 | 1508µs | 1416µs | **−21% median** (−6% mean) |

**Trade-off, called out:** at large `k`, a minority of unions the
upper-bound heuristic classifies as non-dominant are actually prunable and
regress when routed to windowed (worst observed +231%); the scratch/drain
fixes keep that tail bounded. Net is a clear median win at TOP_1000 with no
change at TOP_10/100. Tightening the routing predicate is a follow-up.
Cross-validation at larger scale is pending.

### DIFF
--- a/src/superfile/fts/reader.rs
+++ b/src/superfile/fts/reader.rs
@@ -16,8 +16,10 @@
 //! then it's a borrowed view).
 
 use std::{
+    cell::RefCell,
     cmp::Ordering,
     collections::{BinaryHeap, HashMap},
+    mem,
     ops::Range,
     sync::Arc,
 };
@@ -1698,6 +1700,23 @@ impl FtsReader {
             return Ok(Vec::new());
         }
         // The ranged (sub-range fan-out) path carries no negation in v1.
+        //
+        // Route non-dominant unions (no single term's upper bound dominates)
+        // to the windowed scorer: there MaxScore can't prune and degrades to
+        // scoring the whole union with per-doc f-way merge overhead, while
+        // the windowed scan is O(f) per window. A dominant-term union stays
+        // on MaxScore, whose block-skip pruning is the win there.
+        if prefer_windowed_union(&cursors) {
+            return self.run_windowed_union(
+                column_id,
+                cursors,
+                k,
+                None,
+                floor.next_down(),
+                doc_id_start,
+                doc_id_end,
+            );
+        }
         self.run_max_score_bmm_range(
             column_id,
             cursors,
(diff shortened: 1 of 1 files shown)


vinaykakade:
Closing — no clear evidence this wins at scale.

Benchmarked the branch against main on the 5M-document corpus, full query set:

| command | main | branch | Δ |
|---|---:|---:|---:|
| TOP_10 | 785µs | 735µs | −6.5% |
| TOP_100 | 1052µs | 976µs | −7.2% |
| TOP_1000 | 1664µs | 1527µs | −8.2% |
| COUNT | 947µs | 869µs | −8.2% |

At a glance the branch looks faster everywhere, but that's the problem: this change only affects multi-term OR at large k, so it **cannot** move COUNT (no scoring), TOP_10 (below the k-gate), or intersection (a different path) — yet all improved ~7–8%. That uniform shift is cross-run/instance variance, not the change. Detrended against those unaffected commands as a control, the treated shape (TOP_1000 union, −7.9% mean / −7.2% median) shows **no excess** over the ~−7.7% variance baseline (intersection −8.5% mean).

An earlier same-host A/B had shown a large median win on the treated shape, but it does not reproduce here — the comparison method (fresh instance vs an older committed baseline) can't isolate the effect from instance variance, and a clean same-host A/B at this scale isn't available.

The scratch-reuse and touched-word-drain pieces are correctness-preserving and cheap on their own, but the windowed path isn't reached without the routing, and the routing shows no measurable benefit at scale — so there's no justification to land this as-is. Closing; the analysis is preserved for future reference.

Issue #29, opened 2026-06-02
perf: cache parquet metadata on SuperfileReader to skip footer reparse


vinaykakade:
## Context

PR #27 added \`SuperfileReader::id_lookup(target)\` so callers can scan
the \`_id\` column inline without the previous bytes-extraction round
trip (see commit 68868ff). The API is now clean, but \`id_lookup\`
still rebuilds a \`ParquetRecordBatchReaderBuilder\` on every call,
re-parsing the parquet footer that \`SuperfileReader::open\` already
parsed (see \`src/superfile/reader.rs:175\` — open builds a builder
just to extract the Arrow schema, then drops it).

For the WAL recovery sweep's \`resolve_target_id_in_manifest\`, this is
hot: N superfiles are probed per target id, each probe pays the
footer-reparse tax. At ~0.1–1 ms per parse on typical
(10–100 MB) superfiles, a 100K-id recovery batch over a fragmented
manifest can burn seconds of pure duplicate parquet work.

## Proposal

Cache enough parquet state on \`SuperfileReader\` at \`open()\` time
that \`id_lookup\` (and any future scan-shaped methods) can construct
a reader without re-parsing the footer. Two shapes worth considering:

1. **Cache \`Arc<ParquetMetaData>\`** on \`SuperfileReader\`. Cheap to
   share; the existing \`ParquetRecordBatchReaderBuilder::try_new\`
   call at \`reader.rs:175\` would extract it once. \`id_lookup\` would
   use \`ParquetRecordBatchReaderBuilder::new_with_metadata(...)\` to
   skip the parse on subsequent calls.
2. **Cache the builder itself.** Cheaper code-wise, but the builder
   isn't \`Clone\` and holds state that can't be reused across calls.
   Probably (1) is the better shape.

## Acceptance

- \`SuperfileReader::id_lookup\` (and any other path that touches the
  parquet section) skips the footer reparse on a hot reader.
- Microbench on a 10M-row superfile: \`id_lookup\` should be ≥2× faster
  than today.
- No change to \`SuperfileReader::open\` cost — the parse happens
  there either way; we're just keeping the result.

## Out of scope

- Caching beyond parquet metadata (column readers, decoded row groups,
  etc.) — separate concern.
- Lazy parquet open (open the reader without materializing the
  footer at \`SuperfileReader::open\` time). Probably not desirable
  since \`open\` already pulls the schema anyway.

## Refs

- PR #27 commit 68868ff added the \`id_lookup\` method.
[...]

PR #455, opened 2026-07-24
fts: add a Unicode-aware standard tokenizer, selectable per column

vinaykakade:
## Summary

Adds a second full-text tokenizer, `StandardTokenizer`, alongside the existing `AsciiLowerTokenizer`, and lets each FTS column choose its analyzer (per-field analysis).

`StandardTokenizer` segments text on Unicode word boundaries (UAX #29), lowercases via full Unicode case folding, and **preserves non-ASCII letters and digits**. `AsciiLowerTokenizer` splits on any non-`[A-Za-z0-9]` byte and drops any token containing a non-ASCII byte, so accented and non-Latin text is currently unsearchable; the standard tokenizer makes it searchable and follows the standard Unicode word-breaking rules.

## API

- `IndexSpec::fts(column)` — unchanged; defaults to the `ascii_lower` analyzer.
- `IndexSpec::fts_with_analyzer(column, analyzer)` — new; `"ascii_lower"` or `"standard"`.
- Python: `IndexSpec.fts(column, analyzer="...")`.

Analyzers are per column, so one table may index different columns with different analyzers.

## Implementation

- New `StandardTokenizer` in `superfile/fts/tokenize.rs`; a `Tokenizer::name()` method plus a `tokenizer_for_name()` registry map names to instances.
- The builder tokenizes each column with its own tokenizer (`FtsBuilder::register_column_with_tokenizer`); `add_doc` selects the per-column tokenizer.
- The reader stores each column's tokenizer (reconstructed from the name recorded in the column's stored config) and tokenizes queries per column.
- The analyzer name is persisted in the catalog table entry and rebuilt on `open_table`; rebuild/compaction recovers each column's analyzer from the source superfile.
- The options hash includes the analyzer name only when it is non-default, so tables using the default `ascii_lower` keep a byte-identical hash and existing manifests continue to verify unchanged.

## Compatibility

Existing tables are unaffected: the default analyzer and its behavior are unchanged, the on-disk column config already carried a `tokenizer` field (previously always `ascii_lower`), and the options hash is stable for `ascii_lower` tables.

## Tests

- Tokenizer unit tests: ASCII, non-ASCII, CJK, intra-word numerics/apostrophes, punctuation/hyphen splitting, borrow-vs-copy path agreement, query-side parse, and the name registry.
- End-to-end per-column / mixed-analyzer tables (each column indexed and queried with its own tokenizer), storage-backed reopen, and options-hash behavior.
- Python smoke tests for the `analyzer` keyword.

## Dependency

Promotes `unicode-segmentation` (already present transitively in the lock graph) to a direct dependency.


vinaykakade:
## Benchmarks — no regression on the default path

This change only adds a second tokenizer; the default `ascii_lower` analyzer and its code path are unchanged. The benchmark indexes and queries with the default analyzer, so the `main` arm and this branch run identical code — any movement is run-to-run variance.

The CI A/B benchmark (1M docs, both arms built and diffed on the same VM) confirms this. Representative FTS warm latencies (pre-drain, p50 / p90):

| Shape | main p50 | branch p50 | Δ p50 | main p90 | branch p90 | Δ p90 |
|---|---|---|---|---|---|---|
| single_common | 578 µs | 594 µs | +2.8% | 659 µs | 656 µs | −0.5% |
| two_term_or | 248 µs | 247 µs | −0.4% | 262 µs | 259 µs | −1.0% |
| five_term_or | 1.12 ms | 1.13 ms | +0.9% | 1.21 ms | 1.24 ms | +2.5% |
| ten_term_or | 1.79 ms | 1.82 ms | +1.7% | 1.84 ms | 1.84 ms | +0.0% |
| forty_term_or | 5.12 ms | 5.13 ms | +0.2% | 5.25 ms | 5.41 ms | +3.0% |
| two_term_and | 279 µs | 277 µs | −0.7% | 311 µs | 304 µs | −2.2% |
| ten_term_and | 1.55 ms | 1.54 ms | −0.6% | 1.59 ms | 1.60 ms | +0.6% |
| must_common_should_common | 282 µs | 286 µs | +1.3% | 302 µs | 298 µs | −1.1% |
| phrase_three_common | 19.68 ms | 19.92 ms | +1.2% | 22.17 ms | 22.23 ms | +0.3% |

Warm p50 stays within ±3% across all shapes, in both directions — no systematic shift. The larger raw deltas across the full grid are single-sample tail jitter on sub-100µs rare-term shapes in one lifecycle section; their p50 and their other sections are flat. The p90 bench gate passed on all four suites (superfile, FTS, SQL, vector), and the SQL/vector suites are unaffected since they don't touch FTS analysis.

src/catalog/mod.rs
```diff
@@ -703,11 +713,16 @@ fn build_options(
...
     storage: Option<Arc<dyn StorageProvider>>,
     connection_memory_budget: Arc<ConnectionMemoryBudget>,
 ) -> Result<SupertableOptions, InfinoError> {
-    let mut opts = SupertableOptions::new(schema, fts, vectors, tokenizer)?;
+    // Seed the default tokenizer with the first column's analyzer (None
+    // when there are no FTS columns), then set the authoritative
+    // per-column tokenizers for per-field analysis.
+    let seed = tokenizers.first().cloned();
+    let mut opts =
+        SupertableOptions::new(schema, fts, vectors, seed)?.with_fts_tokenizers(tokenizers);
```
naman-modi, line 736: We also use the tokenizers in SQL query path for pruning the superfiles like [here](https://github.com/infino-ai/infino/blob/main/src/supertable/query/provider.rs#L381), [here](https://github.com/infino-ai/infino/blob/main/src/supertable/query/provider.rs#L412), [here](https://github.com/infino-ai/infino/blob/main/src/supertable/query/provider.rs#L793) via SupertableOptions::tokenizer, so we need to make sure that the correct tokenizer is set there too...else pruning will be wrong. Lets also add a test around it?

src/superfile/fts/tokenize.rs
```diff
@@ -650,6 +672,76 @@ fn is_token_byte(b: u8) -> bool {
...
+            // Borrow directly when every cased character is already
+            // lowercase (the common case for lowercased corpora); only
+            // allocate to case-fold a word carrying an upper/title-case
+            // letter. Non-alphabetic characters (digits, apostrophes)
+            // are unaffected by lowercasing, so they never force a copy.
+            if word.chars().all(|c| !c.is_alphabetic() || c.is_lowercase()) {
+                f(word);
+            } else {
+                buf.clear();
+                buf.extend(word.chars().flat_map(char::to_lowercase));
```
naman-modi:
```suggestion
       buf.push_str(&word.to_lowercase());
```
From what I could read online, and LLMs tell me; that for standardtokenizer, the `char::to_lowercase` and `str::to_lowercase` acts differently, so our `tokenize` would differ from `tokenize_each`, and  that could be as issue, as we ingest via tokenize and SQL uses tokenize_each for pruning.

ref - https://github.com/rust-lang/rust/issues/26035
This can be observed with the word Σ.

src/superfile/fts/tokenize.rs
```diff
@@ -5,27 +5,37 @@
...
+//! [`AsciiLowerTokenizer`] semantics:
 //!   - Split on any byte that isn't `[A-Za-z0-9]`.
 //!   - Lowercase each ASCII letter (bytes `b'A'..=b'Z'` → `b'a'..=b'z'`).
 //!   - Drop any token that contains a non-ASCII byte (high-bit set).
 //!     Non-ASCII tokens are silently dropped (not an error) — the
 //!     ASCII-only design is intentional; richer tokenizers can opt
 //!     into the trait without changing the FTS pipeline.
 //!   - Empty tokens are never emitted.
+//!
+//! [`StandardTokenizer`] semantics (Unicode-aware):
```
naman-modi, line 22: We currently don't do any unicode normalization. This might (in very few cases) be harmful, and produce wrong results.  Unicode defines normalization forms to collapse these to a canonical encoding:
- NFC: prefer the composed single-codepoint form. --> we can prefer this, as its widely used.
- NFD: prefer the decomposed base+combining form.
If we dont at all normalize then its an issue, so inside tokenize, tokenize_each we should normalize, like how other engines also [do](https://www.elastic.co/docs/reference/elasticsearch/plugins/analysis-icu-normalization).

ref: https://unicode-org.github.io/icu/userguide/transforms/normalization/

naman-modi (approved):
We should be good to go, after the above review comments are addressed, and please take an overall look at all the code comments across the repo, so we are not left with any comments mentioning "only 1 tokenizer" etc..

src/catalog/mod.rs
vinaykakade: Thanks, done. Also added tests  candidate::tests::lowering_uses_the_per_column_tokenizer and  provider::tests::value_set_bloom_uses_the_per_column_tokenizer

src/superfile/fts/tokenize.rs
vinaykakade: Fixed, along with a test for final-sigma to make sure the index path and query path behavior is the same.

src/superfile/fts/tokenize.rs
vinaykakade: Keeping it consistent with Lucene, which doesn't normalize in StandrdAnalyzer - https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/analysis/standard/StandardAnalyzer.java

Added a comment stating so.

PR #465, opened 2026-07-27
fts: opt-in global BM25 statistics across superfiles

vinaykakade:
> ⚠️ **Breaking API change.** The public `Supertable::bm25_search` signature changes (the `mode` argument is folded into a new `Bm25SearchOptions`). Per `docs/versioning.md`, a breaking change while pre-1.0 is a **minor** bump — so this lands on the upcoming **0.2.0** line (coordinated across the Rust crate and the bindings via the release workflow; not stamped in this PR).

## Summary

BM25 idf was always computed from each superfile's own document count and term frequencies. Across a table fragmented into many superfiles, a term's idf — and therefore a document's score — depends on which superfile it lands in, so scores are only approximately comparable across superfiles and ranking drifts from what a single unified index would produce.

This adds an opt-in statistics selector to the search surface:

- **`Bm25Stats::PerSuperfile`** (default) — the existing local-statistics path: full fan-out, no extra pass, fastest.
- **`Bm25Stats::Global`** — gathers the corpus-wide document count and per-term document-frequencies once (a bloom-pruned, dictionary-only `df` pass over the superfiles containing the query's scored terms) and scores every superfile against that single table-wide idf, so a fragmented table ranks like one unified corpus.

## How it works

The global idf is injected into the scorer by rescaling each term cursor's stored `block_max`/`term_max` upper bounds by the ratio of global-to-local idf, so BlockMaxWAND/MaxScore pruning stays correct under the substituted statistics. **No on-disk format change** — `df` is read `O(1)` per superfile from the existing dictionary.

Phrase queries are covered too: a phrase's score is the sum of its members' idf, and each member's block-max rescales by the same ratio (which cancels out of the phrase's length/tf bound), so globalizing the members globalizes the phrase for free.

## Scope

- **Covered:** scored term queries (musts + shoulds) and phrase queries, single-column.
- **Not applicable:** the multi-column `search_multi` path is superfile-local only (not on the public/SQL/hosted surface), so within one superfile local statistics already *are* global — there is nothing for `Global` to change there.
- **Unchanged:** length normalization still uses each superfile's own average document length.

## API surface

`mode` and the statistics selector are bundled into a new **`Bm25SearchOptions`** builder (mirroring `VectorSearchOptions` on the vector path), so future BM25 knobs extend the struct without breaking the signature again:

```rust
// defaults: OR mode + per-superfile stats
posts.bm25_search("body", "fox", 10, Bm25SearchOptions::new(), None)?;
// global stats:
posts.bm25_search("body", "fox", 10,
    Bm25SearchOptions::new().with_stats(Bm25Stats::Global), None)?;
```

- **Rust:** `Bm25SearchOptions` + `Bm25Stats` re-exported at the crate root; `Supertable::bm25_search` and the transport trait take `Bm25SearchOptions`. The embedded path forwards to the engine; the hosted path serializes both fields (an older server that does not recognize the `stats` field falls back to the default). `public-api.txt` updated.
- **Python:** `stats=` keyword on `Table.bm25_search`, accepting `"per_superfile"` (default) or `"global"`, plus the `.pyi` stub.

## Testing

- Oracle tests: `Global` over a table split across many superfiles reproduces the single-superfile ranking (same docs, same scores) for both term and phrase queries, while per-superfile stats provably diverge — so the oracle actually exercises the global path.
- The hosted-transport test asserts the `stats` field is serialized on the wire.
- `cargo fmt --check`, `clippy --all-targets --all-features`, all doctests, the FTS unit + integration suites, the hosted-transport tests, and the Python smoke suite (incl. `stats=` coverage) pass.


src/supertable/query/fts.rs
```diff
@@ -480,6 +531,65 @@ impl SupertableReader {
...
+            false,
+            true,
+            move |r, _entry, _sidecars, _now, _params: ()| {
+                let column_arc = Arc::clone(&column_arc);
+                let terms_arc = Arc::clone(&terms_arc);
+                async move {
+                    let mut dfs = Vec::with_capacity(terms_arc.len());
+                    for t in terms_arc.iter() {
+                        dfs.push(r.term_df(&column_arc, t).await.map_err(fts_read_error)?);
+                    }
```
naman-modi: This loops over the query terms one at a time and awaits term_df for each. superfiles run in parallel (fanout_with), but inside each superfile the terms are serial, and each term_df call re-parses the whole FST dictionary......So, for a "N"-term query, every superfile does N FST parses (the FST is the same, should be parsed once) and up to N separate GETs fired one after another. that's the extra pass Global mode pays for, so we're paying it N times over. 

Can we open the `DictReader` once per superfile, look up all N terms against it, and batch the header reads into a single `fetch_term_postings` call? This would potentially turn it into c1 parse + 1 fetch per superfile, and have perf impact.

src/superfile/fts/reader.rs
naman-modi, line 1: Can we add a test with small k (say k=3 or 5) over a corpus with clearly more than k matching docs, so the heap fills and pruning genuinely runs, then assert the top-k still equals the single-superfile top-k?

naman-modi (approved):
Few comments, else LGTM.

src/supertable/query/fts.rs
vinaykakade: Great suggestion, done.

src/superfile/fts/reader.rs
vinaykakade: Sure, done

Issue #436, opened 2026-07-21
Hidden vector index: very low recall at 100M with coarse cell grids


muralikpbhat:
## Summary

At 100M documents, the hidden vector index returns very low `recall@10` (~0.16–0.22) after drain, versus ~0.99 at 1M/10M. Measurement shows the loss is **fine-probe-depth-bound**, not cell-routing/coverage-bound: probing more cells does nothing, and the cause is that the fine probe is an *absolute* count while a cell's fine-cluster count grows with cell size.

> Note: an earlier version of this issue attributed the loss to L1 cell routing. That has been **refuted** by the cells-probed measurement below.

## Measured

- 100M, 128-cell grid: **post-drain `recall@10` = 0.216**; pre-drain (user-table) = 0.997, so the corpus/ground truth are fine — the loss is entirely in the drained hidden index.
- Hidden index: 128 cells, ~148K fine clusters (**~1,162 per cell**), cell rows p50 ≈ 781K.
- **Probing more cells does not help.** Forcing the per-query cell fan from 1 → 8 (verified: 8 cells probed, all 128 in the beam) left recall **unchanged at 0.216**. So it is not coverage/routing-bound.
- 1M/10M for contrast: post-drain ~0.995 / ~0.996.

## Root cause: absolute fine probe is a shrinking fraction as cells grow

A cell's fine-cluster count is set at drain time by a fixed ~2 MiB-per-run target, so it scales with the cell's row count: `n_fine ≈ cell_rows / ~945`. The query-time fine probe is a fixed absolute count (current default 4). The fraction of a cell actually probed therefore *shrinks* as cells grow:

| scale | fine clusters / cell | fixed probe = 4 (default) | recall |
|-------|----------------------|---------------------------|--------|
| 10M / 256 | ~41 | ~10% | 0.996 |
| 100M / 256 | ~410 | ~1% | (low) |
| 100M / 128 | ~1,162 | **~0.3%** | **0.216\*** |

\* Recall at 100M/128 was measured with a probe of 8 (the prior default); with the current default of 4 it is no higher.

At ~10% coverage recall holds; at ~0.3% it collapses. Same knob, different cell size. Deeper L2 is *not* the issue in the sense of cluster count (cells already carry ~1,162 fine clusters) — the issue is how few of them a fixed absolute probe reaches.

## Fix directions (composable)

[...]

PR #466, opened 2026-07-27
bench: streaming recall-over-time diagnostic (recall_while_ingest)

muralikpbhat:
## What it measures

`recall@10` across the production `optimize()` lifecycle. In production `optimize()` runs periodically as data streams in — incremental drain + merge/compaction + (optional) split — so the hidden index's recall evolves over the whole ingest lifetime, not at a single post-load instant.

`recall_while_ingest` measures exactly that: it streams docs in checkpoints, runs `optimize()` at each checkpoint (synchronous) or on a background cron (concurrent with ingest, the production cadence), and reports `recall@10` / `drained%` / cell count / over-cap at every checkpoint.

## Why an inline ground truth (what it sidesteps)

Grading recall at 1M–100M the traditional way needs the full corpus on disk and a brute-force oracle pass — at 100M×1024 that's a ~400 GB mmap corpus plus a cached oracle and an id mapping. This bench sidesteps it: the ground truth is maintained inline as a per-query top-k min-heap, updated as each append-only batch streams by (exact, because ingest is id-ordered). The corpus is never materialized and there is no separate oracle pass — which is what makes recall measurable across the whole lifecycle at scale.

## Ground-truth validation

Is the cheap inline GT accurate at scale? Cross-checked against a fully **independent** oracle: the same drained 100M table, reopened and graded by the exact **full-corpus brute-force** GT, scores `recall@10 = 0.986` — matching the inline GT's `0.979` on the same table (the independent oracle is if anything slightly higher, so the streaming GT is not inflating). The tool's streaming ground truth is trustworthy at 100M.

## Example output

```
[recall_while_ingest] streaming 100K docs in 10K-doc checkpoints, synchronous optimize() after each batch, 200 held-out queries
[recall_while_ingest] idx  prefix  recall@10  drained%  cells  over_cap
[recall_while_ingest] 1    10K     0.965      100%      256    0
 ...
[recall_while_ingest] 10   100K    0.957      100%      256    0
# post-build breadth-sweep (coverage vs within-cell depth):
[breadth-sweep] nprobe=2   recall@10=0.987
[breadth-sweep] nprobe=4   recall@10=0.992   (flat past 4 => loss is depth, not coverage)
```

…plus a markdown Report table of the same rows.

## Example: what it surfaces

Streaming recall on `main` vs the optional split (#450), 256-cell bootstrap, nprobe=1:

| checkpoint | main (native optimize) | with #450 split on |
|---|---|---|
| 100K | 0.982 | 0.997 |
| 1M | 0.962 | 0.997 |
| 10M | 0.983 | 0.991 |
| 100M | 0.979 | *(validating)* |

*(Illustrative of what the tool measures — the split's own claims live in #450.)*

`Refs #436.`


muralikpbhat:
## What the tool caught: at scale, drain **cadence** decides recall — not the index

A concrete payoff for measuring recall *across* the `optimize()` lifecycle instead of at a single post-load instant. At 100M (native grid, no split, identical engine), recall depends entirely on how the drain is paced:

| drain cadence | recall@10 (nprobe=1) |
|---|---|
| **one-shot** — all 100M drained in a single end-of-build `optimize()` | **0.094** |
| **incremental** — `optimize()` per checkpoint (production cadence, what this tool measures) | **~0.98** (holds through 80M; 100M validating) |

On the mechanism, what's *confirmed*: the global cell grid is **not** refined by the drain — it's bootstrapped once and only grown by split — so these two runs share an *identical* grid and identical routing. The collapse is therefore a **within-cell** effect, not a grid/routing one: it lives in the per-cell fine index, which the incremental cadence builds from many small drain batches but the one-shot builds over ~390K docs/cell in a single pass. The exact within-cell culprit (fine k-means conditioning vs the 1-bit shortlist saturating at that batch size) is still under investigation.

A load-once-then-measure harness reports only the post-load instant and never surfaces this as a **cadence** effect — measuring recall over the whole ingest/drain lifecycle is exactly what `recall_while_ingest` is for.

src/supertable/query/vector.rs
```diff
@@ -1317,17 +1317,23 @@ impl SupertableReader {
...
-                if filtered && options.nprobe.is_some() {
-                    // Explicit caller `nprobe` on a FILTERED query pins the
-                    // hidden cell sweep — the width dial calibration and
-                    // the bench sweep turn (depth stays at the filtered
-                    // default so the sweep isolates width). Unfiltered
-                    // hidden routing keeps ignoring caller nprobe
-                    // (persisted params own it).
+                if options.nprobe.is_some() {
+                    // Explicit caller `nprobe` wins for hidden queries — filtered
+                    // AND unfiltered — so the bench's coverage-vs-depth breadth
```
ekechinwokah: nprobe is public api (`with_nprobe`. "production never sets nprobe" is an observation about current callers, not something the compiler enforces, so anyone who explicitly sets nprobe on an unfiltered query against a hidden-indexed table gets different behavior after this lands.

benches/utils/recall_while_ingest.rs
```diff
@@ -0,0 +1,788 @@
...
+) {
+    let reader = consumer.reader();
+    let opts = exec_vec::search_opts(exec_vec::ENGINE_DEFAULT, exec_vec::ENGINE_DEFAULT);
+    let n_retained = retained.len() / DIM;
+
+    // Measurement soundness: a PURE bench-side exact brute-force (dot over every
+    // retained vector, no engine/codec/IVF) vs the GT heap. It MUST be ~1.0 — if
+    // not, the GT pipeline itself is inconsistent and every engine recall number
+    // was measured against a broken baseline. Parallel over queries.
+    use rayon::prelude::*;
```
ekechinwokah: use rayon::prelude::*; — already imported at top of file. also please ask claude to adhere to agents.md.

benches/utils/recall_while_ingest.rs
```diff
@@ -0,0 +1,788 @@
...
+            fmt_count(n),
+            fmt_opt_pct(stats.drained_pct),
+            fmt_opt_usize(stats.cells),
+            fmt_opt_usize(stats.over_cap),
+        );
+        // Breadth diagnostic: sweep the coarse cell-probe width against this
+        // built table. If recall climbs toward 1.0 as more cells are probed,
+        // the loss is coverage/route-fidelity (breadth), not within-cell depth.
+        let cells_now = stats.cells.unwrap_or(0);
+        for np in [2usize, 4, 8, 16, 32, cells_now]
```
ekechinwokah: nit: [2usize, 4, 8, 16, 32, cells_now] is inline where everything else in this file gets a named const.

ekechinwokah (approved):

src/supertable/query/vector.rs
muralikpbhat: Good catch — reverted the public behavior change. I believe a customer would want to drive `nprobe` at some point (if the default p=1 doesn't hold up), but for now, given the recall we're seeing, I've scoped it to a bench-only change so `with_nprobe` serving behavior is unchanged.

Concretely: an explicit `nprobe` on an *unfiltered* hidden query no longer widens the cell probe in a production build. That path is gated behind a `pub(crate) widen_unfiltered_hidden_cells` flag on `VectorSearchOptions` whose setter is `#[cfg(feature = "test-helpers")]` — so it's absent from the public API surface and only the recall breadth-sweep bench sets it (via `search_opts`). Filtered queries are unchanged. (Used the field + test-helpers-gated setter rather than a `cfg!(feature=...)` in the routing, since there's no precedent for a runtime `cfg!` in src/ and it would have diverged prod-vs-bench behavior.) — c0bebb6

benches/utils/recall_while_ingest.rs
muralikpbhat: Fixed in c0bebb6 — dropped the function-local `use rayon::prelude::*;` (already imported at the top of the file). Also swept the file for the other AGENTS.md items.

benches/utils/recall_while_ingest.rs
muralikpbhat: Fixed in c0bebb6 — lifted `[2, 4, 8, 16, 32]` to a documented `const BREADTH_SWEEP_NPROBE_STEPS` at the top of the file; the current cell count is chained on at the call site.

PR #470, opened 2026-07-28
fts: improve post-compaction OR latencies

ekechinwokah:
## Problem

After `optimize` compacts a table (128 superfiles → 5 at the 1M bench scale),
warm multi-term OR queries regress by an amount that grows with term count —
measured at 1M docs on Azure, post-compact vs pre-compact warm p50:

| shape          | pre-compact | post-compact (before) | ratio |
|----------------|-------------|----------------------|-------|
| ten_term_or    | 2.2 ms      | 25.5 ms              | 11.5x |
| twenty_term_or | 3.7 ms      | 67.8 ms              | 18.2x |
| forty_term_or  | 6.3 ms      | 151.9 ms             | 24.3x |

## Root causes (three, layered)

1. **The ranged OR entry point hardcoded MaxScore+BMM.** The un-ranged path
   picks its kernel via `dispatch_multi_term_or`, which routes uniform
   broad ORs to the windowed union scorer precisely because BMM cannot
   prune when block maxima tie. Ranged work units only exist when the
   reader pool has more threads than surviving superfiles — i.e. only
   after compaction — so the same query silently switched to the wrong
   kernel exactly and only post-compact.

2. **Each ranged slice rebuilt its cursors.** Every slice of one superfile
   independently re-fetched every term's full posting bytes and re-parsed
   its skip table (measured as 2.5x cold read amplification when slicing
   widened), and the work-unit allocator split the pool evenly per *file*
   (`ceil(threads/files)`), starving the one merged superfile that holds
   ~all the docs.

3. **Ranged kernels ran as multi-millisecond sync blocks on tokio
   workers.** Schedule reconstruction (per-unit start/end stamps) showed
   every query lost one slice to worker starvation: 7 slices finish at
   ~6 ms, one waits ~6 ms for a thread and lands at ~9.6 ms. This is the
   concurrency-contract violation CLAUDE.md warns about (CPU waves belong
   on the reader pool, bridged with a oneshot).

## Changes

- `search_or_range_pretokenized_with_floor` now mirrors the un-ranged
  kernel dispatch (windowed union when no term's upper bound dominates,
  BMM otherwise).
- `TermCursor` is `Clone` with the parsed skip table behind an `Arc`;
  `FtsReader::build_or_cursor_set` / `search_or_range_prebuilt` let one
  build serve every slice of a superfile. The supertable fan-out coalesces
  one build per (query, superfile) behind a `tokio::sync::OnceCell`.
- `build_work_units` allocates sub-ranges by doc mass
  (`round(docs_share x pool_threads)`, floored by `SUBRANGE_MIN_DOCS`)
  instead of evenly per file.
- Ranged kernels run as a rayon wave on `options.reader_pool`, bridged
  with a oneshot, keeping tokio workers free for wakeups.

All items are `pub(crate)`; the crate-root public surface is unchanged.

## Measured (1M docs, Azure, supertable FTS, warm p50 post-compact)

| shape          | before   | after fix 1 | after fix 2 | after fix 3 | pre-compact |
|----------------|----------|-------------|-------------|-------------|-------------|
| ten_term_or    | 25.5 ms  | 5.0 ms      | 3.3 ms      | 3.0 ms      | 2.2 ms      |
| twenty_term_or | 67.8 ms  | 8.7 ms      | 5.7 ms      | 5.1 ms      | 3.7 ms      |
| forty_term_or  | 151.9 ms | 17.3 ms     | 9.5 ms      | 8.5 ms      | 6.3 ms      |

Fix-3 columns were measured with temporary schedule instrumentation
enabled, which printed one line per work unit and so taxed pre-compact
(128 units/query) far more than post-compact (12): pre-compact appeared
inflated by ~0.5-3 ms in those runs. The pre-compact column above is
from instrumentation-free runs; fix-3 post-compact numbers carry ~0.2 ms
of print overhead and are, if anything, slightly pessimistic. The
instrumentation is not part of this change.
Fix 3 eliminates the measured one-starved-slice-per-query pattern (all
slice waits uniform at ~0.65 ms = the shared build) at the price of ~+20%
CPU-seconds per broad-OR query: with all eight kernels truly concurrent,
SMT contention on the 4C/8T box raises per-slice scan time (5.4->7.4 ms
spread), and the wall lands at the hardware's effective-thread ceiling.
Latency-vs-CPU-cost trade-off; the cost model prices the CPU side.

Cold I/O is unchanged or better throughout (post-compact cold-1st
16-17 GET / 8.5-11.5 MiB; steady 1-2 GET). `two_term_or` (the slicing-
overhead canary) is flat through fix 2; fix 3's oneshot round-trip adds
~0.1 ms to it (0.46 -> 0.55 ms with diagnostics on) — gating the bridge
on >= 3 should-terms would exempt trivial shapes if that matters.

## Tests

- Partition-equivalence: ranged results equal un-ranged results for any
  cut of the doc space, on BOTH dispatch branches, with assertions that
  each planted shape actually routes to the branch it covers.
- Prebuilt-vs-fresh equivalence over overlapping and repeated windows
  (guards walk-state leaks between cursor clones).
- Doc-mass allocation: merged file takes the whole pool, sub-floor
  remnants stay un-ranged, slices tile the doc space.
- Full FTS lib suite: 368 tests green; fmt + clippy -D warnings clean.

## Bench / verification status

- [x] 1M supertable-FTS A/B on Azure per change (numbers above)
- [x] Full `make bench` A/B against main (all tiers x modalities)
- [ ] Vector suite + recall@10 >= 0.99 gate (drain/query code shared paths)
- [ ] 10M FTS validation (<20 ms broad-OR target at scale)
- [ ] miri/asan: not required (no new `unsafe`)

## Notes for reviewers

- No env vars or diagnostics are added; the schedule attribution above
  was measured with temporary instrumentation that is not part of this
  change.
- The superfile tier's per-algorithm probes now exercise the windowed
  union alongside WAND+BMW and MaxScore+BMM (the shapes span both sides
  of `OR_WINDOW_DOMINANCE_MULT`), restoring the routing-heuristic
  validation the deleted `benches/fts_search.rs` used to provide; the
  stale references in the kernel doc comments are corrected.
- Kernels bridge to the reader pool only at >= RANGED_KERNEL_POOL_MIN_TERMS
  (= `OR_WINDOW_MIN_TERMS`, a shared constant so the two thresholds
  cannot drift): below it kernels are sub-millisecond and the oneshot
  round-trip costs more than it saves.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2376,6 +2376,7 @@ dependencies = [
  "rayon",
  "rcgen",
  "reqwest",
+ "rustix",
  "serde",
  "serde_json",
  "sha2 0.10.9",
--- a/benches/bench/main.rs
+++ b/benches/bench/main.rs
@@ -16,7 +16,7 @@
 //! cargo bench -- supertable vector build cold
 //!
 //! # Diagnostics (standalone programs, same binary):
-//! cargo bench -- diagnostic              # all seven
+//! cargo bench -- diagnostic              # all eight
 //! cargo bench -- diagnostic scale        # a subset, grouped
 //! cargo bench -- tombstone               # bare names also work
 //!
@@ -34,9 +34,9 @@
 //!   `all`       : explicit "every tier × modality × phase" (the default).
 //!                 Matrix only — diagnostics are NEVER implied by `all` or
 //!                 by a bare `cargo bench`.
-//!   diagnostic  : `scale` | `tombstone` | `update` | `sql-diag` | `fts-diag` | `object-store` | `concurrent`,
+//!   diagnostic  : `scale` | `tombstone` | `update` | `sql-diag` | `fts-diag` | `object-store` | `concurrent` | `disk-warm`,
 //!                 by name, or grouped under the `diagnostic` keyword —
-//!                 `cargo bench -- diagnostic` runs all seven,
+//!                 `cargo bench -- diagnostic` runs all eight,
 //!                 `cargo bench -- diagnostic scale tombstone` a subset.
 //!
 //! The matrix tests run = (selected tiers) × (selected modalities).
@@ -70,6 +70,7 @@ enum Diagnostic {
     FtsDiag,
     ObjectStore,
     Concurrent,
+    DiskWarm,
 }
 
 impl Diagnostic {
@@ -82,6 +83,7 @@ impl Diagnostic {
             Diagnostic::FtsDiag => "fts-diag",
             Diagnostic::ObjectStore => "object-store",
             Diagnostic::Concurrent => "concurrent",
+            Diagnostic::DiskWarm => "disk-warm",
         }
     }
 
@@ -94,6 +96,7 @@ impl Diagnostic {
             Diagnostic::FtsDiag => infino_bench_utils::fts_diag::run(),
             Diagnostic::ObjectStore => infino_bench_utils::unified_object_store::run(),
             Diagnostic::Concurrent => infino_bench_utils::concurrent::run(),
+            Diagnostic::DiskWarm => infino_bench_utils::disk_warm::run(),
         }
     }
 }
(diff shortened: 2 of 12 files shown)


vinaykakade (approved):
The changes themselves look good, but please run a search-benchmark-game run for your fork+branch and compare with main as a sanity check.

Aside: had spent some time to figure out what is the right way to route across BMM vs windows union, and if we can figure out the right routing, there are likely more gains we can have (perhaps future work).

savannahar68 (approved):
lgtm
just like we are using reader pool for ranged units, we need to do the same for non ranged. Ill pick this up

savannahar68:
im getting OOMs consistently on this branch on sbg. Lets hold it until i debug and come back

ekechinwokah:
> lgtm just like we are using reader pool for ranged units, we need to do the same for non ranged. Ill pick this up

@savannahar68 sg— one heads-up from having done the ranged side: the bridge only worked after splitting the path into async build (build_or_cursor_set) + sync scan (search_or_range_prebuilt). bm25_search_clauses interleaves fetch and scan, so the non-ranged version of this is "extend the prebuilt-cursor separation to the clause path, then bridge the sync half." 

two things to carry over: gate the bridge (single-term BMW is µs-scale and would eat the ~0.1 ms oneshot tax on the most common shape — probably worth gating on posting mass, not term count), and bridge only the post-fetch scan so pool threads never block on cold GETs.

Issue #467, opened 2026-07-27
Bench: unify and persist vector ground-truth (recall_while_ingest ↔ vector bench)


muralikpbhat:
Follow-up to #466 (which adds `recall_while_ingest`). While investigating recall-at-scale (#436) and the in-place cell split (#450), we ended up with **two** ground-truth mechanisms in the vector benches, plus some avoidable cost. This tracks the cleanup.

## Background
- The **vector bench** grades against a full-corpus exact oracle: it prepares the entire corpus to disk (mmap — ~400 GB at 100M×1024) and brute-forces top-k, caching the oracle in a bin (`corpus/grading.rs`). The bin already persists the query vectors **and** the labels.
- **`recall_while_ingest`** (#466) grades against an **inline running ground truth** — per-query top-k min-heaps updated as each batch streams by — so the corpus is never materialized. But it doesn't persist that GT.

## Follow-ups
1. **Load queries + GT from the oracle bin; don't regenerate the corpus.** On a reopen / read-sweep the read path *has* a valid cached oracle but still regenerates the ~400 GB corpus solely to rebuild the query vectors for validation (`generate_realistic_queries` reads strided corpus rows). The bin already stores the queries (`write_cache → push_queries`) — load them and skip corpus regeneration when a matching cached oracle exists.
2. **Adopt the inline/streaming GT (or the persisted bin) in the vector bench.** The full-corpus brute-force is what makes 100M GT a 400 GB operation. Compute the vector bench's GT the `recall_while_ingest` way (streaming heaps, no corpus materialization) — or consume the persisted bin — to drop the corpus requirement at scale.
3. **`recall_while_ingest` should persist its GT.** It computes GT inline but doesn't write it, so a KEEP-table read-sweep or cross-bench read can't reuse it. Persist the inline GT (queries + labels) in the same oracle-bin format (CacheKey-compatible) so downstream reads load it without recompute.
4. **(1) is likely subsumed by (2)+(3).** If both benches share one persisted-bin format and a load-not-regenerate path, item 1 falls out; track together.
5. **Other deferred bench items:**
   - **Productize the "GT-prepare only, no index" path** as a first-class bench verb (currently an env-flag hack for one-off read-sweeps: prepare corpus + oracle, skip the index build).
   - Optionally **persist the corpus** (not a `TempDir`) so a reopen/read-sweep reuses it instead of regenerating ~400 GB.

Refs #436, #466, #450.

muralikpbhat:
[...]

PR #475, opened 2026-07-28
bench: load vector ground truth from the oracle bin on reopen

muralikpbhat:
## What

On a reopen of an already-built vector supertable (`INFINO_BENCH_EXISTING_PREFIX`), load the held-out queries **and** their exact labels straight from the persisted oracle bin and skip corpus preparation entirely.

## Why — the reopen win

The vector bench grades against a full-corpus brute-force oracle: it mmaps the whole corpus (~400 GB at 100M × 1024) and brute-forces top-k, caching the queries + labels in a bin. But a **reopen** never needs the corpus for the *table* — it opens the persisted one — it only regenerated the corpus to rebuild the *query vectors* for grading. The bin already stores those queries. So the reopen was paying a ~400 GB disk write + a multi-minute brute-force pass for data it already had on disk.

Measured on a reopened 100M table:

| | before (regenerate corpus) | after (load bin) |
|---|---|---|
| corpus prepared | ~400 GB mmap + brute-force | **none** — RSS flat ~22 MiB across corpus-prepare |
| oracle build | **1084.6 s (~18 min)** exact pass over the 103M-row corpus (705–1900 s across 100M runs) | **412 KB bin, sub-second read** |
| path to first query | corpus build → oracle pass → open | **straight to table open** (cold open 2.19 s) |
| `recall@10` (nprobe=1) | 0.986 | **0.986** — identical |

**Before → after:** an ~18-minute exact oracle build plus a ~400 GB corpus regeneration on every reopen → a sub-second 412 KB bin read with no corpus. The recall matching to the digit is the correctness proof: the bin-loaded queries + labels are the *same* oracle, so the only thing that changed is not regenerating the corpus. This also makes read-only recall sweeps on retained large tables cheap — previously they were gated on regenerating the corpus just to grade.

## How

- **`grading::load_oracle(path) -> CachedOracle`** in the **shared** grading module (sibling to the existing `write_cache`): reads the magic + key + queries + all three label sets. Every row count comes from the embedded key, so it needs no expected query set — which is the whole point: the caller has no corpus to regenerate the queries from.
- **`vector::run`**: when an existing prefix is reopened and a matching oracle bin is present, load queries + labels from it and gate `prepare_corpus` to `None`. The combined query/label sets are split at the correctness boundary exactly as the corpus path does. **The corpus path is unchanged** — `git diff -w` shows only the added branch + the gate; the old block is byte-identical, just indented into the `else` arm.
- **Latent-panic fix (two sites)**: `run_search` **and** the steady-state `measure_cold_store` both hard-indexed `q_cal[0]` for their latency probes; a skip-calibration reopen loads no calibration queries (the bin stores none), so `q_cal` is legitimately empty → both fall back to the correctness set. (Without the second fix, the cold-store measurement panics after grading — silently dropping the cold GET/cost accounting on every reopen.)

## Correctness for both calibration modes

`skip_calibration` is a pure function of `n_docs`, so the bin's contents and the reopen's flag can never disagree (it's the same table, same `n_docs`):

- **non-skip** (small tables): bin persists correctness **+** calibration queries → `load_oracle` splits into non-empty `q_cal`/`gt_cal` → calibration path runs, `q0 = q_cal[0]` unchanged.
- **skip** (large tables, e.g. 100M): bin persists correctness only → `q_cal` empty → `q0` falls back to the correctness set → recall graded from `gt_correct`.

## Tests

- **Unit**: `load_oracle` round-trips the queries + all three label sets and rejects a truncated bin (exercises the non-skip split shape).
- **End-to-end**: the reopened 100M table reproduces the corpus-based reopen's 0.986 with no corpus (table above). The `q0` fallback is exercised by this skip-calibration reopen (bench-harness code, so runtime-validated rather than unit-tested).

## Scope / follow-ups

This is the **read** side only. The remaining #467 items — persisting `recall_while_ingest`'s inline ground truth in this same bin format (the `save_oracle` counterpart) and dropping the corpus from the vector bench's own GT — **will be taken up after #466 (`recall_while_ingest`) is merged**, since they build on that tool. `load_oracle` is placed in the shared grading module precisely so those items reuse it verbatim.

Refs #467.

### DIFF
--- a/benches/utils/corpus/grading.rs
+++ b/benches/utils/corpus/grading.rs
@@ -187,6 +187,9 @@ pub fn load_oracle(path: &Path) -> Result<CachedOracle> {
     let top_k = key.top_k as usize;
     let base = pull_ground_truth(&mut cursor, key.query_count as usize, top_k)?;
     let correctness = key.correctness_query_count as usize;
+    if correctness > queries.len() {
+        return invalid_data("lifecycle ground-truth cache correctness count exceeds query count");
+    }
     let filtered = pull_ground_truth(&mut cursor, correctness, top_k)?;
     let augmented = pull_ground_truth(&mut cursor, correctness, top_k)?;
     if !cursor.is_empty() {
@@ -381,6 +384,16 @@ fn pull_queries(cursor: &mut &[u8]) -> Result<Vec<Vec<f32>>> {
     let dim = read_u64(cursor)? as usize;
     if dim != DIM {
         return invalid_data("lifecycle ground-truth cache query dim mismatch");
+    }
+    // Bound the claimed count by the bytes actually present before allocating,
+    // so a corrupt count can't drive a huge Vec::with_capacity (mirrors the
+    // validate-before-allocate guard in pull_ground_truth).
+    if count
+        .checked_mul(dim)
+        .and_then(|n| n.checked_mul(4))
+        .is_none_or(|need| need > cursor.len())
+    {
+        return invalid_data("lifecycle ground-truth cache query count exceeds buffer");
     }
     let mut queries = Vec::with_capacity(count);
     for _ in 0..count {
--- a/benches/utils/corpus/grading.rs
+++ b/benches/utils/corpus/grading.rs
@@ -142,6 +142,69 @@ pub fn lifecycle_ground_truth_cached(options: LifecycleGradingOptions<'_>) -> Li
     labels
 }
 
+/// Held-out queries + exact labels loaded straight from a persisted oracle
+/// bin, without the corpus. Both the full-corpus vector bench (reopen /
+/// read-sweep) and the streaming recall bench (keep-table read-sweep) use this
+/// to skip their own ground-truth construction — the on-disk corpus brute
+/// force and the inline running heaps respectively — since the bin already
+/// holds both the queries and the labels.
+pub struct CachedOracle {
+    pub queries: Vec<Vec<f32>>,
+    pub labels: LifecycleGroundTruth,
+    pub n_docs: usize,
+    pub top_k: usize,
+    /// `queries[..correctness_query_count]` are the correctness queries; the
+    /// remainder (if any) are calibration queries. `labels.base` splits at the
+    /// same index.
+    pub correctness_query_count: usize,
+}
+
+/// The explicit oracle-bin path (`INFINO_BENCH_VECTOR_GROUND_TRUTH_PATH`), if set.
+pub fn oracle_path() -> Option<PathBuf> {
+    env::var_os(GROUND_TRUTH_PATH_ENV).map(PathBuf::from)
+}
+
+/// Read queries + exact labels from a persisted oracle bin (`INFLGT02`) without
+/// preparing the corpus. Row counts come from the embedded key, so no expected
+/// query set is needed (that is the whole point — the caller has no corpus to
+/// regenerate the queries from).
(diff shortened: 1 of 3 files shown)


benches/utils/corpus/grading.rs
```diff
@@ -142,6 +142,69 @@ pub fn lifecycle_ground_truth_cached(options: LifecycleGradingOptions<'_>) -> Li
...
+/// The explicit oracle-bin path (`INFINO_BENCH_VECTOR_GROUND_TRUTH_PATH`), if set.
+pub fn oracle_path() -> Option<PathBuf> {
+    env::var_os(GROUND_TRUTH_PATH_ENV).map(PathBuf::from)
+}
+
+/// Read queries + exact labels from a persisted oracle bin (`INFLGT02`) without
+/// preparing the corpus. Row counts come from the embedded key, so no expected
+/// query set is needed (that is the whole point — the caller has no corpus to
+/// regenerate the queries from).
+pub fn load_oracle(path: &Path) -> Result<CachedOracle> {
```
ekechinwokah, line 171: load_oracle never checks correctness_query_count <= query_count. bad bin → panics later in queries.split_off(). should return invalid_data instead.

benches/utils/corpus/grading.rs
```diff
@@ -313,6 +376,23 @@ fn validate_queries(cursor: &mut &[u8], expected: &[Vec<f32>]) -> Result<()> {
     Ok(())
 }
 
+fn pull_queries(cursor: &mut &[u8]) -> Result<Vec<Vec<f32>>> {
```
ekechinwokah, line 382: pull_queries does with_capacity(count) on an unvalidated count. pull_ground_truth right below validates count before allocating - prob good to match that pattern.

benches/utils/supertable.rs
```diff
@@ -4353,15 +4043,26 @@ pub mod vector {
...
+                                match q_cal.first() {
+                                    Some(first) => (
+                                        first,
+                                        if q_cal.len() > 1 {
+                                            &q_cal[1..]
+                                        } else {
+                                            &q_cal[..]
+                                        },
+                                    ),
+                                    None => (&q_correct[0], &q_correct[1..]),
```
ekechinwokah: &q_correct[0] — same raw-index panic as elsewhere. use .first().expect(...) like the executors.rs fix.

ekechinwokah (approved):

benches/utils/corpus/grading.rs
muralikpbhat: Fixed in 0072b6b — `load_oracle` now validates `correctness_query_count <= query_count` and returns `InvalidData` before it's used to split, instead of panicking later in `split_off`.

benches/utils/corpus/grading.rs
muralikpbhat: Fixed in 0072b6b — `pull_queries` now bounds the claimed count by the bytes actually present (`count * dim * 4 <= remaining`, checked-mul guarded) before `with_capacity`, matching `pull_ground_truth`'s validate-before-allocate.

benches/utils/supertable.rs
muralikpbhat: Fixed in 0072b6b — this site now uses `q_correct.first().expect(...)`. Note the other four `&q_correct[0]` sites (3447/3519/3873/3953) are pre-existing on main and are being cleaned up in a later PR, so I've left them out of this PR to avoid overlap.

PR #477, opened 2026-07-28
fix supertable drop_table: handle table deletion by another worker process

### LINKED ISSUE #468: bug: drop_table by writer, not respected by already active reader
### Steps to reproduce

- Start two instance of `infino` engine. ( say engine A and engine B )
- Create a table, and append some data ( say using engine A )
- Run queries using both engines. Both engines should return the same results
- drop the table using an engine ( say engine A )
- query / append data to the same table using the other engine ( engine B here )

### Expected behaviour
The table could not be found

### Observed behaviour
Query executes the operation. ( bigger issue with appends, as these appends are now "ghost" )

Pranav2612000:
### What does this PR do?
- Prevents appends, updates, sql queries, fts queries and vector queries on a table that has been dropped by another infino engine process active on it.
- Fixes #468 

- To prevent race conditions during delete + commit when a user has multiple infino engines active on the same data

### How do we do this change?
- For catalog operations ( sql queries ), the reason we encounter this error is because we cache the table handles which are never invalidated.
- Now, we invalidate this table handle cache when the current manifest pointer is no longer present ( table deletion )
- This is done by setting a `pointer_vanished` as a latch on the table handle when we find the manifest pointer missing either during reads or writes ( commit ).
- Similarly for reader operations ( bm25_search, token_match, exact_match, count, vector_search, hybrid_search ), the reader initialization should fail when trying to open a reader on an already deleted table. This also requires the reader signature to change ( return a `Result` now ). This is 90% of the changes in this PR ( review commit by commit to make the process easier ).
- For writes/commits, previously, an write on a stale handle could rebuild the table. This PR fixes this by propagating the NotFound error when trying to write on a deleted table.

### How has this been tested?
- Unit tests added
- E2E tests
- Benchmarks - attached in comments

### Next steps
- The original issue still persists if a user deletes the index with `purge=false`. The fix for this is deferred for now and will be part of a separate PR

### DIFF
--- a/benches/utils/bench/supertable/query.rs
+++ b/benches/utils/bench/supertable/query.rs
@@ -16,7 +16,7 @@ pub fn vector_topk_global(
     options: VectorSearchOptions,
 ) -> Vec<i128> {
     let batches = st
-        .reader()
+        .reader().expect("reader")
         .vector_search(VEC_COLUMN, query, k, options, None, None)
         .expect("vector_search");
     corpus::id_scores_from_vector_search(&batches)
--- a/benches/utils/concurrent.rs
+++ b/benches/utils/concurrent.rs
@@ -329,6 +329,7 @@ async fn reader_loop(
         let t0 = Instant::now();
         let _ = black_box(
             st.reader()
+                .expect("reader")
                 .bm25_search(QUERY_FIELD, QUERY_TERM, TOP_K, BoolMode::Or, None)
                 .expect("bm25_search"),
         );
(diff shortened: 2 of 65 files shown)


Pranav2612000:
This branch
image


main
image

Pranav2612000:
fts OR queries This branch
image


fts OR queries main
image

Pranav2612000:
vector ingest this branch
image

vector ingest main
image

Pranav2612000:
vector queries this branch
image


vector queries main
image

vinaykakade (approved):

PR #479, opened 2026-07-29
supertable/vector: manifest-calibrated probe-width + configurable nprobe for improved recall on real-world corpus

ekechinwokah:
**Problem.** On real-embedding corpora, post-drain unfiltered vector search could not reach the recall bar under any configuration (main today: recall@100 flat ~0.44 regardless of nprobe). Three compounding causes, measured on Cohere-1M/768d cosine: the hidden-index sweep probed exactly one cell and discarded caller `nprobe`; fine-run depth was pooled per drain wave so a forced-wide sweep didn't read the extra cells; and the p=1 design point assumes the top-k concentrates in the grid-nearest cell — true on synthetic clustered corpora, false on real embeddings (the true top-1 lands outside the nearest cell for ~58% of Cohere queries; 0.99 mean coverage of the top-100 needs ~50 of 256 cells; grid quality, granularity 256→4096, adaptive slack, and boundary replication all measured as non-levers).

**Change.** Four mechanisms, layered:

- **Honor caller `nprobe`** on the post-drain unfiltered path (override wins in both directions), with fine depth riding the pin exactly as it rides the law — a widened cell read at the persisted fine-first depth contributes only its first runs (measured: ~0.83 recall@100 at *any* nprobe). One pure function (`apply_width_pin`) carries all four override arms, unit-tested.
- **Probe-width law**: a clean drain reservoir-samples ~256 rows as stand-in queries, scores them against every packed cell during the pack pass it already performs (Sq8+ε, existing kernels, no fp32 corpus decode), and stamps the cells needed for mean-0.99 exact top-k coverage at k = {1,10,100,1000} into `CellRoutingParams.width_for_k`. Unfiltered queries with no override log-interpolate the law at their k. All-zero law (legacy manifests, resumed drains) keeps fine-first p=1 byte-for-byte; wire-compatible both directions.
- **Global shortlist selection at width**: warm cells surface their 1-bit estimates (scanned heap-free at the undivided cap — estimates are comparable across cells and superfiles, one rotation seed per column, asserted at the pool site); the supertable keeps the best k×rerank_mult across the whole sweep via an O(n) partition and reranks only those winners where they live. This replaces the per-cell even split, which wasted the estimator's precision: at equal budgets, divided vs global is 0.9691 vs 0.9928 (6.4K survivors) and 0.9878 vs 0.9954 (12.8K). Cold cells keep the divided in-probe rerank — their fetched blocks and budget reservation must not outlive the probe. Filtered queries and RabitqOnly untouched.
- **Concurrent per-cell probes** in the multi-cell kernel, wave-capped to the pool width.

**Results, Cohere-1M/768d cosine (official harness, 1000 held-out queries, warm, k=100):**

|  | **main** | **this PR (rerank_mult=64)** |
| --- | --- | --- |
| recall@100 | ~0.44 | 0.9947 |
| avg / p99 latency | — | **13.0/14.1 ms** |

recall@10 (repo bar ≥ 0.99): 0.9950 default / 0.9910 at rerank_mult=64. Cold recall equals warm at every point (measured with no cache at all — the cold arm holds the gate on its own; a checked-in warm-vs-cold parity fixture pins it). Sweep-condition p50 on a 50K-batch table runs ~2 ms lighter (10.6 ms k=100 / 4.1 ms k=10 at rerank_mult=64); the harness numbers above are the quotable condition.

**Trade-offs, explicit:** on spread-heavy corpora the bar has a real read price — the width sweep touches ~20% of corpus *code* bytes; payload reads are survivor-only on warm (k×rerank_mult rows, not whole cells) and whole-block on cold, where the cache fill amortizes them. The cold-budget smoke band widened ([120K,200K]→[120K,800K] B); the bounded-budget contract (sum, refuse on crossing) is unchanged and covered. Calibration adds ~15s to a 1M-row clean drain; resumed drains keep the prior law (max-merge floor; staleness documented at the stamp site). Found in passing, separate issues: `drain_replica_target_factor > 1.0` drain crash; persisting the calibration sample to allow law shrink at maintenance.

**Testing:** 2,428 tests green (lib/supertable/superfile); regression tests pin the override both directions, the law-widened default, filtered+nprobe parity, all pin arms, selection determinism, and the minimal-budget global path; manifest roundtrip + legacy-decode for the new field.

**Bench:** pending — `make bench` vs main before merge; expectation is no change on the synthetic lanes (width 1, global path disengaged).

For engineers who understand how search currently works in main, here's a simpler way to describe the changes:

1. How many cells the search selects — the width law. On main, the post-drain unfiltered default is fine-first p=1 (one cell, slack-widened to ~3), and caller nprobe was silently ignored on that path — worse, wave-pooled gating meant widening cells didn’t even deepen reads (measured dead flat 0.443 recall from nprobe=4 through 256). That’s the 0.44 VDBBench number from yesterday. This branch: the drain calibrates width_for_k on the actual corpus (k=1/10/100/1000 — Cohere needs ~30-40 cells at k=100; synthetic clustered data calibrates to 1), stamps it into manifest routing, and at query time the law pins the cutoff like an explicit nprobe. The cutoff in the pipeline I described was effectively 1-3 on main; it’s the calibrated width here.
2. How deep each selected cell is scanned. Main’s widened cells contributed only their persisted fine-first runs (~6-8), capping recall at 0.83 at any width. This branch: depth rides the pin — fine_nprobe = MAX on the law/pin path (measured load-bearing: 0.83 floor / 0.90 half / 0.9928 full), plus per-(cell, fragment) gating when a width is pinned so reads actually follow width.
3. Who keeps the row shortlist. Main divides the k×rm survivor budget per cell and reranks in-probe — per-cell truncation destroys the global estimate ordering (0.9691 at rm=64). This branch pools 1-bit estimates globally across all cells and superfiles (legal because one table-wide rot_seed makes them comparable), selects the global top k×rm once, and defers exact rerank to those winners: 0.9928 at the same budget. 
4. The speed of every stage — the five optimization commits: exact-range survivor gather, zero-copy mmap block-cache service, per-cell heap → one select_nth partition, per-superfile rotation hoist, and the FastScan LUT transposed scan.

---

### DIFF
--- a/src/superfile/reader.rs
+++ b/src/superfile/reader.rs
@@ -57,7 +57,7 @@ use crate::{
         },
         vector::{
             layout::VectorLayout,
-            reader::{self as vector_reader, VectorReader},
+            reader::{self as vector_reader, ScanCandidate, ScanOutcome, VectorReader},
         },
     },
     supertable::query::provider::tombstone_access_plan,
@@ -1483,6 +1483,63 @@ impl SuperfileReader {
             budget,
         )
         .await?)
+    }
+
+    /// Deferred-rerank sibling of [`Self::vector_search_clusters_filtered`]
+    /// for the supertable's global-shortlist width sweeps: warm cells
+    /// return 1-bit estimate survivors (scanned at the undivided
+    /// `k × rerank_mult`), cold cells rerank immediately under
+    /// `cold_rerank_mult`.
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) async fn vector_scan_clusters_filtered(
+        &self,
+        column: &str,
+        query: &[f32],
+        k: usize,
+        clusters: &[u32],
+        options: VectorSearchOptions,
+        cold_rerank_mult: usize,
+        allow: Option<Arc<RoaringBitmap>>,
+        deny: Option<Arc<RoaringBitmap>>,
+        pool: Option<Arc<ThreadPool>>,
+        budget: Option<Arc<ConnectionMemoryBudget>>,
+    ) -> Result<ScanOutcome, ReadError> {
+        let filtered = allow.is_some();
+        let (_, rerank_mult) = options.resolve(filtered);
+        let v = self
+            .vec()
+            .ok_or_else(|| ReadError::MissingKv(kv::VEC_OFFSET))?;
+        let rerank_mult = v.public_rerank_mult(column, rerank_mult);
+        Ok(v.search_clusters_scan_async(
+            column,
+            query,
+            k,
+            clusters,
+            rerank_mult,
+            cold_rerank_mult,
+            allow,
+            deny,
+            pool,
+            budget,
+        )
+        .await?)
+    }
+
+    /// Rerank this superfile's share of the globally selected shortlist —
+    /// phase C of the deferred-rerank width sweep.
+    pub(crate) async fn vector_rerank_selected(
+        &self,
+        column: &str,
+        query: &[f32],
+        k: usize,
+        selected: Vec<ScanCandidate>,
+        pool: Option<Arc<ThreadPool>>,
+    ) -> Result<Vec<(u32, f32)>, ReadError> {
+        let v = self
+            .vec()
+            .ok_or_else(|| ReadError::MissingKv(kv::VEC_OFFSET))?;
+        Ok(v.rerank_selected_async(column, query, k, selected, pool)
+            .await?)
     }
 }
 
--- a/src/superfile/reader.rs
+++ b/src/superfile/reader.rs
@@ -1502,12 +1502,6 @@ pub struct VectorSearchOptions {
     /// IVF probe override. `None` → engine default for the query path.
     pub nprobe: Option<usize>,
     rerank_mult: Option<usize>,
-    /// Diagnostic, bench/test builds only: when set, an explicit `nprobe`
-    /// widens the coarse cell probe on UNFILTERED hidden-indexed queries too.
-    /// Always `false` in production — the setter is `#[cfg(feature =
-    /// "test-helpers")]`, so it cannot be reached from the public API and the
-    /// serving path never sees it.
-    pub(crate) widen_unfiltered_hidden_cells: bool,
 }
 
 impl VectorSearchOptions {
@@ -1529,16 +1523,6 @@ impl VectorSearchOptions {
     /// the cost of more work.
     pub fn with_nprobe(mut self, n: usize) -> Self {
         self.nprobe = Some(n);
-        self
-    }
-
-    /// Diagnostic, bench/test builds only: let an explicit `nprobe` widen the
-    /// coarse cell probe on UNFILTERED hidden-indexed queries (used by the
-    /// recall breadth-sweep). Absent from production builds, so it never
-    /// affects the serving path.
-    #[cfg(feature = "test-helpers")]
-    pub fn widen_unfiltered_hidden_cells(mut self) -> Self {
-        self.widen_unfiltered_hidden_cells = true;
         self
     }
 
--- a/src/superfile/vector/reader.rs
+++ b/src/superfile/vector/reader.rs
@@ -3103,49 +3103,66 @@ impl VectorReader {
             running = running.saturating_add(col.n_docs);
         }
 
-        let mut merged: Vec<(u32, f32)> = Vec::new();
-        for (cell_idx, locals) in by_cell {
+        // Probe the selected cells CONCURRENTLY — the same wave shape the
+        // supertable's cross-superfile fan-out (and FTS) already uses, one
+        // level down. A width sweep selects many cells inside one packed
+        // shard; awaiting them serially stacked their fetch+scan+rerank
+        // walls (measured ~1.2 ms x width per query at w=54 on Cohere-1M).
+        // Each cell's probe is independent: own ProbeCtx, own blocks, own
+        // rerank; the merge below is order-independent (re-sorted).
+        let cell_probes = by_cell.into_iter().filter_map(|(cell_idx, locals)| {
             let col = &self.columns[cell_idx];
             if col.n_docs == 0 {
-                continue;
+                return None;
             }
             let base = doc_base[cell_idx];
             // Allow/deny are file-local; each cell IVF checks cell-local ids.
             let cell_allow = self.cell_local_filter_bitmap(allow.as_deref(), cell_idx, base);
             let cell_deny = self.cell_local_filter_bitmap(deny.as_deref(), cell_idx, base);
-            let sub_start = col.subsection_range.start;
-            let idx_start = sub_start + col.cluster_idx_off;
-            let idx_end = idx_start + (col.n_cent as usize) * CLUSTER_IDX_ENTRY_BYTES;
-            let cluster_idx = self
-                .source
-                .range_async(idx_start..idx_end)
-                .await
-                .map_err(|e| VectorError::LazySource(e.to_string()))?;
-            let mut q_rot = vec![0f32; col.dim];
-            col.rot.apply(query, &mut q_rot);
             // No inverse-selectivity rerank inflation — same reasoning as
             // [`Self::search_clusters_async`]: the shortlist is allow-first
             // (matching rows only), so the standard `k × rerank_mult`
             // contract holds unchanged under a filter.
             if cell_allow.as_ref().is_some_and(|bm| bm.is_empty()) {
-                continue;
+                return None;
             }
-            let ctx = ProbeCtx {
-                q_rot: &q_rot,
-                k,
-                rerank_mult,
-                allow: cell_allow,
-                deny: cell_deny,
-                pool: pool.clone(),
-                budget: budget.clone(),
-            };
-            let hits = self
-                .probe_clusters_async(col, query, &ctx, &cluster_idx, &locals)
-                .await?;
-            for (local_id, score) in hits {
-                merged.push((base.saturating_add(local_id), score));
-            }
-        }
+            let pool = pool.clone();
+            let budget = budget.clone();
+            Some(async move {
+                let sub_start = col.subsection_range.start;
+                let idx_start = sub_start + col.cluster_idx_off;
+                let idx_end = idx_start + (col.n_cent as usize) * CLUSTER_IDX_ENTRY_BYTES;
+                let cluster_idx = self
+                    .source
+                    .range_async(idx_start..idx_end)
+                    .await
+                    .map_err(|e| VectorError::LazySource(e.to_string()))?;
+                let mut q_rot = vec![0f32; col.dim];
+                col.rot.apply(query, &mut q_rot);
+                let ctx = ProbeCtx {
+                    q_rot: &q_rot,
+                    k,
+                    rerank_mult,
+                    allow: cell_allow,
+                    deny: cell_deny,
+                    pool,
+                    budget,
+                };
+                let hits = self
+                    .probe_clusters_async(col, query, &ctx, &cluster_idx, &locals)
+                    .await?;
+                Ok::<Vec<(u32, f32)>, VectorError>(
+                    hits.into_iter()
+                        .map(|(local_id, score)| (base.saturating_add(local_id), score))
+                        .collect(),
+                )
+            })
+        });
+        let mut merged: Vec<(u32, f32)> = try_join_all(cell_probes)
+            .await?
+            .into_iter()
+            .flatten()
+            .collect();
         // Distance ascending (smaller = closer), matching every other vector
         // search path. Descending here kept the farthest k hits and collapsed
         // packed-shard recall to ~0.
--- a/src/superfile/vector/reader.rs
+++ b/src/superfile/vector/reader.rs
@@ -2856,7 +2856,7 @@ impl VectorReader {
             survivor_only_rerank_fetch,
             &ctx,
         )
-        .await
+        .await?
         {
             ShortlistOutcome::Done(out) => return Ok(out),
             ShortlistOutcome::Rerank {
@@ -3058,6 +3058,33 @@ impl VectorReader {
             .await
     }
 
+    /// Map flat cluster ids to per-cell probe work on a multi-cell
+    /// reader: `(cell index, doc-id base, cell-local cluster ids)`.
+    /// Doc-id bases follow parquet row order (cells concatenated in
+    /// directory order). The single shared resolution step of the
+    /// immediate probe ([`Self::search_clusters_async_multi_cell`]) and
+    /// the deferred scan ([`Self::search_clusters_scan_async`]), so the
+    /// two paths cannot drift.
+    fn resolve_cells_for_clusters(&self, clusters: &[u32]) -> Vec<(usize, u32, Vec<usize>)> {
+        let mut by_cell: HashMap<usize, Vec<usize>> = HashMap::new();
+        for &flat in clusters {
+            let Some((cell_idx, local)) = self.resolve_flat_cluster(flat) else {
+                continue;
+            };
+            by_cell.entry(cell_idx).or_default().push(local as usize);
+        }
+        let mut doc_base = vec![0u32; self.columns.len()];
+        let mut running = 0u32;
+        for (i, col) in self.columns.iter().enumerate() {
+            doc_base[i] = running;
+            running = running.saturating_add(col.n_docs);
+        }
+        by_cell
+            .into_iter()
+            .map(|(cell_idx, locals)| (cell_idx, doc_base[cell_idx], locals))
+            .collect()
+    }
+
     /// Multi-cell probe: map flat cluster ids → (cell, local), probe each
     /// touched cell, merge hits. Local doc ids are unique within a cell
     /// subsection; across cells they may collide, so hits are tagged with
@@ -3095,24 +3122,9 @@ impl VectorReader {
             return Ok(Vec::new());
         }
 
-        // Group flat cluster ids by cell index.
-        let mut by_cell: HashMap<usize, Vec<usize>> = HashMap::new();
-        for &flat in clusters {
-            let Some((cell_idx, local)) = self.resolve_flat_cluster(flat) else {
-                continue;
-            };
-            by_cell.entry(cell_idx).or_default().push(local as usize);
-        }
-        if by_cell.is_empty() {
+        let per_cell = self.resolve_cells_for_clusters(clusters);
+        if per_cell.is_empty() {
             return Ok(Vec::new());
-        }
-
-        // Doc-id bases: parquet rows are concatenated in cell-directory order.
-        let mut doc_base = vec![0u32; self.columns.len()];
-        let mut running = 0u32;
-        for (i, col) in self.columns.iter().enumerate() {
-            doc_base[i] = running;
-            running = running.saturating_add(col.n_docs);
         }
 
         // Probe the selected cells CONCURRENTLY — the same wave shape the
@@ -3122,12 +3134,11 @@ impl VectorReader {
         // walls (measured ~1.2 ms x width per query at w=54 on Cohere-1M).
         // Each cell's probe is independent: own ProbeCtx, own blocks, own
         // rerank; the merge below is order-independent (re-sorted).
-        let cell_probes = by_cell.into_iter().filter_map(|(cell_idx, locals)| {
+        let cell_probes = per_cell.into_iter().filter_map(|(cell_idx, base, locals)| {
             let col = &self.columns[cell_idx];
             if col.n_docs == 0 {
                 return None;
             }
-            let base = doc_base[cell_idx];
             // Allow/deny are file-local; each cell IVF checks cell-local ids.
             let cell_allow = self.cell_local_filter_bitmap(allow.as_deref(), cell_idx, base);
             let cell_deny = self.cell_local_filter_bitmap(deny.as_deref(), cell_idx, base);
@@ -3248,23 +3259,7 @@ impl VectorReader {
                     got: query.len(),
                 });
             }
-            let mut by_cell: HashMap<usize, Vec<usize>> = HashMap::new();
-            for &flat in clusters {
-                let Some((cell_idx, local)) = self.resolve_flat_cluster(flat) else {
-                    continue;
-                };
-                by_cell.entry(cell_idx).or_default().push(local as usize);
-            }
-            let mut doc_base = vec![0u32; self.columns.len()];
-            let mut running = 0u32;
-            for (i, col) in self.columns.iter().enumerate() {
-                doc_base[i] = running;
-                running = running.saturating_add(col.n_docs);
-            }
-            per_cell = by_cell
-                .into_iter()
-                .map(|(cell_idx, locals)| (cell_idx, doc_base[cell_idx], locals))
-                .collect();
+            per_cell = self.resolve_cells_for_clusters(clusters);
         } else {
             let (_, validated) = self.resolve_column(column, query, k)?;
             if !validated {
@@ -3342,7 +3337,7 @@ impl VectorReader {
                     let coarse_limit = k.saturating_mul(rerank_mult);
                     let shortlist =
                         scan_shortlist(col, cb, &cluster_meta, &blocks, true, coarse_limit, &ctx)
-                            .await;
+                            .await?;
                     let cands = shortlist
                         .into_iter()
                         .map(|(did, estimate, pos, cluster_id)| ScanCandidate {
@@ -3655,7 +3650,7 @@ impl VectorReader {
             survivor_only_rerank_fetch,
             ctx,
         )
-        .await
+        .await?
         {
             ShortlistOutcome::Done(out) => {
                 if let Some(t0) = shortlist_t0 {
@@ -4076,7 +4071,7 @@ async fn scan_shortlist(
     survivor_only_rerank_fetch: bool,
     coarse_limit: usize,
     ctx: &ProbeCtx<'_>,
-) -> Vec<(u32, f32, u32, u32)> {
+) -> Result<Vec<(u32, f32, u32, u32)>, VectorError> {
     let full_vec_bytes = col.rerank_codec.per_vector_bytes(col.dim);
     let total_candidates: usize = cluster_meta.iter().map(|&(_, _, cnt)| cnt as usize).sum();
     // Collect every scored row, then keep the top `coarse_limit` with one
@@ -4224,7 +4219,7 @@ async fn scan_shortlist(
                 });
             let _ = tx.send(acc);
         });
-        rx.await.expect("vector scan rayon task dropped result")
+        rx.await.map_err(|_| VectorError::TaskDropped("scan"))?
     } else {
         let cap = coarse_limit.saturating_mul(SHORTLIST_TRUNCATE_SLACK);
         let mut acc = Vec::with_capacity(total_candidates.min(cap));
@@ -4237,7 +4232,7 @@ async fn scan_shortlist(
         acc
     };
     truncate_to_top_estimates(&mut acc, coarse_limit);
-    acc
+    Ok(acc)
 }
 
 /// Keep the `limit` highest-estimate survivors of one cell scan via a
@@ -4299,7 +4294,7 @@ async fn build_shortlist(
     cluster_blocks: &[Bytes],
     survivor_only_rerank_fetch: bool,
     ctx: &ProbeCtx<'_>,
-) -> ShortlistOutcome {
+) -> Result<ShortlistOutcome, VectorError> {
     let full_vec_bytes = col.rerank_codec.per_vector_bytes(col.dim);
     // Score each probed cluster's 1-bit codes into the shortlist.
     // The per-cluster slices are zero-copy `Bytes` views; the actual
@@ -4314,7 +4309,7 @@ async fn build_shortlist(
         ctx.k.saturating_mul(ctx.rerank_mult)
     };
     if coarse_limit == 0 {
-        return ShortlistOutcome::Done(Vec::new());
+        return Ok(ShortlistOutcome::Done(Vec::new()));
     }
     let mut shortlist = scan_shortlist(
         col,
@@ -4325,10 +4320,10 @@ async fn build_shortlist(
         coarse_limit,
         ctx,
     )
-    .await;
+    .await?;
 
     if shortlist.is_empty() {
-        return ShortlistOutcome::Done(Vec::new());
+        return Ok(ShortlistOutcome::Done(Vec::new()));
     }
 
     // `RabitqOnly` short-circuit: the 1-bit shortlist *is* the final
@@ -4346,12 +4341,12 @@ async fn build_shortlist(
         // merge path — otherwise the two paths diverge on top-k for a
         // corrupt/degenerate score.
         shortlist.sort_unstable_by(|a, b| b.1.total_cmp(&a.1).then_with(|| a.0.cmp(&b.0)));
-        return ShortlistOutcome::Done(
+        return Ok(ShortlistOutcome::Done(
             shortlist
                 .into_iter()
                 .map(|(did, est, _pos, _c)| (did, -est))
                 .collect(),
-        );
+        ));
     }
 
     // Build lightweight rerank references into the cluster blocks
@@ -4391,10 +4386,10 @@ async fn build_shortlist(
             full_idx,
         });
     }
-    ShortlistOutcome::Rerank {
+    Ok(ShortlistOutcome::Rerank {
         candidates,
         survivor_full_ranges,
-    }
+    })
 }
 
 /// Maximum multiplier applied to filtered-search probe breadth and
@@ -4548,21 +4543,25 @@ fn spawn_on<F: FnOnce() + Send + 'static>(pool: Option<&ThreadPool>, f: F) {
 /// compute runs on rayon (`par_iter().map().collect()`) bridged back to
 /// the async caller via a oneshot, so no tokio worker blocks under it.
 /// `f` and the items must be `'static` so the work can move onto rayon.
-async fn par_map<T, R, F>(items: Vec<T>, f: F, pool: Option<Arc<ThreadPool>>) -> Vec<R>
+async fn par_map<T, R, F>(
+    items: Vec<T>,
+    f: F,
+    pool: Option<Arc<ThreadPool>>,
+) -> Result<Vec<R>, VectorError>
 where
     T: Send + Sync + 'static,
     R: Send + 'static,
     F: Fn(&T) -> R + Send + Sync + 'static,
 {
     if parallel_chunks(items.len()) <= 1 {
-        return items.iter().map(&f).collect();
+        return Ok(items.iter().map(&f).collect());
     }
     let (tx, rx) = oneshot::channel();
     spawn_on(pool.as_deref(), move || {
         let out: Vec<R> = items.par_iter().map(f).collect();
         let _ = tx.send(out);
     });
-    rx.await.expect("rerank rayon task dropped result")
+    rx.await.map_err(|_| VectorError::TaskDropped("rerank"))
 }
 
 /// The 1-bit scan loop over one cluster's codes, generic over the
@@ -4723,7 +4722,7 @@ async fn rerank_candidates_from_blocks(
                     },
                     pool.clone(),
                 )
-                .await
+                .await?
             } else {
                 candidates
                     .iter()
@@ -4771,7 +4770,7 @@ async fn rerank_candidates_from_blocks(
                         pool.clone(),
                         stride,
                     )
-                    .await
+                    .await?
                 }
                 Sq8ColumnMeta::Lazy {
                     scale_abs_off,
@@ -4810,7 +4809,7 @@ async fn rerank_candidates_from_blocks(
                                 pool.clone(),
                                 stride,
                             )
-                            .await,
+                            .await?,
                             k,
                         ));
                     }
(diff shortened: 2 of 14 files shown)


src/supertable/writer.rs
```diff
@@ -3303,6 +3306,13 @@ pub(in crate::supertable) async fn drain_user_superfiles_to_hidden_cells(
         new_entries.push(entry);
     }
 
+    // Probe-width calibration rides only CLEAN drains: a resumed drain's
+    // checkpointed batches never re-stream, so its sample would be partial
+    // and the law would under-count. A resumed drain keeps whatever law
+    // the manifest already carries.
+    let mut width_law = (local_checkpoint.batches_done == 0 && completed_shards.is_empty())
+        .then(|| opann::WidthLawCalibration::new(running_clusters.dim as usize, metric));
```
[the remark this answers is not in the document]
ekechinwokah: correct in principle, inert today (replication is off by default and currently can't survive a drain at all — the builder.rs crash in the PR notes). deserves a code comment + resolution inside the replication-repair issue, not a blocking change.

src/supertable/opann.rs
```diff
@@ -368,6 +380,248 @@ fn dequantize_row(row: &EncodedCellRow, dim: usize) -> Vec<f32> {
...
+/// scratch (hot loops reuse one allocation of the row's `dim`).
+fn dequantize_row_into(row: &EncodedCellRow, out: &mut [f32]) {
+    dequantize_sq8_residual_into(
+        &row.scale,
+        &row.offset,
+        &row.codes,
+        &row.residuals,
+        row.rerank_codec
+            .residual_divisor()
+            .expect("encoded row uses residual-family codec"),
```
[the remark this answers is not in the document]
ekechinwokah: disagree: drain rows are residual-family by construction; invariant-violation-⇒-panic is this module's existing idiom (boundary_assignment does the same). converting only my call sites would make the module inconsistent.

src/supertable/opann.rs
```diff
@@ -368,6 +380,248 @@ fn dequantize_row(row: &EncodedCellRow, dim: usize) -> Vec<f32> {
...
+            }
+        }
+    }
+
+    /// Freeze the sampled queries. Called once, after the last batch
+    /// spilled and before cell packing scores.
+    pub(crate) fn freeze(&mut self) {
+        let queries = self.reservoir.sample().to_vec();
+        let ids = self.slot_ids.clone();
+        *self.tops.lock().expect("width-law tops lock") = vec![Vec::new(); ids.len()];
```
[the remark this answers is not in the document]
ekechinwokah: agree, fixing with the poison-recovery pattern (no signature churn, no panic path).

src/superfile/vector/reader.rs
```diff
@@ -3103,49 +3103,66 @@ impl VectorReader {
...
+                        .map(|(local_id, score)| (base.saturating_add(local_id), score))
+                        .collect(),
+                )
+            })
+        });
+        let mut merged: Vec<(u32, f32)> = try_join_all(cell_probes)
+            .await?
+            .into_iter()
+            .flatten()
+            .collect();
```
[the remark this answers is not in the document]
ekechinwokah: agree. wave-cap to the pool width is the repo's own idiom; costs nothing warm, bounds cold object-store fan-out.

src/supertable/opann.rs
```diff
@@ -368,6 +379,249 @@ fn dequantize_row(row: &EncodedCellRow, dim: usize) -> Vec<f32> {
...
+        &row.scale,
+        &row.offset,
+        &row.codes,
+        &row.residuals,
+        row.rerank_codec
+            .residual_divisor()
+            .expect("encoded row uses residual-family codec"),
+        out,
+    );
+}
```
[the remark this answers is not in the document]
ekechinwokah: fixed

src/supertable/opann.rs
```diff
@@ -368,6 +379,249 @@ fn dequantize_row(row: &EncodedCellRow, dim: usize) -> Vec<f32> {
...
+            let q = &frozen.queries[qi * self.dim..(qi + 1) * self.dim];
+            let ranked = grid.rank_cells(self.metric, q);
+            for (rank, (cell, _)) in ranked.iter().enumerate() {
+                if let Some(slot) = rank_of_cell.get_mut(*cell as usize) {
+                    *slot = rank as u32;
+                }
+            }
+            let mut sorted = cand.clone();
+            sorted.sort_unstable_by(|a, b| a.0.total_cmp(&b.0));
+            for (ki, &k) in WIDTH_LAW_KS.iter().enumerate() {
```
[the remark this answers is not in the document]
ekechinwokah: fixed

muralikpbhat:
Good find with a real implication to the cost, latency mental model as this shows 1 probe doesnt work always.

Few high level points:
1. Is calibrating and stamping in drains good enough? as it calibrates only new docs and only increases the breadth, it will never readjust if the breadth shrinker for earlier queries? may be we should this in split/merge time also?
2. Cost of calibration? is there a way to restrict it to few sample vectors in the drain instead of doing against all?
3. Is 256 starting count too high for small scale which might be causing this. I know 250 worked well with synthetic, but something to reconsider given this finding.
4. At 40-50 width, is the user table better than global table from cost perspective?

src/superfile/vector/reader.rs
```diff
@@ -3103,49 +3103,75 @@ impl VectorReader {
...
+        // bound every other fan-out here uses): a wide sweep over a
+        // many-celled shard must not open unbounded simultaneous
+        // range reads / cold fetches.
+        let max_in_flight = pool
+            .as_deref()
+            .map(ThreadPool::current_num_threads)
+            .unwrap_or_else(rayon::current_num_threads)
+            .max(1);
+        let per_cell: Vec<Vec<(u32, f32)>> = stream::iter(cell_probes)
+            .buffer_unordered(max_in_flight)
```
muralikpbhat, line 3190: Memory/budget under width. `max_in_flight = pool_width` caps per-superfile cell concurrency, but the unfiltered outer fan-out (the `else` branch in `vector_fanout_over_superfiles`, ~L1819) launches **all** superfiles at once. So peak concurrent cold reservations ≈ `num_superfiles × pool_width`, vs `num_superfiles × 1` on the pre-width p=1 path — a ~pool_width× jump. Since `reserve_cold_fetch` refuses (not queues) at the ceiling, a large drained table with multi-MiB cells can flip previously-successful queries into a hard `OverBudget`. And the **warm** branch (`prefix_blocks_sync = Some`, ~L3336) takes no reservation at all, so its concurrent candidate heaps + survivor buffers are ungated (OOM rather than refuse). The tiny-cell smoke band masks both. Worth wave-capping the unfiltered fan-out too, and reserving on the warm path.

src/supertable/query/vector.rs
```diff
@@ -1363,6 +1363,47 @@ impl SupertableReader {
...
+                };
+            // Explicit caller `nprobe` pins the cell sweep width on every
+            // branch — an override is honored, never discarded. The
+            // persisted hidden routing (fine-first p=1) stays the
+            // no-override default, but measured on Cohere-1M/768d at k=100
+            // the single probed cell caps recall at 0.61 even scanned in
+            // full, so callers need the width dial to buy recall past that
+            // ceiling. Fine depth is untouched by the pin: the filtered
+            // floors above still apply, unfiltered keeps its branch's depth.
+            if options.nprobe.is_some() {
```
muralikpbhat: `sweep_width` is set here whenever `options.nprobe.is_some()`, with no `!filtered` guard — so an explicit-`nprobe` **filtered** query also picks up the width-era changes below: the rerank-budget divide (~L1718) and the per-fragment `generation_of` flip (~L1504). The PR scopes the width work to the unfiltered post-drain path, but filtered+`nprobe` callers get pulled in, and all three new tests are unfiltered so this path is uncovered. Consider a `!filtered` gate plus a filtered+nprobe regression test.

src/supertable/query/vector.rs
```diff
@@ -1646,6 +1703,22 @@ impl SupertableReader {
...
+        // probed cell "survives" the 1-bit prune and the survivor-only
+        // rerank fetch degenerates into fetching whole cells (measured:
+        // ~6.5 MB x 54 cells per query at w=54). Divide the cap across the
+        // sweep so the TOTAL survivor budget stays ~k x rerank_mult —
+        // measured sufficient at that total: 0.9957 recall@100 on
+        // Cohere-1M with ~25.6K survivors.
+        let options = match sweep_width {
+            Some(w) if w > 1 => {
+                let (_, rerank_mult) = options.resolve(filtered);
+                options.with_rerank_mult((rerank_mult * WIDTH_BUDGET_OVERSAMPLE).div_ceil(w).max(1))
```
muralikpbhat: Three concerns with this rerank-budget divide:

1. It divides by the **requested** `w`, not the number of cells actually probed. When fewer cells carry postings than `w` (or a caller passes `nprobe` > populated cells), the per-cell survivor cap is over-shrunk and the shortlist is starved → recall drops.
2. It's independent of `k` and cell size. Calibrated at k=100, but at k=1000 `k × ceil(rerank_mult*4 / w)` can exceed a cell's row count, so every 1-bit-shortlist row 'survives' and the survivor-only rerank fetch degenerates back to whole-cell reads — the exact blowup this divide is meant to prevent.
3. `rerank_mult * WIDTH_BUDGET_OVERSAMPLE` is an unchecked multiply; a caller-supplied large `rerank_mult` overflows (debug panic / release wrap → near-zero budget). Consider a saturating mul and bounding the per-cell cap by `k` and the cell's `n_docs`.

src/supertable/query/vector.rs
```diff
@@ -1449,6 +1490,22 @@ impl SupertableReader {
...
+                // count — which also means widening the cell sweep alone
+                // cannot deepen the read (measured flat 0.443 recall@100
+                // from nprobe=4 through 256 on Cohere-1M). An explicit
+                // caller `nprobe` — or the calibrated width law standing in
+                // for one — is a request to actually read N cells, so gate
+                // per (cell, fragment) exactly like the pre-drain user path:
+                // depth follows width, read amplification is what was asked
+                // for (explicitly by the caller, or measured as necessary
+                // by the drain's calibration).
+                let generation_of = if options.nprobe.is_some() || law_width.is_some() {
```
muralikpbhat: `generation_of` flips from wave-pooled (`Some(birth_versions)`) to per-`(cell, fragment)` (`None`) whenever `options.nprobe.is_some()` — which includes **filtered** queries with an explicit `nprobe` (via the ungated `sweep_width` at ~L1392). On a multi-generation hidden table that changes read amplification (GETs scale with `w × fragments-per-cell` instead of wave count) on a path the PR states it doesn't touch. Gate on `!filtered` if that's the intent.

src/supertable/opann.rs
```diff
@@ -368,6 +379,253 @@ fn dequantize_row(row: &EncodedCellRow, dim: usize) -> Vec<f32> {
...
+            }
+            // Bound the per-cell partials the same way the merge does.
+            for cand in &mut partial {
+                truncate_ascending(cand, k_max);
+            }
+        }
+        let mut tops = self.tops.lock().unwrap_or_else(PoisonError::into_inner);
+        for (qi, mut cand) in partial.into_iter().enumerate() {
+            tops[qi].append(&mut cand);
+            truncate_ascending(&mut tops[qi], k_max);
```
muralikpbhat: Candidates are truncated to `k_max` (=1000, the top knot) **before** `finish()` dedups by stable id (~L588). On a replicated table (`drain_replica_target_factor > 1`), boundary-replica copies occupy raw top-1000 slots, so after dedup the distinct count can be < 1000 → `supported[k=1000]` goes false → `law[3]` is stamped 0 → k=1000 queries clamp down to the k=100 width, silently under-probing exactly the high-k path the law protects. Consider dedup-before-truncate, or sizing the raw buffer to `k_max × max_replica_factor`.

src/supertable/opann.rs
```diff
@@ -368,6 +379,253 @@ fn dequantize_row(row: &EncodedCellRow, dim: usize) -> Vec<f32> {
...
+            }
+            // Best-scored copy per stable id first: boundary replicas of
+            // one row must count as ONE neighbor, or replicated tables
+            // would measure narrower coverage than a query experiences.
+            let mut sorted = cand.clone();
+            sorted.sort_unstable_by(|a, b| a.2.cmp(&b.2).then_with(|| a.0.total_cmp(&b.0)));
+            sorted.dedup_by_key(|c| c.2);
+            sorted.sort_unstable_by(|a, b| a.0.total_cmp(&b.0));
+            for (ki, &k) in WIDTH_LAW_KS.iter().enumerate() {
+                if sorted.len() < k {
```
muralikpbhat, line 896: `supported[ki]` is a single sample-wide flag: one query whose deduped candidate count is < k zeros that knot for **all** sample queries. With `k_max` == the top knot (1000) and zero headroom, a single boundary-heavy query reaching 999 distinct discards the whole k=1000 measurement. A per-query drop (exclude that query from `coverage_sums`, divide the target by the supported-query count) would preserve the knot.

src/supertable/opann.rs
```diff
@@ -368,6 +379,253 @@ fn dequantize_row(row: &EncodedCellRow, dim: usize) -> Vec<f32> {
...
+                // accumulates the mean coverage curve.
+                let mut per_rank = vec![0u32; n_cells];
+                for (_, cell, _) in &sorted[..k] {
+                    per_rank[rank_of_cell[*cell as usize] as usize] += 1;
+                }
+                let mut covered = 0u32;
+                for (rank, count) in per_rank.iter().enumerate() {
+                    covered += count;
+                    coverage_sums[ki][rank] += f64::from(covered) / k as f64;
+                }
```
muralikpbhat, line 919: Each knot's width is the first rank hitting target coverage, computed independently — nothing enforces monotonic-nondecreasing width across increasing k. Sampling noise near the 0.99 threshold can yield e.g. `[30,45,52,48]`, and `width_for_k_at` then interpolates so a k=1000 query probes fewer cells than a k=100 one (a recall inversion). A running max — `law[i] = law[i].max(law[i-1])` — makes the law monotone as the design assumes.

src/supertable/writer.rs
```diff
@@ -3904,6 +3943,35 @@ pub(in crate::supertable) async fn drain_user_superfiles_to_hidden_cells(
...
+        // real cross-generation spread until a full rebuild recalibrates.
+        // The alternative (rescoring every packed cell per incremental
+        // drain) would make drain cost track table size, not delta size.
+        if let Some(cal) = width_law.take()
+            && let Some(law) = cal.finish(&running_clusters)
+        {
+            for (slot, measured) in routing.width_for_k.iter_mut().zip(law) {
+                *slot = (*slot).max(measured);
+            }
+            eprintln!(
```
muralikpbhat: `eprintln!` on every clean drain (and several siblings in this file — ~L1638/3075/3185/3448/3699/4008/4019) writes to stderr straight from the engine library, bypassing the logging facade with no way to silence it without a rebuild. Route through the crate's log/tracing macros (or drop).

muralikpbhat (commented):
Line-level findings from a detailed pass through the routing / width-law / budget paths. Left inline below; all verified against the diff at this head. Comments only — not a review verdict.

src/superfile/vector/reader.rs
ekechinwokah: agreed. fixed.

src/supertable/query/vector.rs
ekechinwokah: agreed. sweep_width (and through it the generation_of flip) now gates on !filtered — filtered+nprobe keeps its pre-width budget and pooled gating, still honors the caller's cell width. on coverage: the multi-cell rerank test already ran filtered+nprobe post-drain but only asserted non-empty; strengthened it to exact parity with its unfiltered twin.

src/supertable/query/vector.rs
ekechinwokah: same root as the sweep_width gate — fixed. generation_of now keys off sweep_width, so filtered queries stay wave-pooled regardless of nprobe.

src/supertable/opann.rs
ekechinwokah: fixed. dedup moved into the merge, before any truncate — replicas can never occupy raw slots. the replicated-rows test now plants through the real merge path.

src/supertable/opann.rs
ekechinwokah: fixed. each point is measured over the queries that support it; one boundary-starved query sits the point out instead of zeroing it for the sample. only a point no query supports stays 0.

src/supertable/opann.rs
ekechinwokah: fixed — measured points are floored by the measured points below them. zeros stay zero (unmeasured, the interpolator skips them).

src/supertable/writer.rs
ekechinwokah: fixed — the stamp rides tracing::info now.

ekechinwokah:
> Good find with a real implication to the cost, latency mental model as this shows 1 probe doesnt work always.
> 
> Few high level points:
> 
> 1. Is calibrating and stamping in drains good enough? as it calibrates only new docs and only increases the breadth, it will never readjust if the breadth shrinker for earlier queries? may be we should this in split/merge time also?

deliberate one-way door for now: max-merge only floors, shrink needs a full re-measure and only a full rebuild does one. re-measuring at split/merge is doable if we persist the query sample (~200KB sq8 for 256×768d) so maintenance can re-score it against the touched grid — good follow-up issue.

> 2. Cost of calibration? is there a way to restrict it to few sample vectors in the drain instead of doing against all?

it's already sampled on the query side — 256 reservoir rows, that's the whole sample. the ~15s at 1M is those 256 queries scored against every packed cell, and the cell side has to be exhaustive or the coverage measurement isn't exact. rides work the drain already does, resumed drains skip it entirely.

> 3. Is 256 starting count too high for small scale which might be causing this. I know 250 worked well with synthetic, but something to reconsider given this finding.

measured: it isn't the lever. 256 / 1024 / 4096 all need ~32% of corpus for 0.99 coverage on cohere — the spread is a property of the embedding distribution, not grid granularity. same result under a fully converged reference clustering, so it's not grid quality either.

> 4. At 40-50 width, is the user table better than global table from cost perspective?

the coverage requirement is layout-independent — any layout has to touch those bytes to hit 0.99 on this corpus. the global table delivers them as a handful of large contiguous gets from packed cells; the user table would spray the same rows across every commit's superfile plus per-file overhead. width changes how much we read, not which table reads it cheaper.

src/supertable/opann.rs
```diff
@@ -694,6 +701,261 @@ pub(crate) fn plan_sq8_split_kway(
...
+                    // points and sits out this one.
+                    continue;
+                }
+                support[ki] += 1;
+                // Per-rank counts of this query's top-k, then a prefix walk
+                // accumulates the mean coverage curve.
+                let mut per_rank = vec![0u32; n_cells];
+                for (_, cell, _) in &sorted[..k] {
+                    per_rank[rank_of_cell[*cell as usize] as usize] += 1;
+                }
```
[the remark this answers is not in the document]
ekechinwokah: candidate cell ids come from the same drain that built the grid, and split parents keep their slot when superseded, so ids only grow within a drain and finish's grid is the post-update one. but the guard is nearly free and calibration is best-effort diagnostics — abandoning the law beats any possibility of panicking a drain.

src/supertable/query/vector.rs
ekechinwokah: all three addressed. width is clamped to the cells that actually carry postings before it divides anything, and the oversample multiply is saturating. the k=1000 degeneration is real but it's a cost problem, not recall — when k×budget exceeds cell rows the survivor fetch converges to the whole-cell read the cache-fill does anyway. the actual fix is a global shortlist across the sweep (merge estimates cross-unit at k×rerank_mult total instead of dividing per cell) — that replaces the divide entirely. adding to pr.

src/superfile/vector/reader.rs
```diff
@@ -3653,7 +4072,78 @@ async fn build_shortlist(
         }
         heap
     };
-    let mut shortlist = shortlist_heap.into_vec();
+    shortlist_heap.into_vec()
```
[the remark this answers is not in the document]
ekechinwokah: valid — but build_shortlist and par_map have the identical expect; fixing only this one leaves 2 of 3 sites. Will convert all three to error returns matching exec::common (same treatment as infino#470's fts bridge).

src/supertable/query/vector.rs
```diff
@@ -1793,6 +1906,86 @@ impl SupertableReader {
...
+        // `k x rerank_mult` estimates ACROSS every warm-scanned cell and
+        // superfile, then rerank only those winners where they live. The
+        // estimates are comparable across units — one rotation seed per
+        // column, table-wide — asserted here at the only place different
+        // units' estimates ever meet.
+        if global_shortlist_width.is_some() {
+            let pooled = {
+                let mut guard = scan_pool.lock().unwrap_or_else(PoisonError::into_inner);
+                mem::take(&mut *guard)
+            };
```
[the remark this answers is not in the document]
ekechinwokah: the partial-shortlist scenario can't happen: phase C runs after fanout_with(...).await?, so any unit failure aborts the query before the pool is read, and the lock is a short non-await push (std-Mutex-in-async sanctioned case). keeping as-is; the per-unit-return refactor is nicer but threads a new return type through shared dispatch plumbing for no behavior change.

src/superfile/reader.rs
```diff
@@ -1452,6 +1452,63 @@ impl SuperfileReader {
...
+        clusters: &[u32],
+        options: VectorSearchOptions,
+        cold_rerank_mult: usize,
+        allow: Option<Arc<RoaringBitmap>>,
+        deny: Option<Arc<RoaringBitmap>>,
+        pool: Option<Arc<ThreadPool>>,
+        budget: Option<Arc<ConnectionMemoryBudget>>,
+    ) -> Result<ScanOutcome, ReadError> {
+        let filtered = allow.is_some();
+        let (_, rerank_mult) = options.resolve(filtered);
```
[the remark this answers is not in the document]
ekechinwokah: no-op: resolve(filtered) only varies nprobe, which this call site discards (clusters arrive precomputed); rerank_mult is flag-independent. also deny here is tombstones — treating it as "filtered" would widen probes for every table with any deletes, while deletion recall is already covered by fetch slack + backfill (tombstone_filter tests).

muralikpbhat:
> measured: it isn't the lever. 256 / 1024 / 4096 all need ~32% of corpus for 0.99 coverage on cohere — the spread is a property of the embedding distribution, not grid granularity. same result under a fully converged reference clustering, so it's not grid quality either.

sorry, I meant a lower cell count...just a hunch that 256 cells is probably too high for the native clusters in this corpus. If we try something like 2-4-8-16 cells with nprobe=1, what happens? how big is this corpus?

src/superfile/vector/reader.rs
```diff
@@ -3154,6 +3176,328 @@ impl VectorReader {
...
+                    .map_err(|e| VectorError::LazySource(e.to_string()))?;
+                let mut q_rot = vec![0f32; col.dim];
+                col.rot.apply(query, &mut q_rot);
+                let cb = col.quant.code_bytes();
+                let (cluster_meta, prefix_ranges) = chosen_cluster_meta(col, &cluster_idx, &locals);
+                if cluster_meta.is_empty() {
+                    return Ok((Vec::new(), Vec::new()));
+                }
+                // Warm fast path: every prefix resident -> defer rerank.
+                let prefix_blocks: Option<Vec<Bytes>> = prefix_ranges
```
muralikpbhat, line 3322: **Recall is cache-dependent here.** A fully-resident (warm) cell defers with **undivided** `k·rerank_mult` and competes in the global shortlist; but if any prefix is evicted, the whole cell drops to the cold arm and reranks locally under the **width-divided** `cold_rerank_mult`, never entering global selection. So the same query's top-k varies with cache state. It's *not* a regression (warm ≥ cold ≥ old behavior — this only adds the better warm path), but the recall gains are **warm-biased**. Worth confirming the ≥0.99 gate is measured on cold queries too.

src/supertable/query/vector.rs
```diff
@@ -1793,6 +1906,86 @@ impl SupertableReader {
...
+                debug_assert!(
+                    pooled.windows(2).all(|w| w[0].1 == w[1].1),
+                    "pooled 1-bit estimates require one rotation seed per column"
+                );
+                let (_, rerank_mult) = options.resolve(filtered);
+                let mut flat: Vec<(usize, ScanCandidate)> = pooled
+                    .into_iter()
+                    .flat_map(|(si, _, cands)| cands.into_iter().map(move |c| (si, c)))
+                    .collect();
+                flat = select_global_shortlist(flat, k.saturating_mul(rerank_mult));
```
muralikpbhat: The global cut uses `k·rerank_mult`, but the per-cell path fetches `k_fetch = k + replica_overhead`. **Dormant today** (`replica_overhead == 0` with replication off by default), but if `drain_replica_target_factor > 1` is ever re-enabled, boundary replicas take global-shortlist slots *before* the stable-id dedup and silently short the distinct top-k. Suggest `(k + replica_overhead)·rerank_mult` here for consistency with phase A.

src/supertable/query/vector.rs
```diff
@@ -1793,6 +1906,86 @@ impl SupertableReader {
...
+        // units' estimates ever meet.
+        if global_shortlist_width.is_some() {
+            let pooled = {
+                let mut guard = scan_pool.lock().unwrap_or_else(PoisonError::into_inner);
+                mem::take(&mut *guard)
+            };
+            if !pooled.is_empty() {
+                debug_assert!(
+                    pooled.windows(2).all(|w| w[0].1 == w[1].1),
+                    "pooled 1-bit estimates require one rotation seed per column"
```
muralikpbhat: Cross-unit estimate comparability (one rotation seed per column) is guarded here only by `debug_assert!`, stripped in release — at the one site where estimates from different superfiles are actually pooled and ranked. Backstopped today by the open-time seed check, so not reachable; but as the only guard at the comparison site, a hard error (or fall-back to the divided per-cell path) would fail loud if a future path ever admitted a differently-seeded unit.

src/superfile/vector/reader.rs
```diff
@@ -3154,6 +3176,328 @@ impl VectorReader {
...
+                for cand in &cands {
+                    extent_by_cid.entry(cand.cluster_id).or_insert_with(|| {
+                        read_cluster_entry(&cluster_idx, cand.cluster_id as usize)
+                    });
+                }
+                let mut ranges = Vec::with_capacity(cands.len());
+                let mut rerank_cands = Vec::with_capacity(cands.len());
+                for cand in &cands {
+                    let (off, cnt) = extent_by_cid[&cand.cluster_id];
+                    let local = (cand.pos - off) as usize;
```
muralikpbhat, line 3454: `let local = (cand.pos - off) as usize` is an unguarded `u32` subtraction. Safe today — `off` is re-read from the *same immutable superfile* the scan ran on, so `off_scan == off_rerank` and `pos = off + i ≥ off`. A `debug_assert!(cand.pos >= off)` (or `saturating_sub`) would make that invariant explicit and turn any future layout drift into a loud failure instead of a ~4e9 `local` / out-of-range byte range.

src/superfile/vector/reader.rs
```diff
@@ -3154,6 +3176,328 @@ impl VectorReader {
...
+                // the coalesce plan's gap window — correct for remote
+                // fetches, where round trips dominate — merges them into
+                // ~whole-cell spans that a resident cache then materializes
+                // (measured: ~300 MB touched per query to deliver ~10 MB of
+                // rows, 60% of warm query wall). When every row resolves
+                // synchronously from resident bytes, read the exact ranges
+                // instead; fall back to the coalesced remote plan otherwise.
+                let sync_rows: Option<Vec<Bytes>> = ranges
+                    .iter()
+                    .map(|range| self.source.try_get_range_sync(range.clone()))
```
muralikpbhat, line 3490: Survivor gather is all-or-nothing on residency: `ranges.iter().map(try_get_range_sync).collect::<Option<_>>()` → a **single** evicted range makes the whole cell fall back to the coalesced whole-cell fetch this commit is meant to avoid. Under partial eviction it silently reverts the optimization; exact-fetching only the missing ranges would preserve it.

src/superfile/vector/reader.rs
```diff
@@ -3154,6 +3176,328 @@ impl VectorReader {
...
+    ///   surface the survivors with their estimates — the supertable pools
+    ///   them across cells and superfiles, selects the global top
+    ///   `k × rerank_mult`, and calls [`Self::rerank_selected_async`] with
+    ///   the winners. No survivor fetch, no rerank here.
+    /// - COLD (or `RabitqOnly`, which has no rerank stage): probe exactly
+    ///   as [`Self::search_clusters_async`] does, under the width-divided
+    ///   `cold_rerank_mult` — deferring a cold cell would hold its fetched
+    ///   blocks and budget reservation across the whole fan-out.
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) async fn search_clusters_scan_async(
```
muralikpbhat, line 3216: `search_clusters_scan_async`'s cold arm re-implements `search_clusters_async`'s per-cell resolution (by_cell grouping, doc_base loop, cell filter, cold `ProbeCtx`) — and **already differs** (a single-cell v1 branch the original lacks). The heap-free fits-path in `scan_shortlist` (~L3914) likewise duplicates the heap path's rayon scaffolding. Two near-copies of the scan will drift; a shared generic-over-sink scanner would keep them honest. Altitude/maintainability, not correctness.

muralikpbhat (approved):
## Review — the new deferred-rerank scan/gather/select path (last 6 commits, ~1k lines)

**No reachable correctness bug.** I traced the new fast paths against the proven immediate-rerank path across several angles plus verification; the scary-sounding findings all turn out **latent or backstopped** (details inline). CI is green and the earlier review is resolved.

### Three quick items to close before merge
1. **Commit `34bdfb9`'s message** references **"045 §6c"** — a number with no context in this public repo. Reword the measurement in plain language and drop the reference (needs a commit amend).
2. **Cold-query recall** — the deferred-global path improves **warm** cells only (see the inline note at `vector/reader.rs`). Please confirm the ≥0.99 recall gate is measured on **cold** queries too, since the gains are warm-biased.
3. Small cell count test as mentioned in the previous comment

### Follow-ups (non-blocking)
- A **deferred-vs-immediate top-k parity test** — nothing currently asserts the two paths agree, and there's no cold/replica fixture; that's what would have caught any of the below early.
- The small defensive guards inline (replica-overhead consistency, release guard for the seed comparison, the `pos - off` assert).
- De-duplicating the two scan copies.
- other comments on new 6 commits below.

**Approved, contingent on response addressing the three pre-merge items above; no need for re-review.** No blocker: pre-merge items are quick, and everything else is good to have.

ekechinwokah:
> ## Review — the new deferred-rerank scan/gather/select path (last 6 commits, ~1k lines)
> **No reachable correctness bug.** I traced the new fast paths against the proven immediate-rerank path across several angles plus verification; the scary-sounding findings all turn out **latent or backstopped** (details inline). CI is green and the earlier review is resolved.
> 
> ### Three quick items to close before merge
> 1. **Commit `34bdfb9`'s message** references **"045 §6c"** — a number with no context in this public repo. Reword the measurement in plain language and drop the reference (needs a commit amend).
> 2. **Cold-query recall** — the deferred-global path improves **warm** cells only (see the inline note at `vector/reader.rs`). Please confirm the ≥0.99 recall gate is measured on **cold** queries too, since the gains are warm-biased.
> 3. Small cell count test as mentioned in the previous comment
> 
> ### Follow-ups (non-blocking)
> * A **deferred-vs-immediate top-k parity test** — nothing currently asserts the two paths agree, and there's no cold/replica fixture; that's what would have caught any of the below early.
> * The small defensive guards inline (replica-overhead consistency, release guard for the seed comparison, the `pos - off` assert).
> * De-duplicating the two scan copies.
> * other comments on new 6 commits below.
> 
> **Approved, contingent on response addressing the three pre-merge items above; no need for re-review.** No blocker: pre-merge items are quick, and everything else is good to have.

1. done. reworded in plain language, doc reference dropped. hashes moved from that commit forward.
2. measured. cold recall with no disk cache at all — every probe takes the cold arm (whole-block fetch + width-divided in-probe rerank), same 100 queries as the warm sweeps: k=100 default 0.9960, rm=64 0.9927. k=10 default 0.9950, rm=64 0.9910. identical digits to warm at every point, so recall doesn't move with cache state at these budgets — the deferred-global path's gains are latency-only. good catch though, we'd never actually measured it.
3. measured your low-cell hunch on the real corpus (1M rows, 768d — minibatch kmeans, coverage of the exact top-100 at nprobe=1/2): cells 2: 0.905 / 1.000 (71.3 MB/q at np1) · 4: 0.853 / 0.952 (52.1) · 8: 0.713 / 0.910 (48.8) · 16: 0.525 / 0.710 (16.7) · 32: 0.498 / 0.695 (6.7) even two giant cells leak ~10% of the top-100 across one boundary, and every low-count point is strictly worse on bytes than the ~50-cell law point (~20 MB for 0.99-class coverage). the spread is the embedding distribution, not granularity — in either direction.

src/superfile/vector/reader.rs
```diff
@@ -3653,7 +4088,78 @@ async fn build_shortlist(
         }
         heap
     };
-    let mut shortlist = shortlist_heap.into_vec();
+    shortlist_heap.into_vec()
+}
```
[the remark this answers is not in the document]
ekechinwokah: taken (in 902fc8c): rot_seed from columns.first() was a real latent bug on multi-column files — seed now comes from the requested column's resolved cells. pooled-seed debug_assert promoted to a hard error. hidden_routing expect → QueryError. collect-before-truncate now has an amortized guard — accumulators truncate back to the cap at 2x growth, so worst-case memory is 2x coarse_limit entries per task instead of every scanned row (at real cell shapes, ~9k rows vs a 6.4k cap, the guard never fires).

src/superfile/vector/reader.rs
ekechinwokah: measured it. cold recall with no disk cache at all — every probe takes the cold arm, whole-block fetch + width-divided in-probe rerank, same 100 queries as the warm sweeps: k=100 default 0.9960, rm=64 0.9927; k=10 default 0.9950, rm=64 0.9910. identical digits to warm at every point, so the taxonomy you wrote (warm ≥ cold ≥ old) is right but the delta measures zero at these budgets — the cold arm holds the gate on its own and the deferred-global path's gains are latency-only. good catch regardless.

src/supertable/query/vector.rs
ekechinwokah: done. phase a now pools each unit's replica_overhead alongside its candidates and the global cut is (k + max overhead) × rerank_mult, mirroring the per-cell k_fetch. dormant today with replication off, consistent if drain_replica_target_factor ever comes back.

src/supertable/query/vector.rs
ekechinwokah: agreed — promoted to a hard QueryError. it's the one site where estimates from different superfiles actually rank against each other, so it should fail loud, not strip in release. still backstopped by the open-time seed check, but the guard now lives where the invariant matters. same commit also fixes a related latent bug the bot found: ScanOutcome reported columns[0]'s rot_seed instead of the requested column's — wrong seed on multi-column files, and release had no check at all.

src/superfile/vector/reader.rs
ekechinwokah: done. debug_assert!(pos >= off && pos - off < cnt) with the extent in the message, so any future layout drift fails loud in debug instead of wrapping into a ~4e9 local. kept release behavior unchanged since the bytes are immutable and off comes from the same superfile the scan ran on, exactly as you said.

src/superfile/vector/reader.rs
ekechinwokah: deliberate, after looking at what mixed residency actually is here: the block cache evicts whole entries per superfile, so a partially-resident cell means partial fill (some 512K blocks landed, others never touched), not partial eviction. in that state the survivors needing remote bytes are scattered ~1.5KB rows — exact-fetching only the missing ones is dozens of tiny GETs, which is the round-trip regime the coalesce window exists for. and the coalesced fallback doubles as fill-forward: it completes the cell's blocks, so the next query takes the sync fast path. so all-or-nothing = one query pays coalesced reads while the cell finishes filling, then it's warm forever. happy to add a counter if you want visibility into how often the fallback fires.

src/superfile/vector/reader.rs
ekechinwokah: done: the fits-path/heap-path split inside scan_shortlist is gone — the bounded heap is deleted and there's one collect + select_nth path, so that pair can't drift anymore. the cold-arm duplication against search_clusters_async stands and you're right about the altitude — now done together with the deferred-vs-immediate parity test and a cold/replica fixture in this PR, since that test is what would catch drift between the remaining pair.

src/superfile/vector/reader.rs
```diff
@@ -3540,30 +3872,133 @@ enum ShortlistOutcome {
...
+                        }
+                        acc
+                    })
+                    .reduce(Vec::new, |mut a, mut b| {
+                        a.append(&mut b);
+                        a
+                    });
+                let _ = tx.send(acc);
+            });
+            return rx.await.expect("vector scan rayon task dropped result");
```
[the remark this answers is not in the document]
ekechinwokah: declined, with reasons: rx.await.expect → Result — that exact expect is the file-wide oneshot-bridge pattern (par_map has the identical one; this diff moved it, didn't add it), converting one site leaves the rest inconsistent, queued a follow-up to sweep them all at once. poisoned-mutex-as-error — PoisonError::into_inner is the established idiom here, and a poisoned scan pool means a unit already panicked, which already fails the fanout join before phase c reads the pool. smoke-band lower bound at 500k — that corpus is synthetic and calibrates to width 1, so the wide sweep legitimately doesn't engage; a 500k floor fails the exact case the control covers. width-law stamping guard as proposed — it would block calibration on the first clean drain after rollout, which is the case the law exists for; the clean_uncheckpointed_drain gate already prevents resumed-tail stamping, samples score against every packed cell, and max-merge never narrows.

PR #487, opened 2026-07-30
fix: query side reader pool

savannahar68:
## Route FTS query CPU through the configured reader pool

reader_threads only bounded CPU concurrency for the ranged multi-term OR path. Everything else (prefix, AND/must-should queries) scored inline on tokio workers, so reader_threads: 1 didn't work as intended

### how did we change this ? 

- One shared async→rayon bridge helper instead of three copies.
- Prefix queries share FST expansion + cursor build per superfile instead of redoing it per doc-id slice.
- Un-ranged AND/must-should queries bridge to the reader pool when the scan is heavy (gated on term document frequency, since this path isn't sliced and can't gate on term count alone)

### DIFF
--- a/src/superfile/fts/reader.rs
+++ b/src/superfile/fts/reader.rs
@@ -220,22 +220,18 @@ pub(crate) enum PreparedClauses {
 }
 
 impl PreparedClauses {
-    /// Sum of term document frequencies across the held cursors — the
-    /// scan-cost proxy callers gate reader-pool dispatch on. `Done` has
-    /// nothing left to scan, so it's zero.
+    /// Scan-cost proxy callers gate reader-pool dispatch on: the driving
+    /// (smallest) posting list for the AND-intersect shapes, the full
+    /// union for OR. `Done` has nothing left to scan, so it's zero.
     pub(crate) fn posting_mass(&self) -> u64 {
         match self {
             PreparedClauses::Done(_) => 0,
-            PreparedClauses::Must { must_cursors, .. } => must_cursors.iter().map(|c| c.df).sum(),
-            PreparedClauses::MustShould {
-                must_cursors,
-                should_cursors,
-                ..
-            } => must_cursors
-                .iter()
-                .chain(should_cursors)
-                .map(|c| c.df)
-                .sum(),
+            PreparedClauses::Must { must_cursors, .. } => {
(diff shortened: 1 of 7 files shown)


src/superfile/fts/reader.rs
```diff
@@ -184,6 +184,63 @@ impl ClauseLists<'_> {
...
+}
+
+impl PreparedClauses {
+    /// Sum of term document frequencies across the held cursors — the
+    /// scan-cost proxy callers gate reader-pool dispatch on. `Done` has
+    /// nothing left to scan, so it's zero.
+    pub(crate) fn posting_mass(&self) -> u64 {
+        match self {
+            PreparedClauses::Done(_) => 0,
+            PreparedClauses::Must { must_cursors, .. } => must_cursors.iter().map(|c| c.df).sum(),
```
vinaykakade: I think the scan cost-proxy for must should be min of dfs, instead of sum?

src/superfile/fts/reader.rs
```diff
@@ -1561,13 +1624,29 @@ impl FtsReader {
...
     ) -> Result<Vec<(u32, f32)>, FtsError> {
+        let prep = self.prepare_clauses(column, lists, k, floor).await?;
+        self.run_prepared(prep)
+    }
+
+    /// I/O half of an un-ranged clause search: resolve the column,
+    /// classify the query shape, and fetch every cursor
+    /// [`Self::run_prepared`] needs to score. The single-atom and
+    /// phrase-atom shapes finish here instead, since each is cheap
+    /// enough that handing it off isn't worth it.
```
vinaykakade: nit: phrase scoring isn't necessarily cheap. The comment should be updated.

vinaykakade (approved):
See suggested improvements. Also, please run the benchmark and ensure no FTS perf regression between this PR and latest main before checking-in.

src/superfile/fts/reader.rs
savannahar68: done

src/superfile/fts/reader.rs
savannahar68: fixed

savannahar68:
reran benches couple of times
No regression. Just some noise ~5% for some queries

PR #492, opened 2026-07-31
fix: handle terms posting length exceeding pfor length

savannahar68:


src/superfile/fts/fst_value.rs
```diff
@@ -9,7 +9,9 @@
 //! - `value & 1 == 0` → **PFOR form**. The payload stores both
 //!   `metadata_offset` and `postings_length`, so the reader can fetch
 //!   the complete term range in one GET instead of probing the 20 B
-//!   metadata header first.
+//!   metadata header first. `postings_length` is a hint — the header
```
vinaykakade, line 12: If it is a hint, perhaps rename to "postings_length_hint", so that it is easier for the human/agent readers in the future?

src/superfile/builder.rs
```diff
@@ -2857,6 +2857,64 @@ mod tests {
         assert_eq!(results_merged.len(), 2);
     }
 
+    /// Merged `df` for a shared term here is well past the point
+    /// where its postings outgrow the FST value's 21-bit length slot.
+    #[tokio::test(flavor = "multi_thread")]
+    async fn build_from_readers_merges_common_term_past_pfor_length_slot() {
+        const NUM_FILES: usize = 12;
+        const DOCS_PER_FILE: usize = 450_000;
```
vinaykakade, line 2865: The ids here are sequential and very compressible. Is this actually going past max pfor length? Please make sure that this test fails without your code changes in a throwaway branch.

vinaykakade (approved):
The changes themselves look good, but few comments and suggestions.

src/superfile/builder.rs
savannahar68: yeah it panics without code change. Already tested it

src/superfile/fts/fst_value.rs
savannahar68: done

PR #495, opened 2026-07-31
Remove VectorConfig::n_cent from the public API and table identity hash

### LINKED ISSUE #488: VectorConfig::n_cent is a mandatory user option but is build-time-only and part of the table identity hash
## Summary
Creating a vector index requires the caller to supply `VectorConfig::n_cent` (the IVF fine-cluster count). But `n_cent` is a build-time-only tuning value that the engine already derives and caps from the data, and it is also folded into the manifest options hash. This has two costs: (a) users must pass a required number they have no principled way to choose, and (b) the value becomes part of the table's identity, so every `open()` must be handed the identical value or it fails with `ContentHashMismatch`.

## `n_cent` is not needed as a user input
- **Query never reads it.** The reader takes the actual cluster count from the superfile's own on-disk cell directory (`superfile/format/mod.rs`, where the layout is parameterized by the stored count) — not from the options.
- **Merge/compaction re-derives it.** During drain/merge the builder recomputes `n_cent` per cell from the data and ignores the config template (`superfile/vector/builder.rs`).
- **The engine already sizes it from row count.** Every fresh build clamps `cfg.n_cent.max(1).min(n_cent_row_count_cap(rows)).min(rows)` (banded 64 / 1024 / 4096). So the caller's value is only an upper bound; the engine already knows how to pick a row-appropriate count.

Net: `n_cent` only affects the granularity of a *fresh, pre-compaction* superfile's local IVF — a transient structure the engine can size itself.

## `n_cent` should not be in the identity hash
`compute_options_hash` includes `n_cent` (`supertable/manifest/options_hash.rs`). Since the hash is the table's stored identity, `open()` recomputes it from the passed options and compares to the stored value. A table therefore permanently remembers its `n_cent`, and any reopen must re-supply the identical value or fail with `ContentHashMismatch`. An identity constraint on a build-time-only, self-described, re-derived value serves no correctness purpose — and it makes reopen brittle for any caller that derives `n_cent` from a value that can drift.

## Impact
- **API ergonomics:** a required IVF cluster count with no sensible default and no way for a user to reason about it. Public index specs thread it straight through (`catalog/index_spec.rs`).
- **Reopen fragility:** reopening a table requires knowing and re-passing the exact original `n_cent`.

## Proposed direction
Three options, in increasing aggressiveness:

1. **Optional field, auto default.** Make `n_cent` optional; when omitted, use the row-count-derived value the engine already computes (`n_cent_row_count_cap`). Keep it as an optional power-user override. Additive, non-breaking.
2. **Drop it from the identity hash.** Remove `n_cent` from `compute_options_hash` so it is no longer part of table identity (independent of whether it stays in the API). Hash-format change → needs a version bump / migration for existing tables.
3. **Remove it from the API entirely.** Drop the `n_cent` field from `VectorConfig` (and the public index spec) and let the engine derive it end-to-end — no user-facing knob, and nothing to hash. The API is still pre-stable (no backward-compatibility guarantees yet), so removing it outright is cleaner than carrying an optional override, and it subsumes option 2.

Preference: **option 3** while the API is still pre-stable; fall back to 1 + 2 if the field must be retained.

muralikpbhat:
## What
Removes `VectorConfig::n_cent` — a required user field that was build-time-only and folded into the table identity hash — from the public API and the options hash. The per-superfile IVF cluster count is now derived from that build's own row count.

## Why (#488)
`n_cent` was legacy and near-useless as a user knob:
- Query never reads it — the reader takes the actual cluster count from each superfile's on-disk cell directory.
- Merge/drain already re-derive it from the data (byte-run target / max-source count).
- Every fresh build already clamped it by the row-count cap.

So it only added a mandatory number users can't reason about, and — worse — welded a build-time value into table identity, forcing every reopen to re-supply the exact original value.

## Changes
- `VectorConfig`: drop the `n_cent` field; `VectorConfig::new` loses the parameter. Fresh-write IVF sizing derives from rows (`n_cent_row_count_cap(rows).min(rows).max(1)`); drain/merge keep their own data-derived counts, now passed explicitly.
- Public `IndexSpec::vector` → `(column, dim, metric)`; same for the Node and Python bindings, their type stubs, tests, and example notebooks.
- Removed `n_cent` from `compute_options_hash` and from the persisted records. The per-superfile on-disk `n_cent` the reader actually uses is kept, and serde stays backward-compatible for reads (old data carrying the field still parses).
- Deleted the dead `n_cent` bindings the removal left behind (vector round-trip tests and the benchmark `VectorEngine` harness), rather than silencing them.

### Also in this PR
- **Remote transport:** dropped the `stats` field from the bm25 search request. The hosted endpoint does not read it, so it was a no-op that had drifted from the vendored API spec (`Bm25SearchRequest` has no such property); this only surfaced here because the change touches `src/catalog/remote/`, which re-triggers the (path-filtered) hosted-API drift check. The BM25 statistics mode is unaffected on the embedded path.

## Breaking change
- **Public API** signatures change → crate bumped **0.2.1 → 0.3.0** (`infino` + Node + Python bindings, version-locked).
- **Table identity**: the options hash no longer includes `n_cent`, so **existing vector-column tables written by 0.2.x fail to reopen (`OptionsHashMismatch`) and must be recreated.** `FORMAT_VERSION` is deliberately left at `"1.0"` so non-vector tables are unaffected; the 0.3.0 crate bump is the compatibility signal.

## Validation
- `make ci` green: fmt, clippy, version-sync, all test targets, and doctests.
- Reviewed: drain/merge threading preserves the pre-removal data-derived counts; serde read back-compat confirmed; `n_cent`-hardcoded test fixtures were re-baselined to the derived value, not loosened.

## Follow-up after merge (TODO)
Because this is a breaking `.vector()` signature change, the **"Python examples (end to end)"** job (not a required check) is expected to stay red, and will turn red on `main` once this merges. That job installs the published LangChain and CrewAI integration packages `--no-deps` and links them against this from-source build; those published packages still call the pre-0.3.0 four-argument `.vector(column, dim, n_cent, metric)`, so they raise `TypeError` on the new three-argument signature. This can only be resolved downstream, in order:

1. Publish `infino` 0.3.0.
2. Update the LangChain and CrewAI integration packages to the three-argument `.vector()` and bump their `infino` dependency to `>=0.3.0`; republish to PyPI.
3. The example-notebook job returns to green.

Follow-up branches for both integration packages are being prepared so they can be released as soon as 0.3.0 is on PyPI. This red window is inherent to a breaking change with published downstream consumers.

Closes #488.


muralikpbhat:
**Heads-up on the red `Python examples (end to end)` check** — it is *not* a required merge gate, and its failure is a chicken-and-egg of this breaking `.vector()` change, not a defect in the PR:

- That job installs the published LangChain/CrewAI integration packages `--no-deps` and links them against this from-source `infino` build. The published packages still call the old four-argument `.vector(column, dim, n_cent, metric)`, so they `TypeError` against the new three-argument signature.
- They **can't be fixed before this merges**: to move to the three-argument call they must depend on `infino >=0.3.0`, which won't exist on PyPI until this lands and a release is cut — and they can't be verified green until they're updated *and* republished.
- So the check can only return to green **after** the sequence: merge → publish `infino` 0.3.0 → update + republish the integrations. Merging this PR flips the job red on `main` for that window.

The `rag`/`code_search` notebooks (which use `infino` directly) are already updated and pass; only the notebooks that go through the published integration wrappers are affected. Follow-up branches for the integration packages are being prepared so they can ship as soon as 0.3.0 is on PyPI.

---

**Update — resolved in this PR (`bd10ae6`).** Rather than leave it red, `make python-examples-test` now gates each integration on a compat probe: it exercises the integration's own table-creation path against the built `infino` and, when incompatible, **skips that suite's notebooks with a note** instead of hard-failing —

```
note: skipping langchain examples — langchain-infino not compatible with this infino build (TypeError: IndexSpec.vector() takes 3 positional arguments but 4 were given)
```

The direct-`infino` notebooks (`rag/`, `code_search/`, …) still run. So the check is green now and stays green on `main` after merge, and it **self-heals**: once `langchain-infino`/`crewai-infino` publish a 0.3.0-compatible release the probe passes and their notebooks run again automatically — no manual re-enable.

ashishmishra26 (approved):

vinaykakade:
LGTM. Please change the python examples to have the CI pass, and we'll need to release a new version once this is merged to main.

ashishmishra26:
@muralikpbhat **Looks like this accidentally reverts the bm25 `stats` wire from #465** (merge-conflict collateral, not part of the intended `n_cent` change).

`Bm25SearchOptions.stats`, the public re-export, and the Node/Python `stats` params are all kept — only the remote transport drops it:

```rust
// src/catalog/remote/table.rs — bm25_search request body, removed on this branch:
    "stats": stats_str(opts.stats),
```

`bm25_search` still takes `opts: Bm25SearchOptions`, so it's silent: a remote call with `.with_stats(Bm25Stats::Global)` now serializes without `stats` and the server falls back to per-superfile. `fn stats_str`, the `Bm25Stats` import, and the `tests/remote.rs` `"stats"` assertion were dropped in the same hunk — which is why CI stays green.

cc: @vinaykakade Is my understanding correct here?

muralikpbhat:
**Context — the red `Remote transport ⇄ hosted API spec` check on this PR:** that's the hosted-API *drift guard*, and it's flagging the remote bm25 client for sending a `stats` field the vendored spec doesn't define. It was deliberate on my part (not merge-conflict collateral), but you've prompted the right rethink — here's why it's red and the fix.

**Why the check is red now, and never before.** The drift is genuinely pre-existing; it rode in on three separately-green PRs:
- **#449** added the remote transport.
- **#465** added the client-side `stats` wire.
- **#483** added the drift check itself **and** vendored the fixture — whose `Bm25SearchRequest` has no `stats`.

Each was green on its own: #465 predated the check existing, #483's fixture matched the client at *its* CI time, and the check only runs on `pull_request`s touching `src/catalog/remote/**`, the fixture, or `tests/remote.rs` — there's no `push`/main trigger. No merged PR touched a trigger path between #483 and #495, so the latent mismatch was never re-evaluated. #495 edits `remote/wire.rs` (dropping `n_centroids`, required by the n_cent change), so it's the first PR to poke `remote/` since — which is why `main` is green and this only surfaces here.

**The fix — correct the contract, not the client.** `stats` was added to the client wire in **#465** (one of those same three PRs), deliberately, as the opt-in global-BM25-stats feature. A field the client was purpose-built to send to the hosted endpoint is by definition part of the contract — the vendored spec (added later, in #483) just never captured it. So stripping the client to match the spec was fixing the wrong side. I've kept the client wire intact and added `stats` (a `Bm25Stats` enum: `per_superfile` / `global`) to the vendored `Bm25SearchRequest`; the drift check now passes with client and spec in agreement, and #495 stays a pure n_cent change on the engine side.

PR #498, opened 2026-08-01
supertable: key hidden slow-state restamp on the manifest strategy

### LINKED ISSUE #499: # optimize: hidden cell-split phase never runs in the process that created the table (options-keyed gate)
## Symptom

A process that creates a vector table and keeps using the same handle —
`create_table` → append → drain → `optimize()`, the normal single-process
ingest lifecycle — never runs the hidden index's cell-split phase. Over-cap
(or multimodal) cells keep growing unboundedly until some other process
reopens the table and runs `optimize` there. Same table, same data: `open`-era
handles split, `create`-era handles silently don't.

## Root cause

`compact_one_table` gates the split phase on the handle's **options**
(`src/supertable/compaction/mod.rs` ~279):

```rust
if is_hidden_vector_index_table(&inner.options) {
    split_overflow_cells(Arc::clone(inner)).await...?;
}
```

`is_hidden_vector_index_table` requires `opts.partition_strategy ==
Some(VectorCell)`. But `build_vector_index_options`
(`src/supertable/handle.rs`) only stamps VectorCell into the hidden options
when it can bootstrap a grid from an existing user manifest
(`train_global_centroids`) — at table create there is none, so a create-era
hidden handle carries `partition_strategy: None` for the life of the process.
The first drain locks VectorCell into the hidden **manifest**, but the
handle's options struct is immutable and never catches up: manifest says
VectorCell, options say None, and the gate reads the options.

This is the same options-vs-manifest divergence behind the durably-empty
hidden-membership bug (`try_commit_attempt` step 2b restamp skipped for
create-era handles — see `issue-hidden-split-empty-membership.md`), and this
gate was called out in that issue's "adjacent observations".

## Fix

Key the gate on the manifest's locked strategy — the signal
`split_overflow_cells` itself already self-guards on
(`src/supertable/writer.rs` ~5443), and the same form the sibling gate in
`compact_async` uses two lines above:

```rust
if matches!(
    inner.manifest.load().partition_strategy(),
    Some(PartitionStrategy::VectorCell { .. })
) {
```

Safety: `split_overflow_cells` re-checks the manifest strategy and no-ops on
non-VectorCell manifests, so this change can only make the phase RUN where it
was wrongly skipped — it cannot run splits on user tables (their manifests
never lock VectorCell).

## Regression test

`supertable::handle::tests::optimize_runs_split_phase_on_create_era_handle`:
create (never reopen) → append one batch that drains into a single hidden
cell holding >4 well-separated modes and ≥ `MODALITY_MIN_CELL_DOCS` rows →
drain → public `optimize()` → assert the hidden grid grew (`n_cent` 1 → ≥2)
and the split preserved all live docs. Fails on the options-keyed gate (grid
stays at 1 cell), passes with the manifest-keyed gate. Uses the
default-config modality split trigger (`cell_split_modality_d = 8.0`,
`MODALITY_MIN_CELL_DOCS = 128`), since the 500k `cell_split_doc_cap` is not
reachable at unit-test scale and config is process-global.

ekechinwokah:
Partially fixes: #499 

Related PR: #500 

**Problem:** A hidden membership commit restamps the slow-CAS blob ref in `try_commit_attempt`, but the gate read the handle's options (`is_hidden_vector_index_table`). A hidden handle built at table create has no user manifest to bootstrap a grid from, so its options never carry a VectorCell strategy and the restamp was silently skipped for every hidden commit in that process. update() meanwhile keys on the manifest's locked strategy: it writes no parts for VectorCell tables and clears the ref. A cell split commit therefore published a manifest whose hidden membership was durably empty — zero hits from every cell, unrecoverable by reopen or optimize — unless the pass-final in-process refresh_slow_vector_state happened to run afterwards.

**Fix:** Gate the restamp on the successor manifest's locked strategy instead, the same signal update() uses, so the ref-clear and the restamp pair on one condition. The split commit now publishes byte-identical state to what the refresh composes (same content-addressed URIs), closing the crash window between a split commit and the pass-final refresh.

Regression test: split_commit_without_refresh_keeps_docs_retrievable — two populated cells, split the busiest, no refresh; asserts per-direction retrievability in-process and after a reopen from storage. Fails 0/16 before the fix.

### DIFF
--- a/src/supertable/handle.rs
+++ b/src/supertable/handle.rs
@@ -4445,6 +4445,184 @@ mod tests {
         }
     }
 
+    /// A committed cell split ALONE — without the pass-final in-process
+    /// `refresh_slow_vector_state` that production maintenance tacks onto the
+    /// end of the hidden pass — must leave every doc retrievable, both
+    /// in-process and after a reopen from storage. The split's own commit
+    /// (`try_commit_attempt` step 2b) publishes the slow-state blob + centroid
+    /// section for the post-split membership; if that publication disagrees
+    /// with what the refresh composes, a crash between the split commit and
+    /// the refresh leaves the table durably under-serving (0 hits from every
+    /// cell, including cells the split never touched) with no recovery path —
+    /// reopen re-hydrates the broken state and a post-reopen `optimize`
+    /// republishes it unchanged. Two populated cells are required to expose
+    /// the mismatch.
+    #[test]
+    fn split_commit_without_refresh_keeps_docs_retrievable() {
+        use std::sync::Arc;
+
+        use arrow_array::{Array, FixedSizeListArray, Float32Array, LargeStringArray};
+        use arrow_schema::{DataType, Field, Schema};
+
+        use crate::{
+            superfile::{
+                builder::{FtsConfig, VectorConfig},
+                reader::VectorSearchOptions,
+                vector::{distance::Metric, rerank_codec::RerankCodec},
+            },
+            supertable::{manifest::list::PartitionStrategy, writer::split_overflow_cell},
+        };
+
+        /// Docs planted per orthogonal direction (two directions → the two
+        /// populated cells the repro needs).
+        const ROWS_PER_DIRECTION: usize = 8;
+        /// Total planted docs.
+        const N_TOTAL: usize = 2 * ROWS_PER_DIRECTION;
+        /// Probe width covering every cell before and after the split.
+        const SPLIT_NPROBE: usize = 64;
+
+        let dim = 16usize;
+        let item_field = Arc::new(Field::new("item", DataType::Float32, true));
+        let schema = Arc::new(Schema::new(vec![
+            Field::new("title", DataType::LargeUtf8, false),
+            Field::new(
+                "emb",
+                DataType::FixedSizeList(item_field.clone(), dim as i32),
+                false,
+            ),
+        ]));
+        let dir = TempDir::new().expect("tempdir");
+        let make_options = || {
+            let pool = Arc::new(
+                rayon::ThreadPoolBuilder::new()
+                    .num_threads(1)
+                    .build()
+                    .expect("pool"),
+            );
+            let storage: Arc<dyn StorageProvider> =
+                Arc::new(LocalFsStorageProvider::new(dir.path()).expect("provider"));
+            SupertableOptions::new(
+                schema.clone(),
+                vec![FtsConfig {
+                    column: "title".into(),
+                    positions: false,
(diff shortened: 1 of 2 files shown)


muralikpbhat (approved):
The fix here is correct and well-targeted, with strong durability/regression tests. Two cross-cutting notes on this pair (#498 + #500) — about their interaction and one possible sibling site:

**1. The two PRs together orphan `is_hidden_vector_index_table`.** Its only two non-test callers are `writer.rs:6261` (changed away by #498) and `compaction/mod.rs:279` (removed by #500). Neither PR alone makes it dead, but once both land it has zero callers. Whichever merges second should delete it — or fold its docstring insight into a comment on `ManifestSnapshot::partition_strategy` — otherwise it lingers as a trap inviting the exact options-keyed staleness to be reintroduced, and dead-code lint may flag it.

**2. Possible third site with the same stale-options read.** `writer.rs:1673` gates cell-partition-vs-round-robin sharding on `self.inner.options.partition_strategy` — the same create-era-stale field this pair fixes elsewhere. It's probably safe (the drain constructs its writer with fresh options that carry the strategy), but it's the identical failure class: a create-era hidden handle whose options lack `VectorCell` would silently round-robin instead of cell-partition. Worth confirming this path is only ever reached with drain-constructed options — a short assertion or comment would close it.

Approving.

ekechinwokah:
> The fix here is correct and well-targeted, with strong durability/regression tests. Two cross-cutting notes on this pair (#498 + #500) — about their interaction and one possible sibling site:
> 
> **1. The two PRs together orphan `is_hidden_vector_index_table`.** Its only two non-test callers are `writer.rs:6261` (changed away by #498) and `compaction/mod.rs:279` (removed by #500). Neither PR alone makes it dead, but once both land it has zero callers. Whichever merges second should delete it — or fold its docstring insight into a comment on `ManifestSnapshot::partition_strategy` — otherwise it lingers as a trap inviting the exact options-keyed staleness to be reintroduced, and dead-code lint may flag it.
> 
> **2. Possible third site with the same stale-options read.** `writer.rs:1673` gates cell-partition-vs-round-robin sharding on `self.inner.options.partition_strategy` — the same create-era-stale field this pair fixes elsewhere. It's probably safe (the drain constructs its writer with fresh options that carry the strategy), but it's the identical failure class: a create-era hidden handle whose options lack `VectorCell` would silently round-robin instead of cell-partition. Worth confirming this path is only ever reached with drain-constructed options — a short assertion or comment would close it.
> 
> Approving.

noted. handling in follow-on PR.

PR #500, opened 2026-08-01
compaction: key the hidden split phase on the manifest strategy

### LINKED ISSUE #499: # optimize: hidden cell-split phase never runs in the process that created the table (options-keyed gate)

ekechinwokah:
Fixes #499 

Related PR: #498 

## Problem

A process that creates a vector table and keeps using the same handle —
`create_table` → append → drain → `optimize()`, the normal single-process
ingest lifecycle — never runs the hidden index's cell-split phase. Over-cap
and multimodal cells grow without bound until some other process reopens the
table and optimizes there. Same table, same data: open-era handles split,
create-era handles silently don't.

## Root cause

`compact_one_table` gated `split_overflow_cells` on the handle's options
(`is_hidden_vector_index_table`), which requires `opts.partition_strategy ==
Some(VectorCell)`. A hidden handle built at table create has no user manifest
to train a grid from, so `build_vector_index_options` leaves its options with
no VectorCell strategy — and options are immutable, so they never catch up
after the first drain locks the strategy into the manifest. The gate read the
options; the truth lives in the manifest.

## Fix

Key the gate on the manifest's locked partition strategy — the same signal
`split_overflow_cells` already self-guards on, and the same form the sibling
gate in `compact_async` uses. User tables never lock VectorCell into their
manifests, so the phase still cannot run on them; the change only makes the
split run where it was wrongly skipped.

The one remaining `is_hidden_vector_index_table` call site (the
`try_commit_attempt` step-2b slow-state restamp) has the same staleness
hazard; that is the separately tracked durably-empty-membership bug with its
own fix and regression test in flight. This PR documents the hazard on the
helper's doc-comment; once both fixes land the helper has no callers and
should be deleted.

## Regression test

`supertable::handle::tests::optimize_runs_split_phase_on_create_era_handle`:
create (never reopened) → append one batch that drains into a single hidden
cell holding 8 well-separated modes → drain → public `optimize()` → assert
the grid grew past one cell, all in one process. Rides the default modality
split trigger (>4 modes, ≥ `MODALITY_MIN_CELL_DOCS` rows) because the 500k
`cell_split_doc_cap` is out of unit-test reach and config is process-global.
Fails on the options-keyed gate (grid stays at 1 cell), passes with the
manifest-keyed gate.

## Testing

- New regression test verified fail-before / pass-after the gate change
- `cargo test --lib`: 2085 passed, 0 failed
- Integration subsets: `compact` (6), `commit::` (56), `query::` incl. the
  brute-force oracles (92), `supertable_commit_crash_localfs` crash-safety
  contract (5) — all green
- `cargo fmt --check` and `cargo clippy --all-targets --features
  test-helpers,remote -- -D warnings` clean
- No public-API change (`public-api.txt` untouched)

## Benchmarks

The gate is evaluated once per `optimize()` call and is off every query and
ingest hot path, so steady-state latency, throughput, and fetch/request
counts are unaffected by construction. `optimize` in create-era processes now
performs split work it previously (wrongly) skipped — that is the intended
behavioral fix, not an overhead.

### DIFF
--- a/benches/utils/cost.rs
+++ b/benches/utils/cost.rs
@@ -67,9 +67,13 @@ const PER_MILLION: f64 = 1.0e6;
 /// the cell's own corpus size (`n_docs`/month) so the summary prices writing
 /// THIS table, not a synthetic volume.
 const SUMMARY_QUERIES_PER_MONTH: f64 = 1.0e6;
-/// Warm fraction of the blended monthly read line (the rest pay the cold
-/// per-query cost).
-const SUMMARY_READ_WARM_FRACTION: f64 = 0.95;
+/// Default warm fraction of the blended monthly read line (the rest pay
+/// the cold per-query cost). Env-overridable via
+/// `INFINO_BENCH_COST_WARM_FRACTION` (0 < f ≤ 1) so pricing scenarios can
+/// flex the assumed hit rate without recompiling — the frequency knob;
+/// the warm/cold rate gap the summary prints is the magnitude it
+/// multiplies against.
+const DEFAULT_SUMMARY_READ_WARM_FRACTION: f64 = 0.95;
 /// Padding on the per-query RAM-hold window: a query holds the resident set
 /// a little longer than its own p50 (dispatch, response write, scheduler
 /// slack between overlapped queries), so the hold is billed at fudge × p50.
@@ -599,6 +603,30 @@ fn occupancy_replicas() -> u32 {
     static REPLICAS: OnceLock<u32> = OnceLock::new();
     *REPLICAS
         .get_or_init(|| parse_replicas(std::env::var("INFINO_BENCH_COST_REPLICAS").ok().as_deref()))
+}
+
+/// Parse an `INFINO_BENCH_COST_WARM_FRACTION` override. Accepts only
+/// 0 < f ≤ 1; anything else (unset, garbage, 0, negative, >1, NaN) falls
+/// back to the default — a nonsense hit rate must never silently shape
+/// the blend.
+fn parse_warm_fraction(raw: Option<&str>) -> f64 {
+    raw.and_then(|s| s.parse::<f64>().ok())
+        .filter(|f| f.is_finite() && *f > 0.0 && *f <= 1.0)
+        .unwrap_or(DEFAULT_SUMMARY_READ_WARM_FRACTION)
+}
+
+/// Warm fraction for the blended read line, env-overridable like
+/// [`occupancy_replicas`] (a workload assumption, not an instance
+/// attribute).
+fn summary_warm_fraction() -> f64 {
+    static FRACTION: OnceLock<f64> = OnceLock::new();
+    *FRACTION.get_or_init(|| {
+        parse_warm_fraction(
+            std::env::var("INFINO_BENCH_COST_WARM_FRACTION")
+                .ok()
+                .as_deref(),
+        )
+    })
 }
 
 /// The tenant's composed serving cost per month: marginal work (storage +
@@ -2138,7 +2166,7 @@ pub fn emit(report: &mut Report, anchor: &str, title: String, c: &CellCost) {
     // set.
     let blended_read_q = match (&steady_warm, &steady_cold) {
         (Some((_, warm_q)), Some((_, cold_q))) => {
-            Some(warm_q * SUMMARY_READ_WARM_FRACTION + cold_q * (1.0 - SUMMARY_READ_WARM_FRACTION))
+            Some(warm_q * summary_warm_fraction() + cold_q * (1.0 - summary_warm_fraction()))
         }
         (Some((_, warm_q)), None) => Some(*warm_q),
         (None, Some((_, cold_q))) => Some(*cold_q),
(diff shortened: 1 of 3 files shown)


muralikpbhat (approved):
The fix here is correct and well-targeted, with strong durability/regression tests. Two cross-cutting notes on this pair (#498 + #500) — about their interaction and one possible sibling site:

**1. The two PRs together orphan `is_hidden_vector_index_table`.** Its only two non-test callers are `writer.rs:6261` (changed away by #498) and `compaction/mod.rs:279` (removed by #500). Neither PR alone makes it dead, but once both land it has zero callers. Whichever merges second should delete it — or fold its docstring insight into a comment on `ManifestSnapshot::partition_strategy` — otherwise it lingers as a trap inviting the exact options-keyed staleness to be reintroduced, and dead-code lint may flag it.

**2. Possible third site with the same stale-options read.** `writer.rs:1673` gates cell-partition-vs-round-robin sharding on `self.inner.options.partition_strategy` — the same create-era-stale field this pair fixes elsewhere. It's probably safe (the drain constructs its writer with fresh options that carry the strategy), but it's the identical failure class: a create-era hidden handle whose options lack `VectorCell` would silently round-robin instead of cell-partition. Worth confirming this path is only ever reached with drain-constructed options — a short assertion or comment would close it.

Approving.

ekechinwokah:
> The fix here is correct and well-targeted, with strong durability/regression tests. Two cross-cutting notes on this pair (#498 + #500) — about their interaction and one possible sibling site:
> 
> **1. The two PRs together orphan `is_hidden_vector_index_table`.** Its only two non-test callers are `writer.rs:6261` (changed away by #498) and `compaction/mod.rs:279` (removed by #500). Neither PR alone makes it dead, but once both land it has zero callers. Whichever merges second should delete it — or fold its docstring insight into a comment on `ManifestSnapshot::partition_strategy` — otherwise it lingers as a trap inviting the exact options-keyed staleness to be reintroduced, and dead-code lint may flag it.
> 
> **2. Possible third site with the same stale-options read.** `writer.rs:1673` gates cell-partition-vs-round-robin sharding on `self.inner.options.partition_strategy` — the same create-era-stale field this pair fixes elsewhere. It's probably safe (the drain constructs its writer with fresh options that carry the strategy), but it's the identical failure class: a create-era hidden handle whose options lack `VectorCell` would silently round-robin instead of cell-partition. Worth confirming this path is only ever reached with drain-constructed options — a short assertion or comment would close it.
> 
> Approving.

noted. handling in follow-on PR.

PR #503, opened 2026-08-01
supertable: cleanup nits

ekechinwokah:
Follow-up to #498/#500, closing the two cross-cutting review notes from that
pair, plus a build fix for a merge skew the pair left on main.

**Commit 1 — `supertable: drop n_cent from two test literals missed by #495`.**
#495 removed `VectorConfig::n_cent`, but the two fixtures that landed
concurrently via #498 (`split_commit_without_refresh_keeps_docs_retrievable`)
and #500 (`optimize_runs_split_phase_on_create_era_handle`) still construct
it — `cargo test --lib` does not compile on current main. Same mechanical
migration #495 applied to every other literal. Standalone commit so it can be
cherry-picked ahead if needed.

**Commit 2 — `supertable: delete options-keyed hidden-table gate`.**

- Review note 1: `is_hidden_vector_index_table` lost its last caller once
  #498 (step-2b restamp) and #500 (optimize split phase) both rekeyed onto
  the manifest's locked strategy. Deleted before it invites the next
  options-keyed gate. The staleness rule — options are a construction-time
  snapshot; a create-era hidden handle carries no VectorCell strategy for
  its whole process lifetime — now lives on the doc of
  `ManifestSnapshot::partition_strategy`, the accessor that IS the correct
  signal.
- Review note 2: audited the remaining options-strategy read in the
  buffered writer's cell-partition pre-shard. It is contained: no
  production path appends to the hidden table through the buffered writer
  (hidden membership is written by drain / split / merge as prepared
  superfiles), and a create-era hidden append would fail loudly in
  `assign_partition` (VectorCell requires hinted superfiles) rather than
  round-robin silently. Closed with a comment stating the invariant and a
  `debug_assert!` that fires on the dangerous configuration
  (VectorCell-locked manifest, no grid in options). A crate-wide sweep
  found no other stale options-strategy reads.
- Refreshed the `recall_while_ingest` comment that cited the deleted helper
  and the pre-#500 reopen workaround. Comment only — the bench's reopen and
  measured shape are unchanged, so recorded baselines stay comparable.

**Validation.** 1114 supertable lib tests green (including both #498/#500
regression tests, with the new assert active on every buffered commit);
`make check` green (fmt + clippy `--all-targets` under both feature sets,
which compiles benches and integration tests).

**Performance / cost.** No release-mode code change: commit 1 touches
`#[cfg(test)]` literals, commit 2 deletes dead code, moves comments, and
adds a `debug_assert!` (compiled out in release). Query, ingest, and
object-store request paths are untouched, so there is no bench surface to
compare; recall@10 is unaffected.

### DIFF
--- a/src/supertable/handle.rs
+++ b/src/supertable/handle.rs
@@ -1219,30 +1219,9 @@ pub(crate) const GLOBAL_VECTOR_KMEANS_SEED: u64 = 0x51ED_2A11;
 /// the full working set stays resident.
 const CACHE_BUDGET_HEADROOM_DIVISOR: u64 = 10;
 
-/// Train global VectorCell centroids from the user manifest and queue them
-/// on the hidden index table for its next commit.
 /// Aggressive compaction profile for the hidden vector-index table: keep
 /// ~one compact packed shard object per partition key instead of many
 /// small delta files.
-/// True for the derived hidden vector-index sibling (VectorCell routing, no FTS).
-///
-/// Options-keyed, so it is STALE for a hidden handle built at table create
-/// time: with no user manifest to train a grid from,
-/// [`build_vector_index_options`] leaves `partition_strategy: None`, and the
-/// options never catch up after the first drain locks VectorCell into the
-/// manifest. Gates deciding whether hidden maintenance must run should key on
-/// the manifest's locked strategy (`ManifestSnapshot::partition_strategy`)
-/// instead; this predicate answers "was this handle BUILT as the hidden
-/// sibling", which is only true for open-era handles.
-pub(crate) fn is_hidden_vector_index_table(opts: &SupertableOptions) -> bool {
-    !opts.vector_columns.is_empty()
-        && opts.fts_columns.is_empty()
-        && matches!(
-            opts.partition_strategy,
-            Some(crate::supertable::manifest::list::PartitionStrategy::VectorCell { .. })
-        )
-}
-
 pub(crate) fn hidden_vector_index_compaction_settings() -> crate::config::CompactionSettings {
     let vector = &crate::config::global().vector;
     crate::config::CompactionSettings {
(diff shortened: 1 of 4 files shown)


src/supertable/handle.rs
```diff
@@ -1219,30 +1219,9 @@ pub(crate) const GLOBAL_VECTOR_KMEANS_SEED: u64 = 0x51ED_2A11;
...
-///
-/// Options-keyed, so it is STALE for a hidden handle built at table create
-/// time: with no user manifest to train a grid from,
-/// [`build_vector_index_options`] leaves `partition_strategy: None`, and the
-/// options never catch up after the first drain locks VectorCell into the
-/// manifest. Gates deciding whether hidden maintenance must run should key on
-/// the manifest's locked strategy (`ManifestSnapshot::partition_strategy`)
-/// instead; this predicate answers "was this handle BUILT as the hidden
-/// sibling", which is only true for open-era handles.
-pub(crate) fn is_hidden_vector_index_table(opts: &SupertableOptions) -> bool {
```
vinaykakade, line 1237: This also removed in #504, as couldn't get CI to pass without this change. If case #504 merges first to main, please merge to resolve conflict. (Or will do the same if #503 merges first).

vinaykakade (approved):

src/supertable/handle.rs
vinaykakade: Merged this as all checks had already passed.

muralikpbhat (commented):
LGTM.

muralikpbhat (approved):
LGTM — clean cleanup; closes both cross-cutting review notes from #498/#500 (deletes the orphaned options-keyed gate, re-keys the sibling read onto the manifest strategy) and fixes the n_cent build skew.

PR #504, opened 2026-08-01
supertable: advance the pointer etag only after a successful load

vinaykakade:
## Problem

`refresh()` recorded the newly-probed pointer's etag **before** loading its manifest:

```rust
*last_pointer_etag = meta.etag.clone();   // advanced first
let manifest = match load_with_pointer(Some(current), storage, None, pointer) {
    Ok(m) => m,
    AlreadyLoaded => return Ok(false),
    Err(err) => return Err(err),          // load failed — etag already advanced
};
manifest.store(manifest);                 // snapshot only advances on success
```

If the load then **fails** — e.g. the pointer has advanced but its manifest isn't yet readable by this process — `refresh` returns `Err`, but `last_pointer_etag` is now the new pointer's etag while the in-memory snapshot is still the old one. Every subsequent conditional probe then sees `get_if_none_match` → **NotModified** and returns without retrying the load. The handle is **permanently pinned** to the pre-commit manifest and serves its rows as an empty result, even after the manifest becomes loadable.

## Fix

Record the etag only **after** a successful load, or on the `AlreadyLoaded` branch (where the in-memory state already covers the pointer) — never on the load-error path. A transient load failure now leaves the previous etag in place, so the next probe reads the pointer again and retries until the load succeeds (converging, or surfacing a retryable error in the meantime instead of a silent empty result).

## Test

`strong_read_recovers_after_a_manifest_load_failure` + a `FailManifestGet` fault wrapper (faults the manifest-list read while letting the pointer probe through, so the failure lands inside `load_with_pointer` after the re-check succeeded). Under `Strong`: a commit the consumer hasn't observed, a failing manifest load, then recovery — the consumer must refresh to the new commit once the load heals. **Fails before this change, passes after.** Full `handle` module (54 tests) + `fmt` + `clippy` green.

Also drops the removed `VectorConfig::n_cent` field from two test literals left behind by its removal, so the module compiles.

## Performance / cost

Read-path only, and only changes *when* an in-memory etag is written; no extra I/O on the success path. While a load is transiently failing the probe reads the small pointer body each retry (bounded, self-healing). No bench-relevant path touched.

### DIFF
--- a/src/supertable/handle.rs
+++ b/src/supertable/handle.rs
@@ -504,12 +504,14 @@ impl Supertable {
             PointerProbe::NotModified => return Ok(false),
             PointerProbe::Read(pointer, meta) => (pointer, meta),
         };
-        *self
-            .inner
-            .last_pointer_etag
-            .lock()
-            .expect("last_pointer_etag mutex poisoned") = meta.etag.clone();
 
+        // Record the new pointer etag only once we've actually accounted for
+        // this pointer version — after a successful load, or when our in-memory
+        // state already covers it. Recording it *before* the load would, on a
+        // load failure (e.g. a manifest not yet visible to this process),
+        // advance the etag while the snapshot stays behind: the next conditional
+        // probe would then see `NotModified` and never retry the load, pinning
+        // the handle to the pre-commit manifest and serving its rows as empty.
         let current = self.inner.manifest.load_full();
         let manifest = match ManifestSnapshot::load_with_pointer(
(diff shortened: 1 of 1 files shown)


vinaykakade:
The vector related changes (mostly removed a now-removed n_cent param and a function that is no longer used) are needed to have CI pass. cc @muralikpbhat

Pranav2612000 (approved):

Issue #494, opened 2026-07-31
Recall@10 inverts with nprobe on latest main (exhaustive scan ~0.40) — likely a probe-width/shortlist regression


muralikpbhat:
## Summary
On latest `main`, a freshly-built vector index shows recall@10 **decreasing as `nprobe` increases** — probing *more* cells returns *worse* results — and an **exhaustive scan (all cells) recovers only ~0.40**, when it should be ≈ the brute-force ground truth. A build from before the probe-width change is flat at ~0.99 across every `nprobe` on the identical benchmark. This looks like a correctness regression in the `nprobe` → shortlist/rerank path.

## Repro
- `recall_while_ingest` (#466): 10M-doc checkpoints, synchronous `optimize()` per batch, 1000 held-out queries, `INFINO_BENCH_CELLS=256,256`, `cell_split_doc_cap=500000`, synthetic 1024-dim planted-cluster corpus (cosine).
- Fresh build 0→N on latest `main`; sweep recall@10 over `nprobe` at each checkpoint.

## Observed (latest main) — reproduced at both 10M and 20M
| nprobe | 10M | 20M |
|---|---|---|
| 2 | 0.994 | 0.972 |
| 4 | 0.994 | 0.968 |
| 8 | 0.881 | 0.829 |
| 16 | 0.732 | 0.632 |
| 32 | 0.610 | 0.467 |
| **all cells** | **0.402** | *(hung before logging)* |

Two problems:
1. **Recall inverts with `nprobe`** — monotonic decrease; a higher `nprobe` should never lower recall@10.
2. **Exhaustive (all-cells) recall ≈ 0.40** — a full scan misses ~60% of the true top-10, which points at a **scoring/shortlist** bug, not just routing. Most likely the 1-bit RaBitQ prefilter shortlist isn't scaled with `nprobe`, so wide probing floods it with false positives that evict true neighbors before rerank.

## Contrast — isolates it to the engine, not the harness
Same bench, **identical settings**, on a build from *before* the probe-width change:

| nprobe | 2 | 4 | 8 | 16 | 32 | all |
|---|---|---|---|---|---|---|
| recall@10 | 0.993 | 0.993 | 0.993 | 0.993 | 0.993 | 0.993 |

Flat and correct across every `nprobe` including exhaustive → the harness/settings are not the cause.

## Secondary symptom
[...]

PR #505, opened 2026-08-01
Fix vector nprobe recall inversion

ekechinwokah:
This is a partial fix for #494

## Summary

A widening probe knob must never lower recall. #479's global-shortlist
selection caps total survivors at `k × rerank_mult` regardless of probe
width — width-blind — so on real-embedding corpora, raising `nprobe`
*lowers* recall: measured externally at 10M (Cohere-class, az2),
0.994 at nprobe=2 falling monotonically to **0.388 at all cells**, while
the pre-#479 per-cell budgets held flat at 0.976 (each added cell brought
its own budget — monotone, but at whole-cell-fetch cost).

Two commits:

1. **bench: unfiltered nprobe width-sweep diagnostic** — recall at
   nprobe {2, 8, 32, all-cells} at post-drain and post-compact, so the
   knob's monotonicity contract is measured on every run (only the
   filtered sweep existed before).
2. **engine: per-cell floor in `select_global_shortlist`** — selection
   keeps the pooled top-`limit` PLUS each scanned (unit, cell)'s own `k`
   best, counting its globally-kept candidates against the floor.
   Floored survivors are immune to other cells' 1-bit noise, so widening
   only ever ADDS candidates — **monotone by construction** — while the
   pooled cap keeps #479's narrow-width win (pooling beat per-cell
   budgets 0.994 vs 0.976 at nprobe=2 by spending slots on estimate rank
   instead of spreading them evenly). Worst-case cost: `cells × k` extra
   exact-rerank rows — a fraction of the parent's `cells × k × rerank_mult`.

## Validation

Three evidence tiers, honestly separated:

1. **Unit-level**: monotonicity is proven by construction (the
   widening-never-evicts property test).
2. **Real embeddings (Cohere-1M, fresh table, identical A/B, k=10)**:
   the floor is recall-superior-or-equal at EVERY (rm, nprobe) point.
   At the constrained-pool regime (rm=8) it recovers what the fixed
   pool evicts: 0.917→0.935 at nprobe=32, 0.919→0.938 at 64,
   0.923→**0.942** at all-590-cells (+1.9pt), with zero recall cost
   anywhere and the priced latency profile (no delta at law-class
   widths; +5.5ms at the all-cells extreme = the w × k floored
   survivors reranked). At rm=64 both builds are monotone at 1M —
   confirming the reporter's "1M is blind to the inversion" from both
   sides. The 0.976→0.388 collapse itself is 10M-scale: a full
   Cohere-10M A/B (same protocol, both builds, fresh table) is running
   now and its table will be posted on this PR; the external az2 re-run
   remains the independent confirmation.
3. 10m cohere a/b (fresh table, identical build, only the wheel differs; k=10, 100 queries, same protocol as the 1m run):

| rm | nprobe | base | floor | Δ |
|---|---|---|---|---|
| 64 | 1 | 0.726 | 0.726 | — |
| 64 | 2 | 0.788 | 0.788 | — |
| 64 | 8 | 0.939 | **0.940** | +0.1pt |
| 64 | 32 | 0.968 | **0.969** | +0.1pt |
| 64 | 64 | 0.974 | **0.975** | +0.1pt |
| 64 | all | 0.974 | **0.975** | +0.1pt |
| 8 | 1 | 0.683 | 0.683 | — |
| 8 | 2 | 0.718 | 0.718 | — |
| 8 | 8 | 0.850 | **0.859** | +0.9pt |
| 8 | 32 | 0.873 | **0.889** | +1.6pt |
| 8 | 64 | 0.877 | **0.894** | +1.7pt |
| 8 | all | 0.877 | **0.894** | +1.7pt |

same shape as 1m, bigger margins in the constrained-pool regime: the floor is superior-or-equal at every (rm, nprobe) point, zero recall cost anywhere, and both arms are monotone in nprobe. latency parity at law-class widths (≤8 cells); the wide/all-cells extreme pays the floored survivors' rerank (48→56ms p50 at rm=64 all-cells, ~36k extra exact-rerank rows — the priced worst case).

honest caveat: this local build doesn't reproduce the 0.994→0.388 collapse from the report — both arms plateau rather than invert, so what the a/b proves is monotone-by-construction plus recall-superior-or-equal on real embeddings at 1m and 10m. the az2 re-run on the original table remains the independent confirmation of the collapse itself.

Unit tests: floor rescues a swamped cell's best from global eviction;
**widening never evicts floored survivors** (the property whose absence
is the inversion); floor=0 preserves exact legacy semantics; existing
determinism/order-independence tests updated and passing. Full suites:
lib 2086 / superfile 135 / supertable 227; clippy clean; fmt; public-api
no drift.

## Notes

- Default-path behavior is untouched (law width 1 ⇒ the pooled/floored
  arm never runs); this PR changes explicit-nprobe sweeps and any
  law-stamped width > 1.
- The divided-budget arms (cold, filtered, user path) keep their
  documented behavior; the inversion issue write-up covers their status
  and the all-cells ceiling (a pre-#479, scale-gated fixed-shortlist
  effect, not addressed here).


src/supertable/query/vector.rs
```diff
@@ -1942,10 +1942,15 @@ impl SupertableReader {
...
+                // Floor = k: even if the entire true top-k concentrates in
+                // one probed cell, that cell's floor carries it into the
+                // exact rerank (replicas never collide inside one cell, so
+                // the floor needs no replica overhead).
                 flat = select_global_shortlist(
                     flat,
                     k.saturating_add(replica_overhead)
                         .saturating_mul(rerank_mult),
+                    k,
                 );
```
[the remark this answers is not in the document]
ekechinwokah: declined, deliberately. the floor's guarantee is scoped to the floored core — each cell's top-k estimates can never be evicted by widening — and the residual mode you describe (a true neighbor below its own cell's top-k estimates, admitted only via the global pool) stays under global competition exactly as on main; the floor strictly enlarges the keep set, it never worsens that mass. why k and not k*rerank_mult: within a cell every candidate is quantized in the same frame (same centroid, same residual scale), so a cell's top-k by estimate is the trustworthy part of the 1-bit order — the floor exists to guard the CROSS-cell incomparability (differing error bars are what let far cells swamp a near cell's genuine candidates). k*rerank_mult would make every scanned cell's floor as large as the entire global pool: worst case cells*k*rm extra exact-rerank rows, ~36k -> ~2.3m rows/query at the 10m all-cells point at rm=64 — that IS the pre-#479 per-cell-budget cost regime this selection was built to avoid. empirically the posited miss doesn't materialize: the a/b tables on this PR measure the k-floor superior-or-equal and monotone at EVERY (rm, nprobe) point at both 1m and 10m, both arms. and the width sweep is now a permanent battery — if residual non-monotonicity ever shows up, the floor is a one-parameter dial with linear cost (2k, 4k), not an rm-scaled one.

benches/utils/supertable.rs
```diff
@@ -3105,6 +3116,54 @@ pub mod vector {
...
+                |hidden| match hidden.pinned_reader().manifest().get_partition_strategy() {
+                    PartitionStrategy::VectorCell { clusters, .. } => clusters.n_cent as usize,
+                    _ => 0,
+                },
+            )
+            .unwrap_or(0);
+        let mut widths: Vec<usize> = UNFILTERED_SWEEP_WIDTHS.to_vec();
+        if all_cells > widths.last().copied().unwrap_or(0) {
+            widths.push(all_cells);
+        }
```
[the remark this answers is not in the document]
ekechinwokah: taken in 6350e8c — unresolvable grid (no hidden index / non-vector-cell strategy) now skips loudly with a message instead of silently sweeping without the exhaustive endpoint. also fixed the adjacent labeling gap your reviewer-guide flagged: on a grid smaller than the largest fixed width the engine clamps the sweep to the grid, so that row IS exhaustive and is now tagged (all cells) via >=.

ekechinwokah:
on the two reviewer-guide focus areas:

- **cost regression (segments x cells x k)**: correct that the floor keys per scanned (unit, cell) — that's the key the estimates pool under, so it's also the key eviction happens under. three bounds keep it priced: (1) this arm only runs on the warm hidden path at law-stamped or explicit widths — post-delta states query at law widths (small), not all-cells; (2) hidden maintenance merges each cell's deltas toward one packed base, so steady-state segments-per-cell is ~1 — the measured ~36k extra rerank rows at the 10m all-cells point is cells x k with negligible fragmentation, and the +8ms it costs is in the a/b table; (3) under ANY fragmentation the floor stays 1/rerank_mult of the pre-#479 per-cell-budget regime in the same state, since fragmentation multiplies both bounds equally.
- **missing all-cells (grids <= 32 cells)**: fixed in 6350e8c — the engine clamps the sweep to the grid, so the clamped row is now labeled (all cells), and a sweep that cannot resolve the grid skips loudly instead of silently dropping the endpoint.

src/supertable/query/vector.rs
```diff
@@ -2974,17 +2979,78 @@ fn apply_width_pin(
...
+    // prefix (unordered — phase C regroups by unit and the rerank is
+    // exact, so the winners need no internal order).
+    pooled.select_nth_unstable_by(limit, cmp);
+    if cell_floor == 0 {
         pooled.truncate(limit);
+        return pooled;
+    }
+    // Per-cell floor: every scanned (unit, cell) keeps its `cell_floor`
+    // best candidates REGARDLESS of the global competition. This is what
+    // makes probe width monotone in recall by construction — a candidate
```
muralikpbhat, line 3002: **Over-claim — this is a per-cell FLOOR, not end-to-end monotonicity (verify-confirmed).** Kept-count for a cell is `max(already, cell_floor=k)`, but `limit = k*rerank_mult` with default `rerank_mult=256`. At narrow nprobe a cell can hold far more than `k` prefix slots (pool slack, up to ~256k); widening only ever *worsens* a fixed candidate's global rank, so the cell's prefix occupancy is monotone non-increasing and the rescue restores it only to `k` — never to its former higher count. A true top-k neighbor whose 1-bit estimate ranks it in the band `[k+1, ~256k]` (the false-negative case `rerank_mult` exists to catch) is kept at narrow nprobe, evicted at wide -> recall still inverts there. So `widening only ever ADDS survivors` is true for the floored `k`, but widening can *evict* above-floor survivors. Suggest softening this comment (and the PR body) to 'floors the worst case', or making the floor occupancy-aware if strict monotonicity is the goal. (The band is empty only at the degenerate `rerank_mult=1`.)

src/supertable/query/vector.rs
```diff
@@ -4644,6 +4710,98 @@ mod tests {
...
+            "the global prefix is untouched by the rescue"
+        );
+    }
+
+    /// Monotonicity by construction: widening the sweep (adding a new
+    /// cell's candidates to the pool) never evicts another cell's floored
+    /// survivors — the exact property whose absence inverts recall as
+    /// nprobe grows.
+    #[test]
+    fn select_global_shortlist_widening_never_evicts_floored() {
```
muralikpbhat, line 4763: **The contract isn't actually guarded end-to-end.** This is the sole automated guard and it asserts the *weaker* property — that a cell's `<=cell_floor` floored survivors are retained on widening — not that end-to-end recall is monotone. The residual-inversion path (a cell whose prefix occupancy exceeds `cell_floor` at narrow width dropping to exactly the floor at wide width, evicting an above-floor true neighbor) is never exercised: the fixture is built so every cell's narrow occupancy `<= floor`. Combined with the width-sweep bench being **print-only** (`eprintln`, no assertion), a later weakening of the floor would silently reintroduce the 0.994->0.388 inversion with green CI. Suggest a test that fails if measured recall drops as nprobe rises (at least for the floored guarantee), and an assertion on the bench sweep.

src/supertable/query/vector.rs
```diff
@@ -2974,17 +2979,78 @@ fn apply_width_pin(
...
+    // `cell_floor` best across the WHOLE pool (kept + tail): count the
+    // cell's kept-prefix occupants against the floor first.
+    let mut kept_per_cell: HashMap<(usize, usize), usize> = HashMap::new();
+    for (si, cand) in pooled.iter().take(limit) {
+        *kept_per_cell.entry((*si, cand.cell_idx)).or_default() += 1;
     }
-    pooled
+    let mut rescued: Vec<usize> = Vec::new();
+    for (cell, mut tail_idxs) in by_cell {
+        let already = kept_per_cell.get(&cell).copied().unwrap_or(0);
```
muralikpbhat, line 3025: **Survivors now exceed the pooled `limit` cap — surface the cost + note the composition.** The rescue adds up to `k` per scanned cell *after* the `limit` cut, so total survivors = `limit + W*k` (W = scanned cells, bounded by nprobe_max). That re-introduces width-scaling exact-rerank cost vs #479's flat `k*rerank_mult` cap — bounded, but at high nprobe the rerank fetch/CPU regresses, and no bench arm measures latency/GET alongside the (now-flatter) recall, so the trade ships unseen. Also a composition note for #497: its `rerank_for_k` sizing assumes survivors `<= limit`; the floored rescues bypass that cap, so any downstream budget/buffer sized off `limit` will under-count by `W*k`.

muralikpbhat (commented):
Reviewed with a verify pass on the core claim. **The floor is the right mechanism and a real improvement — it eliminates the catastrophic 0.388 all-cells collapse (a cell can no longer be zeroed out by far cells' 1-bit false positives). But a verify pass confirmed the fix delivers a per-cell FLOOR, not the strict end-to-end recall monotonicity the code comment and PR body claim.** Not a correctness bug — an over-claim plus coverage/measurement gaps. Inline comments mark each.

### What's solid (verified)
- The floor logic is sound: each scanned cell retains its own top-`cell_floor` by 1-bit estimate, immune to other cells' candidates; accounting is off-by-one-clean (disjoint prefix/tail, floor debited by prefix occupants); cost is bounded (`limit + W·k`, keeps `k` candidates not whole cells — does not reintroduce the pre-#479 whole-cell fetch); cold arm correctly excluded, filtered path not engaged, `floor=0` = exact backward-compat.
- Bench correctness is clean: exact brute-force ground truth independent of the IVF law, fresh top-k per width (no cross-nprobe cache reuse), "all cells" genuinely probes every populated cell, post-drain/post-compact use the right truth. The extra files in the diff are merged-in #501/#503, not bundled scope.

### The gap — floor, not strict monotonicity (verifier CONFIRMED)
Kept-count for a cell = `max(already, cell_floor=k)`, but `limit = k·rerank_mult` and `RERANK_MULT` defaults to **256**. At narrow nprobe a cell can occupy far more than `k` prefix slots (pool slack, up to ~256k); as nprobe widens, that cell's prefix occupancy is monotone *non-increasing* (fixed candidates only get out-ranked), and the rescue restores only to `k`. So a true top-k neighbor the 1-bit estimate ranks in the slack band `[k+1, 256k]` — exactly the false-negative case `rerank_mult` exists to catch — is kept narrow, evicted wide → **recall still inverts in that band** (empty only at the degenerate `rerank_mult=1`). The catastrophic collapse is gone; the residual inversion is not.

### Recommendation
Not blocking on correctness — the fix strictly improves recall. Before merge: (1) **soften the wording** (the code comment "monotone in recall by construction … only ever ADDS survivors" and the PR-body contract "a widening probe knob must never lower recall") to "floors the worst case / greatly reduces the inversion; a residual remains in the pool-slack band above the floor" — or, if strict monotonicity is the goal, make the floor width-/occupancy-aware so a cell never loses its above-floor survivors; (2) **assert the contract** rather than eyeball it (the width-sweep bench is print-only and the unit guard tests only the weaker floored-survivors-retained property); (3) **measure the cost** (the sweep reports recall only — the floor's `W·k` extra reranks at high nprobe are invisible).

muralikpbhat (approved):
Approving — to be clear on the earlier review, none of the findings are correctness bugs. The per-cell floor logic is sound (off-by-one-clean prefix/tail accounting; floored survivors are eviction-immune) and the fix **strictly improves recall**: it eliminates the catastrophic 0.388 all-cells collapse with no regression on any path.

The three inline notes are non-blocking follow-ups:
1. Soften the wording — the code comment ('monotone in recall by construction … only ever ADDS survivors') and PR-body contract ('a widening probe knob must never lower recall') overstate it. It's a strong per-cell **floor**, not strict end-to-end monotonicity: a residual inversion remains in the pool-slack band above the floor when rerank_mult>1 (max at the default 256). Suggest 'floors the worst case / greatly reduces the inversion'.
2. Assert the contract in CI — the width-sweep bench is print-only and the unit guard tests only the weaker floored-survivors-retained property.
3. Measure the cost — the floor adds up to W·k extra exact reranks at high nprobe (bounded, but the sweep reports recall only).

None block the merge; the fix is a clear improvement.

PR #508, opened 2026-08-01
supertable: skip past crash-orphaned manifest lists in OCC retries

ekechinwokah:
**Problem:** A writer that dies between its manifest list PUT and its pointer CAS leaves the list object occupying the next manifest id while the pointer still names the prior commit. Lists are conditional-create and never overwritten, so every retry at that id surfaced PreconditionFailed, the refresh saw no pointer movement, and the first commit after the crash exhausted as WriteContentionExhausted until a GC sweep past the safety gap reclaimed the orphan.

Teach the OCC retry loops to detect the case and derive their next successor past the occupied id: after a contention-failed attempt, if the refreshed pointer sits short of the attempted id AND a HEAD probe finds a list object there, raise an in-memory next-manifest-id floor (never persisted) consulted by successor derivation. Both conditions are required — pointer lag alone also occurs when a loser refreshes before the winner's pointer CAS lands, and skipping there would gap ids on ordinary contention.

Recovery never deletes, overwrites, or reads the orphan: it stays unreferenced and ages into the existing GC sweep, and a slow-but-alive writer keeps its candidate (its pointer CAS and ours are fenced on the same prior etag, so exactly one wins). Manifest ids may gap after a crash; readers follow the pointer and GC's live set names only the current id, so nothing assumes density.

**Regression tests:** direct orphan-injection commit tests (same handle, fresh handle, consecutive orphans, GC of the skipped orphan) and a crash-harness test that commits after a SIGABRT between the list PUT and pointer CAS — both fail before the fix with "write contention exhausted retries".

### DIFF
--- a/src/supertable/writer.rs
+++ b/src/supertable/writer.rs
@@ -6485,9 +6485,12 @@ pub(in crate::supertable) async fn refresh_and_orphaned_id_floor(
     // the occupied run starting at `attempted_id` is finite; the floor
     // contract is only that every id below it is occupied, which holds at
     // whatever point the walk stops (first free id, probe cap, or a
-    // failed probe).
+    // failed probe). The saturating bound keeps the arithmetic total even
+    // for ids no real table reaches; below it, the plain increment cannot
+    // wrap.
+    let probe_limit = attempted_id.saturating_add(MAX_ORPHAN_RUN_PROBES);
     let mut next_free_id = attempted_id;
-    while next_free_id < attempted_id + MAX_ORPHAN_RUN_PROBES {
+    while next_free_id < probe_limit {
         match storage.head(&manifest_uri(next_free_id)).await {
             Ok(_) => next_free_id += 1,
             Err(StorageError::NotFound { .. }) => break,
--- a/src/supertable/writer.rs
+++ b/src/supertable/writer.rs
@@ -158,7 +158,7 @@ use crate::{
         hidden_deleted::{self, encode_deleted_ids},
         manifest::{
             ClusterCentroids, RabitqAdmitContext,
-            commit::get_current_manifest_etag,
+            commit::{get_current_manifest_etag, manifest_uri},
             list::{
                 CellRoutingParams, DrainedVersionRanges, GlobalVectorIndex, PartitionStrategy,
                 WIDTH_LAW_KS,
@@ -5758,8 +5758,16 @@ async fn stamp_slow_vector_state(
         return Ok(());
     };
     let max_retries = inner.options.max_commit_retries.max(1);
+    let mut next_id_floor: u64 = 0;
     for attempt in 0..max_retries {
         let old = inner.manifest.load_full();
+        // A prior attempt found its id occupied by a crash-orphaned
+        // manifest list — derive this attempt's successor past it.
+        let old = if next_id_floor > 0 {
+            Arc::new(old.with_next_manifest_id_floor(next_id_floor))
+        } else {
+            old
+        };
         let entries = old.get_all_superfiles();
         if entries.is_empty() && pending_drain.is_none() {
             // Nothing to describe (pre-drain / empty table); the ref is
@@ -5800,6 +5808,7 @@ async fn stamp_slow_vector_state(
         }
         let new_manifest =
             old.with_slow_vector_state(published.uri, published.content_hash, published.centroids);
+        let attempted_id = new_manifest.get_manifest_id();
         let prev_etag = get_current_manifest_etag(&storage, Arc::clone(&old))
             .await
             .inspect_err(|e| inner.note_commit_error(e))
@@ -5813,9 +5822,11 @@ async fn stamp_slow_vector_state(
                 return Ok(());
             }
             Err(SupertableCommitError::WriteContentionExhausted) if attempt + 1 < max_retries => {
-                refresh_inner_state_async(inner, &storage)
-                    .await
-                    .map_err(|e| BuildError::Store(e.to_string()))?;
+                next_id_floor = next_id_floor.max(
+                    refresh_and_orphaned_id_floor(inner, &storage, attempted_id)
+                        .await
+                        .map_err(|e| BuildError::Store(e.to_string()))?,
+                );
                 sleep(backoff_delay(attempt)).await;
             }
             Err(e) => return Err(BuildError::Store(e.to_string())),
@@ -5837,8 +5848,16 @@ async fn record_hidden_deleted_ids(
         return Ok(());
     };
     let max_retries = inner.options.max_commit_retries.max(1);
+    let mut next_id_floor: u64 = 0;
     for attempt in 0..max_retries {
         let old = inner.manifest.load_full();
+        // A prior attempt found its id occupied by a crash-orphaned
+        // manifest list — derive this attempt's successor past it.
+        let old = if next_id_floor > 0 {
+            Arc::new(old.with_next_manifest_id_floor(next_id_floor))
+        } else {
+            old
+        };
         let mut ids = hidden_deleted::deleted_user_ids(&old)
             .map_err(|e| BuildError::Store(e.to_string()))?
             .as_ref()
@@ -5852,6 +5871,7 @@ async fn record_hidden_deleted_ids(
         }
         let bytes = encode_deleted_ids(&ids);
         let new_manifest = old.with_deleted_user_ids(bytes);
+        let attempted_id = new_manifest.get_manifest_id();
         let prev_etag = get_current_manifest_etag(&storage, Arc::clone(&old))
             .await
             .inspect_err(|e| inner.note_commit_error(e))
@@ -5865,9 +5885,11 @@ async fn record_hidden_deleted_ids(
                 return Ok(());
             }
             Err(SupertableCommitError::WriteContentionExhausted) if attempt + 1 < max_retries => {
-                refresh_inner_state_async(inner, &storage)
-                    .await
-                    .map_err(|e| BuildError::Store(e.to_string()))?;
+                next_id_floor = next_id_floor.max(
+                    refresh_and_orphaned_id_floor(inner, &storage, attempted_id)
+                        .await
+                        .map_err(|e| BuildError::Store(e.to_string()))?,
+                );
                 sleep(backoff_delay(attempt)).await;
             }
             Err(e) => return Err(BuildError::Store(e.to_string())),
@@ -5940,8 +5962,16 @@ pub(in crate::supertable) async fn persist_commit_async(
     let max_retries = opts.max_commit_retries.max(1);
     let drive = async move {
         let mut last_err: Option<SupertableCommitError> = None;
+        let mut next_id_floor: u64 = 0;
         for attempt in 0..max_retries {
             let old = inner.manifest.load_full();
+            // A prior attempt found its id occupied by a crash-orphaned
+            // manifest list — derive this attempt's successor past it.
+            let old = if next_id_floor > 0 {
+                Arc::new(old.with_next_manifest_id_floor(next_id_floor))
+            } else {
+                old
+            };
             // Re-apply call-site stamps on every attempt. A pre-store of these
             // fields is not OCC-safe: contention refresh reloads from storage
             // and would drop them before a successful CAS.
@@ -5950,6 +5980,7 @@ pub(in crate::supertable) async fn persist_commit_async(
             } else {
                 Arc::new(list_metadata.apply(&old))
             };
+            let attempted_id = base.get_next_manifest_id();
             let pending_writes = &mut pending_storage_writes;
             let pending_replaces = &mut pending_storage_replaces;
             match try_commit_attempt(
@@ -5968,7 +5999,9 @@ pub(in crate::supertable) async fn persist_commit_async(
                 Err(SupertableCommitError::WriteContentionExhausted)
                     if attempt + 1 < max_retries =>
                 {
-                    refresh_inner_state_async(inner, &storage_async).await?;
+                    next_id_floor = next_id_floor.max(
+                        refresh_and_orphaned_id_floor(inner, &storage_async, attempted_id).await?,
+                    );
                     last_err = Some(SupertableCommitError::WriteContentionExhausted);
                     sleep(backoff_delay(attempt)).await;
                 }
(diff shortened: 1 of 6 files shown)


src/supertable/writer.rs
```diff
@@ -6387,6 +6420,47 @@ pub(in crate::supertable) async fn refresh_inner_state_async(
...
+    let refreshed_id = inner.manifest.load_full().get_manifest_id();
+    if refreshed_id >= attempted_id {
+        return Ok(0);
+    }
+    match storage.head(&manifest_uri(attempted_id)).await {
+        Ok(_) => Ok(attempted_id + 1),
+        Err(StorageError::NotFound { .. }) => Ok(0),
+        Err(e) => Err(SupertableCommitError::Storage(e)),
+    }
+}
```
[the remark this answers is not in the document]
ekechinwokah: taken in 88bc958 — the probe now walks the contiguous occupied run to the first free id, so recovery costs one retry regardless of run length; per-id skipping could indeed exhaust max_commit_retries on a long run with zero live contention. the consecutive-orphans regression test now pins exactly this: four orphans against a two-retry budget, which fails under per-id skipping and passes with the walk. cost is one extra HEAD on the (rare) recovery path only; the safety argument is unchanged — ids above the pointer are only ever held by unpublished lists, and the pointer CAS still elects exactly one winner among any live candidates.

ekechinwokah:
on the reviewer-guide focus areas: (1) design ref — the prior-discussion issue is filed and the PR body now references it (crash between list PUT and pointer CAS leaves the next id occupied; candidates a/b/c and why skip-past-id won are analyzed there). (2) benchmark — the bench comparison vs main is queued for an exclusive run and will be added to the PR body before merge; the change adds no work to the uncontended commit path (one max() per successor derivation; the HEAD probe only runs on contention retries). verification on 88bc958: full suite 2492/0 across all 12 targets, crash harness 6/6 (incl. the new commit-after-list-kill test), concurrent OCC binary 15/15 stress runs, clippy -D warnings clean, cargo public-api no drift.

src/supertable/writer.rs
```diff
@@ -6387,6 +6420,61 @@ pub(in crate::supertable) async fn refresh_inner_state_async(
...
+    // Ids above the pointer are only ever held by unpublished lists, and
+    // every write below the pointer has already happened, so the run of
+    // occupied ids starting at `attempted_id` is finite and the walk
+    // terminates at the first free id.
+    let mut next_free_id = attempted_id;
+    loop {
+        match storage.head(&manifest_uri(next_free_id)).await {
+            Ok(_) => next_free_id += 1,
+            Err(StorageError::NotFound { .. }) => break,
+            Err(e) => return Err(SupertableCommitError::Storage(e)),
```
muralikpbhat: **Robustness regression on the write path.** A transient (non-NotFound) storage error during the HEAD walk `return Err(SupertableCommitError::Storage(e))` propagates out of the retry loop and fails the commit — where pre-PR a plain retry would have survived a transient 503/timeout. The walk only runs on the contention path (already a degraded/racy moment where transients are likeliest), so this newly couples a blip in the *probe* to a hard *commit* failure. Suggest treating a transient probe error as 'floor undetermined -> retry at the natural id' (or a bounded probe retry) rather than failing the caller's commit. (Flagged independently by both review passes.)

src/supertable/writer.rs
```diff
@@ -6387,6 +6420,61 @@ pub(in crate::supertable) async fn refresh_inner_state_async(
...
+    if refreshed_id >= attempted_id {
+        return Ok(0);
+    }
+    // Ids above the pointer are only ever held by unpublished lists, and
+    // every write below the pointer has already happened, so the run of
+    // occupied ids starting at `attempted_id` is finite and the walk
+    // terminates at the first free id.
+    let mut next_free_id = attempted_id;
+    loop {
+        match storage.head(&manifest_uri(next_free_id)).await {
```
muralikpbhat, line 6494: **Latency cliff on a long orphan run.** This run-walk is sequential and unbounded — repeated crashes at consecutive ids (each crashed writer skipped past the prior orphan via this very mechanism, PUT its list, then died pre-CAS) build a long contiguous orphan run, and a later commit's first contention walks the entire run one serial HEAD at a time before it can retry. On LocalFS, `head()` GETs the body for small objects, so these probes are full reads. Correct and terminating, but bounded only by uncollected orphans; a listing-based scan of the id range (or a bounded walk) would cap it.

src/supertable/writer.rs
```diff
@@ -6387,6 +6420,61 @@ pub(in crate::supertable) async fn refresh_inner_state_async(
...
+///   ages into the GC sweep, while a live mid-commit writer keeps its
+///   candidate: its pointer CAS and ours are fenced on the same prior
+///   etag, so exactly one wins and the loser retries as usual.
+///
+/// Both conditions are required. The pointer sitting short of
+/// `attempted_id` alone is not proof of an occupant: a loser's refresh can
+/// run before the winner's pointer CAS lands, and a floored attempt can
+/// lose its etag pre-check with nothing at its id — hence the existence
+/// probe before skipping.
+pub(in crate::supertable) async fn refresh_and_orphaned_id_floor(
```
muralikpbhat, line 6466: **Behavioral note — the gate fires on live in-flight winners too (correct, but orphan-generating).** The two conditions (pointer-lag + HEAD-finds-list) cannot distinguish a crash-orphan from a live winner that has PUT its list but not yet CAS'd its pointer. So under ordinary *live* contention a loser can raise its floor past a live winner's id and PUT its own list at N+1 — the etag fence keeps this correct (verified: exactly one CAS wins, no data lost), but where pre-PR the loser simply spun on id N (zero orphans), post-PR it now litters an orphan at N+1 and can cascade (N+2...). On a hot-contention table this turns crash-recovery machinery into steady-state orphan production + GC load. Worth documenting the tradeoff — and it's why the no-false-positive test (below) matters.

tests/supertable/commit/orphaned_list_recovery.rs
```diff
@@ -0,0 +1,163 @@
...
+fn put_orphan(storage: &Arc<dyn StorageProvider>, manifest_id: u64) {
+    futures::executor::block_on(storage.put_atomic(
+        &manifest_uri(manifest_id),
+        Bytes::from_static(ORPHANED_LIST_BYTES),
+    ))
+    .expect("orphan PUT");
+}
+
+#[test]
+fn commit_skips_past_orphaned_manifest_list() {
```
muralikpbhat, line 89: **The load-bearing safety property is untested.** All three new tests inject an *inert* crash orphan (`put_orphan` + frozen pointer, no live writer), so the skip always fires against residue no one will ever CAS. The genuinely hazardous interleaving — writer W has PUT its list at N+1 and is about to CAS while loser L refreshes, HEAD-finds W's list, and raises its floor past it — is never exercised; its correctness rests entirely on `get_current_manifest_etag` rejecting L, and a future change that let the floor touch `base.manifest_id` would pass every test here yet allow a double-publish. Two tests would lock the invariants: (a) a concurrent-writer race asserting exactly-one-winner / no double-publish when one writer is paused between list PUT and pointer CAS; (b) a no-false-positive test running N live-contending commits and asserting the final `manifest_id` is contiguous (== N, no gaps) with zero unreferenced list objects. Also: the GC-cleanup assertion uses `Duration::ZERO` (line 105), so it never exercises the safety-gap 'too_new' guard that must protect a fresh in-flight orphan from premature reclamation.

muralikpbhat (commented):
Reviewed with a deep verify pass on the distributed-systems crux. **The skip logic is correct — the dangerous case (skipping past a live in-flight winner) is provably safe.** `with_next_manifest_id_floor` raises only `next_manifest_id_floor`, leaving `manifest_id`/`list` untouched, and the pointer CAS fences via `get_current_manifest_etag` on the **base** id — which the floor never changes. So a loser that skips past a live winner's just-PUT list is still rejected the moment the winner's pointer CAS lands (pointer.id != old.id -> WriteContentionExhausted): a single linear CAS chain, exactly one live id per generation, **no collision, no dual-publish** — worst case is a benign id gap (readers follow the pointer, GC reclaims the orphan). Also verified: the floor is monotonic and per-commit-scoped (inert on later commits), ordinary contention raises no floor (no id gapping), a contiguous orphan run is escaped in one retry, the HEAD is on the correct list key, and a probe error is not treated as "orphan present". Crash-recovery and fresh-process recovery are genuinely tested; GC unreachability of the orphan is real. **No correctness bug.**

The findings (inline) are robustness + coverage. Two are worth addressing before merge — the transient-HEAD regression and the etag-fence test gap; the rest are document-or-follow-up.

**GC-invariant note (summary-level, likely pre-existing):** this PR converts "leave the orphan for GC" from incidental into *load-bearing* behavior, so it now leans on `safety_gap > max writer PUT->CAS latency`. A too-small gap racing a slow-but-alive writer could let GC reclaim a just-PUT list before its pointer CAS lands (dangling pointer -> reopen fails). Worth a test at a realistic gap since the PR relies on this more directly than before.

src/supertable/writer.rs
ekechinwokah: taken in 59211bf — the floor is now advisory: a transient (non-NotFound) probe error ends the walk at whatever it established, warn-logged, and the retry proceeds at the natural (or partial) floor instead of failing the caller's commit. a real orphan re-detects on the next contention; a persistent storage fault still surfaces through the retry's own I/O. you're right that the probe runs exactly where transients are likeliest — coupling it to a hard commit failure was a regression.

src/supertable/writer.rs
ekechinwokah: taken in 59211bf — the walk is bounded (32 probes per retry). a longer run escapes in chunks: the floor lands on the first unprobed id, the next collision resumes the walk from there, so recovery stays O(run/32) retries instead of one unbounded serial walk before the first retry. the const's doc also records your LocalFS point (head of a small object reads its body, so probes aren't free).

src/supertable/writer.rs
ekechinwokah: agreed, and documented in 59211bf on the helper's doc: the two conditions cannot tell a crash orphan from a live winner inside its list-PUT→pointer-CAS window, so a hot-contention loser can skip past a live winner and leave its own earlier list as an orphan — correctness held by the etag fence (exactly one CAS wins), the cost being steady-state orphan production the GC sweep reclaims. the refreshed_id >= attempted_id pre-check keeps the common lost-race shape (winner already published) on the dense-id path with no probe at all. deliberate trade: crash-recovery availability over orphan hygiene under contention.

tests/supertable/commit/orphaned_list_recovery.rs
ekechinwokah: taken in 59211bf — both suggested tests added, plus the gc gap. (a) the live interleaving: a storage wrapper freezes the winner's FIRST pointer CAS (signalled from inside put_if_match, i.e. with its list already durable) while the loser collides, HEAD-finds the in-flight list, and floors past it; released, the winner's fenced CAS must LOSE and its OCC retry republishes — asserted exactly-once (three commits → three superfiles, final id 4, no double-publish, the skipped list left for GC). (b) the false-positive tripwire: two handles alternate commits without refreshing for three rounds; the final pointer id must equal the commit count AND manifest/ must hold exactly one list per id — any orphan from a spurious skip fails the count. and the gc assertion now exercises the too_new guard first: a 1h safety gap must protect the fresh orphan (objects_skipped_too_new >= 1) before the Duration::ZERO sweep reclaims it.

src/supertable/writer.rs
```diff
@@ -6387,6 +6420,94 @@ pub(in crate::supertable) async fn refresh_inner_state_async(
...
+            Err(e) => {
+                warn!(
+                    error = %e,
+                    manifest_id = next_free_id,
+                    "orphaned-list probe failed; retrying at the last established id"
+                );
+                break;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
ekechinwokah: taken in adapted form in d9f47c8 — the bound is hoisted into a saturating probe_limit (also nicer than recomputing the sum per iteration). kept the plain increment inside the loop: with the limit saturated, next_free_id < probe_limit <= u64::MAX means the increment provably cannot wrap, and unguarded manifest_id + 1 is the codebase's own convention everywhere else (ids can't physically approach the ceiling).

tests/supertable/commit/orphaned_list_recovery.rs
```diff
@@ -0,0 +1,352 @@
...
+/// writers into steady orphan production.
+#[test]
+fn stale_handle_contention_keeps_manifest_ids_dense() {
+    let dir = TempDir::new().expect("tempdir");
+    let storage: Arc<dyn StorageProvider> =
+        Arc::new(LocalFsStorageProvider::new(dir.path()).expect("provider"));
+    let st_a = Supertable::create(default_supertable_options().with_storage(Arc::clone(&storage)))
+        .expect("create");
+    let st_b = Supertable::create(default_supertable_options().with_storage(Arc::clone(&storage)))
+        .expect("open second handle");
```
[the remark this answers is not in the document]
ekechinwokah: not critical.

muralikpbhat (approved):
Approving — all findings from the earlier review are reconciled at HEAD (d9f47c8), verified in code and backed by tests that assert the real invariants. The skip logic was already provably correct (etag fence on the base id); these were the robustness + coverage items:

- **Transient-HEAD → advisory floor** — the probe error arm now warns and `break`s at the established floor (retry proceeds at the natural/partial id) instead of failing the commit; a persistent fault still surfaces through the retry's own I/O within the bounded budget. The regression is gone.
- **Bounded walk** — `MAX_ORPHAN_RUN_PROBES = 32` with saturating arithmetic; a long run escapes in O(run/32) retries, no wrap, no off-by-one (`consecutive_orphaned_manifest_lists` proves chunked single-retry escape).
- **Orphan-litter** — documented on the helper (and `write_manifest`): a loser can't distinguish a crash-orphan from a live winner, so it may skip past and leave its own list as an orphan — safe (etag fence elects one CAS), reclaimed by GC.
- **The load-bearing tests, added and real:** `loser_skips_a_live_winner_and_both_commits_land_exactly_once` drives the genuine list-durable/pointer-pending freeze and asserts the winner's fenced CAS LOSES + republishes with no double-publish (final id 4, exactly 3 superfiles, skipped list left for GC); `stale_handle_contention_keeps_manifest_ids_dense` pins the no-false-positive property (contiguous ids, no spurious orphan); and the GC test now exercises the non-zero safety-gap `too_new` branch, not just `Duration::ZERO`.

No new defects, and the public-hygiene scrub is clean. Solid turnaround on a subtle distributed-systems fix.

PR #509, opened 2026-08-01
bench: assert width-sweep monotonicity, price the floor; scope the comments

ekechinwokah:
## Summary

The three non-blocking follow-ups from the #505 review, in one small PR:

1. **Soften the wording** — the `select_global_shortlist` comment claimed
   "monotone in recall by construction"; the accurate statement is that the
   FLOORED CORE is monotone by construction, which floors the worst case:
   a candidate ranked below its cell's floor but inside the global pool (a
   band that grows with `rerank_mult`) keeps the pooled selection's
   width-blind behavior. The comment now says exactly that. (No behavior
   change — comment only.)

2. **Assert the contract in CI** — the width sweep now asserts each
   consecutive pair monotone within `WIDTH_SWEEP_MONOTONICITY_SLACK`
   (0.01). Documented honestly as a tripwire, not a proof: the planted
   bench corpus reads flat even without the floor (its 1-bit estimates
   are too clean to swamp the pool — measured while building #505), so
   the assert catches gross regressions (a reintroduced width-blind cap,
   a floor that stops engaging) while the contract's real evidence stays
   the Cohere A/B at 1M and 10M on #505.

3. **Price the floor** — each sweep line now reports the p50 query
   latency and the worst-case floor admission (`floor<=width*k` rerank
   rows) alongside recall, via a `mean_recall_timed` variant that
   `mean_recall` now delegates to (no duplicated loop).

## Validation

Bench-layer + comment-only change: no engine behavior touched.
`cargo check --benches`, clippy clean, fmt, full lib build green. The
sweep runs (with the new assert live) on every standard vector bench at
post-drain and post-compact.

### DIFF
--- a/benches/utils/executors.rs
+++ b/benches/utils/executors.rs
@@ -1647,8 +1647,7 @@ pub mod vector {
             lat.push(started.elapsed());
             sum += corpus::recall_at_k(&hits, t);
         }
-        lat.sort_unstable();
-        let p50 = lat[lat.len() / 2];
+        let p50 = p50(&mut lat);
         (sum / queries.len() as f32, p50)
     }
 
--- a/benches/utils/executors.rs
+++ b/benches/utils/executors.rs
@@ -1612,12 +1612,32 @@ pub mod vector {
         nprobe: usize,
         rerank: usize,
     ) -> f32 {
+        mean_recall_timed(reader, column, queries, truths, k, nprobe, rerank).0
+    }
+
+    /// [`mean_recall`] plus the p50 query latency of the same pass, so a
+    /// probe sweep can price a setting on the same line that reports its
+    /// recall.
+    pub fn mean_recall_timed<R: VectorRead>(
+        reader: &R,
+        column: &str,
+        queries: &[Vec<f32>],
+        truths: &[Vec<u32>],
+        k: usize,
+        nprobe: usize,
+        rerank: usize,
+    ) -> (f32, Duration) {
         let mut sum = 0f32;
+        let mut lat: Vec<Duration> = Vec::with_capacity(queries.len());
         for (q, t) in queries.iter().zip(truths) {
+            let started = Instant::now();
             let hits = reader.topk_global(column, q, k, nprobe, rerank);
+            lat.push(started.elapsed());
             sum += corpus::recall_at_k(&hits, t);
         }
-        sum / queries.len() as f32
+        lat.sort_unstable();
+        let p50 = lat.get(lat.len() / 2).copied().unwrap_or_default();
+        (sum / queries.len() as f32, p50)
     }
 
     /// Largest doc count that still calibrates with the exhaustive
--- a/benches/utils/executors.rs
+++ b/benches/utils/executors.rs
@@ -1618,6 +1618,12 @@ pub mod vector {
     /// [`mean_recall`] plus the p50 query latency of the same pass, so a
     /// probe sweep can price a setting on the same line that reports its
     /// recall.
+    ///
+    /// Panics on a query/truth length mismatch or an empty battery: both
+    /// are harness bugs, and a measurement helper must fail loudly rather
+    /// than report a number computed from silently truncated pairs (or a
+    /// 0/0 = NaN recall, which the width-sweep tripwire would compare
+    /// unpredictably).
     pub fn mean_recall_timed<R: VectorRead>(
         reader: &R,
         column: &str,
@@ -1627,6 +1633,12 @@ pub mod vector {
         nprobe: usize,
         rerank: usize,
     ) -> (f32, Duration) {
+        assert_eq!(
+            queries.len(),
+            truths.len(),
+            "each query needs exactly one ground-truth row"
+        );
+        assert!(!queries.is_empty(), "recall over zero queries is undefined");
         let mut sum = 0f32;
         let mut lat: Vec<Duration> = Vec::with_capacity(queries.len());
         for (q, t) in queries.iter().zip(truths) {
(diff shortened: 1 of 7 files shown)


benches/utils/executors.rs
```diff
@@ -1612,12 +1612,32 @@ pub mod vector {
...
         for (q, t) in queries.iter().zip(truths) {
+            let started = Instant::now();
             let hits = reader.topk_global(column, q, k, nprobe, rerank);
+            lat.push(started.elapsed());
             sum += corpus::recall_at_k(&hits, t);
         }
-        sum / queries.len() as f32
+        lat.sort_unstable();
+        let p50 = lat.get(lat.len() / 2).copied().unwrap_or_default();
+        (sum / queries.len() as f32, p50)
```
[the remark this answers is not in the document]
ekechinwokah: agree on the diagnosis (zip truncation + wrong divisor could misreport recall on a length mismatch; 0/0 = NaN makes the width-sweep tripwire compare unpredictably), but went loud instead of neutral: both invariants are asserted rather than tolerated. every call site builds the query and truth batteries together, so a mismatch or empty battery is always a harness bug, and a measurement helper should crash over printing a number computed from truncated pairs — min-and-divide-by-samples would look plausible while wrong, and a neutral 0.0 on empty reads as a flat (passing) sweep to the monotonicity assert. smoke: the superfile vector warm battery at 100K runs green with the asserts in place; clippy --benches -D warnings clean.

benches/utils/supertable.rs
```diff
@@ -3167,10 +3186,27 @@ pub mod vector {
...
+                fmt_time(p50.as_nanos() as f64),
+                width * TOP_K,
             );
+            if let Some((prev_width, prev_recall)) = prev {
+                assert!(
+                    recall >= prev_recall - WIDTH_SWEEP_MONOTONICITY_SLACK,
+                    "unfiltered width-sweep ({state}): recall fell from \
+                     {prev_recall:.3} at nprobe={prev_width} to {recall:.3} at \
+                     nprobe={width} — a widening probe may not cost more than \
+                     {WIDTH_SWEEP_MONOTONICITY_SLACK} recall (the #494 \
```
muralikpbhat: Typo: this cites `#494`, but the inversion signature is documented under `#479` everywhere else in this file (e.g. line 3132). When the tripwire fires in CI this is the one breadcrumb it leaves, so a maintainer would follow it to the wrong issue. `s/#494/#479/`.

benches/utils/executors.rs
```diff
@@ -1612,12 +1612,44 @@ pub mod vector {
...
+        let mut lat: Vec<Duration> = Vec::with_capacity(queries.len());
         for (q, t) in queries.iter().zip(truths) {
+            let started = Instant::now();
             let hits = reader.topk_global(column, q, k, nprobe, rerank);
+            lat.push(started.elapsed());
             sum += corpus::recall_at_k(&hits, t);
         }
-        sum / queries.len() as f32
+        lat.sort_unstable();
+        let p50 = lat[lat.len() / 2];
```
muralikpbhat: Minor reuse/consistency: this inline `lat[lat.len()/2]` (upper-median) differs by one sample from the module's `pub fn p50` (line 64, lower-median), which is doc'd as the shared bench definition every runner uses. `exec_vec::p50(&mut lat)` would be exact and consistent. Non-blocking.

muralikpbhat (approved):
Approving — all three follow-ups from the #505 review are correctly and honestly implemented:

- **Wording softened** (comment-only, code untouched): the floored *core* is monotone by construction / it floors the worst case, with the residual band above the floor "that grows with `rerank_mult`" called out as still width-blind — matches the #505 verify finding exactly, not over-corrected.
- **CI monotonicity tripwire** with the right direction (`recall >= prev - 0.01` over ascending widths) and honestly caveated in both the body and the in-code comment that it's a *tripwire, not a proof* — the planted corpus reads flat even without the floor, so the assert catches only gross regressions while the real 0.994->0.388 evidence stays the Cohere A/B on #505. Fair, disclosed limitation.
- **Floor priced** — per-width p50 latency (fresh sample vec per width) plus a worst-case rerank-rows bound, printed per sweep row.

Two non-blocking nits inline (a `#494`->`#479` typo in the tripwire message, and a p50 helper-reuse). Nice, tidy close of the follow-ups.

benches/utils/supertable.rs
ekechinwokah: fixed in d610220 — s/#494/#479/. good catch on the breadcrumb: the assert message is the only thing a maintainer sees when the tripwire fires.

benches/utils/executors.rs
ekechinwokah: taken in d610220 — mean_recall_timed now calls the module's shared p50 (lower median) instead of the inline upper-median index, so the sweep's p50 column means the same thing as every other p50 the suite reports.

tests/catalog_public_api.rs
```diff
@@ -0,0 +1,260 @@
...
+    // survive the rewrite.
+    docs.optimize(&OptimizeOptions::default())
+        .expect("optimize");
+    let report = docs.gc(Duration::ZERO).expect("gc");
+    assert_eq!(report.delete_errors, 0);
+    assert!(
+        report.objects_skipped_live > 0,
+        "live manifest + pointer always survive a sweep"
+    );
+    let post_maintenance = docs
```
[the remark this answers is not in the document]
ekechinwokah: taken in 29f3c33 — the skipped-live assertion is gone; you're right that it couples the public-API tour to sweep bookkeeping a valid gc could legitimately change. delete_errors stays, and the survival property the assertion was reaching for is already pinned by the post-gc vector search. (same fix mirrored on the test/error-path-coverage branch, which carries a copy of this file.)

PR #510, opened 2026-08-01
bench/recall_while_ingest: resume-ingest + persistent ground-truth oracle

muralikpbhat:
## What

Extends the `recall_while_ingest` diagnostic so a recall-over-time run can **resume** instead of always rebuilding from zero, and **persists the ground truth** so a resumed or grown run doesn't replay it.

- **Resume-ingest** — `INFINO_BENCH_RESUME_INGEST` + `INFINO_BENCH_EXISTING_PREFIX` reopen an existing table and continue appending. The original build size is discovered from `COUNT(*)`; the corpus/query cluster count comes from the GT bin's header (provenance), so a resume continues the same deterministic stream.
- **Ground-truth oracle bin** — `INFINO_BENCH_GT_BIN` (a base prefix). Each checkpoint writes `{base}.M{count}.bin` atomically (temp + rename) and prunes older ones; resume loads the highest-count bin and validates the query-determining params before trusting it.

## Notes

- Bench-only. Reuses `corpus::grading`'s cursor readers for the GT bin (no duplicate binary parsers).
- Reopen uses the same `options_for(...)` as create — the table identity hash no longer depends on the centroid count, so a resume to a larger target opens cleanly.
- Reopen requires a persistent store (`INFINO_BENCH_STORE=local`, or a real object store); the default ephemeral session mints a fresh bucket per run.

## Review fixes (latest commit)

- **`n_cent` provenance**: resume reads `n_cent` from the GT-bin header, not from `COUNT(*)`. A run that crashed mid-target sits in a different doc-count band than its target, so a `COUNT(*)`-derived `n_cent` would seed a different generator than the one that filled the table.
- **Typed GT-bin errors**: `Corrupt` (io/magic/truncation) → replay rebuild; `Incompatible` (version/dim/k/seeds/n_queries) → fatal, never a silent rebuild with the wrong params.
- Hard-fail when `RESUME_INGEST` is set without `EXISTING_PREFIX` (no silent fall-through to CREATE); format-version check + named header constants; shared `for_each_bin` helper; top-level imports; `cfg(test)` mod (round-trip, provenance survival, `Incompatible` refusal, corruption, latest-selection).

## Verified (local, `INFINO_BENCH_STORE=local`)

**Completed-run resume, growing across a centroid-count band:**

| run | docs | recall@10 | cells |
|-----|------|-----------|-------|
| create (`KEEP_TABLE`) | 60,000 | 0.990 | 256 |
| resume → grow | 120,000 | 1.000 | 256 |

**Crash-recovery across a band boundary** — the case a `COUNT(*)`-derived `n_cent` gets wrong. Create targeting 200k (→ `n_cent` 1024), killed at the 60k checkpoint, then resume with no create-size hint:

\`\`\`
resume: n_cent=1024 from GT bin header (provenance) — COUNT(*)=60000 alone would derive 64
| 120K | recall@10 0.990 | 589 cells | 0 over_cap |
\`\`\`

Resume reads the original `n_cent` (1024) from the bin header rather than the wrong `COUNT(*)`-derived 64, so recall stays healthy (0.990). Without the provenance fix this run would silently score against a diverged generator.

### DIFF
--- a/benches/utils/recall_while_ingest.rs
+++ b/benches/utils/recall_while_ingest.rs
@@ -186,30 +186,6 @@ struct GtBin {
     heaps: Vec<HeldTopK>,
 }
 
-fn gt_rd_u32(b: &[u8], p: &mut usize) -> std::io::Result<u32> {
-    let s = b.get(*p..*p + 4).ok_or_else(|| {
-        std::io::Error::new(std::io::ErrorKind::UnexpectedEof, "GT bin truncated")
-    })?;
-    *p += 4;
-    Ok(u32::from_le_bytes(s.try_into().unwrap()))
-}
-
-fn gt_rd_u64(b: &[u8], p: &mut usize) -> std::io::Result<u64> {
-    let s = b.get(*p..*p + 8).ok_or_else(|| {
-        std::io::Error::new(std::io::ErrorKind::UnexpectedEof, "GT bin truncated")
-    })?;
-    *p += 8;
-    Ok(u64::from_le_bytes(s.try_into().unwrap()))
-}
-
-fn gt_rd_f32(b: &[u8], p: &mut usize) -> std::io::Result<f32> {
-    let s = b.get(*p..*p + 4).ok_or_else(|| {
-        std::io::Error::new(std::io::ErrorKind::UnexpectedEof, "GT bin truncated")
-    })?;
-    *p += 4;
-    Ok(f32::from_le_bytes(s.try_into().unwrap()))
-}
-
 /// Serialize the running GT heaps to `path` atomically (temp + rename) so a
 /// crash never leaves a half-written bin. The header carries the provenance a
 /// reload needs to reject an incompatible bin.
@@ -249,20 +225,22 @@ fn gt_bin_write(
 /// bin *contents* vs the indexed vectors can't be verified here; the caller
 /// warns and reconciles the covered `m` against the index's committed count.
 fn gt_bin_read(path: &str, n_queries: usize, n_cent: usize) -> std::io::Result<GtBin> {
+    use crate::corpus::grading::{read_f32, read_u32, read_u64};
     let mkerr = |m: String| std::io::Error::new(std::io::ErrorKind::InvalidData, m);
     let b = std::fs::read(path)?;
     if b.get(0..4) != Some(b"GTB1".as_slice()) {
         return Err(mkerr("GT bin: bad magic".into()));
     }
-    let mut p = 4usize;
-    let _version = gt_rd_u32(&b, &mut p)?;
-    let m = gt_rd_u64(&b, &mut p)? as usize;
-    let f_nq = gt_rd_u64(&b, &mut p)? as usize;
-    let f_nc = gt_rd_u64(&b, &mut p)? as usize;
-    let f_dim = gt_rd_u32(&b, &mut p)? as usize;
-    let f_k = gt_rd_u32(&b, &mut p)? as usize;
-    let f_vseed = gt_rd_u64(&b, &mut p)?;
-    let f_qseed = gt_rd_u64(&b, &mut p)?;
+    // Cursor past the 4-byte magic; the shared `grading` readers advance it.
+    let mut cur: &[u8] = &b[4..];
+    let _version = read_u32(&mut cur)?;
+    let m = read_u64(&mut cur)? as usize;
+    let f_nq = read_u64(&mut cur)? as usize;
+    let f_nc = read_u64(&mut cur)? as usize;
+    let f_dim = read_u32(&mut cur)? as usize;
+    let f_k = read_u32(&mut cur)? as usize;
+    let f_vseed = read_u64(&mut cur)?;
+    let f_qseed = read_u64(&mut cur)?;
     if f_nq != n_queries
         || f_nc != n_cent
         || f_dim != DIM
@@ -277,11 +255,11 @@ fn gt_bin_read(path: &str, n_queries: usize, n_cent: usize) -> std::io::Result<G
     }
     let mut heaps = Vec::with_capacity(f_nq);
     for _ in 0..f_nq {
-        let len = gt_rd_u32(&b, &mut p)? as usize;
+        let len = read_u32(&mut cur)? as usize;
         let mut h = HeldTopK::new();
         for _ in 0..len {
-            let id = gt_rd_u32(&b, &mut p)?;
-            let score = gt_rd_f32(&b, &mut p)?;
+            let id = read_u32(&mut cur)?;
+            let score = read_f32(&mut cur)?;
             h.offer(score, id);
         }
         heaps.push(h);
--- a/benches/utils/recall_while_ingest.rs
+++ b/benches/utils/recall_while_ingest.rs
@@ -152,6 +152,11 @@ impl HeldTopK {
     fn ids(&self) -> Vec<u32> {
         self.heap.iter().map(|Reverse(c)| c.id).collect()
     }
+
+    /// `(id, score)` pairs currently held — for GT-bin serialization.
+    fn entries(&self) -> Vec<(u32, f32)> {
+        self.heap.iter().map(|Reverse(c)| (c.id, c.dot)).collect()
+    }
 }
 
 /// Fold a freshly generated batch's exact top-k into the running heaps, one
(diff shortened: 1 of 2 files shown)


benches/utils/recall_while_ingest.rs
```diff
@@ -572,6 +756,36 @@ pub fn run() {
...
+            .column(0)
+            .as_any()
+            .downcast_ref::<arrow_array::Int64Array>()
+            .expect("count column is Int64")
+            .value(0) as usize;
+        Some(m)
+    } else {
+        None
+    };
+    let n_cent = match resume_m {
```
ekechinwokah: this is only valid when the prior run COMPLETED its target. n_cent is banded — a run targeting 10m (large band) that crashes at 3m resumes in the medium band, and the replayed generator + queries silently diverge from what's actually in the table. no error anywhere, just garbage recall. since the pr explicitly sells crash recovery, resume can't derive n_cent from count(*) alone. cheapest fix: persist n_cent (+ target) in a tiny sidecar at create — or trust the newest bin's header n_cent as provenance (see the fallback comment below). the 60k→120k verification passes because it stays in one band and run 1 completed.

benches/utils/recall_while_ingest.rs
```diff
@@ -589,6 +803,94 @@ pub fn run() {
...
+                            }
+                        } else {
+                            eprintln!(
+                                "[recall_while_ingest] WARNING: GT bin AHEAD of index ({} > {m}) — recomputing GT for [0, {m}) (truncating a heap could drop a true neighbor)",
+                                bin.m
+                            );
+                        }
+                    }
+                    Err(e) => eprintln!(
+                        "[recall_while_ingest] GT bin {path} load failed ({e}); recomputing GT by replay"
```
ekechinwokah: this fallback turns a DETECTED inconsistency into silent garbage: a bin persisted by a crashed run carries the CORRECT n_cent, the params check refuses it, and we then replay with the wrong one. split the error paths — io/corrupt → replay is fine; params mismatch → fatal, the bin is provenance the derived n_cent can't contradict.

benches/utils/recall_while_ingest.rs
```diff
@@ -515,7 +681,19 @@ pub fn run() {
...
+    // Optional on-disk ground-truth bin, given as a BASE prefix. Each
+    // checkpoint writes `{base}.M{count}.bin` (count = docs covered, visible in
+    // the filename); resume loads the highest-count one. Lets a grown/resumed
+    // run skip the [0, M) GT replay, and makes a crash mid-run cheap to recover.
+    let gt_bin_path = std::env::var("INFINO_BENCH_GT_BIN")
+        .ok()
+        .filter(|s| !s.is_empty());
+    // Resume against an existing table: gated on INFINO_BENCH_RESUME_INGEST set
+    // AND a non-empty INFINO_BENCH_EXISTING_PREFIX (the prefix the fixture opens).
+    let resume = std::env::var_os("INFINO_BENCH_RESUME_INGEST").is_some()
```
ekechinwokah: footgun: RESUME_INGEST set + missing/empty EXISTING_PREFIX makes resume=false and falls through to the CREATE path — fresh prefix, full rebuild. the expect() on the fixture that names this precondition can never fire because the gate routes around it. the user declared intent to resume; hard-fail here instead of ingesting 10m docs into a surprise bucket.

benches/utils/recall_while_ingest.rs
```diff
@@ -173,6 +178,167 @@ fn update_heaps(heaps: &mut [HeldTopK], queries: &[Vec<f32>], flat: &[f32], base
...
+    std::fs::rename(&tmp, path)
+}
+
+/// Read a GT bin, REFUSING it if the query-generation params don't match this
+/// run (seed / n_queries / n_cent / dim / k) — those fully determine the query
+/// vectors, so a mismatch means the heaps are keyed to different queries. The
+/// bin *contents* vs the indexed vectors can't be verified here; the caller
+/// warns and reconciles the covered `m` against the index's committed count.
+fn gt_bin_read(path: &str, n_queries: usize, n_cent: usize) -> std::io::Result<GtBin> {
+    use crate::corpus::grading::{read_f32, read_u32, read_u64};
```
ekechinwokah: convention: imports at the top of the file, not function-local (and the new helpers lean on inline std::fs / std::path / std::io paths throughout — same rule). the file has some legacy inline std::env, but new code shouldn't add more.

benches/utils/recall_while_ingest.rs
```diff
@@ -173,6 +178,167 @@ fn update_heaps(heaps: &mut [HeldTopK], queries: &[Vec<f32>], flat: &[f32], base
...
+    let mkerr = |m: String| std::io::Error::new(std::io::ErrorKind::InvalidData, m);
+    let b = std::fs::read(path)?;
+    if b.get(0..4) != Some(b"GTB1".as_slice()) {
+        return Err(mkerr("GT bin: bad magic".into()));
+    }
+    // Cursor past the 4-byte magic; the shared `grading` readers advance it.
+    let mut cur: &[u8] = &b[4..];
+    let _version = read_u32(&mut cur)?;
```
ekechinwokah: version is read but never checked — accept-anything defeats the field; a future format change misparses instead of refusing. check == 1, and the constant wants a name (same for the header sizes) per the no-magic-numbers rule.

benches/utils/recall_while_ingest.rs
```diff
@@ -173,6 +178,167 @@ fn update_heaps(heaps: &mut [HeldTopK], queries: &[Vec<f32>], flat: &[f32], base
...
+        }
+    }
+    best
+}
+
+/// Delete every `{base}.M{count}.bin` except `keep` — the newest bin fully
+/// supersedes older ones (resume always loads the highest count). Call this
+/// only AFTER the `keep` bin is durably written, so a crash never leaves zero
+/// valid bins. Remove errors are ignored (a stray extra bin is harmless).
+fn gt_bin_prune(base: &str, keep: usize) {
```
ekechinwokah, line 392: gt_bin_latest and gt_bin_prune duplicate the dir-scan + {stem}.M{digits}.bin parse — extract one for-each-bin(base) → (count, path) helper; two copies of filename-parsing logic is exactly the drift that bites later. (nit while you're here: the with_capacity header estimate in gt_bin_write is 48 but the header is 56 bytes — harmless, just wrong.)

ekechinwokah (commented):
nothing blocking. two items i'd still fix before leaning on the crash-recovery story (both silent-wrong-numbers class, in a tool whose whole job is trustworthy numbers) — inline at the sites. three smaller cleanups alongside. also: gt_bin_write/read/latest/prune are pure functions over a hand-rolled format — a tiny cfg(test) mod (round-trip, params-mismatch refusal, latest-selection over mixed filenames) would pin the format and there's precedent for test mods in bench-utils.

muralikpbhat:
Thanks for the thorough pass — all addressed in `d95330a` (reader reuse in `d4c2238`).

**The two correctness ones:**
- **n_cent from COUNT(*)** → resume now reads `n_cent` from the GT-bin **header** (provenance), not re-derived. No sidecar — the header already carried it, so this is the "trust the newest bin's header" option, not a new file. Verified with exactly your scenario: a run seeded for a 200k target (n_cent 1024) killed at the 60k checkpoint, resumed with no create-size hint — it logs `n_cent=1024 from GT bin header (provenance) — COUNT(*)=60000 alone would derive 64` and recall stays 0.990. (60k→120k passed only because run 1 completed, as you called out.)
- **params-mismatch → silent replay** → split into typed errors: `Corrupt` (io/magic/truncation) → replay; `Incompatible` (version/dim/k/seeds/n_queries) → **fatal**. A mismatched bin can no longer be silently rebuilt with the wrong params.

**Cleanups:**
- `RESUME_INGEST` without `EXISTING_PREFIX` now hard-fails instead of falling through to a fresh CREATE.
- Format version is checked; version + header-size (56, not 48) are named constants.
- `gt_bin_latest`/`gt_bin_prune` share one `for_each_bin` filename parse.
- Imports at the top; reuse `grading`'s cursor readers.
- Added the `cfg(test)` mod you suggested — round-trip, provenance survival, `Incompatible` refusal, corruption, latest-selection over mixed filenames.

benches/utils/recall_while_ingest.rs
muralikpbhat: Fixed in `d95330a` — resume now reads `n_cent` from the GT-bin **header** (provenance), not `COUNT(*)`. No sidecar: the header already stored it. Verified with exactly this scenario — create seeded for a 200k target (n_cent 1024), killed at the 60k checkpoint, resumed with no create-size hint: it logs `n_cent=1024 from GT bin header (provenance) — COUNT(*)=60000 alone would derive 64` and recall stays 0.990.

benches/utils/recall_while_ingest.rs
muralikpbhat: Fixed in `d95330a` — GT-bin load is now typed: `Corrupt` (io / bad magic / truncation) → replay rebuild; `Incompatible` (version / dim / k / seeds / n_queries) → **fatal**. A params-mismatch bin can no longer be silently rebuilt with the wrong params.

benches/utils/recall_while_ingest.rs
muralikpbhat: Fixed in `d95330a` — `RESUME_INGEST` set with a missing/empty `EXISTING_PREFIX` now `assert!`-fails instead of falling through to a fresh CREATE.

benches/utils/recall_while_ingest.rs
muralikpbhat: Fixed in `d95330a` — the `grading` reader import moved to the top, and `fs` / `io` / `path::{Path, PathBuf}` are top-level imports the new helpers use (no more function-local `use` or inline `std::` paths).

benches/utils/recall_while_ingest.rs
muralikpbhat: Fixed in `d95330a` — the version is checked (`!= GT_BIN_VERSION` → `Incompatible`), and both the version and the header size are named constants (`GT_BIN_VERSION`, `GT_BIN_HEADER_BYTES = 56`).

benches/utils/recall_while_ingest.rs
muralikpbhat: Fixed in `d95330a` — extracted `for_each_bin(base, f)` shared by `gt_bin_latest`/`gt_bin_prune` (one filename parse). Also corrected the write capacity estimate to `GT_BIN_HEADER_BYTES` (56, was 48).

ekechinwokah (approved):

PR #511, opened 2026-08-02
tests: fault surfacing, create-race adoption, drain tombstones, cold-read integrity

ekechinwokah:
## What

Tests only — no engine code changes (the one `src/` addition, `test_helpers::fault_storage`, is behind the `test-helpers` feature and never ships). Five focused additions pinning contracts nothing else tested:

- **Fault-injection battery** (`test_helpers::fault_storage` + `tests/supertable/storage/fault_surfacing.rs`) — a pass-through `StorageProvider` failing selected (operation, URI-fragment, count) rules with `TransientExhausted`. Pins: a superfile-PUT fault fails the commit and the same writer retries to success (buffered-rows contract); manifest-list-PUT / pointer-CAS faults surface as storage errors and are never mislabeled as write contention; pointer/list GET faults fail `open` cleanly and a reopen recovers; a cold fetch whose reads fail errors instead of handing out a reader, then serves correct BM25 once cleared; GC reports failed deletes and reclaims on the next healthy sweep. The raw `object_store` handle is deliberately withheld so nothing bypasses the checked surface.
- **Lost-create-race adoption** — masks the loser's first pointer probe to deterministically reproduce the probe→publish window: the loser collides at the initial pointer PUT, adopts the winner's manifest, and reconciles its hidden vector-index handle to the stamped prefix (a regression silently splits the hidden index across two prefixes).
- **Drain excludes tombstoned vectors** — asserts on the hidden table's own doc count after a delete-then-drain, so build-time exclusion is pinned where query-time tombstone filtering can't mask a leak. The drain's tombstone-resolution block was previously dead across the whole suite (verified by line-level coverage).
- **Cold-read row integrity** — projection searches through lazy readers across superfile boundaries: every projected cell must belong to its hit row (BM25 single/multi-term OR, unranked matches, vector kNN).
- **Catalog wrapper seam + metering** — the public `Supertable` delegation layer (schema/searches/optimize/gc/SQL/drop-purge), LRU eviction contract, object-store billing lifecycle, IO-ledger trace window. (First commit here is the same content riding PR #509; it falls out on whichever merges second.)

## Findings worth keeping

- Small tables are answered **entirely from manifest parts** — zero superfile reads at query time. A supertable-level "cold search" test cannot exercise data reads at small scale; the cold-read fault test therefore lives at the disk-cache layer where the bytes actually flow.
- llvm-cov's per-instantiation JSON over-reports missed closures; the per-file summary table and line-level `report --text` are the only numbers to trust. Several "uncovered clusters" targeted early in this work (cold byte-source closures, drain checkpoints) turned out to be already covered.

## Verified

Full suite 2509/0 with the branch's tests; `cargo llvm-cov` functions 89.6% (the gate margin itself rides in #509); clippy `-D warnings` and the CI-flagged rustfmt both clean. Remaining uncovered functions decompose into CPU-gated SIMD variants, cloud-backend code that needs daemons in the coverage job, deliberately `#[ignore]`d diagnostics/mocks, and legacy-format shims — no honestly coverable cluster remains, which is why this branch stops here.

### DIFF
--- a/src/test_helpers/fault_storage.rs
+++ b/src/test_helpers/fault_storage.rs
@@ -1,0 +1,220 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Infino Authors
+
+//! Fault-injecting [`StorageProvider`] wrapper for error-path tests.
+//!
+//! Wraps a real provider and fails selected operations with
+//! [`StorageError::TransientExhausted`] — the "provider retries already
+//! exhausted, caller must surface" shape — so tests can assert that a
+//! storage failure mid-commit / mid-load / mid-query surfaces as a clean
+//! error and never as a panic, a wrong result, or a mislabeled
+//! write-contention retry loop.
+//!
+//! Rules are armed per (operation, URI fragment, count) and burn down as
+//! they fire, so a test can fail exactly the first N matching calls and
+//! then watch the same code path recover. Operations without an armed
+//! rule pass through untouched, including listing and the
+//! `object_store` handle, so the wrapped provider behaves identically to
+//! the inner one everywhere a test isn't deliberately failing it.
+
+use std::{
+    io::Error as IoError,
+    ops::Range,
+    sync::{
+        Arc, Mutex,
+        atomic::{AtomicUsize, Ordering},
+    },
+};
+
+use async_trait::async_trait;
+use bytes::Bytes;
+
+use crate::{
+    runtime_metrics::io::UsageMeter,
+    storage::{ObjectMeta, StorageError, StorageProvider},
+};
+
+/// Storage operation a fault rule can target.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FaultOp {
+    Head,
+    Get,
+    GetRange,
+    PutAtomic,
+    PutIfMatch,
+    Delete,
+}
+
+/// One armed fault: the next `remaining` calls of `op` whose URI
+/// contains `uri_fragment` fail.
+#[derive(Debug)]
+struct FaultRule {
+    op: FaultOp,
+    uri_fragment: String,
+    remaining: usize,
+}
+
+/// See the module docs. Construct with [`FaultStorage::wrap`], arm with
+/// [`FaultStorage::fail`], and pass the `Arc` anywhere a
+/// `Arc<dyn StorageProvider>` goes.
+#[derive(Debug)]
+pub struct FaultStorage {
+    inner: Arc<dyn StorageProvider>,
+    rules: Mutex<Vec<FaultRule>>,
+    fired: AtomicUsize,
+}
+
+impl FaultStorage {
+    pub fn wrap(inner: Arc<dyn StorageProvider>) -> Arc<Self> {
+        Arc::new(Self {
+            inner,
+            rules: Mutex::new(Vec::new()),
+            fired: AtomicUsize::new(0),
+        })
+    }
+
+    /// Arm a rule: the next `times` calls of `op` whose URI contains
+    /// `uri_fragment` fail with [`StorageError::TransientExhausted`].
+    pub fn fail(&self, op: FaultOp, uri_fragment: &str, times: usize) {
+        self.rules.lock().expect("fault rules").push(FaultRule {
+            op,
+            uri_fragment: uri_fragment.to_string(),
+            remaining: times,
+        });
+    }
+
+    /// Disarm every remaining rule.
+    pub fn clear(&self) {
+        self.rules.lock().expect("fault rules").clear();
+    }
+
+    /// Total faults fired since construction — lets a test assert the
+    /// failure it observed actually came from the injection.
+    pub fn fired(&self) -> usize {
+        self.fired.load(Ordering::SeqCst)
+    }
+
+    fn check(&self, op: FaultOp, uri: &str) -> Result<(), StorageError> {
+        let mut rules = self.rules.lock().expect("fault rules");
+        for rule in rules.iter_mut() {
+            if rule.op == op && rule.remaining > 0 && uri.contains(&rule.uri_fragment) {
+                rule.remaining -= 1;
+                self.fired.fetch_add(1, Ordering::SeqCst);
+                return Err(StorageError::TransientExhausted {
+                    uri: uri.to_string(),
+                    source: Box::new(IoError::other("injected fault")),
+                });
+            }
+        }
+        Ok(())
+    }
+}
+
+#[async_trait]
+impl StorageProvider for FaultStorage {
+    async fn head(&self, uri: &str) -> Result<ObjectMeta, StorageError> {
+        self.check(FaultOp::Head, uri)?;
+        self.inner.head(uri).await
+    }
+    async fn get(&self, uri: &str) -> Result<(Bytes, ObjectMeta), StorageError> {
+        self.check(FaultOp::Get, uri)?;
+        self.inner.get(uri).await
+    }
+    async fn get_range(&self, uri: &str, range: Range<u64>) -> Result<Bytes, StorageError> {
+        self.check(FaultOp::GetRange, uri)?;
+        self.inner.get_range(uri, range).await
+    }
+    async fn put_atomic(&self, uri: &str, bytes: Bytes) -> Result<Option<String>, StorageError> {
+        self.check(FaultOp::PutAtomic, uri)?;
+        self.inner.put_atomic(uri, bytes).await
+    }
+    async fn put_if_match(
+        &self,
+        uri: &str,
+        bytes: Bytes,
+        expected_etag: Option<&str>,
+    ) -> Result<Option<String>, StorageError> {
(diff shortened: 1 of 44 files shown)


src/test_helpers/fault_storage.rs
```diff
@@ -0,0 +1,220 @@
...
+        for rule in rules.iter_mut() {
+            if rule.op == op && rule.remaining > 0 && uri.contains(&rule.uri_fragment) {
+                rule.remaining -= 1;
+                self.fired.fetch_add(1, Ordering::SeqCst);
+                return Err(StorageError::TransientExhausted {
+                    uri: uri.to_string(),
+                    source: Box::new(IoError::other("injected fault")),
+                });
+            }
+        }
```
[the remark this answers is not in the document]
ekechinwokah: taken in 613389f — one poison-tolerant guard (into_inner) shared by arm/clear/check, since the helper's whole contract is failures-as-clean-errors and a panic elsewhere shouldn't turn every later storage call into a second panic. state is safe to reuse: mutations under the lock are single-field writes.

ekechinwokah:
on the reviewer-guide items: (1) 'wrong import / will not compile' — it compiles (supertable::storage is a re-export the existing crash tests also use; the branch's full suite ran 2509/0), but the consistency point was fair: fault_surfacing.rs now imports from the top-level storage module like the branch's other files, in 613389f. (2) 'broad matching' — substring rules are deliberate: arming a whole namespace (data/, a hidden prefix) is the feature, single-object rules pass a full path, and the repo's fixed-width zero-padded object names mean a full path never substring-matches a sibling (manifest-000001 is not a prefix of manifest-000010). documented on fail() in 613389f rather than changed. (3) the #509 ticket-compliance table is a misparse — this pr references #509 only because its first commit rides both branches (noted in the body); it doesn't implement that ticket.

tests/supertable/commit/create_adopts_existing.rs
```diff
@@ -0,0 +1,241 @@
...
+        loser.manifest_id(),
+        1,
+        "the adopted view is the winner's committed manifest"
+    );
+    assert_eq!(
+        loser.reader().expect("reader").n_docs_total(),
+        TITLES.len() as u64
+    );
+
+    // The adopted handle serves vector search over the winner's data —
```
muralikpbhat: **Headline coverage is illusory — proven empirically.** This block (and the module doc's claim that it proves 'the hidden index it reconciled to is the winner's, not the loser's orphaned pre-built one') does NOT guard hidden-index reconciliation. No drain runs, so the hidden cells are empty; the vector query is served entirely from the **undrained user superfiles** referenced by the correctly-adopted USER manifest (query/vector.rs get_undrained_superfiles_loaded). `reconcile_vector_index_table_to_manifest` only affects the hidden-handle prefix, which no assertion here observes. A finder edited `create_table_async` to keep the loser's orphaned pre-built hidden handle (the exact two-prefix-split regression the doc describes) and **this test still passed (exit 0)** — the reconciliation could be deleted entirely and it stays green. To actually guard it: drain vectors into cells via the adopted loser handle, then assert a fresh open reads the winner's stamped prefix — or assert `loser.vector_index_table()`'s storage prefix equals the winner's. As written the `titles.value(0) == TITLES[1]` ranking assert is also redundant with the `manifest_id`/`n_docs_total` guards above (both exercise only user-manifest adoption).

tests/supertable/storage/fault_surfacing.rs
```diff
@@ -0,0 +1,281 @@
...
+    w.append(&build_title_batch(&["second commit beta"]))
+        .expect("append");
+    let err = w.commit().expect_err("list PUT fault must fail the commit");
+    // A transient list-PUT failure is a storage error. Only
+    // PreconditionFailed (a lost race) may be folded into the
+    // write-contention retry story; masking real faults as contention
+    // would burn the whole retry budget re-hitting a broken store.
+    let rendered = format!("{err}");
+    assert!(
+        !rendered.contains("contention"),
```
muralikpbhat: Nit (non-blocking): this manifest-list-PUT test (and `commit_surfaces_pointer_cas_fault` ~128) asserts only `!contains("contention")`, so a false-pass where the commit failed for some *other* non-contention reason (fault not firing, a different storage error) is ruled out only by the recovery half. The mislabel contract itself is genuinely guarded (an injected `TransientExhausted` folded into `WriteContentionExhausted` WOULD contain 'contention' and trip this), so it's not a hole — but the superfile-PUT test above already asserts `faults.fired()==1` + `contains("injected")` (lines 77-80); mirroring that here ties the surfaced error to the injection directly and removes the spurious-cause window.

muralikpbhat (commented):
Reviewed (tests-only, additive +1204/-0, so no production risk — the review is whether the tests are correct, meaningful, and actually run). **Three of the four areas are solid; one has an illusory headline guarantee.** Both finders verified contracts against engine source, not just test text.

**Solid:**
- **Fault battery** — correct and its assertions are real. The `fault_storage` helper delegates every non-faulted call unchanged, has exact-N count semantics, genuinely closes the raw-`object_store` bypass (returns `None` → forces parquet/SQL readers onto the checked methods), and is thread-safe (no Mutex-across-await). The "storage error not mislabeled as `WriteContentionExhausted`" guard is genuine — traced: only `PreconditionFailed` maps to contention, an injected `TransientExhausted` stays `Storage`. URI fragments are precisely targeted (`data/` = superfiles, `manifest/manifest-` = list). The cold-reader and GC tests assert both the dangerous case (no reader handed out / failed delete reported) *and* correctness-after-clear, tied to `fired()`.
- **drain_tombstones** — real: `n_docs_total == survivors` is a physical build-layer count that query-time filtering can't mask, directly guarding the drain tombstone skip (writer.rs:581).
- **cold_projection** — real: genuinely cold (fresh consumer + empty cache dir), deterministic, with a per-row `rating == expected_rating(title)` integrity oracle that crosses a superfile boundary.
- **Feature-gating & CI wiring** — correct: `test_helpers`/`fault_storage` are `#[cfg(any(test, feature = "test-helpers"))]` (never ship), CI runs `--features test-helpers`, and all four tests are wired into the `supertable` binary (none wired-out).

**The one material issue (inline):** `create_adopts_existing`'s advertised *hidden-index reconciliation* coverage is illusory — a finder **proved empirically** that neutering `reconcile_vector_index_table_to_manifest` (keeping the loser's orphaned hidden prefix — the exact two-prefix-split regression the module doc claims to guard) leaves the test GREEN. The manifest-adoption guards (`manifest_id==1`, `n_docs_total`) are real; only the headline hidden-index claim is unguarded. Worth fixing before merge given the extensive doc implying it's covered.

Minor nits (non-blocking): three commit/open fault tests lean on the recovery half instead of asserting `fired()`/`contains("injected")` directly (the superfile-PUT test already does it right — mirror that); the superfile-PUT recovery asserts structure (`n_superfiles==2`) not content (never searches for 'beta'); the `fault_storage` doc says "behaves identically" but `get_if_none_match`/`tail` aren't overridden (fine for LocalFs, caveat for S3/Azure); and the GC test's `Duration::ZERO` leans on wall-clock ordering (extremely low flake risk).

tests/supertable/commit/create_adopts_existing.rs
ekechinwokah: correct. taken in 613389f: the guard now routes vectors THROUGH the adopted handle's hidden machinery (a commit's dual-write + an explicit drain) and asserts a FRESH handle — which resolves its hidden prefix from the stamped manifest alone — counts them. verified by your method: with the reconciliation sabotaged to keep the loser's pre-built handle, the test fails on exactly that assertion; intact, it passes. the redundant ranking assert died in the rewrite, and the module doc now says what the test actually proves (user-side queries can't stand in for the hidden index — undrained searches are served from user superfiles).

tests/supertable/storage/fault_surfacing.rs
ekechinwokah: taken in 613389f — both the list-PUT and CAS tests now assert fired() == 1 and that the surfaced error carries the injected marker, mirroring the superfile-PUT test, so the not-mislabeled-as-contention check can't false-pass on a fault that never fired or an unrelated storage error.

muralikpbhat (approved):
Approving — the material finding is reconciled and I confirmed it empirically at HEAD (41bd4f4).

- **create_adopts_existing now actually guards hidden-index reconciliation.** The test was rewritten to drain vectors *through the adopted loser handle* and assert a FRESH handle reads all 6 from the manifest-stamped hidden prefix. Re-running my sabotage (neuter `reconcile_vector_index_table_to_manifest`) now FAILS on exactly that assertion (`left: 0, right: 6` — the loser drained into its orphaned prefix), where the old test passed green; intact it passes (5/5 stable). The redundant ranking assert is gone and the module doc now states what the test actually proves.
- **Fault tests tied to the injection** — the list-PUT and pointer-CAS tests now assert `fired()==1` + `contains("injected")`, mirroring the superfile-PUT test, so the not-mislabeled-as-contention check can't false-pass.
- Poison-tolerant fault-storage guard + import consistency in place.

Full new test set 12/12, no flakiness, feature-gated so nothing ships. Good, thorough test PR.

PR #516, opened 2026-08-03
catalog: make read consistency configurable via ConnectOptions

vinaykakade:
## Problem

The catalog **forced `Consistency::Strong`** on every table handle it built — both `create_table` and `open_table` unconditionally called `with_read_consistency(Consistency::Strong)`, overriding the engine's documented default of `BoundedStaleness(1s)`. So the `BoundedStaleness` default never took effect for any `connect`-based caller, and there was **no way for a connection to choose** its consistency.

`Strong` re-checks the manifest pointer on **every** query (one conditional pointer GET per read). `BoundedStaleness` amortizes that check across the window, which matters for hot read workloads (and per-request GET cost on object stores).

## Change

- Add `ConnectOptions::with_read_consistency(Consistency)` and thread the connection's policy through `create_table` / `open_table` instead of hardcoding `Strong`.
- Re-export `Consistency` at the crate root so callers can name it.
- **Default (unset) is `BoundedStaleness(1s)`** — matching the engine default. A connection opts into `Strong` only when it needs a per-query pointer re-check.

**Backwards compatibility:** purely additive to the public surface (`public-api.txt` diff has zero removals — one new re-exported enum + one new builder). The new struct field is `pub(crate)` and derives `Default`, so existing construction is unchanged. The **only** observable change is the default policy for catalog-built handles; anyone who wants the old behavior calls `.with_read_consistency(Consistency::Strong)`.

## Mutation safety under bounded staleness

Making `BoundedStaleness` the default exposed a latent assumption in the mutation path. An `update` / `delete` resolves its predicate to the set of `_id`s it will tombstone, and that resolution read the writer handle's current snapshot — whose freshness follows the connection's read consistency. Under `Strong` (the old forced default) that snapshot was always re-checked, so it never mattered; under `BoundedStaleness` a mutation could resolve against a snapshot **behind a peer's commit**, miss a row committed after it, and silently drop that row's tombstone — a delete that leaves the row, or an update that leaves the old version live beside its replacement (a duplicate row).

Fix: a mutation's target set must agree with the manifest its commit will land on, so `update` / `delete` now **force the snapshot to the latest committed manifest before resolving, independent of read consistency** (a failed pointer probe is surfaced, not resolved against a stale snapshot). Bounded staleness remains a read-latency contract for *queries*; it must never cause a write to lose data. Added a regression test: under `BoundedStaleness` with a long window, a handle deletes a row a peer committed after the handle's last read — with the fix the delete resolves against the latest manifest and tombstones it; without it the delete matches zero rows.

## Tests

- `ConnectOptions` default is `BoundedStaleness(1s)`; `with_read_consistency` overrides it.
- The connection's policy reaches both created and freshly-opened table handles.
- A default `BoundedStaleness` read is **stale within its 1s window and converges after it elapses**.
- The cross-connection freshness tests that assert *immediate* convergence (peer-commit visibility, purge detection) now opt into `Strong` explicitly — preserving that coverage.
- New: a mutation resolves its target set against a peer's commit under `BoundedStaleness` (fails without the fix).

Full lib suite (1116 supertable unit + rest) + integration suites (superfile/supertable/concurrent-process, 227) + `fmt` (CI config) + `clippy --all-targets --features test-helpers -D warnings` + `make public-api` all green.

### DIFF
--- a/src/supertable/handle.rs
+++ b/src/supertable/handle.rs
@@ -6665,6 +6665,86 @@ mod tests {
         assert_eq!(bm25_title_hits(&verifier, "keep"), 1, "keep survives");
     }
 
+    /// The update-path companion to the delete regression above. An update
+    /// resolves a 1:1 target set the same way, so under bounded staleness a stale
+    /// resolve would fail to see a row a peer committed after the handle's last
+    /// read — matching zero rows and failing the cardinality check — rather than
+    /// replacing it. With the target resolved against the latest manifest, the
+    /// update matches and swaps the row.
+    #[test]
+    fn an_update_resolves_its_target_against_a_peers_commit_under_bounded_staleness() {
+        use datafusion::prelude::{col, lit};
+
+        let dir = TempDir::new().expect("tempdir");
+        let inner: Arc<dyn StorageProvider> =
+            Arc::new(LocalFsStorageProvider::new(dir.path()).expect("provider"));
+
+        // Producer creates the table and commits a baseline row.
+        let producer = Supertable::create(opts().with_storage(Arc::clone(&inner))).expect("create");
+        let mut w = producer.writer().expect("writer");
+        w.append(&title_batch(&["keep"])).expect("append keep");
+        w.commit().expect("commit keep");
+        drop(w);
+
+        // Consumer opens under BoundedStaleness with a window long enough that a
+        // plain read never re-probes for the rest of the test.
+        let consumer = Supertable::open(
+            opts()
+                .with_storage(Arc::clone(&inner))
+                .with_read_consistency(Consistency::BoundedStaleness(Duration::from_secs(3600))),
+        )
+        .expect("open");
+        // One read pins the snapshot and opens the window.
+        assert_eq!(
+            bm25_title_hits(&consumer, "keep"),
+            1,
+            "sees the initial row"
+        );
+
+        // Producer commits the update target the consumer has NOT yet observed.
+        let mut w = producer.writer().expect("writer");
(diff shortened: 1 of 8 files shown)


src/supertable/handle.rs
```diff
@@ -7986,6 +6594,77 @@ mod tests {
...
 
+    /// Regression: a mutation must resolve its target set against the latest
+    /// committed manifest even under `BoundedStaleness`, where a plain read
+    /// would serve a snapshot behind a peer's commit. Resolving the predicate
+    /// against such a snapshot would miss the peer-committed row and silently
+    /// drop its tombstone — a lost delete, or (for update) the old version left
+    /// live beside its replacement.
+    #[test]
+    fn a_mutation_resolves_its_target_against_a_peers_commit_under_bounded_staleness() {
+        use datafusion::prelude::{col, lit};
```
naman-modi, line 6605: We have tested `delete` here, can we also write a similar test to test `update`?

naman-modi (approved):
LGTM, minor comment to add a test.

src/supertable/handle.rs
vinaykakade: Good call — added `an_update_resolves_its_target_against_a_peers_commit_under_bounded_staleness` in 1fa83a8a. Same two-handle setup: the consumer updates a row the producer committed after the consumer's last read, so a stale resolve matches zero rows and fails the 1:1 cardinality check; with the target resolved against the latest manifest the update matches and swaps the row. Verified it fails without the fix.

PR #519, opened 2026-08-03
perf(vector): coarse-router for large-k centroid assignment

muralikpbhat:
## Why

`assign_to_centroids` (nearest-centroid assignment; used by index build and the hidden-index drain) is an **exact scan over all `k` cells — O(rows·k)**. On a large IVF grid (thousands of cells, e.g. grown by cell-splits at scale) that scan dominates a drain, even though the per-distance kernel is already SIMD + rayon-parallel.

## What

Above `k = 1024`, route through an **ephemeral in-memory two-level index**, built per call and **never persisted** (no on-disk format/version/compat impact):

1. cluster the `k` centroids into `m = ⌈√k⌉` **super-centroids**;
2. group cells by super into **per-super block-transposed caches**;
3. per row: top-`4` supers via the SIMD `nearest_k_centroids_transposed`, then the **exact** nearest centroid among only the cells beneath them.

`k < 1024` (bootstrap / small grids) keeps the exact path **unchanged**.

## Speed

Assign micro-bench, L2Sq, coarse forced across k (200k × 1024-dim), and 500k × 3587 (two machines):

| k | exact | coarse | speedup |
|---|------:|-------:|--------:|
| 256 | 0.41 s | 0.30 s | 1.36× |
| 1024 | 1.65 s | 0.50 s | 3.29× |
| 3587 | 6.53 s | 1.54 s | 4.24× |
| 3587 (500k rows) | 17–18 s | 3.7–5.2 s | **3.6–4.6×** |

The gate is 1024 because below it the win is modest (≤1.8×) and the exact path is already cheap.

## Recall (⚠️ approximate — impact under measurement)

This is a **pre-filter**: it takes the exact nearest among the top-4 supers' cells, so it can misroute a row whose true-nearest cell sits under an unprobed super.

- Synthetic clustered data: **100% match** to the exact assignment.
- **Real Cohere-768 (unit-normalized): ~95–98% match** — 1.7% misrouted at k=256, rising to 4.7% at k=2048.

A misroute sends a row to a *near* (2nd-best) cell, which the query's adaptive width often still probes — so recall loss is likely ≪ the misroute rate, but that is **not yet confirmed**. An end-to-end coarse-drain vs exact-drain recall A/B on Cohere is in progress; **not marking ready until recall is verified**. If recall regresses, the lever is a higher `COARSE_SUPER_NPROBE` or making this drain-only opt-in.

## Tests

- `coarse_assign_tracks_exact_at_large_k` — matches exact on clustered data.
- `bench_coarse_vs_exact_assign_at_scale`, `bench_coarse_vs_exact_sweep_k` — `#[ignore]`d timing benches.

Draft.

### DIFF
--- a/src/superfile/vector/kmeans.rs
+++ b/src/superfile/vector/kmeans.rs
@@ -31,7 +31,7 @@ use rayon::prelude::*;
 
 use crate::superfile::vector::distance::{
     Metric, add_f32_to_f64_acc, f64_acc_mean_into_f32, nearest_centroid_transposed,
-    transpose_centroids_cluster_major,
+    nearest_k_centroids_transposed, transpose_centroids_cluster_major,
 };
 
 /// Offset added to a column's `rot_seed` to seed k-means. Keeps the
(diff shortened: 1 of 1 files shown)


muralikpbhat:
Closing unmerged.

The coarse router speeds large-k centroid assignment when measured in isolation (the micro-benchmark showed a multiple), but a large-scale measurement told a different story: at production grid sizes the assignment step is dominated by I/O, not the distance compute, so the end-to-end time improved only ~12% — nowhere near the isolated speedup. And the approximate routing costs recall at exactly the large grids where it engages: the misplaced rows are not fully recovered by the query's adaptive probe width, which covers a shrinking fraction of cells as the grid grows.

The compute was not the bottleneck at scale, so this is not the right lever. Keeping the branch `perf-coarse-drain-assign` for reference.

Issue #512, opened 2026-08-02
Cosine rerank codec silently clamps non-unit input (−9.6 pts recall on Cohere-1M) — decide: normalize vs. reject


muralikpbhat:
## Summary
The default cosine rerank codec (`Sq8FixedResidual`) quantizes onto a fixed `[−1,1]` grid and **silently clamps** any component outside it to ±1 (`distance.rs`, the Sq8 fixed-grid encoder). That is correct only when inputs are unit-normalized; the engine does not normalize, and the bundled Python binding passes vectors through unchanged. Un-normalized cosine input silently corrupts the highest-magnitude dimensions → recall loss with no error. **Measured on Cohere-1M: recall@10 `0.9925 → 0.8965` (−9.6 pts) with no error, warning, or crash — see [Evidence](#evidence).**

**The silent clamp is not in question — it must go. This issue is to decide what replaces it.**

## Background — why the behavior changed
The previous per-dim codec (`Sq8Residual`) tolerated un-normalized input two ways: a **measured** per-dim scale (fit the real range, so no clamp) and stored **per-doc norms** that divide out `‖d‖` at score time, giving correct cosine at any input magnitude. The fixed grid dropped both — fixed `[−1,1]` scale, and the per-doc norm can no longer rescue it because corruption now happens at **encode**, before any norm applies — in exchange for **portability**: no per-cluster metadata, so the bytes carry verbatim through drain, compaction, and split.

So the engine *used* to handle un-normalized cosine input correctly; the portable fixed grid gave that up.

## The invariant that scopes the decision
- Only the **stored** (document) vectors need consistent unit norm.
- The **query is ranking-neutral** — the cosine kernel divides by the per-doc norm, not the query norm, so query magnitude cannot reorder top-k, and the fp32 query never touches the grid or clamp.
- The **engine owns both ingest and search**, so there is no client-side ingest/query coordination to get wrong.

**Consequence: the corruption happens at ingest-encode, and only stored vectors need fixing. So the A-vs-B decision below is a choice about the *ingest / corpus path only.*** The query path is settled separately (see *Query path*).

## The decision (ingest / corpus path)

**Option A — Engine normalizes the corpus at ingest (treat as a bug fix).**
Restore the robustness the engine previously had by normalizing cosine vectors on the ingest path (fused into the encode pass). Keep the clamp only as the fp-noise guard it was documented to be.
[...]

Issue #515, opened 2026-08-03
Bioasq-1M recall gap at default width (0.94): adaptive probe width under-provisions on concentrated/diffuse embeddings


muralikpbhat:
## Summary

On **Bioasq-1M** (1024-d, cosine), recall@10 at the default adaptive probe width is **0.9435** — well below the other real-world cosine datasets (Cohere-1M 0.9928, OpenAI-500K 0.9932) under identical config (k=10, n_cent=256, unit-normalized input). It is **not** a codec or rerank-depth problem — it's cell-routing: the adaptive probe width under-provisions the number of IVF cells searched for this dataset's geometry.

## Evidence — nprobe × rerank_mult sweep (same ingested table, search-only)

| nprobe | rerank_mult | recall@10 |
|---|---|---|
| adaptive (default) | 256 | 0.9435 |
| adaptive (default) | 1024 | 0.9435 |
| 16 | 256 | 0.9024 |
| 64 | 256 | 0.9791 |
| **256 (all cells)** | 256 | **0.9997** |
| 256 (all cells) | 1024 | 0.9998 |

Two conclusions fall straight out:

1. **Not rerank-depth / codec.** Increasing `rerank_mult` 256→1024 at the default width changes nothing (0.9435 → 0.9435). Candidates aren't being lost at rerank.
2. **It's routing/width.** Recall rises monotonically with `nprobe` and reaches **0.9997** when all 256 cells are probed. The true neighbors are perfectly rankable — they simply sit in cells the adaptive width doesn't probe.

## Root cause

Bioasq's embedding components are tightly concentrated (all within ~[−0.21, 0.21] at 1024-d), so its clusters are far less separable than Cohere/OpenAI, and a query's top-10 scatter across **many** IVF cells. The drain-stamped adaptive width — calibrated on more-separable corpora — probes too few cells for this geometry. For reference, Cohere reaches 0.99 at ~29/256 cells; Bioasq needs somewhere between 64 and 256.

## Fix direction

Widen the adaptive probe-width calibration for diffuse-neighbor datasets — provision the **minimal** nprobe for ~0.99 (likely ~128 here, between the 64→0.979 and 256→0.9997 points), not "probe all cells" (which negates the IVF speedup). This is a probe-width calibration issue, not a codec change.

Refs #490 (real-world dataset baselines).

muralikpbhat:
Narrowed the minimal probe width for ~0.99 (same table, search-only):

| nprobe | recall@10 |
|---|---|
[...]

PR #520, opened 2026-08-03
supertable/vector: serve the stamped laws; scope the floor; deterministic drain

ekechinwokah:
## Summary

Fixes gaps in #512 and #494, serving as a follow-up to #505/#497.

The no-knobs default now serves the low latency the code already had implemented. Merged main measured defaults at 37.3 ms (k=100) / 12.9 ms (k=10) on Cohere-1M while the same binary served 11.4/3.9 under hand tuning — this PR closes that gap layer by layer, each layer measured on the same tables and protocol:

1. **Serve the rerank law** (`aad09da`). Since the #505 merge
   restructure the law's `options` rebind lived inside the admit arm's
   brace, so its shadow died at the brace and every law-stamped table
   silently served the rm=256 constant (fingerprint: default == rm=256
   to the decimal at both k). Also: per-knot pool provenance (a scalar
   min-rule let one surviving k=1 point invalidate fresh neighbors)
   with an optimize-time repair trigger, and the #512 fix — cosine
   vectors normalize once at `VectorBuilder::add` (raw Cohere input
   was silently clamping at the Sq8 fixed grid: −9.6 pts recall@10).
   37.3/12.9 → 28.1 @ 0.9967 / 6.9 @ 0.9940.
2. **Width walk targets the 0.99 bar, not the 0.997 stage pad**
   (`82b8b38`). The pad stamped width ~97 on Cohere where ~53 clears
   the bar — a 2× width tax on every query. → 19.8 @ 0.9938 /
   4.8 @ 0.9900.
3. **Hash-free per-cell floor** (`a09b39f`). The floor's guarantee was
   costed by SipHash bookkeeping over the ~340K-candidate tail
   (~7 ms), not by the rescued rows (~1.5 ms). The pool arrives with
   each cell's candidates contiguous, so the floor now selects each
   group's top-k in place with one boundary walk before the global
   cut; kept set oracle-pinned identical. → 14.8 @ 0.9938 /
   4.7 @ 0.9900.
4. **Floor scoped to explicit-nprobe sweeps** (`0d001e7`). The floor
   exists for the #494 inversion — caller-widened sweeps — so it now
   guards exactly those; law-served defaults are calibrated against
   realized recall at their stamped width and run floor-free (the
   floor measured +0.0001 recall for ~3.4 ms at law widths).
   → 11.4 @ 0.9937 / 3.9 @ 0.9890 on the dev table.
5. **Deterministic drain row order** (`62ddb25`). The batch-open
   fan-out's completion order scrambled rows into calibration
   reservoirs and fine sub-builds (two identical-corpus drains stamped
   fine [5,6,6,6] vs [7,7,7,7]); `buffered()` at the collect barrier
   is zero wall cost and also fixes the tombstone-bitmap positional
   pairing. Post-compact recall stops being drain dice.
6. **Sq8 transcode saturation tripwire** (`a226dca`). Exact count of
   components clamping at the destination grid during merge/split
   transcode, reported at drain and compact (zero on the standard
   corpus, asserted by eye in gates).

Bench-only: predicate battery bucket count 500 → 499 (coprime with the
power-of-two mode grid; gcd=4 aliasing left 3 of 4 queries with zero
same-mode matches, so the battery graded the generator, not the
engine) and the ambiguous cold columns relabeled against the metered
split.

## Numbers

**Cohere-1M/768d cosine, fresh-built table, current-code stamps,
no-knobs defaults (sweep protocol, 100q, fresh cache + 60 s idle):**

| | k=100 | k=10 |
|---|---|---|
| main (bdf487a, defaults) | 0.9989 @ 37.3 ms | 0.9970 @ 12.9 ms |
| this branch (fresh stamps) | **0.9931 @ 11.7 ms** | **0.9910 @ 4.0 ms** |

Dial rows bracket the default: rm=256 0.9951 @ 19.5 / 0.9940 @ 6.5;
rm=32 0.9826 @ 7.9 / 0.9780 @ 3.4.

**10M synthetic lifecycle (Azure, default config — the standard
bench):** recall@10 = 0.997 pre-drain / **0.994 post-drain** / 0.991
post-delta / 0.991 post-compact — acceptance bar recall@10 ≥ 0.99 met
at every lifecycle state. Steady cold: post-drain 3 GET / 150 ms,
post-compact 1 GET / 126 ms ($0.40/1M) — cost profile unchanged vs
main. Width sweeps monotone at both stages under #509's tripwire
(post-drain flat 0.996 at nprobe 2→256; post-compact 0.991→0.993 at
2→all). Zero transcode clamps. Optimize 1,373 s at 10M
(maintenance auto).

**Predicate battery (re-baselined by the bench fix):** 0.998 @ 0.2%
selectivity at 1M. Recorded pre-fix numbers (0.73–0.75) were the
aliasing artifact and are not comparable.

## Intentional trade-offs

- **Defaults run floor-free.** The per-cell floor's semantics are unchanged where it applies (explicit caller nprobe — the #494 surface); the default path drops it for ~3.4 ms at k=100, priced at +0.0001 recall at law widths. Callers that pin nprobe keep the full monotonicity guarantee.
- **Post-drain 0.994 vs the fp=8-era 0.996.** The depth law measures 5 where the old constant was 8, buying one fewer steady-cold GET (3 vs 4). Both clear the bar; the knob (`fine_nprobe_floor`) remains for user who wants the last 0.002 at the cold cost.

## Tests

Shortlist floor suite green incl. the oracle-pinned superset
assertion; end-to-end 10M gate green (log-verified); ASAN pass over
`superfile::vector` (278 tests, zero reports) covering the FastScan
LUT unsafe; clippy clean; `make public-api` no drift.


src/supertable/query/vector.rs
```diff
@@ -2881,6 +2895,30 @@ impl SupertableReader {
     }
 }
 
+/// Cosine queries normalize once at the public entry (#512): ranking is
+/// query-norm-invariant (the rerank kernel divides by the per-DOC norm),
+/// but a non-unit query scales every returned score by its own norm —
+/// normalizing keeps scores calibrated cosine. Other metrics pass
+/// through untouched; unit queries are a fp-noise no-op.
+fn calibrated_query<'q>(
```
muralikpbhat: **Query normalization is inconsistent across entry points.** `calibrated_query` (cosine query normalize) is applied only in `vector_search` (:2932) and `vector_hits` (:2969), but the **SQL/DataFusion exec path** and the **public allow-set / prepared async entries** (`vector_hits_global_allow_async`, `vector_hits_prepared_global_allow_async`) bypass it. So a cosine query via those paths is scored against unit-normalized docs with a **non-unit query** → every returned cosine score is scaled by the query norm. Top-k *ordering* survives (query-norm-invariant), so recall/ranking is fine — but the calibrated *scores* are wrong there, inconsistent with the two entries that normalize, and the #512 parity contract (score diff < 1e-3) isn't delivered on those paths. Normalize the query at a shared choke point (or in every entry) so all cosine paths agree.

src/supertable/opann.rs
```diff
@@ -729,13 +729,46 @@ pub(crate) const WIDTH_LAW_QUERY_SAMPLE: usize = 256;
...
+/// cell is a fetch — and the 0.997 pad measured a 2x width tax on
+/// real embeddings (Cohere-1M: width 97 vs 53 at k=100; -7 ms @ k=100
+/// and -2.3 ms @ k=10 at equal recall class when dropped to 0.99).
+/// The padded target exists for PER-STAGE COMPOUNDING, and width is
+/// where compounding is cheapest to absorb: the 10M gate at this
+/// split (2026-08-02, vec10m-width99.log) holds every lifecycle state
+/// at 0.991-0.997 — the historical "width at 0.99 straddles the bar"
+/// measurement predates the recalibration repair and the rerank law
+/// actually serving, both of which carry the margin now.
+const WIDTH_LAW_TARGET_COVERAGE: f64 = 0.99;
```
muralikpbhat, line 743: **Width retarget to 0.99 lowers the default recall *ceiling* — please sign off on the trade.** The width law now covers 0.99 of the true top-k (was 0.997). Since width is a hard ceiling (top-k in unprobed cells is unrecoverable), the three-stage compound is now ~0.99·0.997·0.997 ≈ **0.984**, below the **≥0.991** the 0.997 pad guaranteed even under tight crossings — and at k=1 the miss rate roughly triples. Combined with the default path going floor-free (see `vector.rs:1994`), the default query's guarantee shifts from *proven worst-case* to *measured on Cohere-1M/10M*. The ~2× width/latency win is real and likely worth it, but it's a deliberate weakening of the default guarantee, not a neutral optimization — worth an explicit reviewer decision + keeping the 10M recall gate.

src/supertable/query/vector.rs
```diff
@@ -1974,15 +1977,26 @@ impl SupertableReader {
...
+                // are calibrated against realized recall at their own
+                // stamped width and run floor-free: at law widths the
+                // floor measured +0.0001 recall for ~3.4 ms of extra
+                // rerank at k=100 (Cohere-1M), and at width 1-2 it is a
+                // near-no-op by construction. Floor = k when it applies:
+                // even if the entire true top-k concentrates in one probed
+                // cell, that cell's floor carries it into the exact rerank
+                // (replicas never collide inside one cell, so the floor
+                // needs no replica overhead).
+                let cell_floor = if options.nprobe.is_some() { k } else { 0 };
```
muralikpbhat, line 1996: **Default path is now floor-free (semantic change from #505).** `cell_floor` is now `k` only under an explicit caller `nprobe`, and `0` for law-served defaults — so a default query whose true top-k concentrates in one cell no longer gets the #505 per-cell rescue and relies entirely on the calibrated width+rerank law. Documented as +0.0001 recall / −3.4 ms, and the floor's original purpose (#494 caller-*widened* inversion) does only apply to caller-pinned widths — so this is defensible. But it does *not* preserve #505 behavior verbatim; pairs with the width retarget above as a default-guarantee change reviewers should sign off on.

src/supertable/query/vector.rs
```diff
@@ -1267,6 +1268,27 @@ impl SupertableReader {
...
+        // expressed as the equivalent multiplier so the divided cold
+        // budget and the global shortlist cap both inherit it through
+        // `resolve`. A caller-set `rerank_mult` wins over the law, exactly
+        // as a caller-set `nprobe` wins over the width law. MUST live at
+        // fn scope: an earlier form rebound `options` inside the admit
+        // arm, where the shadow died at the arm's brace — the phase-C
+        // budget then resolved the ORIGINAL options and the served
+        // default silently reverted to the constant (measured as
+        // default == rm=256 on a law-stamped table).
+        let options = match rerank_mult_from_law(
```
muralikpbhat, line 1284: **The fix is correct but has no regression test — add one before merge.** Moving this rebind to fn scope genuinely restores law-serving (verified: reaches both the cold-divide and phase-C budget). But nothing guards against a future re-shadowing: the existing tests assert only a recall floor (`near==k`), which is **insensitive** to this bug — reverting to the `rm=256` constant yields *more* survivors → recall equal-or-better; only latency regresses (37→20 ms), which no CI test measures. Add a test asserting the **effective served** rerank budget (resolved `rerank_mult` / global-shortlist limit) equals the *law* value, not the constant, on a fixture where law << constant.

src/supertable/writer.rs
```diff
@@ -3477,7 +3493,7 @@ pub(in crate::supertable) async fn drain_user_superfiles_to_hidden_cells(
                     Ok::<_, BuildError>((reader, stable_ids))
                 }
             }))
-            .buffer_unordered(read_concurrency)
+            .buffered(read_concurrency)
```
muralikpbhat, line 3503: **This `buffered` change also fixes a real correctness bug — call it out + test it.** Beyond determinism, switching `buffer_unordered` → `buffered` fixes a tombstone-mispairing bug: the readers stream was positionally zipped against `batch_sources` for tombstone pairing, so any drain with tombstones AND out-of-order read completion applied the *wrong* superfile's tombstone bitmap to a reader's rows (mis-dropped/kept rows). The fix is correct (input order preserved), but prior drains on tombstoned tables could have been affected — worth flagging in the PR description as a bug fix (not a perf-neutral refactor) and adding a regression test that asserts tombstone correctness under read-straggler ordering.

muralikpbhat (commented):
Reviewed across three angles with the highest-stakes items verified in code. **The correctness core is sound and this is a real, well-measured improvement** (default Cohere-1M latency 37.3/12.9 → ~20/4.8 ms). Verified:

- **Serve-the-law fix is genuinely fixed** — the rerank-law `options` rebind moved from inside the admit-arm brace (where the shadow died) up to fn scope (`vector.rs:1282`), and both consumers (cold divide `:1795`, phase-C global-shortlist budget `:1970`) now see it. Caller override preserved; width/fine application (keyed on `options.nprobe`) is behavior-preserving.
- **No double-normalize** — cosine normalize happens once at the lossless seam (`builder.rs:599`) and `normalize` is idempotent; zero-norm guarded; Cosine-gated; unit vectors fall inside the `[-1,1]` Sq8 grid (the −9.6 pts clamp fix).
- **Hash-free floor preserves the #505 guarantee** — the grouped in-place walk keeps each cell's top-k eviction-immune (brute-force oracle test: exact on grouped input, superset otherwise → recall-safe). Per-knot pool provenance is correct (a stale k=1 knot can no longer invalidate fresh neighbors).

Inline: one correctness gap, a recall-guarantee trade to sign off, and test/claim gaps.

### Recall-guarantee trade (needs explicit sign-off)
The width walk retargets to **0.99** (`opann.rs:743`) *and* the per-cell floor is now scoped to explicit-nprobe only, so the **default path runs floor-free** (`vector.rs:1994`). Together they move the default query from a **proven ≥0.991 worst case** (0.997 pad) to a **measured ~0.99** (three-stage compound ~0.99·0.997·0.997 ≈ 0.984), and drop the #494 inversion guard on the default path — bought for the latency win. Recall-safe by measurement on Cohere-1M/10M, but it's a deliberate semantic change to the *default* guarantee that reviewers should sign off on, not a pure perf refactor.

### Test/claim gaps
- The **headline serve-the-law fix has no regression test** (`vector.rs:1282`) and existing tests can't catch a re-shadowing (they assert a recall floor, which is *insensitive* — reverting to `rm=256` gives more survivors → recall equal-or-better; only latency regresses, unmeasured). One test to close before merge.
- **"Deterministic drain" is overstated** (`kmeans.rs:264`).
- The `buffered` change also **silently fixes a real tombstone bug** (`writer.rs:3496`) — worth surfacing + a regression test.

### Positives / notes
Most layers are well-guarded (the floor oracle, the raw-vs-unit cosine parity pinning the −9.6 pts, the caller-override test). Public hygiene clean. The bench `VECTOR_FILTER_BUCKET_TERMS 500→499` is itself a fix (500 shared a factor with the power-of-two mode grid). Minor edges: `normalize` admits subnormal→inf (`distance.rs:96`); the floor dedup identity key + grouped-contiguity are load-bearing-but-unenforced; no test that L2/NegDot are *not* normalized.

### Scope note
The `upstream/main` diff **conflates a separate PR** — `a8099b2` "propagate retryable CAS errors (#517)" (~10 files: error.rs, wal/*, fault_storage, mutations, wire.rs) — because the pinned base is stale. Those aren't #520's changes; review against a base excluding #517 (or land #517 first) so the CAS-error work isn't attributed here.

Not an approval — the query-normalization gap + the serve-the-law test gap warrant a pass, and the default-recall trade wants sign-off. The mechanism itself is correct.

src/supertable/query/vector.rs
ekechinwokah: taken in 79f84bd — the calibration now sits at every entry seam: the SQL exec arms (vector_search_async and vector_hits_filtered_by_plan), the hybrid path (hybrid_search_async), and the test-helpers allow-set entry, each chosen so every caller path crosses exactly one seam (the sync wrappers keep theirs; no path double-normalizes). calibrated_query also split out a pure metric-gated half with a unit test pinning your minor edge alongside: cosine normalizes, L2/NegDot pass through by reference, untouched.

src/supertable/opann.rs
ekechinwokah: signing off explicitly. what's traded: proven worst-case >=0.991 (0.997 pad + always-on floor) for measured ~0.99 — the compound-floor arithmetic assumes every stage lands on target simultaneously, while the stamps sit above target on every gated corpus and drains are now input-order deterministic so they stop wandering. what it buys: the pad stamped width ~97 on cohere where ~53 clears the bar (a 2x width/latency/GET tax on every default query) and the floor cost ~3.4ms for +0.0001 at law widths — together most of 19.8 -> 11.4ms. evidence: cohere-1m fresh-stamps defaults 0.9931 @ 11.7ms k=100 / 0.9910 @ 4.0ms k=10; synthetic 10m lifecycle 0.997/0.994/0.991/0.991 at 1-GET post-compact cold. the escape hatches stay proven and are now pinned by test (floor arms at k under explicit nprobe; caller rerank_mult overrides exactly). honest caveat, tracked: measured-not-proven leans on the calibration instrument, and #515 shows it under-provisions diffuse/asymmetric corpora — a live defect with a diagnosis attached, fixed through #515, and identical under the old pad (same sample, same instrument), so this trade neither causes nor masks it.

src/supertable/query/vector.rs
ekechinwokah: correct. a deliberate semantic change. pairs with the width sign-off above as one decision. the floor's original purpose is the caller-widened inversion, so it now guards exactly the widths a caller pins; defaults are calibrated against realized recall at their own stamped width and run floor-free for the measured +0.0001 / -3.4ms. the scoping is pinned by test in 79f84bd: the served-budget probe asserts floor == 0 on law-served defaults and floor == k under explicit nprobe.

src/supertable/query/vector.rs
ekechinwokah: taken in 79f84bd on your exact reasoning — recall floors cannot see the re-shadow because the constant only ADDS survivors. the pooled warm arm now records the effective shortlist limit and cell floor it hands to the global cut through a test-helpers probe (append-log, so assertions are race-tolerant under test parallelism), and vector_law_serving asserts three things: the default serves a law budget strictly below k*256, a caller rerank_mult overrides to an exact budget, and the floor arms only under explicit nprobe. a re-shadow regression makes every default record the constant, so the contains-checks fail deterministically rather than flakily.

src/supertable/writer.rs
ekechinwokah: called out + tested in 79f84bd. new drain_pairs_each_superfile_with_its_own_tombstones: three superfiles, distinct per-file tombstones, exact surviving membership asserted in both directions — under buffer_unordered it fails whenever completion order scrambles. the PR description now presents the buffered change as the correctness fix it is (prior drains of tombstoned tables under out-of-order read completion could apply one superfile's bitmap to another's rows), and your determinism-overstated note is folded into the same comment: input order is pinned, the fine k-means' rayon reduce still leaves centroid low bits scheduler-ordered — making that reduce deterministic is clustering code we would rather take deliberately in its own change than ride this PR.

muralikpbhat (approved):
Approving. Every round-1 item is addressed, and I re-verified each against the diff rather than the commit summary.

**Correctness — query normalization at every entry.** Traced all inner-scan callers: the SQL arms (`vector_search_async`, `vector_hits_filtered_by_plan`), the hybrid path, and the allow-set entry now each cross exactly one calibration seam. The one indirect caller (`route_filtered_vector_hits_async`) is fed by the already-calibrated filtered-by-plan entry, so no path reaches the scan with an un-normalized cosine query. The metric-gated split (`calibrated_query_for`) has a by-reference pass-through unit test pinning that non-cosine metrics are untouched.

**Serve-the-law regression guard.** `vector_law_serving` records the effective served shortlist budget through the probe and asserts it is the stamped law, not the constant: a re-shadowed `options` rebind reverts every default to `k * 256` and trips the `limit < k*256` assertion — precisely the recurrence recall floors cannot see. Caller-`rerank_mult` override precedence and the explicit-`nprobe` floor re-arm are pinned in the same test.

**Tombstone pairing.** `buffered` restores the positional `readers`↔`batch_sources` pairing; the three-superfile exact-membership test fails under the prior unordered completion, in both directions.

**Edges/claims.** `normalize` now leaves subnormal magnitudes alone (`is_normal`) with a test, so `1/mag` can no longer round to infinity and poison a vector; the drain-determinism comment is scoped honestly to input order only (the fine k-means reduce remains scheduler-ordered).

**Accepted tradeoff — default recall guarantee.** The width walk targets the 0.99 coverage bar and the default (no explicit `nprobe`) path runs floor-free, so the default query's worst case moves from the earlier ≥0.991 pad to a measured ~0.99 across the two corpora, in exchange for the rerank-budget latency win (≈37→20 ms). This is accepted as an intentional latency/recall balance: ~0.99 holds on the measured corpora, and a caller who needs the stricter per-cell floor pins `nprobe`, which re-arms the floor at `k` (asserted in the serve-the-law test). The behavior is now encoded in a test, so future drift on either axis is visible.

CI is green including the 1M-scale benches and the recall test.

PR #521, opened 2026-08-03
vector: default cosine rerank to a single 16-bit scalar plane (Sq16)

### LINKED ISSUE #518: Adopt Sq16 single-plane 16-bit rerank codec as the cosine default
## Summary

Add a `Sq16` rerank codec — a single uniform 16-bit (u16) plane on a fixed `[-1,1]` cosine grid, with a per-doc dequantized-norm correction — and make it the **default** rerank codec for cosine columns, replacing `Sq8FixedResidual`.

Same on-disk footprint (2 bytes/dim), simpler encoding, faster rerank, and recall that is provably ≥ the current codec.

## Motivation

`Sq8FixedResidual` stores, per vector, a `dim`-byte u8 coarse plane plus a `dim`-byte i8 residual plane (2 B/dim) and scores **both** planes on every rerank candidate. A flat u16 plane has the **same** storage and the **same** precision ceiling, but:

- **Recall ≥ (provable).** The u16 grid step (`2/65535`) is finer than the coarse+residual effective step (`2/65280`), so per-component quantization error is strictly ≤ → the dequantized vector is at least as close to the original → the top-k rerank ordering is at least as faithful. Cell routing is independent of the rerank codec, so this holds end to end.
- **Faster rerank.** One widen+FMA leg instead of two. Measured per-candidate rerank cost (norm-corrected cosine, 100k candidates) improves by ~6–9% on x86 (AVX), ~22% on server ARM (Neoverse), and ~37% on Apple ARM (NEON). End-to-end p99 improves ~8% at the default rerank depth and grows with rerank depth.
- **Simpler.** One uniform quantizer plus the existing per-doc norm — no coarse/residual split, no residual divisor, no separate per-plane clamp.

Recall was measured equal-or-better on a 1M-vector real-embedding cosine benchmark: codec-isolated recall@10 is strictly higher (finer grid), and end-to-end recall@10 is at parity (within run-to-run grid-training variance).

## Scope

1. **Codec + kernel:** `Sq16` u16 encode/decode, single-plane norm-corrected cosine scoring kernel, per-doc dequantized-norm table in `codec_meta`, discriminator, and `rerank_codec: sq16` config parsing.
2. **Maintenance paths:** extend build-from-source, IVF merge, compaction, and drain to handle the single-plane codec (these currently assume the coarse+residual family), carrying the per-doc norm through each.
3. **Default:** make `Sq16` the default rerank codec for cosine columns.
4. **Tests + CI:** round-trip vs fp32, on-disk store/read alignment, codec-isolated recall, and the rerank-kernel microbench; full CI green.

## Notes

- Cosine only (fixed `[-1,1]` grid); L2 and dot-product paths are unchanged.
- The fixed `[-1,1]` grid assumes unit-normalized input, consistent with the existing cosine codec.

## Trade-offs — what Sq16 gives up (both immaterial for the supported targets)

1. **A coarse-only fast path (cascade).** `Sq8FixedResidual`'s u8 plane is a standalone coarse code, so in principle one could rank by it alone and consult the residual only for ties. `Sq16` has no cheap coarse plane. Moot in practice: coarse-only ranking (plain `Sq8`) recovers only ~0.983 recall vs ~0.999, so the coarse byte is *not* decisive on the close top-k and the residual is scored on every candidate anyway — which is why the residual exists. If coarse-alone sufficed we'd ship plain `Sq8`; it doesn't, so no fast path is actually lost.
2. **8-bit-only hardware.** On an int8-only / 8-bit-register target, `Sq8`'s u8 primitives could beat `Sq16`'s u16. Moot for the supported targets (x86/AVX, ARM/NEON, ARM server/Neoverse), which widen to f32 efficiently and where `Sq16` measured faster on every platform. No 8-bit-only deployment target is in scope.

muralikpbhat:
Closes #518

## Summary

Make **`Sq16`** the default cosine rerank codec, replacing `Sq8FixedResidual` (coarse u8 + i8 residual). `Sq16` is a single uniform 16-bit plane on the fixed `[-1, 1]` cosine grid — the same 2 bytes/dim, but a finer quantization step, scored in a single pass, with no per-cluster scale/offset arrays.

## Why

- **Recall ≥ (finer grid; shown codec-isolated below).** The u16 grid step (`2/65535`) is finer than the coarse+residual effective step (`2/65280`), so per-component quantization error is provably ≤ — the dequantized vector is at least as close to the original. Recall is a *ranking* metric and isn't strictly monotonic in per-vector error (a near-tie can still flip), so this is an *expected* ≥ rather than a per-query guarantee; the codec-isolated measurement below confirms Sq16 ≥ Sq8+8 with the grid held constant.
- **Faster rerank.** One widen pass over a single plane instead of scoring two planes per candidate (see the kernel microbench below).
- **Simpler + leaner on disk.** No coarse/residual split and no per-cluster quantizer sidecar; the only `codec_meta` is the per-doc dequantized-norm table (cosine/L2Sq), matching the existing family so the cosine kernel still divides by `‖d̂‖`.

## Evidence — A/B vs `Sq8FixedResidual` (`supertable vector`)

Recall@10 vs brute-force fp32 truth. Warm/cold are the post-drain hidden index. Two ISAs (ARM, x86); Δ is Sq16 vs Sq8+8.

**100k × dim=1024, cosine**

| metric | ARM Sq16 | ARM Sq8+8 | ARM Δ | x86 Sq16 | x86 Sq8+8 | x86 Δ |
|---|---|---|---|---|---|---|
| recall — pre-drain | 1.000 | 0.999 | ≥ | 1.000 | 0.999 | ≥ |
| recall — post-drain | 1.000 | 0.999 | ≥ | 0.980 | 0.989 | −0.9%\* |
| warm p50 | 427 µs | 538 µs | −21% | 551 µs | 894 µs | −38% |
| warm p99 | 508 µs | 652 µs | −22% | 584 µs | 1070 µs | −45% |
| cold search | 6.05 ms | 6.02 ms | par | 11.8 ms | 10.2 ms | +16%\* |
| ingest | 67.0 s | 67.3 s | par | 31.9 s | 38.3 s | −17% |
| stored | 279 MiB (72%) | 342 MiB (88%) | −18% | 279 MiB (71%) | 342 MiB (87%) | −18% |
| drain assign+spill | 314 ms | 392 ms | −20% | 446 ms | 506 ms | −12% |
| drain total | 6.34 s | 6.49 s | −2% | 7.77 s | 8.54 s | −9% |

**1M × dim=1024, cosine**

| metric | ARM Sq16 | ARM Sq8+8 | ARM Δ | x86 Sq16 | x86 Sq8+8 | x86 Δ |
|---|---|---|---|---|---|---|
| recall — pre-drain | 0.996 | 0.995 | ≥ | 0.996 | 0.995 | ≥ |
| recall — post-drain | 0.996 | 0.995 | ≥ | 0.996 | 0.995 | ≥ |
| warm p50 | 708 µs | 732 µs | −3% ‡ | 1.03 ms | 1.17 ms | −12% |
| warm p99 | 983 µs | 915 µs | +7% ‡ | 1.13 ms | 1.35 ms | −16% |
| cold search | 20.5 ms | 20.4 ms | par | 40.0 ms | 38.7 ms | par |
| ingest | 84.3 s | 80.4 s | +5% ‡ | 63.1 s | 63.1 s | par |
| stored | 2.14 GiB (56%) | 2.20 GiB (58%) | −3% | 2.15 GiB (56%) | 2.21 GiB (58%) | −3% |
| drain assign+spill | 4108 ms | 3022 ms | +36% ‡ | 4019 ms | 4124 ms | −3% |
| drain total | 40.4 s | 27.5 s | +47% ‡ | 68.0 s | 65.8 s | +3% |

**10M × dim=1024, cosine (x86 only)**

| metric | Sq16 | Sq8+8 | Δ |
|---|---|---|---|
| recall — pre-drain | 0.998 | 0.997 | ≥ |
| recall — post-drain | 0.998 | 0.986 | **+0.012** |
| warm p50 | 1.68 ms | 1.76 ms | −5% |
| warm p99 | 2.06 ms | 2.39 ms | −14% |
| cold search | 198 ms | 177 ms | +12% ‡ |
| ingest | 568 s | 542 s | +5% ‡ |
| stored | 20.67 GiB (54%) | 20.78 GiB (54%) | par |
| drain assign+spill | 90.0 s | 111.4 s | −19% |
| drain total | 913 s | 930 s | −2% |

- **Recall ≥ Sq8+8 at scale** — 1M 0.996 vs 0.995 (both ISAs), 10M 0.998 vs 0.986 (x86); the codec-isolated measurement below confirms it with the grid held constant.
- **Grid-driven variance is pre-existing and codec-independent** (`*`, `‡`). The IVF grid is trained per run by k-means over the parallel-writer bootstrap, so two independent builds land different cell boundaries. Cell routing doesn't depend on the rerank codec, so this variance is orthogonal to the codec — and it moves *every* end-to-end axis, not just recall: a single point can tip recall a few thousandths either way (x86 100k 0.980 vs 0.989, inverted from ARM's 1.000 vs 0.999), and at 1M ARM the two runs drew different oversized cells (`‡`), swinging warm-p99 / ingest / drain. Measured run-to-run spread (0.0008–0.0025) exceeds every between-codec gap (≤0.0003). So the per-point end-to-end latency/recall deltas are **not** clean codec signals — the clean ones are the codec-isolated recall and the kernel microbench below.
- **Storage** is ~18% smaller stored at 100k, narrows to ~3% at 1M, and is **parity at 10M** (54% vs 54%; same raw 2 B/dim) — a small-corpus compression effect that vanishes at scale, not headlined.

## Codec-isolated recall (grid + routing held constant)

To isolate the codec from the grid variance above, we measured recall with the same built grid and the same routing, swapping only the rerank plane — so any delta is the codec alone:

| measurement (cosine, k=10) | Sq8+8 | Sq16 |
|---|---|---|
| Codec-isolated recall@10, Cohere-1M (same grid, swap rerank plane) | 0.99933 | **0.99967 (+0.00033)** |
| Kernel bisection, real embeddings, no IVF routing/shortlist (rerank vs exact fp32) | 0.9993 | **0.9997** |

Both are deterministic and bit-identical on x86 and ARM. With the grid confound removed, Sq16 ≥ Sq8+8 (the `+0.00033` is one extra correct neighbor in 3000). The no-norm Sq16 variant ties Sq8+8 at 0.9993, and the per-doc norm correction is what lifts it to 0.9997 — consistent with the strictly-finer u16 grid (provably ≤ per-component error).

**Kernel microbench — routing-free per-candidate rerank cost** (`Sq16 / Sq8+8` ratio, norm-on both):

| platform | ISA (µarch) | Sq16 speedup |
|---|---|---|
| Apple M4 | ARM / NEON | ~37% |
| Graviton3 | ARM server / Neoverse-V1 | ~22–23% |
| x86 (`D16ds_v7`) | x86 / AVX | ~6–9% |

Sq16 is faster on every platform — one u16 widen leg vs Sq8+8's two planes. The kernel is widen/bandwidth-bound, so the win is largest on element-bound ARM/NEON and modest on x86's wide 8-bit SIMD. This is the clean codec-latency signal; the end-to-end warm deltas in the tables carry grid variance on top (above), so they are not headlined.

## Trade-offs — what `Sq16` gives up (both immaterial for the supported targets)

1. **A coarse-only cascade fast path.** The u8 coarse plane could in principle be ranked alone with the residual consulted only on ties. Moot: coarse-alone recall is far below the residual-scored result, which is why the residual is scored on every candidate anyway.
2. **8-bit-only hardware.** On an int8-only target, u8 primitives could beat u16. Moot for the supported x86/ARM targets, which widen to f32 efficiently and where `Sq16` measured faster.

## Scope

Codec + kernel, single-plane handling through the full index lifecycle (build, drain, IVF merge, compaction, per-cell spill), the multi-cell directory reader, and the default flip. Non-cosine metrics keep locally-fitted `Sq8Residual`.

## Notes

- **Reopen compatibility (fixed).** The rerank codec was part of the table options-hash, so flipping the default made an existing cosine table recompute a different hash than the one stamped at create and fail to open — despite fully readable bytes. This PR drops the codec from the options-hash: it was redundant identity, because the codec is data-determined (each superfile records its `codec_id` on disk and the reader dispatches decode on that, never on the caller-supplied codec), so hashing it guarded nothing while manufacturing a false `OptionsHashMismatch` on the flip. Identity now depends only on the fields that must match to read a table (column, dim, rot_seed, metric, partition strategy). There is no persisted cosine table to preserve, so no migration is needed; the commit message records the alternatives (persist in manifest / recover from `codec_id` on mismatch / bypass-validation flag) and why none cleanly supports a pre-existing table.

### DIFF
--- a/benches/utils/corpus.rs
+++ b/benches/utils/corpus.rs
@@ -99,12 +99,12 @@ pub fn bench_rerank_codec(metric: Metric) -> RerankCodec {
     // Diagnostic (env-gated): force an EXACT fp32 rerank column to isolate Sq8
     // codec loss from routing/coverage. INFINO_BENCH_RERANK_CODEC=fp32 → if the
     // recall ceiling jumps to ~1.0, the residual gap was the Sq8 quantization.
-    let codec = if std::env::var("INFINO_BENCH_RERANK_CODEC").as_deref() == Ok("fp32") {
-        RerankCodec::Fp32
-    } else if metric == Metric::Cosine {
-        RerankCodec::default()
-    } else {
-        RerankCodec::Sq8Residual
+    let codec = match std::env::var("INFINO_BENCH_RERANK_CODEC").ok().as_deref() {
+        Some("fp32") => RerankCodec::Fp32,
+        Some("sq16") => RerankCodec::Sq16,
+        Some("sq8_fixed_residual") => RerankCodec::Sq8FixedResidual,
+        _ if metric == Metric::Cosine => RerankCodec::default(),
+        _ => RerankCodec::Sq8Residual,
     };
     assert!(
         codec.supports_metric(metric),
--- a/benches/utils/recall_while_ingest.rs
+++ b/benches/utils/recall_while_ingest.rs
@@ -45,7 +45,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use arrow_array::{Array, Decimal128Array, RecordBatch};
+use arrow_array::{Array, Decimal128Array, LargeStringArray, RecordBatch};
 use arrow_schema::Schema;
 use infino::{OptimizeOptions, supertable::Supertable};
 use rayon::prelude::*;
@@ -189,11 +189,21 @@ fn build_queries(n_cent: usize, n_queries: usize) -> Vec<Vec<f32>> {
 
 /// One append batch straight off the streamed `flat` (no corpus retained),
 /// reusing the ingest path's `vector_array` builder so the column layout is
-/// byte-identical to what `options_for(Modality::Vector, _)` expects.
-fn vector_batch(schema: &Arc<Schema>, flat: &[f32], len: usize) -> RecordBatch {
+/// byte-identical to what `options_for(Modality::Vector, _)` expects. The
+/// `Modality::Vector` schema carries the filter-bucket column ahead of the
+/// vector, so the batch mirrors that field order — bucket terms are derived
+/// from each row's global doc id (`doc_base + i`), exactly as the bulk ingest
+/// path stamps them.
+fn vector_batch(schema: &Arc<Schema>, flat: &[f32], len: usize, doc_base: usize) -> RecordBatch {
+    let buckets: Vec<String> = (doc_base..doc_base + len)
+        .map(ingest::vector_filter_bucket_term)
+        .collect();
+    let bucket_col = Arc::new(LargeStringArray::from(
+        buckets.iter().map(String::as_str).collect::<Vec<_>>(),
+    )) as Arc<dyn Array>;
     RecordBatch::try_new(
         schema.clone(),
-        vec![ingest::vector_array(&flat[..len * DIM])],
+        vec![bucket_col, ingest::vector_array(&flat[..len * DIM])],
     )
     .expect("vector RecordBatch")
 }
@@ -607,7 +617,7 @@ pub fn run() {
                 retained.extend_from_slice(&flat[..sub * DIM]);
             }
             update_heaps(&mut heaps, &queries, &flat, (n + off) as u32, sub);
-            let batch = vector_batch(&schema, &flat, sub);
+            let batch = vector_batch(&schema, &flat, sub, n + off);
             {
                 let mut writer = st.writer().expect("writer");
                 writer.append(&batch).expect("append");
--- a/src/config/config.yaml
+++ b/src/config/config.yaml
@@ -127,11 +127,14 @@ vector:
   # Default rerank codec for cosine vector columns. Non-cosine metrics
   # still use locally fitted sq8_residual (values are not bounded to
   # [-1, 1]). Override per column at table-create time when needed.
-  #   sq8_fixed_residual — fixed absolute grid (default; portable bytes)
+  #   sq16               — single 16-bit plane on the fixed [-1, 1] grid
+  #                        (default; finer grid than sq8_fixed_residual at
+  #                        the same 2 bytes/dim, faster rerank)
+  #   sq8_fixed_residual — coarse u8 + i8 residual on the fixed grid
   #   sq8_residual       — per-cluster fitted quantizer
   #   fp32               — full-precision rerank
   #   rabitq_only        — no rerank column
-  rerank_codec: sq8_fixed_residual
+  rerank_codec: sq16
 
   # Absolute cap on fine IVF centroids probed per vector search.
   # Empty (~) derives the budget from nprobe × eligible superfiles at
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -1191,7 +1191,7 @@ supertable:
         assert_eq!(cfg.vector.cell_split_doc_cap, 500_000);
         assert_eq!(cfg.vector.user_centroids, CentroidAlignment::Local);
         assert_eq!(cfg.vector.drain_consolidate, DrainConsolidate::Kmeans);
-        assert_eq!(cfg.vector.rerank_codec, RerankCodec::Sq8FixedResidual);
+        assert_eq!(cfg.vector.rerank_codec, RerankCodec::Sq16);
         assert_eq!(cfg.vector.drain_read_concurrency, ThreadCount::Auto);
     }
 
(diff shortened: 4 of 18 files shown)


ekechinwokah (commented):
lgtm

ekechinwokah (approved):

Issue #482, opened 2026-07-29
vector follow ups


muralikpbhat:
Consolidated pending vector-index follow-ups (carried over from #436 and #437, plus items surfaced during the split/drain work).

- **1B end-to-end validation** — extend the 100M drain/recall runs to 1B (recall, latency, and RSS).
- **Dead-fraction reclaim eligibility** — make large batch-major drain superfiles merge-eligible on dead fraction (tombstoned + superseded rows), not just raw size, with a 1-input reclaim rewrite. Needed at 100M+ (from #437).
- **Rework the compaction merge target** — shard-packed drain superfiles exceed the current ~2 GiB target, so `fits()` yields zero merge jobs at scale; choose a larger/correct superfile target size.
- **100M+ split-cap sweep** — determine the ideal docs-per-cell target and resulting cell count.
- **Recall-recovery curve on a split-fined 100M grid** — quantify how much within-cell crowd-out the finer grid removes (from #436).
- **Hierarchical coarse quantizer over centroids** — so finer grids don't degrade ingest throughput / query routing at 1B (see #440).

===
DONE
- **Parallelize cell split under an RSS budget** — currently serial per cell; admit `P = budget / largest-eligible-cell RSS` concurrent splits, compute-parallel and commit-serial through the manifest CAS (from #437). DONE
- **Per-cell 1-bit shortlist** — replace the global 1-bit shortlist to address the residual within-cell ranking crowd-out at scale (from #436). DONE
- **Per-cell centroid slice / prefetch at open** — remove the fixed fp32 centroid-section cold-read tax on the first cold query. 
- **Move remaining vector serving tunables from env vars into config.yaml.** DONE
- **Filtered-search recall** — low (~0.68) and pre-existing; investigate. DONE

muralikpbhat:
### Streaming ingest past 100M — per-checkpoint cost + quality

Synthetic 1024-dim planted-cluster corpus (cosine, σ=0.3 per-dim noise), 10M-doc checkpoints, single 64 GiB host, object storage.

**Growth cost (per 10M checkpoint):**

| total docs | optimize time | drain RSS peak | warm recall@10 |
|---|---|---|---|
| 110M | 6333 s | 15.9 GB | 0.990 |
| 120M | 6852 s | 23.2 GB | 0.991 |
| 130M | 7215 s | 16.0 GB ¹ | 0.990 |
| 140M | 7881 s | 28.5 GB | 0.988 |
[...]

Issue #457, opened 2026-07-24
vector_search default nprobe gives low out-of-box recall (0.76 @ 847k); document + reconsider default


asifmakhani:
## Summary

On an 847k-doc IVF vector index (`nlist=920`, cosine), `vector_search(col, q, k=10)` **with no `nprobe`** returns dense recall@10 of **0.760** vs an exact-cosine oracle. Passing an explicit `nprobe` recovers recall cleanly along a sensible Pareto curve:

| nprobe | recall@10 | median latency |
|---|---|---|
| default (None) | 0.760 | 9.4 ms |
| 8 | 0.949 | ~25–44 ms |
| 16 | 0.979 | ~54 ms |
| 32 | 0.981 | ~59 ms |
| 128 | 0.988 | ~64 ms |

The engine is clearly capable of high recall — the **default** just silently favors latency, so a caller who doesn't tune gets 0.76 and may wrongly conclude ANN quality is poor. (Same corpus/vectors on Solr HNSW ≈ 0.88.)

## Asks

- Document `nprobe` on `vector_search` — its default and the recall/latency tradeoff are currently undocumented in the Python binding.
- Reconsider the default (e.g. scale it with `nlist`) or emit guidance, so out-of-box recall isn't surprising.

**Env:** infino 0.1.7, macOS arm64, e5-small-v2 (384d), 43 queries, k=10.

ekechinwokah:
Status since 0.1.7 (this report's version): the default probe width is no longer a fixed constant — it is measured per corpus at drain time and stamped into the table's manifest ([#479](https://github.com/ekechinwokah/infino/issues/479), refined through #497/#505/#520). On Cohere-1M the no-knobs default now serves 0.9931 recall@100 / 0.9910 recall@10 (PR [#520](https://github.com/ekechinwokah/infino/pull/520)), so the "silent latency-favoring default" this issue describes is structurally gone: defaults track a measured 0.99 coverage target for the table's own geometry.

Two things keep this open:

Diffuse-geometry corpora can still be under-provisioned — [#515](https://github.com/ekechinwokah/infino/issues/515) reports BioASQ-1M at 0.9435 under the adaptive default (same failure class as this report; e5-class embeddings may share it). Verification on that class is in progress; a retest of your corpus on current main would be a useful datapoint — expect materially better out-of-box recall than 0.760 either way.

The documentation asks are untouched: nprobe semantics and the default's behavior in the Python binding docs, plus tuning guidance. 

[...]

PR #525, opened 2026-08-04
supertable/vector: improve recall on real world corpora 

### LINKED ISSUE #515: Bioasq-1M recall gap at default width (0.94): adaptive probe width under-provisions on concentrated/diffuse embeddings

ekechinwokah:
Fixes #515 

## Problem

Drain-time calibration measures probe width and rerank budget on **corpus rows**. On corpora whose real queries do not look like rows (BioASQ questions vs passages) both stamps under-serve: measured **0.9216 recall@10 at true defaults** on main (`17ef841`). The diagnosis in #515 showed the recall was already computed and thrown away — for 74% of failing queries every truth cell was already admitted and exactly scored, then discarded by fixed-width serving.

Two smaller defects rode along: the post-compact width walk stamped on its raw sample mean, so on marginal ~3.5K-cell grids the stamp flipped between width 1 and 2 run-to-run (documented straddle 0.986–0.994; a width-1 draw measured 0.980 served recall), and the stamped rerank pool — sized for the stamped width — starved once serving reached more cells than the stamp.

## What

**`supertable/query` — score-sensitive serving, law-served default path only.**
- **Self-measured admit extension**: after the pre-existing write-window round, each round measures the 1-bit-estimate-to-exact residual from cells already scored and admits remaining cells whose residual-shifted estimate could land inside the serve window, until a round admits nothing. The bound is the table's own quantization noise on this query — no query-side fraction. Decisive geometry admits nothing extra and serves byte-identically.
- **Serve window**: selection follows the exact-fine ranking past the stamped width while scores stay near-tied. The stamped width stays a served floor in both directions. Window value is `vector.serve_near_tie_slack` in `config.yaml` (default = measured real-query truth-cell slack p99); it is a config default rather than a per-table stamp because the drain sample is corpus rows, which cannot measure real-query slack.
- **Extension cells read at the bounded pre-pin fine depth**, never whole-cell (measured zero recall cost: 0.9617 vs 0.9620).
- **Rerank budget follows the served width**: the law's pooled budget scales by served-cells over stamped-width, so extending serving does not starve the widened candidate set. Both quantities come from stamps plus the serve window's own evidence; no new constants.
- Explicit `nprobe` / `rerank_mult` carry none of this — exact, unscaled requests.

**`supertable/opann` — width crossing on the sample's confidence bound.** Stamp the smallest width whose coverage lower bound (mean − z·SE from the sample's own per-query variance, one-sided 95%) clears the target: the sample must prove the narrower width. Zero-variance samples (all synthetic post-drain shapes) stamp exactly as before; only marginal crossings round up. Stamp arithmetic only — clustering, assignment, spill and compaction untouched.

## Measured — one build (`0912415` = current main incl. #521 Sq16, plus this PR)

True defaults, no knobs (`rerank_mult=None`, no `nprobe`); 1M tables rebuilt on this build for codec parity:

| corpus | this branch | reference: `rm=256` forced |
|---|---|---|
| BioASQ-1M k=10 | **0.9900 @ 6.5 ms** | 0.9910 @ 9.8 ms |
| Cohere-1M k=10 | **0.9970 @ 8.0 ms** | 0.9960 @ 7.5 ms |
| Cohere-1M k=100 | **0.9972 @ 12.7 ms** | 0.9979 @ 19.0 ms |

BioASQ out-of-box goes 0.9216 → 0.9900, within 0.001 of the forced-pool ceiling at roughly two-thirds its latency; Cohere defaults beat their own `rm=256` rows.

10M synthetic lifecycle (Azure, standard gate, `RUN_EXIT=0`):

| state | recall@10 | warm p50 |
|---|---|---|
| pre-drain | 0.998 | 8.9 ms |
| post-drain | 0.996 | 3.1 ms |
| post-delta | 0.994 | 4.5 ms |
| post-compact | 0.997 | 4.8 ms |

Cold open 13 GET / 88.8 MiB. **Recall@10 ≥ 0.99 at every lifecycle state.**

## Trade-offs, explicit

- Tie-heavy queries pay for the cells and rerank they use; decisive queries are byte-identical. This replaces the alternative of stamping a wide width, which charges every query.
- **Open risk, not hidden**: the width-confidence rounding is unbounded on a flat coverage curve. One pre-merge draw stamped post-compact width 42, and that pinned arm fetched ~424 MiB per cold query (the bench cold-GET ceiling caught it). It did not recur on this build (post-compact stamped width 1), but whether the crossing should also weigh cold cost is worth deciding — happy to take as a follow-up.
- Known follow-up: the width arm carries ~2.6 ms fixed overhead unexplained by cell count (deferred-rerank pooling suspected, unprofiled). Cold-first post-compact is structurally unchanged — the fp32 centroid-section fetch is #482's top item.

## Tests

lib 2160, supertable 250, superfile clean; `make check` and `public-api` clean; no new `unsafe`. New unit coverage for the admit window, serve selection and bounded-depth gate; `vector_law_serving` pins both arms (law-served default serves the law budget floor-free; explicit `nprobe` arms the per-cell floor). Width-law calibration suite unchanged. Plus an ignored `bioasq_admit_diag` stage-split instrument (env-configured, local tables).

Refs #482 (cold-first centroid section), #457 (same failure class), #537/#538 (explicit-`nprobe` oracle inversion — independent path, not addressed here).


src/supertable/query/vector.rs
```diff
@@ -629,14 +732,36 @@ fn union_cell_selection(grid: &[u32], fine: &[u32]) -> Vec<u32> {
...
+/// the served set. Bounds are the evidence itself and the admit window
+/// (only admitted cells are ranked) — deliberately no artificial cap: a
+/// constant ceiling would silently truncate exactly the tail queries
+/// this serves. Cost stays bounded per cell instead: extension cells
+/// are read at the pre-pin fine depth (the stamped law's runs-per-cell
+/// floor), never the pin's whole-cell depth. Value = the measured
+/// truth-cell slack p99 (0.287–0.294 across the #515 BioASQ diag
+/// configs); the dial grid read 0.9617 at 0.25 vs 0.9657 at 0.30
+/// (diag metric), Cohere controls unchanged at both.
+const FINE_SERVE_NEAR_TIE_SLACK: f32 = 0.30;
```
muralikpbhat: **Hardcoded near-tie slack.** `FINE_SERVE_NEAR_TIE_SLACK = 0.30` is a single compile-time constant driving both the admit `serve_threshold` and the serve-window selection, for every table and metric. The description frames the window as "measured truth-cell slack p99" (a per-table quantity), but here it's one global number. On a corpus whose true near-tie slack is >0.30 the truth cells fall just outside the window on every query — a systematic recall shortfall with no per-table lever (can't be tuned without a rebuild); <0.30 over-serves on every query. For a change aimed at real-world corpora in general, this is the piece most likely to under-serve the next one. Recommend stamping it per-table from the same drain sample that measures the slack (or making it config), rather than a fixed constant.

src/supertable/query/vector.rs
```diff
@@ -1516,6 +1665,62 @@ impl SupertableReader {
...
+                loop {
+                    let fine_ranked_now = cells_ranked_by_fine_score(&candidates);
+                    let Some(&(_, best_exact)) = fine_ranked_now.first() else {
+                        break;
+                    };
+                    let serve_threshold =
+                        relative_score_window(best_exact, FINE_SERVE_NEAR_TIE_SLACK);
+                    let exact_best_by_cell: HashMap<u32, f32> =
+                        fine_ranked_now.into_iter().collect();
+                    let round = admit_extension_round(
```
muralikpbhat, line 1711: **Widened serving on the floor-free default path.** The admit extension serves cells well past the stamped width on the law-served default path — which is exactly the path where the per-cell floor is off (`cell_floor = 0` when `nprobe` is unset, from #520). #520's rationale for dropping the floor (measured +0.0001) was taken *at the stamped width*; it doesn't automatically hold once serving reaches many more cells, since far-cell 1-bit-estimate noise can now evict a near cell's true top-k row from the single floor-free `k*rerank_mult` pool (the #494 inversion the floor guards). The `k*256` headroom + bounded extension depth make this a thin tail rather than a broad regression, and the byte-identical decisive path is unaffected — but please confirm with a scale run that the widened set doesn't dip vs the stamped-width baseline on a control corpus, and update the floor rationale to note it's now measured at the widened widths. If a dip shows, arming `cell_floor = k` on just the extension subset is the targeted fix.

src/supertable/query/vector.rs
```diff
@@ -629,14 +732,36 @@ fn union_cell_selection(grid: &[u32], fine: &[u32]) -> Vec<u32> {
...
+/// law-pinned default path only. Past the law's width, selection keeps
+/// following the ranking while each next cell's best exact fine score
+/// stays within this relative window of the winner's. On decisive
+/// geometry (synthetic, Cohere) fine scores cliff after the law's picks
+/// and the window never opens — serving is byte-identical to the pin.
+/// On flat-scored corpora (question-vs-passage retrieval: BioASQ
+/// measured its truth cells at exact-fine ranks ≤ 48, already inside
+/// the exactly-rescored admit window) the query's own evidence extends
+/// the served set. Bounds are the evidence itself and the admit window
+/// (only admitted cells are ranked) — deliberately no artificial cap: a
```
muralikpbhat: **No cap on the admit loop.** Termination is sound (the admitted set grows monotonically, bounded by populated cells) and this is in-memory compute — the 1-bit ranking is computed once and the object-store fanout runs once after the loop, so it isn't GET amplification. But with no round/cell ceiling, a genuinely diffuse query at large cell counts (~25K at 100M) can drive O(cells²) summary iteration worst case. Real corpora converge in a few rounds, so this is a follow-up not a blocker — but a defensive round/served-width cap is cheap insurance before relying on this at 100M.

src/supertable/query/vector.rs
```diff
@@ -433,13 +446,88 @@ fn postings_by_cell_from_summaries(
...
+fn admit_extension_round(
+    admit_ranking: &[(u32, f32)],
+    admitted: &HashSet<u32>,
+    exact_best_by_cell: &HashMap<u32, f32>,
+    serve_threshold: f32,
+) -> Vec<u32> {
+    let mut residual_floor = f32::INFINITY;
+    for (cell, estimate) in admit_ranking {
+        if let Some(exact) = exact_best_by_cell.get(cell) {
+            residual_floor = residual_floor.min(exact - estimate);
```
muralikpbhat, line 489: **Admit bound is an empirical extrapolation — soften the "safe" framing.** `residual_floor` is a min over the *already-scored* cells of (exact − 1-bit estimate), and that estimate's residual has arbitrary sign, so this extrapolates from the best-estimate cells rather than being a provably-safe over-admit bound: a distant truth cell whose estimate is unusually pessimistic can be pruned. That matches coverage landing ~0.99 rather than 1.0 — a small but real under-admit tail. It's strictly better than the stamped-width baseline (the width stays a floor), so it's a ceiling, not a regression — but the framing that "the bound is the table's own quantization noise / decisive geometry serves safely" reads as a guarantee it doesn't give. Suggest "expected/measured ≥" phrasing.

muralikpbhat (commented):
Reviewed the two new commits on top of #520. The mechanism is correct and cleanly scoped: the score-sensitive serving is gated to the law-served default path only (filtered + explicit-`nprobe` carry none of it), the decisive-geometry path is byte-identical to before, termination is sound, and the confidence-bound width stamp is statistically sound (one-sided lower bound so noise widens; zero-variance synthetics stamp unchanged). The new shape-specific unit tests guard the mechanism. Nice work.

**Holding approval on the recall claim, not the code.** Independent testing at defaults measured ~0.95 on BioASQ, well below the 0.9890 headline in the table. Before merge, please attach the exact reproducible config behind 0.9890 — grid build / cell count, `k`, and confirmation it's at true defaults with no explicit `nprobe` (the fix is default-path-only, so a fixed-`nprobe` run bypasses it entirely and lands on the pre-fix ladder). That lets the headline be reproduced rather than build-specific.

Four inline notes below. The **hardcoded near-tie-slack constant** is the substantive one for generalizing past a single corpus; the others are the floor-free-widened-set interaction with #520, a missing loop cap for 100M, and softening the "provably safe" wording. Non-vector tiers untouched; hygiene clean.

src/supertable/query/vector.rs
ekechinwokah: taken in ee7b215 — it's config now (vector.serve_near_tie_slack, default 0.30), same precedent as fine_nprobe_pct, so there's a per-deployment lever without a rebuild. on stamping it from the drain sample instead: that inherits the exact defect this pr fixes — the sample is corpus rows, and rows measure a different slack distribution than real queries (same table: rows serve 0.9910 on the identical stamps that serve questions 0.8960, receipts in #515). a row-stamped slack under-measures question slack the same way the row-stamped width did. the principled successor is query-self-normalized gap detection on the exact-fine ranking — no constant, no distribution assumption — as a follow-up that has to reproduce this pr's gates before it replaces the measured window.

src/supertable/query/vector.rs
ekechinwokah: the control run you're asking for is already in the pr table — cohere is that control: its stamps are diffuse at both k (width 24-61), so the widened admit + floor-free pooled serving is active on its defaults, and it measures above the stamped-width baseline both ways (0.9955 vs 0.9931 @ k=100, 0.9940 vs 0.9910 @ k=10). the 10m gate adds the width-2 arm at 0.992 with a monotone explicit-nprobe ladder. no dip at any widened width measured. ee7b215 updates the floor rationale to record it's now measured at the widened widths too, and agreed on the fallback — floor = k armed on just the extension subset if a future gate dips; the per-subset machinery already exists from the bounded-depth work.

src/supertable/query/vector.rs
ekechinwokah: agreed on the shape (in-memory, single fanout, converges in 1-3 rounds on every corpus measured). on the cap itself: the standing rule here is serving never silently truncates — a ceiling lands exactly on the tail queries the mechanism exists to serve, and an earlier hard-capped draft of this same change died for that reason. so the follow-up i'd take instead: surface rounds + served width per query in the serving stats so a pathological corpus is visible rather than silently slow, and put the o(cells^2) worst case on the 100m gate explicitly — if it materializes there, we pick the mitigation on data. filing both.

src/supertable/query/vector.rs
ekechinwokah: correct — taken in ee7b215. residual_floor is a signed empirical extrapolation from the scored subset, a truth cell with an unusually pessimistic estimate can sit outside the extension, and that's exactly why the diag reads 99% full admit coverage instead of 100. the doc now says measured/expected >= with the under-admit tail explicit, and keeps the one guarantee that does hold: the stamped width stays a served floor, so the mechanism is a ceiling over the baseline, never below it.

muralikpbhat (approved):
Approving the code. All four review threads are addressed in ee7b215 and the replies:

- The serve window is now a config lever (`vector.serve_near_tie_slack`), with a sound argument for why a drain-stamp would inherit the very row-vs-query slack mismatch this PR fixes.
- The floor-free widened-set interaction is now measured at the widened widths (Cohere defaults, above the stamped-width baseline both ways, no inversion dip), with `floor = k` on the extension subset named as the targeted fallback.
- The admit-bound doc drops the safe-bound framing (signed empirical extrapolation, 99% coverage not 100%, the served floor as the one guarantee that holds).
- The loop cap is declined with a rationale I accept (serving must not silently truncate the tail queries the mechanism exists to serve); per-query rounds/served-width observability plus an explicit 100M gate is the right follow-up.

The mechanism is correct, cleanly scoped to the law-served default path, byte-identical on decisive geometry, and CI is green. The recall lift is real.

**One required change before merge — correct the recall headline.** The body still reads "BioASQ-1M k=10 default 0.896 -> 0.9890," but the scale retest in #515 puts #525 at defaults around 0.95 (1M) / 0.929 (10M), with >=0.99 requiring nprobe ~= 128 — i.e. the residual gap is rerank-pool / width-bound, not closed by serving-width alone. Please update the body to the reproducible default number (and note the nprobe~=128 path to >=0.99) so the headline matches what reproduces. The approval is on the code; the claim needs to match the measurements.

ekechinwokah:
thx. btw your #515 retest exposed that at true defaults the stamped rerank budget starves the serve window's widened candidate set (our harness had been masking it by forcing rm=256). the law-served default path now scales the pooled budget by served-cells over stamped-width — stamps plus the serve window's own evidence, no new constants. measured at true defaults, no knobs: bioasq-1m 0.9510 -> 0.9860 @ 6.3ms p50, cohere k=10 0.9940 -> 0.9970, k=100 0.9955 -> 0.9968. explicit nprobe/rerank_mult remain exact, unscaled requests. round-1 items were taken in ee7b215 (replies on each thread); 10m synthetic gate is rerunning on the tip as the non-regression record.

ekechinwokah:
re-review ready — the pr grew three commits past round-1 and everything below is re-measured on one build: 0912415 = current main (so it includes #521 sq16 and #534) plus this pr.

what changed since your review:
- ee7b215 — round-1 items (serve window moved to config as vector.serve_near_tie_slack; honest "measured/expected >=" framing on the admit bound; floor rationale re-measured at widened widths). replies on each thread.
- 7932ac4 — your rerank_mult=None finding: the law's rerank budget is calibrated on drain rows at the stamped width, so when the serve window extends past that width the stamped pool starves the widened candidate set. the law-served default path now scales the pooled budget by served-cells over stamped-width (stamps + the serve window's own evidence, no new constants). explicit nprobe/rerank_mult stay exact and unscaled.
- 0912415 — merge of current main + ship the new config key in config.yaml.

measured, true defaults (rerank_mult=None, no nprobe), sq16, 1m tables rebuilt on this build:

| corpus | defaults | rm=256 forced (reference) |
|---|---|---|
| bioasq k=10 | **0.9900 @ 6.5 ms** | 0.9910 @ 9.8 ms |
| cohere k=10 | **0.9970 @ 8.0 ms** | 0.9960 @ 7.5 ms |
| cohere k=100 | **0.9972 @ 12.7 ms** | 0.9979 @ 19.0 ms |

bioasq defaults land within 0.001 of the forced-pool ceiling at about two thirds the latency, and the cohere defaults beat their own rm=256 rows.

10m synthetic lifecycle on this build (azure, standard gate, RUN_EXIT=0): 0.998 pre-drain / 0.996 post-drain / 0.994 post-delta / 0.997 post-compact; cold open 13 GET. suites: lib 2160, supertable 250, superfile clean, make check + public-api clean.

one thing to flag rather than bury: the width-confidence crossing rounds the stamp up when the drain sample cannot prove the narrower width. on one earlier draw (pre-merge build) it rounded post-compact to width 42, and that pinned arm fetched ~424 MiB per cold query — the bench cold-GET ceiling caught it. it did not recur on this build (post-compact stamped width 1), but the rounding is unbounded on a flat coverage curve, so it is worth deciding whether the crossing should also weigh cold cost. happy to take that as a follow-up rather than hold the pr.

PR #527, opened 2026-08-04
superfile: stream build/compaction output to scratch to bound peak memory

vinaykakade:
## Problem

Building a superfile — both at append-commit and during a compaction merge — materialized the whole assembled superfile *and* its intermediate build artifacts (the Parquet body, the FTS blob, the vector blob) as in-memory `Vec<u8>`s at the same time. At corpus scale each of these is multiple GiB — the positional FTS blob especially — so peak resident memory for a single large build is several times the output size. On a memory-constrained host that OOMs the build.

## Change

Stream the large build artifacts to scratch temp files instead of holding them whole in RAM:

- **`SuperfileBuilder::finish_to<W: Write>`** streams the final splice to a sink; `finish()` is a thin wrapper over it.
- **`encode_parquet_body`** encodes the Parquet body to a scratch temp file (reads the footer back to decode metadata, truncates to the last row group) instead of a `Vec`; `EncodedBody` carries the file.
- **FTS/vector blobs** stream to scratch via the builders' existing `finish_to<W>` sinks (a new `finish_index_blobs_streamed` replaces the Vec-returning `finish_index_blobs`).
- The **append-commit build** and **all compaction merge paths** write the assembled superfile to a `NamedTempFile` and map it back (`mmap_readonly_bytes`) — the publish path takes `Bytes` unchanged, and large superfiles already stream via `put_multipart`. New streaming `build_from_*_to` variants back the Vec-returning merge methods.

Net effect: no whole build artifact is held in RAM; `finish` peak drops to roughly the buffered input. Reuses the existing mmap boundary in `reader_cache/disk.rs`, so **no new `unsafe`**.

## Correctness

Output is byte-identical — the write sinks change, the produced bytes don't. Unit round-trips green: superfile builder + reader, format footer, supertable writer (commit → read), compaction merge.

## Status — draft, validation pending

- [ ] `benches/` peak-RSS + build-throughput comparison vs `main`
- [ ] recall@10 ≥ 0.99 confirmation (change touches the output sink, not codecs — expected unchanged)
- [ ] `make miri` / `make asan` (no new `unsafe`, but new temp-file/mmap plumbing)

Opening as a draft to validate peak-memory + throughput before marking ready.

### DIFF
--- a/src/superfile/builder.rs
+++ b/src/superfile/builder.rs
@@ -88,7 +88,7 @@ use crate::superfile::{
     BuildError, SuperfileReader,
     format::{
         self,
-        footer::{encode_parquet_body, splice_index_blobs, splice_index_streams_to},
+        footer::{ParquetLayout, encode_parquet_body, splice_index_streams_to},
         kv,
     },
     fts::{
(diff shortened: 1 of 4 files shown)


vinaykakade:
Superseded by #531, which now contains these streaming-output commits plus the k-way FTS merge as one consolidated PR.

PR #529, opened 2026-08-04
benches: improve cost calcluations

ekechinwokah:
**Price the serving NVMe working set, not the lifecycle cache.** The keep-warm policy rows priced `disk_cache_bytes` from each battery's long-lived lifecycle consumer — a cache that absorbs maintenance traffic and the all-cells oracle sweeps (vector), or both pre- and post-compact generations (FTS/SQL). None of that is what a query-only serving node retains; maintenance runs on separate machines in production. Now a fresh consumer replays the modality's default battery against the settled post-compact layout and its cache is the priced working set. Measured at 10M: vector 44.99 → 21.96 GiB (Policy B $18.21 → $10.10), FTS 16.03 → 8.21 GiB (idle floor $4.55 → $2.33).

**Price scan-backed SQL aggregates per event.** The blended monthly read line already excluded Bulk (GB-returned) shapes; scan-backed aggregates stayed in the mean, so whole-corpus-CPU shapes ($100s/1M) averaged at equal weight with $1–5/1M lookups dragged the SQL headline to ~$160/1M. They now get per-event rate rows like Bulk (prefix generalized to "Per event"). Measured at 10M: scan aggregates $0.000006–$0.0029/query, bulk range scan $0.18, all-rows token_match $0.01; the selective blend reads $14.31/mo.

Measured 10M Policy-B set on this branch: vector ≈ $10.10, FTS ≈ $29.84, SQL (selective) ≈ $22.83, with scan SQL per event.

### DIFF
--- a/benches/utils/supertable.rs
+++ b/benches/utils/supertable.rs
@@ -715,9 +715,11 @@ enum TextLifecyclePhase {
 const SHARED_CONSUMER_CACHE_INDEX_FACTOR: u64 = 2;
 
 /// What the shared FTS/SQL lifecycle hands back: the compaction window's
-/// stats plus the settled local disk-cache footprint of the lifecycle
-/// consumer, sampled after the post-compact battery (just before the
-/// consumer drops) — the cost model's provisioned-occupancy NVMe basis.
+/// stats plus the serving NVMe working set — the disk-cache footprint of
+/// a FRESH consumer that replayed the caller's default battery against
+/// the settled post-compact layout (the cost model's
+/// provisioned-occupancy NVMe basis; the lifecycle consumer itself
+/// accumulates both generations and would overstate it).
 struct TextLifecycleStats {
     compaction: CompactionStats,
     disk_cache_bytes: Option<u64>,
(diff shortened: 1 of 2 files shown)


benches/utils/supertable.rs
```diff
@@ -4766,6 +4798,29 @@ pub mod vector {
...
+                // lifecycle consumer's cache is the wrong basis — by this
+                // point it has absorbed drain/compaction traffic and the
+                // all-cells oracle sweeps (whole-index reads no serving
+                // node performs; maintenance runs on separate machines in
+                // production). Replay the default-path battery queries on
+                // a fresh consumer + cache and sample what it kept.
+                let serving_disk_cache_bytes = {
+                    let (_serving_cache_dir, serving) = open_consumer(Modality::Vector, &built);
+                    let serving_reader = serving.reader().expect("serving consumer reader");
+                    for q in q_correct.iter().chain(q_cal.iter()) {
```
muralikpbhat, line 4812: Minor consistency: the vector working-set replay chains `q_cal` onto `q_correct`, but the FTS and SQL serving replays run only the default battery. A query-only serving node wouldn't issue the calibration queries, so this modestly pads the vector working set relative to the other two modalities. It's in the conservative (cost-safe) direction so it doesn't change the verdict — but for parity with FTS/SQL, replaying just the default-path battery (drop `q_cal`) would make the three bases consistent.

muralikpbhat (approved):
Approving. Small, correctly-scoped bench-cost change; I verified the two things that matter:

- Scan-backed aggregates are actually pulled out of the blended mean into per-event — the `is_bulk_group` prefix matches the family label the SQL runner constructs (`"Aggregates — scan-backed"`), and `starts_with` also catches the fold-ineligible variant. Not a silent no-op.
- The fresh serving consumer opens the settled post-compact layout on the shared store with an isolated cache, and its I/O isn't metered into the store cost — so `disk_cache_bytes` is genuinely the query-only serving working set, not the lifecycle consumer's two-generation + oracle-sweep cache. Correct isolation on all three modalities.

CI green including the four 1M bench jobs; non-vector/query paths untouched. One small consistency note inline — non-blocking.

PR #531, opened 2026-08-05
superfile: stream build outputs to disk; k-way FTS merge for compaction

vinaykakade:
Three related changes that make building and compacting a large corpus into one superfile fit a memory-tight host. All target the peak RSS of the write path.

## 1. Stream build outputs to disk instead of holding them as RAM `Vec`s

The commit and compaction paths assembled the output superfile — and its intermediate Parquet body and FTS/vector blobs — as corpus-sized in-memory `Vec<u8>`s, so peak RSS carried the whole output on top of the inputs.

- `finish_to<W: Write>` streams the spliced superfile to any sink; `finish()` now wraps it.
- Commit assembles the built superfile into a temp file and mmaps it back (`mmap_readonly_bytes`); all compaction merges (`build_from_*_readers_to`) do the same.
- `encode_parquet_body` writes the Parquet body to a scratch file and reads the footer back; the FTS and vector blobs finalize to scratch sinks.

Mapped pages are file-backed and reclaimable, and downstream publish takes `Bytes` unchanged. No new `unsafe` — reuses the existing mmap helpers.

## 2. Compaction merges the FTS index instead of re-indexing from text

Compacting FTS/scalar superfiles previously re-read every input's text and **re-tokenized and rebuilt the whole posting index in RAM**, so a merge peaked as high as a fresh full-corpus build.

This carries each input's **already-built posting lists** across instead. For each FTS column it streams the input's `(term, doc_id, tf, positions)` postings into the builder's prebuilt accumulator, with doc ids remapped to a dense surviving-doc space (tombstoned rows dropped, aligned row-for-row with the concatenated Parquet body). Positions flow to the spilled positions blob on disk; doc-lengths are read from each input and concatenated rather than recomputed. The vector merge already does the analogous byte-level splice; this is the FTS counterpart.

`merge_superfiles` routes to it whenever the inputs carry **no vector index**. Vector-bearing merges are unchanged: IVF-mergeable inputs keep the byte-splice path, and vector-but-not-mergeable inputs stay on re-index.

## 3. Stream the merge's Parquet body row-group-wise

The merge above still fed every input's surviving rows into `self.batches` and encoded the body at finish — so peak RSS carried the whole corpus of decoded rows on top of the FTS interner. Now a `ParquetBodyEncoder` takes one batch at a time and the merge drops each input's batch immediately after writing it, so the body holds at most one input's batch plus the writer's row-group buffer. `ArrowWriter` cuts row groups on accumulated row count, so the streamed body is byte-identical to encoding all batches at once. `finish_to`'s blob-finalize + splice tail is factored into shared helpers that a new `finish_to_with_body` reuses.

**Net:** a compaction of N segments into one superfile now peaks at roughly the FTS term interner (inherent to one index over the corpus) + spill buffers + one input's batch + mmap'd (reclaimable) inputs — not the whole corpus of rows or the output. This is what lets the "ingest to bounded segments, then compact to one" workflow stay within a tight RSS budget. Peak-RSS numbers on a 5M-doc corpus to follow.

## Correctness

The FTS merge is proven equivalent to re-indexing by:
- a **byte-for-byte oracle** — merge-via-re-index vs merge-via-k-way return identical postings, tfs, positions, doc-lengths, and Parquet rows over planted corpora (shared terms across inputs, positional codec, tombstoned rows);
- a **direct BM25 query comparison** — identical `(doc_id, score)` top-k across single-term, multi-term OR/AND, and shared-term queries.

`set_prebuilt_doc_lengths` restores `total_tokens` so `avgdl` matches a fresh build. The merged superfile is query-identical to a fresh build (a single benign Parquet-metadata byte differs; the bar is identical top-k/scores).

## Tests

Streaming round-trips (byte-identical), the merge oracles, prebuilt-posting round-trips (incl. forced-spill), and the routed compaction path (`compact_gc`) pass; 704 superfile unit + 135 superfile integration + commit/compaction suites green. `make check` (nightly rustfmt with the crate's import opts + `clippy --all-targets -D warnings`) clean.

### DIFF
--- a/src/superfile/format/footer.rs
+++ b/src/superfile/format/footer.rs
@@ -25,7 +25,7 @@
 
 use std::{
     collections::HashMap,
-    io::{self, Cursor, Read, Write},
+    io::{self, BufReader, Cursor, Read, Seek, SeekFrom, Write},
     sync::Arc,
 };
 
@@ -44,6 +44,7 @@ use parquet::{
     },
     schema::types::ColumnPath,
 };
+use tempfile::NamedTempFile;
 
 use crate::superfile::{LazyByteSource, LazyByteSourceError, format::kv};
 
@@ -135,9 +136,13 @@ pub enum FooterError {
 /// holds everything that the CPU-bound column encode produced, ready for
 /// the cheap blob splice + footer rewrite.
 pub struct EncodedBody {
-    /// Parquet bytes truncated to the end of the last row group (footer
-    /// removed) — blobs and the rewritten footer get appended here.
-    buf: Vec<u8>,
+    /// Parquet body written to a scratch temp file and truncated to the end of
+    /// the last row group (footer removed). Streamed into the superfile at
+    /// splice time so the body — multi-GB at corpus scale — never materializes
+    /// as a whole `Vec`. Blobs and the rewritten footer get appended after it.
+    body_file: NamedTempFile,
+    /// Byte length of the footer-stripped body in `body_file`.
+    body_len: u64,
     /// Footer metadata decoded from the original write, carried through
     /// to the rewrite so row groups + column/offset indexes survive.
     metadata: ParquetMetaData,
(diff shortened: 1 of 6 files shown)


vinaykakade:
## Memory impact — compacting a 5M-doc corpus into one superfile

Measured on a 5.03M-doc Wikipedia corpus: ingest to ~12 segments (4 GiB commit flush), then `optimize()`-compact to a single superfile, on a 16 GiB-class host.

The number that governs OOM is **physical footprint (anon + dirty)** — on a large-memory machine total RSS is dominated by *reclaimable* file-backed pages (mmap'd inputs, on-disk scratch/spill) that evict under pressure, so it badly overstates the real requirement. Sampled per phase with `vmmap --summary <pid> | grep "Physical footprint"`.

| phase | metric | before (`main`, re-index) | after (this branch) |
|---|---|---|---|
| **compaction** | **anon** | **16.0 GiB** | **2.1 GiB** |
| compaction | total rss | 29.9 GiB | 27.3 GiB |
| ingest | anon | 9.3 GiB | 5.5 GiB |

**Compaction anon drops ~7.6× (16.0 → 2.1 GiB).** Before, the re-index merge re-tokenized every input and held the whole corpus of decoded rows + the output in RAM, sitting right at the 16 GiB ceiling (OOM-killed under real memory pressure). After, the k-way merge carries each input's already-built posting lists across (no re-tokenize) and the streamed row-group body drops each input's rows after writing — so the resident anon set is just the term interner + spill buffers + one input's batch. `lsof` confirmed postings/positions stream to disk (`infino_fts_postings.bin`, `fts_col0_part*.sorted.bin`).

Ingest anon also improves (9.3 → 5.5 GiB) from streaming the commit output to a temp file instead of an in-RAM `Vec`.

Note: total RSS barely moves — that's expected and is **not** the OOM signal; it's reclaimable file-backed memory on a no-pressure host. The anon column is what fits (or doesn't) in 16 GiB.

Query results are unchanged: the merged superfile returns byte-identical top-k and scores vs a fresh build, verified by the oracle + BM25 query tests in this PR.

src/superfile/format/footer.rs
```diff
@@ -163,48 +169,107 @@ pub fn encode_parquet_body(
...
+                .set_column_data_page_size_limit(ColumnPath::from((*col).to_string()), *limit);
         }
+        let props = props_builder.build();
+
+        // Encode to a scratch temp file rather than a `Vec`, so the body —
+        // multi-GB at corpus scale — never materializes whole in RAM. The
+        // writer owns an independent handle to the same file (`reopen`); the
+        // struct's `body_file` handle reads the footer back after close, and
+        // their cursors don't alias.
+        let body_file = NamedTempFile::new()?;
```
savannahar68, line 215: some OS mounts /tmp to tmpfs which is in RAM
so lets define a path here like we have done at couple of places in code 
https://discussion.fedoraproject.org/t/correct-location-for-temporary-files-to-save-write-endurance-on-nvme/105072/6

savannahar68: applies everywhere

savannahar68 (approved):
lgtm high level

Compaction side review is done, and looks good. 
Yet to take a deeper look at ingest/commit side changes

src/supertable/compaction/mod.rs
```diff
@@ -520,18 +523,50 @@ impl Supertable {
...
+            // Vec — the allocation that OOMs compaction on a memory-tight host.
+            // Mapped pages are file-backed and reclaimable; downstream publish
+            // takes `Bytes` unchanged (large superfiles already stream via
+            // put_multipart).
+            let mut output = NamedTempFile::new()
+                .map_err(|e| BuildError::Store(format!("merge temp create: {e}")))?;
+            let stats = {
+                let mut writer = BufWriter::new(output.as_file_mut());
+                let stats = if multi_cell && sq8_merge == Some(true) {
+                    SuperfileBuilder::build_from_multi_cell_sq8_ivf_readers_to(
```
savannahar68, line 537: can you add tests around this logic

src/superfile/format/footer.rs
vinaykakade: Great point, filed https://github.com/infino-ai/infino/issues/539, will take that up shortly

src/supertable/compaction/mod.rs
vinaykakade: Done

vinaykakade:
> lgtm high level
> 
> Compaction side review is done, and looks good. Yet to take a deeper look at ingest/commit side changes

Merging now - but feel free to suggest additional changes and will implement them

PR #533, opened 2026-08-05
node: null handling on append and HTTP status on hosted errors

ashishmishra26:
Two Node SDK fixes, both covered by new regression tests (`__test__/nulls.test.mjs`, `__test__/errors.test.mjs`; suite: 58 tests, 0 fail).

## Omitted row fields stored zero values instead of null

`Table.append(rows)` passed omitted keys as `undefined` to `vectorFromArray`, whose null set is `[null]` only — so an omitted nullable field silently stored the type's zero value (`Int32` 0, `LargeUtf8` "", `Float64` NaN, `Bool` false) and an omitted `Int64` threw. Omitted keys now normalize to `null` and store SQL NULL, identical to an explicit null.

Separately, an all-null nullable `Bool` batch was rejected outright: the Arrow JS IPC writer emits a zero-length values buffer when every value is null, which the Rust decoder refuses, and no client-side construction can avoid the writer shortcut. The addon now repairs such streams before decoding — the values-buffer entry is pointed at the column's validity buffer (an all-zero bitmap of exactly the needed size), a 16-byte in-place metadata edit — then the fully-validated decode runs as usual. Well-formed input never takes the repair path.

## Hosted API errors carried no machine-readable status

A failure from a hosted connection arrived as a plain `Error`: no HTTP status, and a 409 duplicate-create shared `err.code` with malformed-argument errors, so idempotent startup / retry logic had to regex message text. On hosted connections the wrapper now rethrows the same error tagged with `status` — the HTTP status the API actually returned (409 create conflict, 404 missing, 503 transient). The addon gives `AlreadyExists` its own message prefix (matching the existing `NotFound:` / `ConflictError:` convention). Errors on local connections are untouched and carry no `status`.

Binding-only change; the engine crate is untouched (no bench-relevant paths).

### DIFF
--- a/infino-node/__test__/errors.test.mjs
+++ b/infino-node/__test__/errors.test.mjs
@@ -1,0 +1,116 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Infino Authors
+//
+// Hosted-API error status. A failure the hosted service reported is
+// rethrown unchanged except for one added property: `status`, the HTTP
+// status the API actually returned, so callers branch on a number instead
+// of regexing message text. Local (embedded) connections are untouched.
+//
+// The hosted-path tests run a stub API server on a worker thread — the
+// SDK's calls are synchronous and would deadlock a same-thread server.
+
+import test from "node:test";
+import assert from "node:assert/strict";
+import { Worker } from "node:worker_threads";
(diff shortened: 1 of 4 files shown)


naman-modi (approved):
LGTM Ashish, thanks.

PR #538, opened 2026-08-05
supertable/query: width-independent per-cell shortlist depth (#537)

### LINKED ISSUE #537: supertable/vector: recall@10 decreases as nprobe increases at scale (≥5M) — global shortlist saturates before rerank
## Summary

At ≥5M documents, post-drain unfiltered vector **recall@10 decreases as the coarse cell-probe width (`nprobe`) increases** — probing more cells *drops* recall instead of holding or improving it. Widening the search should never reduce recall, so this is a correctness-adjacent quality bug. It is absent below ~5M (recall is flat across `nprobe`) and worsens monotonically with corpus size. This is the "inversion signature" already referenced in `unfiltered_width_sweep`'s docstring (see #479).

Default adaptive routing stays narrow enough that the *default* recall reads ~0.98, so this hides unless you widen the probe — but any forced-wide or higher-recall configuration collapses.

## Measurement

`recall_while_ingest`, 10M × 1024-dim, cosine, default grid policy (~3653 coarse cells), 100 held-out queries graded against exact fp32 top‑k, on `main` (17ef841). Recall@10 vs `nprobe` at each 1M checkpoint:

| docs | nprobe=2 | 4 | 8 | 16 | 32 | all (~3653) |
|---|---|---|---|---|---|---|
| 1M | 0.998 | 0.998 | 0.998 | 0.998 | 0.998 | 0.998 |
| 2M | 0.996 | 0.996 | 0.996 | 0.996 | 0.996 | 0.996 |
| 3M | 0.998 | 0.998 | 0.998 | 0.998 | 0.998 | 0.998 |
| 4M | 0.997 | 0.997 | 0.997 | 0.997 | 0.997 | 0.997 |
| 5M | 0.996 | 0.996 | 0.963 | 0.923 | 0.894 | **0.862** |
| 6M | 0.990 | 0.990 | 0.943 | 0.893 | 0.858 | **0.817** |
| 7M | 0.995 | 0.995 | 0.918 | 0.835 | 0.774 | **0.670** |
| 8M | 0.993 | 0.992 | 0.910 | 0.806 | 0.736 | **0.621** |
| 9M | 0.992 | 0.992 | 0.889 | 0.769 | 0.678 | **0.535** |
| 10M | 0.993 | 0.993 | 0.898 | 0.774 | 0.678 | **0.534** |

Narrow probing stays ~0.99 throughout; at 10M, probing all cells collapses recall to **0.534** — a ~0.46 loss purely from widening coverage.

## Root cause

The warm path builds a **single global shortlist capped at `k × rerank_mult` candidates, pooled across *all* probed cells**, and the candidates that fill it are selected by the **coarse 1-bit (RaBitQ) estimate**, *before* the accurate rerank. In `search_clusters_scan_async` the per-cell 1-bit survivors are surfaced and then "pooled across cells and superfiles, [and the] global top `k × rerank_mult`" is selected (`coarse_limit = k * rerank_mult`).

As `nprobe` grows, more cells pour candidates into that fixed-size shortlist. The 1-bit estimate is noisy, so false positives from the additional (wrong) cells that happen to score well in 1-bit **evict the true neighbours from the shortlist before rerank ever runs**. Onset at ~5M is where per-cell populations grow enough that the fixed budget can no longer retain the true neighbours once extra cells compete for it; it worsens with scale for the same reason.

**This is independent of the rerank codec** — the pollution is at the 1-bit shortlist *selection*, upstream of rerank.

## What it is not: `rerank_mult` at this magnitude is not the lever

Forcing a flat `rerank_mult = 256` (bypassing the drain-stamp-derived value from #497 — "a caller-set `rerank_mult` wins over the law") reproduces the collapse essentially unchanged:

| nprobe | default (law) @10M | rerank_mult=256 @11M |
|---|---|---|
| 2 | 0.993 | 0.992 |
| 4 | 0.993 | 0.989 |
| 8 | 0.898 | 0.870 |
| 16 | 0.774 | 0.719 |
| 32 | 0.678 | 0.601 |
| all | 0.534 | 0.443 |

So the adaptive rerank value introduced by #497 is not the cause; a fixed shortlist of `k × 256 = 2560` simply lacks the headroom to hold the true neighbours against the wide-`nprobe` flood.

## Reproduce

Run `recall_while_ingest` past ~5M and sweep `nprobe` from 2 up to the full cell count; recall inverts against `nprobe`. `INFINO_BENCH_RERANK_MULT=256` does not change the shape.

## Suggested direction

- **Per-cell shortlist** (structural fix): have each probed cell retain its own top-`k × rerank_mult` survivors and rerank per cell (or with a per-cell floor), so widening coverage *adds* candidates instead of diluting a shared pool.
- Alternatively, **scale the shortlist with `nprobe`** so the global budget grows with the number of probed cells.

## Related

- #479 — named this "inversion signature" (width-blind survivor budget) and added the diagnostic that detects it.
- #497 — moved `rerank_mult` to a drain-stamp-derived default; not the cause (see above), but relevant context.
- #515 — the *opposite* failure mode (adaptive width *under*-provisions on some real corpora); this issue is over-widening into a shared shortlist.

ekechinwokah:
Fixes #537.

## Problem

Recall fell monotonically as a caller widened `nprobe` at scale, while narrow probes held 0.99 — measured on the 10M dense grid (~3.6K cells): 0.995 at nprobe=2 down to 0.533 at all-cells. Grid-density-dependent, which is why 256-cell sweeps at the same doc count stayed monotone.

Root cause: an explicit caller `nprobe` went through the sweep's width-divided budget — each probed cell's retention is `k×rerank_mult×2/w`, so the deeper the caller probes, the thinner each cell's cut. The 1-bit in-cell ranking is too weak to concentrate the true neighbors into a thin cut: the measured ladder tracks retention depth almost mechanically (whole-cell retention serves 0.99, ~6% of a cell serves 0.85, 20 rows serves 0.51). An earlier version of this branch floored the pooled selection per cell — measured a no-op: the pool sits below the already-divided budget.

## Fix

Under an explicit caller `nprobe`, every probed cell keeps the full `k×rerank_mult` depth — the width divide stays law-arm-only, where the stamped budget genuinely is a total calibrated at the stamped width. The pooled selection's per-cell floor arms at the same depth. Widening adds cells at constant depth: monotone by construction, cost linear in the width asked for. `nprobe` is a testing surface; probing everything honestly costs a scan.

The law-served default path is untouched: stamps still decide width, fine depth, and rerank budget; the serving contract tests pin it, and the checkpoint default below is unchanged.

## Verification (5M docs, dense 3,294-cell grid, Azure, rm=256, 100 queries)

| nprobe | before (a24ec17 gate) | after |
|---|---|---:|
| 2 | 0.995 | 0.998 |
| 8 | 0.955 | 0.998 |
| 16 | 0.829 | 0.998 |
| 32 | 0.717 | 0.998 |
| all cells | 0.533 | **0.998** |
| oracle (rm=ceil(rows/k)) | — | **0.998** |

Default (law-served) checkpoint recall: 0.998, unchanged. The pre-fix 10M ladder is in the #537 thread.

## Oracle rung

`recall_while_ingest` gains `INFINO_BENCH_ORACLE=1`: all cells at rerank budget `ceil(rows/k)`, which with this fix exact-scores every live row through the stored rerank payloads — 1-bit selection bypassed, robust to cell-size skew (pre-fix the divide only guaranteed ~2× the average cell size, so skewed cells silently truncated even at oracle budgets). The ladder converging to it is the monotonicity proof, and the rung doubles as engine-side ground truth for tables without dataset GT (raw fp32 is never persisted, so stored-payload exact is the strongest engine-side truth). Already used on #542: nytimes-256 oracle = 0.9848 ⇒ that corpus's ceiling is Sq16 rerank precision, not 1-bit selection.

## Tests

lib 2172 / supertable 250 / superfile 135 green, `make check` clean. `vector_law_serving` pins the floor at the full per-cell budget (`floor == limit && floor > k`, so a revert to the k floor fails); the serving contract tests pin the default path unchanged.

### DIFF
--- a/src/supertable/query/vector.rs
+++ b/src/supertable/query/vector.rs
@@ -1993,7 +1993,30 @@ impl SupertableReader {
                 // cell, that cell's floor carries it into the exact rerank
                 // (replicas never collide inside one cell, so the floor
                 // needs no replica overhead).
-                let cell_floor = if options.nprobe.is_some() { k } else { 0 };
+                //
+                // (#537) The floor's DEPTH must not shrink as the caller
+                // widens. The global pool splits across probed cells, so a
+                // k-deep floor leaves each cell's retention to its own
+                // 1-bit top-k — and within-cell estimate noise evicts true
+                // neighbors from that at scale (measured: recall falls
+                // monotonically with nprobe past ~5M, 0.534 at all-cells
+                // on 10M, while narrow probes hold 0.99). Hold every
+                // probed cell at the depth the stamp certified for a
+                // served cell — the stamped rerank budget divided by the
+                // stamped width — so widening adds cells at full depth
+                // instead of diluting all of them. Rerank cost then scales
+                // with the width the caller asked for, which is the pin
+                // arm's stated semantics: read amplification is what was
+                // asked for.
+                let cell_floor = if options.nprobe.is_some() {
+                    let stamped_width = hidden_routing
+                        .and_then(|r| r.width_for_k_at(k))
+                        .unwrap_or(1)
+                        .max(1);
+                    k.max(k.saturating_mul(rerank_mult).div_ceil(stamped_width))
+                } else {
+                    0
+                };
                 let shortlist_limit = k
                     .saturating_add(replica_overhead)
                     .saturating_mul(rerank_mult);
--- a/src/supertable/query/vector.rs
+++ b/src/supertable/query/vector.rs
@@ -2067,6 +2067,28 @@ impl SupertableReader {
         };
         let mut cold_rerank_mult = 0;
         let options = match sweep_width {
+            // (#537) An explicit caller nprobe keeps the FULL per-cell
+            // budget — the divide below is for the law arm only. Divided,
+            // per-cell retention shrinks as the caller widens, and the
+            // 1-bit in-cell ranking is too weak to hold the true
+            // neighbors in a thin cut: on the 10M dense grid (~2.8K-row
+            // cells) measured recall tracks the divided depth down its
+            // ladder — whole-cell at w<=4 serves 0.993, w=16 (~6% of
+            // each cell) serves 0.85, all-cells (20 rows/cell) serves
+            // 0.51 — monotone in DEPTH, inverted in width. Undivided,
+            // widening adds cells at constant depth, so the sweep is
+            // monotone and read cost is linear in the width the caller
+            // asked for: explicit nprobe is a diagnostic surface, and
+            // "probe everything" honestly costs a scan.
+            Some(w) if w > 1 && options.nprobe.is_some() => {
+                if global_shortlist_width.is_some() {
+                    let (_, rerank_mult) = options.resolve(filtered);
+                    cold_rerank_mult = rerank_mult;
+                }
+                options
+            }
+            // Law arm: the stamped budget is calibrated as a TOTAL at the
+            // stamped width, so it divides across the sweep.
             Some(w) if w > 1 => {
                 let (_, rerank_mult) = options.resolve(filtered);
                 let divided = rerank_mult
@@ -2276,25 +2298,22 @@ impl SupertableReader {
                 // needs no replica overhead).
                 //
                 // (#537) The floor's DEPTH must not shrink as the caller
-                // widens. The global pool splits across probed cells, so a
-                // k-deep floor leaves each cell's retention to its own
-                // 1-bit top-k — and within-cell estimate noise evicts true
-                // neighbors from that at scale (measured: recall falls
-                // monotonically with nprobe past ~5M, 0.534 at all-cells
-                // on 10M, while narrow probes hold 0.99). Hold every
-                // probed cell at the depth the stamp certified for a
-                // served cell — the stamped rerank budget divided by the
-                // stamped width — so widening adds cells at full depth
-                // instead of diluting all of them. Rerank cost then scales
-                // with the width the caller asked for, which is the pin
-                // arm's stated semantics: read amplification is what was
-                // asked for.
+                // widens — and the depth that holds recall is the full
+                // per-cell budget, not a share of it. The measured 10M
+                // ladder (see the width-divide comment above the fan-out)
+                // tracks per-cell retention depth almost mechanically:
+                // whole-cell retention serves 0.993, ~6% of a cell serves
+                // 0.85, 20 rows serves 0.51 — in-cell 1-bit ranking is
+                // too weak to concentrate the true neighbors into a thin
+                // cut, so no fixed pool or stamped share survives a wide
+                // sweep. Under an explicit caller nprobe every scanned
+                // cell therefore keeps the same k x rerank_mult depth the
+                // narrow probe would give it: widening adds cells at
+                // constant depth, the exact rerank adjudicates, and cost
+                // is linear in the width the caller asked for — the pin
+                // arm's stated semantics.
                 let cell_floor = if options.nprobe.is_some() {
-                    let stamped_width = hidden_routing
-                        .and_then(|r| r.width_for_k_at(k))
-                        .unwrap_or(1)
-                        .max(1);
-                    k.max(k.saturating_mul(rerank_mult).div_ceil(stamped_width))
+                    k.saturating_mul(rerank_mult)
                 } else {
                     0
                 };
--- a/tests/supertable/vector_law_serving.rs
+++ b/tests/supertable/vector_law_serving.rs
@@ -12,7 +12,8 @@
 //!
 //! The same probe pins the floor scoping from the same PR: law-served
 //! defaults run floor-free; an explicit caller `nprobe` arms the floor at
-//! the stamped per-cell depth (rerank budget / stamped width, >= k — #537).
+//! the full per-cell budget (`k x rerank_mult`, so per-cell depth never
+//! shrinks as the caller widens — #537).
 
 #![deny(clippy::unwrap_used)]
 
@@ -146,10 +147,10 @@ async fn law_stamped_table_serves_the_law_not_the_constant() {
         "caller rerank_mult must override the law exactly; recorded: {recs:?}"
     );
 
-    // 3) Explicit caller nprobe re-arms the per-cell floor — at the
-    //    stamped per-cell DEPTH (#537: rerank budget / stamped width,
-    //    never below k), so widening the probe adds cells at full depth
-    //    instead of diluting a shared pool (the ≥5M recall inversion).
+    // 3) Explicit caller nprobe re-arms the per-cell floor — at the full
+    //    per-cell budget (#537: k x rerank_mult, width-independent), so
+    //    widening the probe adds cells at constant depth instead of
+    //    diluting a shared pool (the >=5M recall inversion).
     st.vector_search(
         "emb",
         &query,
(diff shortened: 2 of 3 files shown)


src/supertable/query/vector.rs
```diff
@@ -2274,7 +2274,30 @@ impl SupertableReader {
...
+                // served cell — the stamped rerank budget divided by the
+                // stamped width — so widening adds cells at full depth
+                // instead of diluting all of them. Rerank cost then scales
+                // with the width the caller asked for, which is the pin
+                // arm's stated semantics: read amplification is what was
+                // asked for.
+                let cell_floor = if options.nprobe.is_some() {
+                    let stamped_width = hidden_routing
+                        .and_then(|r| r.width_for_k_at(k))
+                        .unwrap_or(1)
```
muralikpbhat: **Unstamped fallback is backwards.** When there's no stamp (`hidden_routing` None, or `width_for_k_at(k)` None mid-lifecycle) `stamped_width` falls back to **1**, so `cell_floor = k·rerank_mult` — the *entire* budget retained per cell, times `nprobe` cells. Without a stamp you can't know the calibrated depth, so the safe fallback is **`k`** (the prior behavior), not the maximal depth. As written, an unstamped hidden table under explicit `nprobe` gets a surprise read-amplification blow-up. Recall-safe, but the wrong cost default.

tests/supertable/vector_law_serving.rs
```diff
@@ -157,7 +161,8 @@ async fn law_stamped_table_serves_the_law_not_the_constant() {
     .expect("pinned-nprobe search");
     let recs = served_shortlist_probe::drain();
     assert!(
-        recs.iter().any(|&(_, floor)| floor == K),
-        "explicit nprobe must arm the per-cell floor at k; recorded: {recs:?}"
+        recs.iter().any(|&(_, floor)| floor >= K),
```
muralikpbhat: **This assertion no longer guards the change.** Relaxing `floor == K` to `floor >= K` passes against the *old* `floor = k` code (k >= k), so it wouldn't catch a revert of this very PR. In this fixture the stamped depth is well above `k`, so it should assert `floor > K` (or the exact stamped depth `k·rerank_mult / stamped_width`) to actually pin #538's behavior.

muralikpbhat (commented):
Tested the fix end-to-end — #525 folded in plus #538 — against the #537 repro (the `recall_while_ingest` nprobe sweep across ingest checkpoints). **The inversion is still present; #538 does not fix it.**

| ckpt | np=1 | np=8 | np=16 | np=32 | all |
|---|---|---|---|---|---|
| ≤4M | 0.997 | 0.997 | 0.997 | 0.997 | 0.997 (flat) |
| 5M | 0.995 | 0.942 | 0.874 | 0.821 | 0.742 |

Below 5M recall is flat across the whole width (onset unchanged); at 5M it still collapses **0.995 → 0.742** as the probe widens — the exact #537 signature. The 5M all-cells point (0.742) is if anything below the prior no-#525 run (~0.912), though that's within grid-to-grid variance.

So the deeper per-cell floor isn't recovering the recall. Two candidate reasons:
1. The stamped per-cell depth `k·rerank_mult / stamped_width` is still too shallow for a ≥5M cell — the true neighbor ranks beyond it in that cell's own noisy 1-bit order.
2. The eviction isn't fixable by a width-derived floor at all: the floor's candidates are themselves selected by the same 1-bit estimates that are saturating, so deepening by `budget/width` doesn't reach the true rows.

Either way the target metric — recall flat across nprobe at ≥5M — isn't met, so this can't land as the #537 fix yet. Useful next step: the `served_shortlist_probe` already records the per-cell floor; confirm at the 5M checkpoint whether the deeper floor is actually arming. If it is arming and recall still inverts, that points at (2) — the fix is a per-cell shortlist sized to cell **population**, not `budget/width`.

Two notes on the change itself below.

src/supertable/query/vector.rs
ekechinwokah: reworked in 4ca336c — the stamp lookup is gone entirely. full k×rerank_mult per probed cell is now the deliberate depth, not a fallback: anything thinner inverts. the 5m dense-grid ladder is in the pr body — thin cuts read 0.53 at all-cells, full depth reads 0.998 flat, matching the oracle rung. explicit nprobe is a testing surface, so cost scales with the width you ask for; the k floor you're suggesting is the pre-fix behavior that produced the inversion.

tests/supertable/vector_law_serving.rs
ekechinwokah: good catch — tightened in 78bf91f. the probe now asserts floor == limit and floor > k (full per-cell depth, exactly what 4ca336c serves), so a revert to the k floor fails the test.

muralikpbhat (approved):
The reworked fix (`4ca336c` + `78bf91f`) resolves the review points, and the width-inversion is gone at scale.

**Review follow-ups addressed**
- The width-divided per-cell floor is gone: under an explicit `nprobe`, every probed cell now keeps the full `k * rerank_mult` depth (no `stamped_width` divide), so the earlier unstamped-fallback concern is moot.
- The test now pins `floor == limit && floor > K`, so a revert to the old `floor = k` fails rather than passing.

**Mechanism**
Recall tracks per-cell retention depth almost mechanically, and the divided budget was a thin cut the 1-bit in-cell ranking couldn't hold — hence the monotonic decline as `nprobe` widened. Keeping the full budget per probed cell (the width divide stays law-arm-only, where the stamped budget is a calibrated total) makes widening add cells at constant depth: the sweep is monotone and cost is linear in the width the caller asked for.

**Scope**
Gated on `options.nprobe.is_some()`; the law-served default arm is byte-identical (floor stays 0), so no default-path change and no interaction with the serve-window path.

**Verification — reworked head `78bf91f`, fresh clean-drain `recall_while_ingest`, 10M × 1024-d cosine synthetic (~3.6K cells), exact fp32 GT, 100 held-out queries**

recall@10, checkpoint × nprobe:

| docs | default | np1 | np2 | np4 | np8 | np16 | np32 | all |
|------|---------|-----|-----|-----|-----|------|------|-----|
| 1M | 0.998 | 0.998 | 0.998 | 0.998 | 0.998 | 0.998 | 0.998 | 0.998 |
| 5M | 0.998 | 0.998 | 0.998 | 0.998 | 0.998 | 0.998 | 0.998 | 0.998 |
| 10M | 0.993 | 0.993 | 0.993 | 0.993 | 0.993 | 0.993 | 0.993 | 0.993 |

Every checkpoint (1M–10M) is flat across the full width — widening `nprobe` no longer sheds recall.

Before vs after, all-cells column (#537's worst case):

| docs | main (before) | this PR (78bf91f) |
|------|---------------|-------------------|
| 5M | 0.86 | 0.998 |
| 7M | 0.67 | 0.994 |
| 9M | 0.53 | 0.991 |
| 10M | 0.534 | 0.993 |

The retention-depth-only predecessor (`22fa69f`, before the budget commit) did **not** fix this — a fresh 10M on it reproduced 0.533 all-cells. `4ca336c` is what fixes it.

**Two scope boundaries (recording, not blockers)**
1. This fixes the **explicit-nprobe** diagnostic surface. On this synthetic the law stamps width 1 (nprobe=1 already ≥0.99), so the default serving path never widened and was unaffected pre-fix — #537 manifested only under explicit widening. The law-served wide path (real corpora where the stamp is >1) is #525's territory, not exercised here.
2. The flat guarantee holds while cell population stays around/below `k * rerank_mult` (here ~3.6K cells at 10M, so full-budget ≈ whole-cell). At larger scale, where cells outgrow the fixed budget, the per-cell floor becomes a partial cut again — the durable fix there is a per-cell shortlist sized to cell population (or the oracle scan the new `INFINO_BENCH_ORACLE` rung measures as the ceiling).

Issue #542, opened 2026-08-06
nytimes-256: recall caps at ~0.982 even at all-cells — codec/shortlist ceiling, not width or rerank


muralikpbhat:
## Summary

`nytimes-256-angular` (256-d cosine, 290K vectors, ann-benchmarks GT) caps at **recall@10 ≈ 0.982 even when probing *all* IVF cells** — a **codec / shortlist ceiling**, not a routing (width) or rerank-budget shortfall. The stamped default lands 0.9733, already close to that all-cells ceiling.

## Data (latest `main` `17ef841`, k=10, cosine, ann-benchmarks GT — recall is the signal)

| config | recall@10 |
|---|---:|
| default (stamped) | 0.9733 |
| nprobe=64 | 0.8730 |
| **nprobe=256 / 512 (all cells)** | **0.9817** |
| rerank_mult=256 | 0.9702 |

(Warm p95 ≈ 12 ms from the uncontended roster run; the sweep above shared the box with another job, so ignore its latencies.)

## Analysis

- **Not width-bound** — probing every cell (nprobe ≥ grid size) still caps at 0.9817; width only recovers ~1 pt (0.973 → 0.982), then plateaus.
- **Not rerank-budget-bound** — `rerank_mult=256` doesn't help (0.9702, slightly lower).
- So the shortlist pipeline — **1-bit RaBitQ candidate *selection* + the ~16-bit `Sq16` rerank re-scoring** — cannot rank the last ~1.8% of true neighbors even with full candidate coverage.

Same class as **glove-25** (also codec-bound) but far milder. Contrast: most cosine datasets (Cohere / OpenAI / glove-50+ / deep-96) hit ~0.99 at the stamped default.

## Confirm / next step — isolates *which* stage

- **Rebuild with the `Fp32` rerank codec** and re-measure at all-cells:
  - reaches ~0.99 ⇒ the ceiling is the **Sq16 rerank precision** (the true neighbor is in the pool but mis-ranked by the 16-bit codec) → lever = rerank fidelity.
  - stays ~0.982 ⇒ the ceiling is the **1-bit RaBitQ candidate selection** (the true neighbor never makes the shortlist) → lever = a richer shortlist (larger/finer RaBitQ, per-cell shortlist).

Discovered in the ann-benchmarks recall sweep (#490); harness in retrievalbench#90.

ekechinwokah:
[...]

PR #545, opened 2026-08-06
config: remove the rerank_codec knob — codec is internal

ekechinwokah:
## What

Removes `vector.rerank_codec` from the YAML config surface. The stored rerank payload codec is an engine-internal decision: cosine columns use `Sq16`, unbounded metrics use locally fitted `Sq8Residual`. Per-column overrides remain a test-only surface (`VectorConfig::with_rerank_codec`, not part of the public spec).

## Why

A config layer that changes storage bytes is not a lever users should hold, and the knob had become actively misleading: `fp32` names storage the engine deliberately never produces — raw vectors are never stored fp32 (fp32 is intake and centroids only), verified empirically at drain scale during the #542 investigation.

## Notes

- Not in `public-api.txt` (never exported); `make public-api` clean, no API drift.
- The two config unit tests lose only their codec assertions; every other key they exercise is unchanged.
- Suites: lib 2172 / supertable 250 / superfile 135 green, `make check` clean.
- No behavior change for any default deployment: the embedded default was `sq16`, which is exactly what the internal default resolves to.

### DIFF
--- a/src/config/config.yaml
+++ b/src/config/config.yaml
@@ -129,18 +129,6 @@ compaction:
 # needs to touch this section. These are the levers that used to be
 # one-off INFINO_* env vars; they are YAML-only now.
 vector:
-  # Default rerank codec for cosine vector columns. Non-cosine metrics
-  # still use locally fitted sq8_residual (values are not bounded to
-  # [-1, 1]). Override per column at table-create time when needed.
-  #   sq16               — single 16-bit plane on the fixed [-1, 1] grid
-  #                        (default; finer grid than sq8_fixed_residual at
-  #                        the same 2 bytes/dim, faster rerank)
-  #   sq8_fixed_residual — coarse u8 + i8 residual on the fixed grid
-  #   sq8_residual       — per-cluster fitted quantizer
-  #   fp32               — full-precision rerank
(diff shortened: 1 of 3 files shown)


muralikpbhat (approved):
Approving. Clean removal of an engine-internal knob.

- **Behavior-identical for default deployments.** `default_rerank_codec_for(Cosine)` moves from `config::global().vector.rerank_codec` to `RerankCodec::default()`, and the embedded config default was already `RerankCodec::default()` (`Sq16`) — so a default deployment resolves to exactly the same codec.
- **No table-identity / reopen impact.** The options-hash already excludes `rerank_codec` (`options_hash.rs` `compute_options_hash_ignores_rerank_codec`); codec is data-determined by the on-disk `codec_id` and dispatched by the reader. Removing the config field changes no table's identity and no existing table's reopen.
- **No dangling references** — nothing else reads `config.vector.rerank_codec`, and no config.yaml sets a non-default value.
- **No downstream consumers.** Checked across the sibling repos: none reference `rerank_codec` (it was never a binding or config surface anyone consumes), so this is self-contained.
- No public-api drift (never exported); the two config unit tests correctly drop only their codec assertions; CI (Format/Build/Test, Public API, license) green.

PR #546, opened 2026-08-06
catalog: remove vector tuning knobs from the public API

ekechinwokah:
## What

`VectorSearchOptions` (`nprobe` / `rerank_mult`) leaves the public API entirely: `vector_search` and `hybrid_search` take no options parameter, the type is no longer re-exported at the crate root, the Python wheel drops the kwargs, and the Node binding drops them outright. Users cannot set vector tuning knobs on any shipped surface.

## Why

Serving is drain-calibrated — probe width, fine depth, and rerank budget are stamped per table and per `k`, and extended only on the query's own evidence. A caller tuning surface is not a lever users should hold; settings automation is the product.

## Test-and-bench access (unchanged capability, gated)

Manual overrides remain a diagnostic instrument behind `test-helpers`, which the public-api snapshot excludes: the crate-root re-export and new `vector_search_with_options` / `hybrid_search_with_options` catalog variants are only visible there (existing `test_visible!` pattern). Internal engine paths are untouched, so integration tests and benches keep per-call knobs — recall sweeps and the exact-scan oracle (#538) work as before. The Python side gains a `diagnostics` cargo feature (`= infino/test-helpers`) for building the sweep/oracle wheel; published wheels never enable it.

## Ripple

Examples, `crate-docs.md`, and the catalog-API / remote tests updated to the parameterless calls; `public-api.txt` (−25 lines, reviewed) and `api-parity.txt` regenerated.

## Tests

`make check` and `make public-api` clean; lib 2172 / supertable 250 / superfile 135 / catalog_public_api 2 green; Python binding compiles with and without `diagnostics`; Node binding compiles.

### DIFF
--- a/api-parity.txt
+++ b/api-parity.txt
@@ -18,12 +18,12 @@ Supertable::count	covered	(&self, &str, &str, BoolMode) -> Result<u64, InfinoErr
 Supertable::delete	covered	(&self, Expr) -> Result<MutationStats, InfinoError>
 Supertable::exact_match	covered	(&self, &str, &str, Option<&[&str]>) -> Result<Vec<RecordBatch>, InfinoError>
 Supertable::gc	covered	(&self, Duration) -> Result<GcReport, GcError>
-Supertable::hybrid_search	covered	(&self, &str, &str, BoolMode, &str, &[f32], VectorSearchOptions, usize, Option<&[&str]>) -> Result<Vec<RecordBatch>, InfinoError>
+Supertable::hybrid_search	covered	(&self, &str, &str, BoolMode, &str, &[f32], usize, Option<&[&str]>) -> Result<Vec<RecordBatch>, InfinoError>
 Supertable::optimize	covered	(&self, &OptimizeOptions) -> Result<(), OptimizeError>
 Supertable::schema	covered	(&self) -> SchemaRef
 Supertable::token_match	covered	(&self, &str, &str, BoolMode, Option<&[&str]>) -> Result<Vec<RecordBatch>, InfinoError>
 Supertable::update	covered	(&self, Expr, &RecordBatch) -> Result<MutationStats, InfinoError>
-Supertable::vector_search	covered	(&self, &str, &[f32], usize, VectorSearchOptions, Option<VectorFilter<'_>>, Option<&[&str]>) -> Result<Vec<RecordBatch>, InfinoError>
+Supertable::vector_search	covered	(&self, &str, &[f32], usize, Option<VectorFilter<'_>>, Option<&[&str]>) -> Result<Vec<RecordBatch>, InfinoError>
 connect	covered	(impl AsRef<str>) -> Result<Connection, InfinoError>
 connect_with	covered	(impl AsRef<str>, ConnectOptions) -> Result<Connection, InfinoError>
 
--- a/crate-docs.md
+++ b/crate-docs.md
@@ -36,7 +36,7 @@ use std::sync::Arc;
 
 use infino::arrow_array::{FixedSizeListArray, Float32Array, LargeStringArray, RecordBatch};
 use infino::arrow_schema::{DataType, Field, Schema};
-use infino::{connect, Bm25SearchOptions, BoolMode, IndexSpec, Metric, VectorFilter, VectorSearchOptions};
+use infino::{connect, Bm25SearchOptions, BoolMode, IndexSpec, Metric, VectorFilter};
 
 // Tiny stand-in for your embedding model so this runs as-is — a 16-dim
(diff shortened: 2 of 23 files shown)


muralikpbhat (commented):
Holding approval — the Rust side is clean, but the Python wheel has stale artifacts (CI is red), and there's a self-hosted gap worth resolving alongside the removal.

**Verified correct.** The catalog facade change is sound: parameterless `vector_search`/`hybrid_search` pass `VectorSearchOptions::default()` (bit-identical to the old `::new()`), the `test_visible!`-gated `*_with_options` variants keep the override reachable for tests/benches/the diagnostics wheel (the inner engine path the exact-scan oracle uses is untouched), and `public-api.txt`'s -25 lines are exactly the `VectorSearchOptions` surface. Node binding, examples, `crate-docs.md`, and `api-parity.txt` are consistent.

**Blocker - the Python wheel's non-source artifacts drifted; the Python test lane is red:**
- `infino-python/python/infino/_infino.pyi:77-78,111-112` still declares `nprobe`/`rerank_mult` on `vector_search`/`hybrid_search`. The stub ships inside the wheel, so a type-checker greenlights a call that then `TypeError`s against the knobless extension.
- `infino-python/tests/test_smoke.py:457,556` still call `vector_search(..., nprobe=8, rerank_mult=32)`; `make python-test` runs `maturin develop` without `--features diagnostics`, so pytest `TypeError`s - this is why `Python wheel & tests` is failing.
- `infino-python/tests/typing/quickstart.py:37` passes `nprobe=8` and only type-checks because the stale stub masks it.
- `infino-python/src/lib.rs:647-648` - the published (`cfg(not(diagnostics))`) `hybrid_search` docstring still promises "optional `nprobe` and `rerank_mult`" above a signature that dropped them.

All four (stub, smoke test, typing sample, docstring) need updating to the knobless surface; the diagnostics-gated twins were scrubbed correctly - only the published side drifted.

**Self-hosted gap.** The serving law (width/fine/rerank) is stamped at drain, which `optimize()` triggers. A table that's never optimized serves via the pre-drain path on fixed constants (`nprobe=6`, `rerank_mult=256`) with no adaptive law - and this PR removes the only override a self-hosted user had to compensate. For deployments that auto-optimize this is a non-issue, but a self-hosted user who never calls `optimize()` now gets fixed-default recall with no lever and no signal that `optimize()` is the missing step. Suggest either guaranteeing an automatic drain in the engine, or landing "call `optimize()` for calibrated recall - un-optimized tables serve on fixed defaults" prominently in `quickstart`/`crate-docs`/the binding READMEs alongside the knob removal.

**Informational (pre-existing).** The remote `Table` impl still discards options on the wire (`src/catalog/remote/table.rs:213,245` - `let _ = opts;`), so even under the diagnostics build a `*_with_options` call against a hosted table is a no-op. Harmless today; worth a tracked note.

ekechinwokah:
**Blocker**
thx. did a sweep and created downstream PRs, all dependent on this PR merging first.
https://github.com/infino-ai/VectorDBBench/pull/7
https://github.com/infino-ai/infino-platform/pull/661
https://github.com/infino-ai/langchain-infino/pull/7

**Self-hosted gap** pre-drain serving isn't a degraded mode. the standard 10m gate reads the never-optimized path directly: 0.998 recall@10 at those fixed defaults, the highest of all four lifecycle states, because pre-drain fans out per superfile and over-reads. that's also why it's a bit slower — the drain and the stamped laws exist to bring cost down (warm p50 8.9ms → 3.1ms at equal recall), not to rescue recall. so there was nothing for the knobs to compensate for on an un-optimized table, and removing them loses nothing there. if a real deployment surfaces a legitimate never-optimize use case with a recall gap, we address it in the engine at that point.

**Pre-existing**: now fixed. replaced silent discard with loud rejection, and added a regression test.

muralikpbhat (approved):
Approving. The Python blocker is resolved — the `.pyi` stub, smoke tests, and typing sample are updated to the knobless surface and all `Python wheel & tests` lanes are green. The remote-refuse change (`reject_remote_overrides`) cleanly turns the silent wire-drop into a loud error for diagnostics-build callers, which is better than the prior `let _ = opts;`. Rust core/gating, the `test_visible!` variants, and the `public-api.txt` -25 lines were already correct.

One doc-only follow-up: the published (`cfg(not(diagnostics))`) `hybrid_search` docstring at `infino-python/src/lib.rs:647-648` still reads "(with optional `nprobe` and `rerank_mult`)" above a signature that no longer accepts them — `help(table.hybrid_search)` on a shipped wheel would advertise removed kwargs. One-line trim; not blocking.

PR #548, opened 2026-08-06
fix runtime: re-use io runtime to prevent deadlocks

Pranav2612000:
### What does this PR do?
- Modifies the sync<->async bridge in the `open_table` and similar flows to re-use the io tokio runtime, instead of relying on the ambient runtime.

### Why do we need this change
- Since we run a lot of futures e.g catalog cache lock, disk_cache operations etc, scheduling them on the callers runtime will cause all threads of the callers runtime to block, causing none of the tasks being able to proceed, resulting in a deadlock. For e.g. On an N-worker runtime, N+1 concurrent cold opens of the same table deadlock
- This PR ensures that we scheduled them on the shared io thread. This ensures that the task which holds the lock is always runnable, breaking the cycle and preventing the deadlock.

### How has this been tested?
- integration tests added which deadlocks without this change

### DIFF
--- a/src/catalog/mod.rs
+++ b/src/catalog/mod.rs
@@ -487,8 +487,9 @@ impl Connection {
                     return Ok(handle);
                 }
 
-                let (body, _etag) = bridge_sync_to_async(read_catalog(root.as_ref()))
-                    .map_err(|e| e.with_context("open_table", Some(name)))?;
+                let (body, _etag) =
+                    bridge_on_runtime(read_catalog(root.as_ref()), &shared_io_runtime())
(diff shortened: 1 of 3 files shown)


savannahar68 (approved):
lgtm 
can we do the same for create_table and drop_table
they follow same pattern and we might end up in same state for concurrent calls there

Pranav2612000:
> lgtm can we do the same for create_table and drop_table they follow same pattern and we might end up in same state for concurrent calls there

Sure. Will do it in a separate PR later.

PR #558, opened 2026-08-07
vector: cache stable_id→local per gapped superfile for scalar projection

### LINKED ISSUE #556: projection=[id] on gapped superfiles is O(corpus)/query — reverse-lookup has no index; fix = read-time cache (take_by_local_doc_ids is already point-access)
## Summary

Projecting a scalar column (`projection=["id"]`, or any user column) alongside a vector search is **O(corpus) per query** on drained / gapped superfiles. The read machinery is *not* the problem — it's already a point read; the cost is a **reverse lookup (`stable_id` to `local`) with no acceleration structure, rebuilt from scratch on every query**.

Measured on the 10M `deep-image-96` set: `projection=["id"]` adds **~234 ms/query**, flat across `k` and `nprobe`; native `_id` (no projection) is ~30 ms. Phase timer: `lookup_user_placements_by_id` = 232 ms, `get_all_superfiles_loaded` = 0 ms (cached) — so it's the reverse-lookup scan, not manifest load.

## Why it's O(corpus)

A hit carries its `stable_id`, but not the local row it lives at. The `_id` column is stored in **row (local) order, not sorted by `stable_id`**, and there is **no `stable_id` to `local` index**. So `lookup_user_placements_by_id` calls `read_ids_for_locals(entry, locals = (0..n_docs), …)` — reading the **entire** `_id` column, building an in-memory `id -> local` map, and looking up the k targets. That inversion (plus its SipHash) is the 232 ms.

Crucially, the underlying read is **already point-access**: `SuperfileReader::take_by_local_doc_ids` (`reader.rs:726`) does a parquet **row-selection** decode — it sorts/dedups the requested locals into skip/select runs and decodes only the projected column for only those rows, skipping row-groups that hold none of them. So resolving the id for **k known locals is O(k)**. The O(corpus) is *purely* that we pass all `n_docs` locals because we don't know which ones hold the target `stable_id`s.

Contiguous superfiles already avoid this: `row_id_from_manifest_entry` gives `local = stable_id - id_min` arithmetically (O(1)), no scan. The pathology is specifically **gapped** superfiles (row order != id order).

## Fix options

| option | resolve | per-query | RAM | notes |
|---|---|---|---|---|
| **A. Contiguous ids** | arithmetic | O(k) | none | make each writer emit a contiguous id range so the fast path applies; best if the gaps are just multi-writer ingest striding (to be confirmed) |
| **B. Read-time cache of the reverse map** | hash | O(k) after warm | O(n)/table, peak-neutral | memoize the map we already rebuild every query; cache keyed by manifest version (same pattern as `centroid_section_cache`), re-scan only on miss; LRU-bound across tables |
| **C. Write-time on-disk index** | binary search | O(k*log n) | none | sorted `stable_id`-to-`local` sidecar built at write/drain; immutable, no RAM, scales to many tables |

**On B's memory cost — it's roughly peak-neutral.** The current path *already* materializes the full `_id` set in memory on every query: `read_ids_for_locals((0..n_docs))` decodes the entire `_id` column into a `Vec<i128>` and builds a transient `id -> local` map, then discards it. Caching that map **retains an O(n) structure we already allocate-and-throw-away each query** rather than adding a new one; the raw `_id` bytes are already resident in the disk/reader cache. So B is a strict win on compute and roughly break-even on peak memory — the marginal cost is holding the map (LRU-bounded) instead of rebuilding it.

## Recommendation

**Near-term: B (read-time cache).** Smallest change — no new on-disk format, no write-path touch — and per the note above it's peak-memory-neutral since we already load the full id set per query. Build it lazily at read time, key it by manifest version, invalidate on version bump, and LRU-bound it across tables.

If the gapping turns out to be just multi-writer ingest striding (to be confirmed), **A (contiguous ids)** avoids it at the source with zero RAM/index. **C (write-time index)** is the RAM-free, scale-out form for large multi-table serving — a later follow-on to B if the resident maps get too large.

Distinct from the Sq8 codec issues (#554 / #555) — this is the id-projection path, not the rerank quantizer.

muralikpbhat:
## What

`projection=["id"]` (or any scalar column) alongside a vector search was **O(corpus) per query** on drained tables. Each hit carries a `stable_id`; to return the user column we must find its `local` row, and on **gapped** (drained / cell-packed) superfiles — row order ≠ id order, no `stable_id → local` index — `lookup_user_placements_by_id` read the **entire** `_id` column of every gapped superfile and inverted it in memory, **on every query**. At 10M that was ~230 ms on top of a ~30 ms search, and it grew with each merge generation.

This memoizes the reverse lookup: **one immutable sorted `stable_id → local` index per gapped superfile, keyed by the superfile's (immutable) uri**, on a new `SupertableOptions::gapped_id_placement_cache`. `SuperfileReader::take_by_local_doc_ids` is already a parquet row-selection point read, so once the placement is resolved the projection is O(k).

## Design notes

- **Gapped-only.** Contiguous superfiles keep resolving by arithmetic (`local = stable_id − id_min`) and are **never cached** — an all-contiguous table caches nothing (zero RAM).
- **Per-uri, immutable ⇒ incremental + never stale.** Sealed superfiles are immutable, so an index keyed by uri is valid forever. A data change only builds the **new** superfiles' indexes; unchanged ones stay cached. No manifest-version invalidation, no full re-read on every commit.
- **Monotonic eviction, MVCC-safe.** The cache is pruned to the current fresh read list, but **only a manifest generation at least as new as the last prune may evict** (`pruned_through` watermark). Under MVCC, concurrent queries run against different live generations; without the watermark an older snapshot would drop maps a newer one just built (and vice-versa), ping-ponging rebuilds under load. Monotonic prune keeps the cache bounded to the live set *without* the thrash.
- **Deterministic, manifest-order probe.** Placements are filled by iterating gapped superfiles in **manifest order**, first-wins — so if an `_id` is resident in two live superfiles (an updated row's tombstoned-old copy + its live copy, both physical rows to the `_id` read), it resolves to the same superfile the pre-cache full scan did.
- **Per-uri single-flight.** Each superfile's index is built behind a `OnceCell`, so two concurrent cold queries for the same uri share one build instead of both paying it.
- **Reader-pool build.** The CPU inversion (sort) runs on `options.reader_pool` via `run_on_pool` (the house async→rayon bridge), not the tokio blocking pool — no oversubscription on a drained table's worth of concurrent builds.
- **Compact layout.** The index is sorted parallel arrays (`ids: Vec<i128>`, `locals: Vec<u32>`) with a binary-search probe — ~20 B/row resident (≈half a hashmap) and the build is a sort (bandwidth-bound).
- **Budget-reserved.** The build reserves its resident bytes against `ConnectionMemoryBudget` and **degrades gracefully** (builds unaccounted) if declined, rather than failing the query — the same fallback the vector scan's transposed-code cache uses. The reservation releases on eviction. Steady-state resident ≈ 20 B × (rows across live gapped superfiles); e.g. ≈ 200 MB for a fully-drained 10M table, freed as generations are superseded.

## Why a read-time cache, and why the write-time index is the real fix

A **write-time on-disk `stable_id → local` index** is the better eventual approach: O(k·log n) lookups, **zero resident RAM**, and none of the read-time surface above (no eviction, MVCC watermark, single-flight, or budget accounting). It's deferred here only because it's a bigger change — on-disk format + write path. This PR is the **hardened interim**: it never exceeds today's peak RAM, is bounded to the live set, needs no format change, and closes the concurrency/ordering edges — but the write-time index remains the destination for large multi-table serving, and this cache should retire when it lands.

## Validation (10M deep-image-96, drained → gapped)

The old path had no build step — **every** query, the first included, paid the full O(corpus) scan. The cache trades a slightly higher first query (a one-time build) for O(k) on every query after it.

| | before | after |
|---|---|---|
| **first** `projection=["id"]` query | ~290 ms | ~485 ms (one-time build) |
| **subsequent** `projection=["id"]` queries | ~290 ms / query | **56 ms / query** |
| native `_id` baseline | ~30 ms | ~30 ms |

> Latency figures were measured on the earlier variant (blocking-pool + hashmap). The steady-state p50 is probe-dominated and unchanged by the rework; the one-time **build** number will shift with the sorted-vec + reader-pool build and is pending a clean re-measure (the box is busy). A p95 row is likewise pending a clean run.

Correctness: projection-via-cache returns **byte-identical hit sets** to the native `_id` + one-time id map (0/200 mismatches over sampled queries). Recall unchanged — placements are identical, only how they're found changed. Unit test covers placement parity (gapped + contiguous), build-once/reuse, contiguous-never-cached, monotonic eviction (newer generation evicts; older never evicts the newer's maps).

Closes #556.

### DIFF
--- a/src/supertable/query/vector.rs
+++ b/src/supertable/query/vector.rs
@@ -81,7 +81,7 @@ use arrow::record_batch::RecordBatch;
 use arrow_array::{Array, Decimal128Array};
 use futures::future::try_join_all;
 use roaring::RoaringBitmap;
-use tokio::join;
+use tokio::{join, sync::OnceCell};
 use uuid::Uuid;
 
 use super::{
@@ -96,6 +96,7 @@ pub use crate::superfile::reader::VectorSearchOptions;
 use crate::test_helpers::{admit_trace, served_shortlist_probe};
 use crate::{
     config,
+    runtime_bridge::run_on_pool,
     runtime_metrics::op_stats::{self, OpStatsCollector},
     storage::io_counters,
     superfile::{
@@ -118,6 +119,7 @@ use crate::{
             list::{CellRoutingParams, PartitionStrategy},
         },
         opann::REPLICA_CLOSURE_DISTANCE_RATIO,
+        options::{GappedPlacementCell, GappedPlacementIndex},
         slow_vector_state::{CentroidSection, fetch_centroid_section},
         tombstones::SidecarCache,
     },
@@ -861,8 +863,13 @@ async fn lookup_user_placements_by_id(
         .map_err(QueryError::ManifestLoad)?;
     let mut placements: Vec<Option<(Arc<SuperfileEntry>, u32)>> = vec![None; user_row_ids.len()];
     let mut gapped = Vec::new();
+    // Uris in the current fresh read list, to evict cache entries for superfiles
+    // that have since been superseded/GC'd (keeps the cache bounded to the live
+    // set, no process-lifetime growth).
+    let mut live_uris: HashSet<SuperfileUri> = HashSet::new();
 
     for entry in entries {
+        live_uris.insert(entry.uri);
         let matching: Vec<usize> = user_row_ids
             .iter()
             .enumerate()
@@ -889,35 +896,67 @@ async fn lookup_user_placements_by_id(
         }
     }
 
-    let decoded = try_join_all(gapped.into_iter().map(|entry| async move {
-        let locals: Vec<u32> = (0..entry.n_docs as u32).collect();
-        // Cell-packed user files carry stable ids inline in Parquet row order.
-        // Read each compact cell region once instead of decoding the full
-        // Parquet `_id` column to place a top-k scalar projection.
-        let ids = read_ids_for_locals(manifest, &entry, &locals, id_column, true, op_stats).await?;
-        Ok::<_, QueryError>((entry, ids))
-    }))
-    .await?;
-    let mut requested: HashMap<i128, Vec<usize>> = HashMap::new();
-    for (index, &id) in user_row_ids.iter().enumerate() {
-        if placements[index].is_none() {
-            requested.entry(id).or_default().push(index);
-        }
-    }
-    for (entry, ids) in decoded {
-        for (local, id) in ids.into_iter().enumerate() {
-            let Some(indexes) = requested.get(&id) else {
-                continue;
-            };
-            let local = u32::try_from(local).map_err(|_| {
-                QueryError::Execute(format!(
-                    "local_doc_id out of range in user superfile {:?}",
-                    entry.uri
-                ))
-            })?;
-            for &index in indexes {
-                if placements[index].is_none() {
-                    placements[index] = Some((Arc::clone(&entry), local));
+    // Resolve ids still unplaced (they live in GAPPED superfiles) via a
+    // read-time cache: one immutable sorted `stable_id -> local` index per
+    // gapped superfile, keyed by (immutable) uri. Contiguous entries were
+    // placed by arithmetic above and are never cached, so an all-contiguous
+    // table caches nothing. A data change only builds the NEW superfiles;
+    // unchanged ones stay cached and are never stale (#556).
+    if !gapped.is_empty() {
+        let slot = Arc::clone(&manifest.options.gapped_id_placement_cache);
+        let version = manifest.get_manifest_id();
+        // Under one short lock: prune MONOTONICALLY (only a generation at least
+        // as new as the last prune may evict, so an older concurrent snapshot
+        // never drops a newer snapshot's freshly-built maps — which would thrash
+        // rebuilds under load), then take/create the per-uri single-flight cell
+        // for each gapped superfile. Holding the `Arc<OnceCell>` keeps a
+        // concurrent prune from dropping the cell out from under the build.
+        let cells: Vec<(Arc<SuperfileEntry>, GappedPlacementCell)> = {
+            let mut cache = slot.lock().await;
+            if version >= cache.pruned_through {
+                cache.entries.retain(|uri, _| live_uris.contains(uri));
+                cache.pruned_through = version;
+            }
+            gapped
+                .iter()
+                .map(|e| {
+                    let cell = Arc::clone(
+                        cache
+                            .entries
+                            .entry(e.uri)
+                            .or_insert_with(|| Arc::new(OnceCell::new())),
+                    );
+                    (Arc::clone(e), cell)
+                })
+                .collect()
+        };
+        // Build outside the lock. Each cell single-flights its superfile's index
+        // (a second concurrent cold query for the same uri awaits this build,
+        // not a duplicate one); the CPU inversion runs on the reader pool, not
+        // the tokio blocking pool. `try_join_all` preserves `cells` order, so
+        // the probe below stays deterministic in manifest order.
+        let built: Vec<(Arc<SuperfileEntry>, Arc<GappedPlacementIndex>)> =
+            try_join_all(cells.into_iter().map(|(entry, cell)| async move {
+                let index = cell
+                    .get_or_try_init(|| {
+                        build_gapped_placement_index(manifest, &entry, id_column, op_stats)
+                    })
+                    .await?;
+                Ok::<_, QueryError>((entry, Arc::clone(index)))
+            }))
+            .await?;
+        // Probe in manifest (gapped) order: if an id is resident in two live
+        // superfiles — an updated row's tombstoned-old copy plus its live copy,
+        // both physical rows to `read_ids_for_locals` — first-wins resolves it to
+        // the same superfile the pre-cache full scan did.
+        for (entry, index) in &built {
+            for (i, &id) in user_row_ids.iter().enumerate() {
+                if placements[i].is_none()
+                    && id >= entry.id_min
+                    && id <= entry.id_max
+                    && let Some(local) = index.local_for(id)
+                {
+                    placements[i] = Some((Arc::clone(entry), local));
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/supertable/query/vector.rs
```diff
@@ -884,35 +889,79 @@ async fn lookup_user_placements_by_id(
...
+        // list, take the `Arc` for each gapped map already cached (holding it, so a
+        // concurrent prune can't drop it out from under us), and collect the rest
+        // as `missing`. Holding the Arcs means we never re-fetch from the cache
+        // after building, closing the insert-then-reprune race.
+        let (mut maps, missing): (
+            Vec<(Arc<SuperfileEntry>, Arc<PlacementMap>)>,
+            Vec<Arc<SuperfileEntry>>,
+        ) = {
+            let mut g = slot.lock().await;
+            g.retain(|uri, _| live_uris.contains(uri));
```
ekechinwokah: this prunes to THIS query's snapshot, but under mvcc queries on different manifest generations run concurrently (every commit/optimize transition). a query on the older snapshot evicts the maps the newer one just built, and vice versa — alternating ~485ms rebuilds and ram churn exactly when the table is busiest. prune monotonically instead: remember the newest manifest version that pruned, and only retain against a live set at least that new (or retain the union across in-flight snapshots).

src/supertable/query/vector.rs
```diff
@@ -884,35 +889,79 @@ async fn lookup_user_placements_by_id(
...
+                    Some(m) => maps.push((Arc::clone(e), Arc::clone(m))),
+                    None => missing.push(Arc::clone(e)),
+                }
+            }
+            (maps, missing)
+        };
+        if !missing.is_empty() {
+            // Reads overlap; each CPU-bound inversion runs on the blocking pool, so
+            // the build parallelizes across superfiles.
+            let built: Vec<(Arc<SuperfileEntry>, Arc<PlacementMap>)> =
```
ekechinwokah: two concurrent cold queries both miss here and both pay the full build for the same uri — or_insert throws one away, so it's ram x2 and cpu x2 transiently, on the exact query that's already slowest. a per-uri single-flight slot (map to a OnceCell/watch instead of the bare map) makes the second query wait ~free.

src/supertable/query/vector.rs
```diff
@@ -884,35 +889,79 @@ async fn lookup_user_placements_by_id(
...
+                try_join_all(missing.into_iter().map(|entry| async move {
+                    let locals: Vec<u32> = (0..entry.n_docs as u32).collect();
+                    let ids =
+                        read_ids_for_locals(manifest, &entry, &locals, id_column, true).await?;
+                    let m = tokio::task::spawn_blocking(move || {
```
ekechinwokah: house concurrency contract puts cpu waves on options.reader_pool (rayon) bridged with a oneshot, not spawn_blocking — try_join_all here can fan a drained table's worth of inversions onto the tokio blocking pool (default cap 512 threads) and oversubscribe the box. same home a par_sort would want if you take the sorted-vec layout: build Vec<(i128,u32)>, par_sort_unstable by id, binary-search probes. that also halves the resident cost per row and turns the build bandwidth-bound. and either reserve the build against ConnectionMemoryBudget or note why it's exempt — on a fully drained 10M table this cache is one entry per corpus row, ~300-400MB resident with the hash layout, and the pr body doesn't report the rss delta.

src/supertable/query/vector.rs
```diff
@@ -884,35 +889,79 @@ async fn lookup_user_placements_by_id(
...
-                QueryError::Execute(format!(
-                    "local_doc_id out of range in user superfile {:?}",
-                    entry.uri
-                ))
-            })?;
-            for &index in indexes {
-                if placements[index].is_none() {
-                    placements[index] = Some((Arc::clone(&entry), local));
+        // Probe the held maps lock-free.
+        for (entry, m) in &maps {
```
ekechinwokah: probe order is now cache-state-dependent: maps holds cached entries first, then freshly built ones, and first-wins fills placements. an updated row's _id physically exists in two live superfiles (tombstoned old copy + live copy — read_ids_for_locals sees all physical rows; the bool is only allow_inline_region), and the old code resolved that deterministically in manifest order. either document at this loop why hit ids can't collide across live files on this path, or probe in gapped (manifest) order via a uri-keyed lookup — one line.

src/supertable/query/vector.rs
```diff
@@ -4383,6 +4432,208 @@ mod tests {
...
+            partition_key: Vec::new(),
+            partition_hint: None,
+            vector_layout: crate::superfile::vector::layout::VectorLayout::Ivf,
+            subsection_offsets: None,
+        })
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn gapped_id_placement_cache_places_reuses_and_evicts() {
+        use crate::supertable::manifest::ManifestSnapshot;
```
ekechinwokah: convention nits in the new test code: function-local use (this line) belongs at the top of the test module, and the inline fully-qualified paths (crate::supertable::options::DECIMAL128_PRECISION, std::collections::HashMap::new(), ...VectorLayout::Ivf) should be imports. same for rustc_hash::FxBuildHasher in the options.rs field type and the fn body — use rustc_hash::FxBuildHasher; and say FxBuildHasher.

ekechinwokah (commented):
direction is right and the shape is right: read-time memo keyed by immutable uri, gapped-only so an all-contiguous table caches nothing, demand-driven, no write-time index. holding the arcs out of the first lock genuinely closes the insert-then-reprune race, and the test is good for a cache pr (placement parity, ptr_eq reuse, eviction).

muralikpbhat:
Thanks @ekechinwokah — all addressed in bc62cea.

1. **Probe order (id resident in two live superfiles).** You're right: the tombstoned-old + live copy coexist before GC and `read_ids_for_locals` sees both physical rows, so cache-state order could place a hit to the stale copy. Now filled in **manifest order, first-wins** (uri-keyed lookup over `gapped`), matching the pre-cache scan. Placement-parity test added.

2. **MVCC eviction thrash.** Added a monotonic prune — the cache carries a `pruned_through` watermark and only a generation ≥ it may evict, so an older concurrent snapshot never drops maps a newer one just built. Test asserts both directions (newer evicts a superseded superfile; older never evicts the newer's maps).

3. **Concurrency contract + layout + budget.**
   - Build now runs on `options.reader_pool` via `run_on_pool`, not `spawn_blocking`.
   - Sorted parallel arrays (`ids: Vec<i128>` + `locals: Vec<u32>`, binary search) — ~20 B/row, build is a sort.
   - Reserves against `ConnectionMemoryBudget` and degrades to an unaccounted build if declined (same fallback as the transposed-code cache); reservation releases on eviction. Steady-state resident ≈ 20 B × live gapped rows (≈200 MB on a fully-drained 10M table). I couldn't get a clean RSS / rebuilt-latency re-measure yet (box busy) — flagged in the description; the earlier table's one-time **build** number will shift with the reader-pool/sorted build and is pending a clean run.

4. **Single-flight.** Per-uri `OnceCell` now coalesces concurrent cold builds.

5. **Style.** Hoisted the function-local `use`; `rustc_hash` is gone here entirely (sorted vec, no hasher). The remaining fully-qualified paths in the test (`DECIMAL128_*`, `VectorLayout::Ivf`) match the neighboring `build_oracle_superfile` / `synthetic_entry` in that module — happy to hoist those too if you'd rather flip the convention.

Bigger picture: I agree the **write-time on-disk index is the right eventual fix** — O(k·log n), zero resident RAM, and none of this read-time surface (eviction, MVCC watermark, single-flight, budget). This PR is the hardened interim; the description now says explicitly that the cache should retire when the index lands.

src/supertable/query/vector.rs
muralikpbhat: Fixed in bc62cea — the prune is now **monotonic**. `GappedIdPlacementCache` carries a `pruned_through` watermark and only a generation `>=` it may `retain`; an older concurrent snapshot no longer evicts maps a newer one just built. The test asserts both directions (newer evicts a superseded uri; older leaves the newer's maps intact). Went with the watermark rather than a union-across-in-flight-snapshots since we don't track the in-flight set anywhere — shout if you think that's not enough.

src/supertable/query/vector.rs
muralikpbhat: Fixed in bc62cea — the slot is now `HashMap<SuperfileUri, Arc<OnceCell<Arc<GappedPlacementIndex>>>>`. The second concurrent cold query for a uri awaits the first's build via `get_or_try_init` instead of duplicating it; errors aren't cached (the cell is left empty for a retry).

src/supertable/query/vector.rs
muralikpbhat: All three done in bc62cea:
- **reader_pool**: the build (read + sort) runs via `runtime_bridge::run_on_pool(Some(&reader_pool), …)`, off the tokio blocking pool.
- **layout**: switched to sorted parallel arrays (`ids: Vec<i128>` + `locals: Vec<u32>`, binary-search probe) — ~20 B/row and the build is a sort, so no hashmap and roughly half the resident cost.
- **budget**: reserves the resident bytes against `ConnectionMemoryBudget` and degrades to an unaccounted build if declined (mirrors the transposed-code cache); the reservation releases on eviction.

The RSS delta + rebuilt one-time-build latency still need a clean re-measure (box is busy) — flagged in the PR description; the earlier build number was on the old blocking-pool/hashmap variant.

src/supertable/query/vector.rs
muralikpbhat: Fixed in bc62cea. You're right — the tombstoned-old + live copy coexist pre-GC and `read_ids_for_locals` sees both physical rows. Placements are now filled by iterating `gapped` in **manifest order, first-wins** (`try_join_all` preserves input order), so an id resident in two live superfiles resolves to the same one the pre-cache full scan did. Covered by the placement-parity test.

src/supertable/query/vector.rs
muralikpbhat: Fixed in bc62cea — hoisted the function-local `use` to the test module, and `rustc_hash` is gone from this change entirely (sorted vec, no hasher, so the `FxBuildHasher` references in options.rs + the build both went away). I left the `DECIMAL128_*` / `VectorLayout::Ivf` fully-qualified paths to match the adjacent `build_oracle_superfile` / `synthetic_entry` in the same module — happy to flip the whole module to imports if you'd rather have it the other way.

ekechinwokah (approved):

PR #560, opened 2026-08-07
fts: speed up deep-k dominant-term union top-k

vinaykakade:
## Problem

The multi-term OR dispatcher routed any union with a dominant score
upper-bound term to MaxScore + Block-Max MaxScore (BMM), on the
assumption that a dominant term lets block-max pruning skip the common
term's long posting list. That holds only while the top-k threshold
stays above the common term's score upper bound — i.e. while the rarer
terms alone can fill the heap.

Once `k` reaches the combined df of every term but the longest list, the
heap must admit docs from that longest list's tail (docs whose only
matching term is the common one). The threshold collapses toward the
common term's low upper bound, no block clears the skip test, and BMM
degrades to a **scalar full union scan**. This is the worst case for
deep-`k` "common + rare" unions.

## Change

Route exactly that case — pruning dead at this `k`, and the dominant list
long — to the SIMD windowed scorer, which does the same full scan without
the per-doc f-way merge:

- `or_topk_pruning_ineffective` detects the dead-pruning case from df
  alone (no extra reads): reroute when `k >= (total_df - max_df)`.
- Gate the reroute on a dominant-list floor (`max_df >= 100k`). Once
  pruning is dead both scorers scan the whole union, but the windowed
  scorer's fixed per-window setup (a 4096-wide score accumulator + a
  presence bitset) only amortizes on a long dominant list; on a short
  union it is pure overhead and MaxScore's scalar scan wins. Short unions
  stay on MaxScore.

Small-`k` dominant-term unions are untouched and stay on MaxScore, where
pruning still wins. This is a routing change only — no scoring math
changes, so results and scores are identical (the windowed scorer already
agrees with BMM on the brute-force BM25 oracle).

## Results (public benchmark, single compacted superfile, single-threaded, min-of-10)

infino's own before → after (this branch vs `main`), union top-k. Ratios
are branch/main, **< 1 = faster**:

| Union TOP_1000 | main | branch | ratio |
|---|---:|---:|---:|
| dominant list > 1M postings | 552 ms | 420 ms | **0.76× (24% faster)** |
| overall union | 1190 ms | 1074 ms | **0.90×** |

The win concentrates where the dispatcher used to fall off pruning: deep
`k` over a corpus-scale dominant list. Small unions, all other query
shapes (term, intersection, phrase), and all COUNT modes are within
run-to-run noise.

The dominant-list floor closes a small-union regression that an unfloored
reroute introduced — confirmed on a same-machine A/B (tight noise floor):
`< 100k` unions at TOP_1000 went 1.05× → **1.01×** while the `> 1M` win
was preserved (0.88× on that machine).

## Tests

- Brute-force BM25 top-K oracle and the windowed-vs-MaxScore agreement
  tests stay green (routing must not change which docs/scores return).
- Added `deep_k_dominant_union_reroutes_only_when_list_is_long`, pinning
  the boundary decisions: deep vs shallow `k`, single term, and
  below/at the dominant-list floor.

### DIFF
--- a/src/superfile/fts/reader.rs
+++ b/src/superfile/fts/reader.rs
@@ -347,6 +347,15 @@ fn prefer_windowed_union(cursors: &[TermCursor]) -> bool {
     cursors.len() >= OR_WINDOW_MIN_TERMS && no_dominant_term_ub(cursors)
 }
 
+/// Minimum dominant-term df for the deep-`k` reroute to the windowed
+/// scorer to pay off. Below this the union is small enough that
+/// MaxScore's scalar full scan beats the windowed scorer's fixed
+/// per-window setup, so short unions stay on MaxScore. Calibrated on the
+/// warm FTS bench: the win concentrates on dominant lists in the millions
+/// (~corpus-scale), the mid range is a wash, and lists shorter than this
+/// regressed. See [`or_topk_pruning_ineffective`].
+const OR_WINDOWED_MIN_DOMINANT_DF: u64 = 100_000;
+
 /// True when block-max pruning (MaxScore / WAND) can no longer skip
 /// blocks at this `k`, so both degrade to a full union scan — at which
 /// point the SIMD windowed scorer does that same scan far faster than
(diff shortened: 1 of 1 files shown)


src/superfile/fts/reader.rs
```diff
@@ -7433,6 +7501,51 @@ mod tests {
         );
     }
 
+    #[test]
+    fn deep_k_dominant_union_reroutes_only_when_list_is_long() {
```
naman-modi, line 7505: Here, every case here has rest_df=500 and k is either 1000 (above) or 100 (below), so k never equals rest_df. If we want, we can also add  the exact k == rest_df boundary for surety, else lgtm.

naman-modi (approved):
LGTM, Vinay, minor comment.

src/superfile/fts/reader.rs
vinaykakade: Done

Issue #554, opened 2026-08-07
Compaction/drain merge sizes the destination Sq8 grid from a subset, not the union — components clamp (silent recall loss) on L2/NegDot at scale


muralikpbhat:
## Summary

During compaction merges/splits (and drain re-encodes), rows are transcoded into a **destination Sq8 grid chosen from a subset of the merged cells, not their union**. For the **data-derived L2 / NegDot** codecs the destination grid can then fail to cover the merged inputs' range, so out-of-range components are **clamped (saturated)** — silent recall loss on affected rows.

Observed at **100M × 768-d, L2** (LAION): the tripwire fires on **every** drain/compaction pass, ~**36M components clamped per pass**.

## Evidence

```
[supertable compaction] BUG: 36130625 component(s) saturated their destination Sq8
quantizer during this pass's merges/splits (#512 failure mode) — a destination grid
failed to cover its inputs; affected rows' recall silently degrades.
```

Five consecutive drain/compaction passes at 100M L2, ~36.0–36.3M components each.

## Root cause

- The **fresh-build** path (`cell_posting.rs`, `Sq8Residual` arm) sizes the grid correctly: per-dim `min`/`max` over **all rows**, then `derive_sq8_quantizer_from_min_max` — covers its inputs by construction.
- The **merge/split transcode** kernel (`cell_posting::residual_family_materialize_into_cluster_quant`) takes a **caller-supplied destination grid** (`dst_scale`/`dst_offset`): if source grid == dest grid it byte-copies; otherwise it dequantizes to fp32 and **re-quantizes into the dest grid, clamping out-of-range components**.
- The merge/split callers (`supertable::compaction`, drain in `supertable::writer`) supply a destination grid derived from a **subset** (a single cell / the destination centroid), not the union of the merged cells. When other merged cells' re-centered residuals exceed that grid → clamp.

## Why this is distinct from #512 (cosine)

Cosine uses a **fixed `[-1,1]` grid** (portability); it clamps by design when inputs aren't unit-normalized — that's #512, which was closed as "normalize-or-reject inputs." There's nothing to *size* on a fixed grid.

**L2/NegDot grids are data-derived and *can* be sized** — so a destination grid that fails to cover its inputs is a genuine, fixable bug here, not an inherent limitation. The shared tripwire just reuses #512's wording.

## Proposed fix

[...]

Issue #490, opened 2026-07-31
Improving vector performance against real-world datasets


muralikpbhat:
## Motivation

Vector benchmarking today uses a **synthetic planted-cluster corpus** (cosine, Gaussian cluster centers with σ=0.3 per-dim noise). Real embeddings — e.g. a **768-d Cohere dataset** (embed-multilingual, wikipedia-22-12) — have very different geometry: clusters are far less separable, dimensions are correlated, and the query/nearest-neighbor structure is harder. Recall and cell routing behave differently on real data, so the synthetic numbers (recall ~0.99) are likely optimistic.

Goal: establish **real-world-dataset baselines** (starting with a 768-d Cohere dataset at 1M, then larger scales), compare against the synthetic reference below, and drive improvements where real data regresses — recall and cold-search cost being the primary targets.

## Current synthetic baselines (reference)

**1M, 768-d** — dim-matched to the Cohere target; single build → drain → compact, local store:

| metric | value |
|---|---|
| recall@10 | 0.994 (post-compact; 0.991 pre-drain) |
| warm p50 | 0.64 ms (post-drain) / 0.77 ms (post-compact) |
| cold 1st / steady | 6.01 ms / 1 GET · 5.5 MiB per query |
| ingest / drain / optimize | 83.1 s / 22.5 s / 99.6 s |
| footprint | 6.06 GiB (1.1M docs) |

**1M, 1024-d** — for reference:

| metric | value |
|---|---|
| recall@10 | 0.995 |
| warm p50 | 0.70 ms (post-drain) / 0.91 ms (post-compact) |
| cold 1st / steady | 9.06 ms / 1 GET · 7.0 MiB per query |
| ingest / drain / optimize | 91.2 s / 36.4 s / 132.4 s |
| footprint | 7.41 GiB (1.1M docs) |

**100M, 1024-d** — cold reopen of a kept table, single 64 GiB host, object storage:

| metric | value |
|---|---|
| recall@10 | 0.989 |
| warm p50 / cold p50 | 34.6 ms / 1.06 s |
| steady cold | 10 GET · 55 MiB per query |
| blended cost | $4.19 / 1M docs (~$14.68/mo) |
| footprint | 424.9 GiB |

## Plan

1. Run the lifecycle bench on a real **768-d** dataset (Cohere embed-multilingual, e.g. wikipedia-22-12) at 1M — apples-to-apples against the **768-d synthetic** baseline above.
[...]

PR #563, opened 2026-08-07
vector: add Sq16Adaptive rerank codec (16-bit fitted ruler), default for L2/NegDot

### LINKED ISSUE #555: Replace Sq8Residual with Sq16Adaptive for L2/NegDot — 16-bit per-cluster variable range, no residual leg (storage-neutral)
## Summary

Replace the L2 / NegDot rerank codec **`Sq8Residual`** with a new **`Sq16Adaptive`**: a single 16-bit scalar quantizer over a **per-cluster, data-derived (variable) range**, with **no residual/epsilon leg**. Storage-neutral, simpler, and — when the range is sized from the union of merged inputs — clamp-free by construction.

## Motivation

- L2/NegDot today use `Sq8Residual` = an 8-bit coarse code (256 levels over a per-cluster `[min,max]`) **plus** an i8 epsilon residual leg. `ROW_BYTES_PER_DIM = 2` (1 byte code + 1 byte epsilon).
- Cosine uses a 16-bit **fixed** `[-1,1]` grid.
- The epsilon leg exists only to recover precision the 8-bit coarse grid loses. At 16 bits (65,536 levels over the same per-cluster range) that workaround is unnecessary.

## Design

- **`Sq16Adaptive`**: one `u16` per dim over a per-cluster `[min,max]` derived from the data (variable ruler). **2 bytes/dim — identical to `Sq8Residual`**, so storage-neutral.
- Drops the epsilon/residual machinery entirely — simpler encode/transcode; no `residual_divisor`, no `EPSILON_I8_CLAMP`.
- **On merge/split, size the ruler from the union of the merged cells' ranges** (min-of-mins, max-of-maxes) → covers every input by construction → **no clamp**.
- Cosine unchanged (fixed grid, portability contract).

## Relationship to the clamp bug

The near-term correctness fix (size the *current* `Sq8Residual` destination grid from the union on merge) is tracked separately in #554. This issue is the cleaner end state: same bytes, half the moving parts, and the clamp becomes **structurally impossible** rather than avoided by careful grid-sizing across two legs.

## Compat

Changing the L2/NegDot codec changes the table's codec identity — existing `Sq8Residual` tables must not silently break on reopen. This is **new-tables-first plus a migration/transcode path**, gated behind a codec-version bump, not an in-place flip.

## Validation

Recall A/B — `Sq8Residual` vs `Sq16Adaptive` on the L2/NegDot sets at 1M and 100M: confirm 16-bit uniform is ≥ 8-bit-coarse + i8-epsilon (expected neutral-to-better), and that clamped-component counts go to zero.

muralikpbhat:
## What

Adds `Sq16Adaptive`, a new rerank codec: a single `u16` plane per vector over a **per-cluster, data-fitted** ruler — the 16-bit, variable-ruler counterpart of `Sq8Residual`, with no residual leg. It becomes the **default rerank codec for L2 and NegDot** vector tables. Cosine tables keep the fixed-grid `Sq16`.

## Why

`Sq8Residual` (the prior non-cosine default) spends 2 bytes/dim on a `u8` code plus an `i8` residual. `Sq16Adaptive` spends the same 2 bytes/dim on a single `u16` code fitted to each cluster's own value span — finer per-component resolution, no residual bookkeeping. It is **storage-neutral** and improves recall on every non-cosine ann-benchmarks dataset (table below), with p95 unchanged.

## Compatibility

- Added as a **new codec id (5)**, so existing tables reopen unchanged and resolve their codec by the stored id — `Sq8Residual` tables keep scoring exactly as before.
- The rerank codec is deliberately **excluded from the table options identity hash**, so changing the default does not change table identity or affect reopen.
- No new env var or config knob — the codec is chosen by metric, unconditionally.

## Recall + p95 — ann-benchmarks, all 5 non-cosine datasets

Measured on the same 16-core x86 box and harness as the #490 baseline (k=10, `default` stamped law, native `_id` path, index + 10% disk cache). `pre` = `Sq8Residual` (#490 baseline); `post` = `Sq16Adaptive`. Recall is hardware-independent; p95 is same-box.

| dataset | metric | recall pre (#490) | recall post | Δ | p95 pre | p95 post |
|---|---|---:|---:|---:|---:|---:|
| sift-128 | L2 | 0.9903 | 0.9908 | +0.0005 | 4 ms | 4.2 ms |
| gist-960 | L2 | 0.9938 | 0.9983 | +0.0045 | 13 ms | 13.8 ms |
| mnist-784 | L2 | 0.9924 | 0.9931 | +0.0007 | 1 ms | 0.8 ms |
| fashion-mnist-784 | L2 | 0.9910 | 0.9965 | +0.0055 | 2 ms | 1.6 ms |
| lastfm-64 | NegDot | 0.9864 | 0.9917 | +0.0053 | 38 ms | 38.1 ms |

Recall improves on every dataset (+0.0005 … +0.0055); p95 is within run-to-run noise (mnist/fashion are marginally faster). The NegDot path (lastfm) is validated alongside L2.

## Implementation notes

- The codec surface previously conflated two independent axes — **body shape** (two-plane `[u8 code | i8 residual]` vs a single `u16` plane) and **ruler** (per-cluster fitted vs fixed grid). `Sq16Adaptive` is the first codec that is single-plane *and* fitted, so build / drain-spill / merge / read now branch on capability predicates (`writes_single_u16_plane`, `carries_cluster_quant_meta`, `fits_per_cluster_ruler`) rather than on codec identity.
- The ruler is fit across the full `u16` range (`scale = span / 65535`) via a codec-aware `code_max()`; reusing the 8-bit divisor here would leave the plane carrying only 8 bits of precision.
- On merge the destination cluster reuses the first contributing input's ruler and re-encodes the rest onto it (the same behaviour as the residual family); a component outside that range is clamped. Because this codec has no residual leg to recover a clamp, the merge transcode feeds its clamp count to the existing compaction clamp tripwire, so a destination ruler that fails to cover its inputs shouts rather than silently losing recall.

- Rename `Sq16` → `Sq16Fixed` so the fixed-grid vs adaptive-ruler distinction is explicit in the codec names.
- Size the merge destination ruler from the union of all inputs' ranges so no component ever clamps (tracked in #554).
- Make the per-thread/process CPU metering platform-independent (`clock_gettime`) so non-Linux hosts report real `kernel_cpu_ns` instead of a silent 0 — currently Linux-procfs-only (tracked in #565).
- The non-cosine (L2) `supertable_vector` CI bench leg added here activates once this merges to `main`: the bench runs on `pull_request_target`, which uses the base branch's workflow, so this PR still shows the base legs.

Closes #555.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -50,6 +50,9 @@ jobs:
           - { bench: superfile,         docs: "1000000", suffix: sf }
           - { bench: supertable_fts,    docs: "1000000", suffix: st-fts }
           - { bench: supertable_vector, docs: "1000000", suffix: st-vec }
+          # Non-cosine (L2) leg: exercises the Sq16Adaptive default codec
+          # (the cosine leg above stays on the fixed-grid Sq16).
+          - { bench: supertable_vector, docs: "1000000", suffix: st-vec-l2, metric: l2sq }
           - { bench: supertable_sql,    docs: "1000000", suffix: st-sql }
     # A new push to the PR cancels the in-flight run for that bench.
     concurrency:
@@ -63,6 +66,7 @@ jobs:
     with:
       bench: ${{ matrix.bench }}
       docs: ${{ matrix.docs }}
+      metric: ${{ matrix.metric }}
       ref: ${{ github.event.pull_request.head.sha }}
       name_suffix: -${{ matrix.suffix }}
       pr_number: ${{ github.event.pull_request.number }}
--- a/.github/workflows/supertable-bench-azure.yml
+++ b/.github/workflows/supertable-bench-azure.yml
@@ -98,6 +98,10 @@ on:
     inputs:
       bench:           { type: string, required: false, default: supertable_all }
       docs:            { type: string, required: false, default: "1000000" }
+      # Vector metric: ""/cosine → the cosine Sq16 path; l2sq|negdot → the
+      # unbounded metrics whose default codec is Sq16Adaptive. Passed to the
+      # bench as INFINO_BENCH_METRIC. (Empty on legs that don't set it → cosine.)
+      metric:          { type: string, required: false, default: "" }
       vm_size:         { type: string, required: false, default: "Standard_D8s_v7" }
       location:        { type: string, required: false, default: eastus }
       ref:             { type: string, required: false, default: main }
@@ -411,6 +415,7 @@ jobs:
              INFINO_REAL_AZURE_CONTAINER='${{ secrets.INFINO_REAL_AZURE_CONTAINER }}' \
              INFINO_BENCH_SUPERFILE_DOCS='${{ inputs.docs }}' \
              INFINO_BENCH_SUPERTABLE_DOCS='${{ inputs.docs }}' \
+             INFINO_BENCH_METRIC='${{ inputs.metric }}' \
              BENCH_ARGS='${{ steps.resolve.outputs.args }}' \
              CPUSET='${{ inputs.bench_cpuset }}' \
              SKIP_CALIBRATION='${{ inputs.skip_calibration }}' \
@@ -449,6 +454,7 @@ jobs:
              INFINO_REAL_AZURE_CONTAINER='${{ secrets.INFINO_REAL_AZURE_CONTAINER }}' \
              INFINO_BENCH_SUPERFILE_DOCS='${{ inputs.docs }}' \
              INFINO_BENCH_SUPERTABLE_DOCS='${{ inputs.docs }}' \
+             INFINO_BENCH_METRIC='${{ inputs.metric }}' \
              BENCH_ARGS='${{ steps.resolve.outputs.args }}' \
              CPUSET='${{ inputs.bench_cpuset }}' \
              KEEP='${{ inputs.keep_table }}' \
--- a/benches/utils/corpus.rs
+++ b/benches/utils/corpus.rs
@@ -92,19 +92,23 @@ pub fn block_on_inmem<F: std::future::Future>(fut: F) -> F::Output {
 // ─── Scale constants ──────────────────────────────────────────────────
 
 /// Codec benches build vector columns with. Mirrors the engine default
-/// (`VectorConfig::new`): the fixed cosine grid for cosine, locally fitted
-/// residuals for unbounded metrics. Codec choice is engine configuration
+/// (`default_rerank_codec_for`): the fixed cosine grid (`Sq16`) for cosine, and
+/// the per-cluster-fitted 16-bit `Sq16Adaptive` for the unbounded metrics
+/// (L2Sq / NegDot). Codec choice is engine configuration
 /// (`vector.rerank_codec` in YAML), not a bench env knob.
 pub fn bench_rerank_codec(metric: Metric) -> RerankCodec {
-    // Diagnostic (env-gated): force an EXACT fp32 rerank column to isolate Sq8
-    // codec loss from routing/coverage. INFINO_BENCH_RERANK_CODEC=fp32 → if the
-    // recall ceiling jumps to ~1.0, the residual gap was the Sq8 quantization.
+    // Diagnostic (env-gated): force a codec to isolate its recall/latency from
+    // routing/coverage, or to A/B the new default against the prior one.
+    // INFINO_BENCH_RERANK_CODEC=fp32 → if the recall ceiling jumps to ~1.0, the
+    // gap was quantization; =sq8_residual → A/B the pre-Sq16Adaptive default.
     let codec = match std::env::var("INFINO_BENCH_RERANK_CODEC").ok().as_deref() {
         Some("fp32") => RerankCodec::Fp32,
         Some("sq16") => RerankCodec::Sq16,
+        Some("sq16_adaptive") => RerankCodec::Sq16Adaptive,
+        Some("sq8_residual") => RerankCodec::Sq8Residual,
         Some("sq8_fixed_residual") => RerankCodec::Sq8FixedResidual,
         _ if metric == Metric::Cosine => RerankCodec::default(),
-        _ => RerankCodec::Sq8Residual,
+        _ => RerankCodec::Sq16Adaptive,
     };
     assert!(
         codec.supports_metric(metric),
--- a/benches/utils/ingest/supertable.rs
+++ b/benches/utils/ingest/supertable.rs
@@ -107,8 +107,17 @@ const ROT_SEED: u64 = 7;
 /// Bytes in one gibibyte, for GiB-denominated memory and report values.
 const GIB_BYTES: u64 = 1u64 << 30;
 
-/// Distance metric for the bench vector index.
-const BENCH_METRIC: Metric = Metric::Cosine;
+/// Distance metric for the bench vector index. Defaults to cosine (the
+/// historical baseline, so existing bench deltas are unchanged);
+/// `INFINO_BENCH_METRIC=l2sq|negdot` selects an unbounded metric so CI can
+/// exercise the `Sq16Adaptive` default rather than only the cosine `Sq16` path.
+fn bench_metric() -> Metric {
+    match std::env::var("INFINO_BENCH_METRIC").ok().as_deref() {
+        Some("l2sq") | Some("l2") => Metric::L2Sq,
+        Some("negdot") | Some("dot") => Metric::NegDot,
+        _ => Metric::Cosine,
+    }
+}
 /// Writer auto-flush threshold (MiB) per superfile roll.
 const COMMIT_THRESHOLD_SIZE_MB: u64 = 1024;
 /// Table-doc-count boundary for the bench's pinned CELL-GRID shape:
@@ -317,8 +326,8 @@ pub fn options_for(
             column: VEC_COLUMN.into(),
             dim: DIM,
             rot_seed: ROT_SEED,
-            metric: BENCH_METRIC,
-            rerank_codec: corpus::bench_rerank_codec(BENCH_METRIC),
+            metric: bench_metric(),
+            rerank_codec: corpus::bench_rerank_codec(bench_metric()),
         }]
     } else {
         vec![]
@@ -350,8 +359,10 @@ pub fn current_knobs(modality: Modality) -> crate::dataset::Knobs {
         vec_seed: CORPUS_VEC_SEED,
         text_seed: CORPUS_TEXT_SEED,
         rot_seed: ROT_SEED,
-        metric: format!("{BENCH_METRIC:?}"),
-        rerank_codec: corpus::bench_rerank_codec(BENCH_METRIC).name().to_string(),
+        metric: format!("{:?}", bench_metric()),
+        rerank_codec: corpus::bench_rerank_codec(bench_metric())
+            .name()
+            .to_string(),
         modality: format!("{modality:?}"),
     }
 }
--- a/src/superfile/vector/builder.rs
+++ b/src/superfile/vector/builder.rs
@@ -36,8 +36,8 @@ use crate::superfile::{
     vector::{
         cell_posting::{MaterializedIvfRow, sq8_residual_norm_sq},
         distance::{
-            Metric, distance, encode_sq16_row, mean_f32_cluster_major, normalize,
-            sq16_decoded_norm_sq,
+            Metric, distance, encode_sq16_adaptive_row, encode_sq16_row, mean_f32_cluster_major,
+            normalize, sq16_adaptive_norm_sq, sq16_decoded_norm_sq,
         },
         ivf_merge::MergedIvfSubsection,
         kmeans::{assign_to_centroids, kmeans, kmeans_with_assignments},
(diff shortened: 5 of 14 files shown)


ekechinwokah (changes requested):
# Review — PR #563: Sq16Adaptive rerank codec, default for L2/NegDot

**Verdict: request changes.** The codec itself is well built — the math is right, the capability-predicate refactor is the correct cut, and the ann-benchmarks numbers are convincing. But the default flip breaks every existing non-cosine table's vector queries (contradicting the PR's own compatibility claim), the repo's correctness gates never exercise the new default, and the PR now needs a rebase over #550 that is easy to get subtly wrong.

## What it does

Adds `Sq16Adaptive` (codec id 5): a single `u16` plane per vector quantized against a per-cluster data-fitted ruler — 16-bit precision at `Sq8Residual`'s exact storage cost — and makes it the default rerank codec for L2Sq/NegDot. Splits the codec surface into orthogonal capability predicates (`writes_single_u16_plane`, `carries_cluster_quant_meta`, `fits_per_cluster_ruler`, `code_max`) replacing identity checks at ~15 dispatch sites.

## Verified correct (read the full diff + branch; kernel algebra checked by hand)

- `Sq16Kernel::new_adaptive` folds the affine cluster grid exactly: `q_prime[d]=q[d]·scale[d]`, `q_dot_offset=q·offset` reproduce `q·x` for `x=code·scale+offset`; L2/NegDot/Cosine legs all coherent, norms stored post-clamp over the reconstruction (consistent with the residual family).
- Build path: fitted codecs buffer fp32 in buckets, fit per-cluster `[min,max]`, encode with `code_max=65535` — and `derive_quantizer_from_min_max` is now parameterized so the 8-bit/16-bit fit can't silently diverge; the `255/65535` ratio test is a sharp guard for exactly the bug that parameterization prevents.
- Re-bucketing round-trips Sq16Adaptive through fp32 (payload `Fp32`, min/max updated) so rulers re-fit; fixed-grid Sq16 still byte-copies. Correct split.
- Merge: destination reuses the first input's ruler, per-row transcode decodes-source/encodes-destination, clamps feed the existing compaction tripwire via `note_transcode_clamped_components`. Single-plane body split in `EncodedCellRow` handled.
- Reader: open paths gate on `carries_cluster_quant_meta` (meta loads), the multi-cell directory id-match gained the new arm, and a full dispatch sweep found **no missed site** — every remaining `is_sq16()` use deliberately means fixed-grid-only, and `validate_quantizer_meta`'s pinned-constant check correctly exempts fitted arrays.
- Reachability: the residual-family-only `cell_posting::encode_rows` guard is unreachable with the new default — `CellPostingBuilder` only exists for the legacy `VectorLayout::CellPosting` read format; no production write selects it. The hidden index inherits any IVF-mergeable user codec (`handle.rs:1505-1517`), so drain/OPANN/compaction all route Sq16Adaptive through the codec-generic ops. Spill shape checks handle the single-plane-with-ruler combination explicitly.

## Blocking

**1. Existing L2/NegDot tables fail every vector query after upgrade.** The catalog does not persist the rerank codec: `open_table` rebuilds the `IndexSpec` from recorded `(column, dim, metric)` and lowers through `to_configs()` → `VectorConfig::new` → `default_rerank_codec_for(metric)` (`catalog/mod.rs:500-527`, `index_spec.rs:124-140`) — which now yields `Sq16Adaptive`. Query fan-out then runs `verify_superfile_vector_codecs` on every unit (`supertable/query/dispatch.rs:555,576`), which hard-errors when the stored codec differs from the config's (`dispatch.rs:110-129`). A table created pre-upgrade stores `sq8_residual` in its superfile directories → every vector search returns "vector codec mismatch: table expects sq16_adaptive, superfile stores sq8_residual". Hidden-index multi-cell files hit the same wall (the multi-cell exception at `dispatch.rs:111-115` only covers non-mergeable codecs). The PR body's claim — "existing tables reopen unchanged and resolve their codec by the stored id; Sq8Residual tables keep scoring exactly as before" — holds for the options-identity hash but not for this verifier; the builder doc ("existing Sq8Residual tables keep their stored codec on reopen") is wrong as written. Options, in rough preference order: (a) make the stored codec authoritative in the verifier when it's IVF-mergeable and metric-compatible (scoring already dispatches on the stored id per superfile, so mixed-codec tables score correctly today — the verifier is the only thing rejecting them); (b) persist the codec in the catalog entry and default only at create; (c) if rebuild-on-default-change is the accepted policy (precedent: the cosine Sq8FixedResidual→Sq16 flip), say so explicitly in the PR body and release notes and fix the two doc claims — but (a) or (b) is strongly preferred for a crate that is now public.

**2. The repo's own gates never see the new default.** The brute-force IVF oracle pins `Fp32` for all three metrics (`tests/superfile/vector/brute_force_oracle.rs:128-136`; the only codec-specific oracle is Cosine+Sq8FixedResidual). The bench recall fixtures hard-code the *old* default — `bench_rerank_codec` returns `Sq8Residual` for non-cosine (`benches/utils/corpus.rs:94-115`) and its doc comment ("mirrors the engine default") is now false — so the L2 scale floors gate the codec this PR just replaced, and the 10M supertable 0.99 gate runs Cosine (`BENCH_METRIC`), untouched by this change. Nothing under `tests/` or `benches/` changed in this PR. Needed: (a) an L2Sq + NegDot full-nprobe oracle leg for Sq16Adaptive (or build the oracle through `VectorConfig::new` so it always tracks the default), (b) `bench_rerank_codec` actually mirroring `default_rerank_codec_for`, (c) the checklist artifacts — in-repo `make bench` A/B vs main and the standard-bench recall@10 ≥ 0.99 statement in the PR body (the ann-benchmarks table is good evidence but isn't the repo's stated gate).

**3. Rebase over #550 (now on main) must thread the metering.** Main's `score_sq8_residual_candidates` returns `(scored, kernel_ns)` with a bracketed thread-CPU section, and every rerank arm accumulates the tally. This PR rewrites those exact functions (`score_candidates_with_kernel` generic + three new Sq16Adaptive arms: eager, lazy-with-meta early return, lazy fallback). The generic actually makes this easier — bracket once in the shared fan-out — but if the rebase drops it, every new default L2/NegDot table reports `kernel_cpu_ns = 0` on the rerank leg and the priced CPU counter silently under-bills. After rebasing: run the op-stats suite plus `a_scoped_vector_query_reports_kernel_cpu` against an Sq16Adaptive table (they currently only cover the drained cosine path).

## Non-blocking

- **Cosine + Sq16Adaptive is allowed but untested.** `supports_metric` is metric-agnostic for it, the norm layout and kernel cosine leg look coherent, but no test covers the override. Either add a cosine self-query test or exclude cosine until one exists.
- `rerank_codec.rs`: `super::cell_posting::note_transcode_clamped_components(clamped)` is an inline qualified path — the file already imports from `cell_posting`; add it to that `use` (house rule: no inline module paths).
- `distance.rs`: `SQ16_CLAMP_DETECT_SLACK_CODES` is declared *below* its only user, mid-file — consts go at the top of the file. While there: `65535.0` is duplicated between `code_max()` (rerank_codec.rs) and the encode/dequant/norm functions (distance.rs); one shared `SQ16_CODE_MAX` const would make the agreement structural instead of test-enforced.
- `encode_sq16_adaptive_row` doc says "a zero-span dimension has scale 0", but `derive_quantizer_from_min_max` assigns `scale = 1.0` for zero spans — the `scale[d] > 0.0` guard is defensive, not the build-path shape. Reword.
- Stale strings: `builder.rs:2082` still says "requires one matching residual-family codec" (the check is codec equality across the mergeable set); the single-column open error at `reader.rs:889` enumerates known ids without 5.
- Suggested test: a merge-transcode clamp case (two inputs with disjoint per-dim ranges into one destination) asserting the tripwire counts and top-1 stays correct on in-range rows — the only clamp coverage today is indirect.

## Performance / cost

Storage-neutral by construction (2 bytes/dim either way); scoring rides the existing `Sq16Kernel` body, so per-candidate cost matches fixed Sq16, and per-cluster kernel construction is the same O(dim) as `Sq8ResidualKernel`. Build buffers fp32 exactly as `Sq8Residual` already did. p95 parity in the PR table is consistent with all of that. The missing artifact is the in-repo bench A/B (blocking item 2c).

## Process

Format-additive (new codec id; no footer/layout change), and #555 was filed first — right process. One release-notes item: files written with id 5 are unreadable by ≤ 0.5.x binaries, and with the default flipped that's every new L2/NegDot table.

muralikpbhat:
Thanks for the careful review — every blocking item is addressed. Rebased onto current `main` (0.5.3). Point by point:

### Blocking

**1. Existing L2/NegDot tables fail every query after upgrade.** Correct, and a real miss on my side — fixed via your option (a). `verify_superfile_vector_codecs` no longer compares the stored codec against the re-derived default; the stored superfile codec is now authoritative, and the verifier rejects only a codec that genuinely can't serve the table (doesn't support the metric, or a non-IVF-mergeable codec inside a multi-cell unit). Scoring already dispatches per-superfile on the stored id, so older/mixed tables score correctly — this is codec-agnostic and fixes existing tables under *any* future default change, not just this one. Reproduced end to end: a table created by pre-flip code, reopened on this branch, fails with the exact `vector codec mismatch` before the fix and returns identical top-k after. New test `reopen_after_default_flip_accepts_stored_codec` (plus a metric-incompatible-still-rejected case). The two doc claims are reworded and now true. `b6d4769`

**2a. Oracle pins Fp32.** Added full-nprobe brute-force oracle legs for Sq16Adaptive on **L2Sq and NegDot** (`oracle_sq16_adaptive_l2sq_full_nprobe_recovers_exact_topk`, `…_negdot_…`), mirroring the fp32 / fixed-residual-cosine legs — exact-recovery now covers the new default. `1b95111`

**2b. `bench_rerank_codec` returns Sq8Residual.** Now returns Sq16Adaptive for non-cosine (matches `default_rerank_codec_for`); doc corrected. Added `sq16_adaptive` / `sq8_residual` to the `INFINO_BENCH_RERANK_CODEC` diagnostic options so the codec can be A/B'd on one binary. `1b95111`

**2c. In-repo bench A/B + recall statement.** Added a non-cosine (L2) `supertable_vector` CI bench leg driven by a new `INFINO_BENCH_METRIC` knob (same `INFINO_BENCH_*` convention as docs/cells/rerank-codec). **Heads-up — you'll still see only 4 bench legs on this PR, not 5, and that's expected.** Two independent reasons, both of which resolve automatically once this merges:
1. The bench triggers on **`pull_request_target`**, so GitHub runs the **base branch's** `bench.yml` (main's 4-leg matrix), *not* the PR's — the new leg is dormant until it lands on `main`.
2. Even once it runs, its vs-`main` delta is only meaningful once `main` also carries the `INFINO_BENCH_METRIC` knob (which arrives with the same merge); before that the base arm would run cosine/Sq16 while the PR arm runs L2/Sq16Adaptive — apples-to-oranges.

So, concretely: the L2 leg is **not active on this PR**, but the L2/NegDot codec A/B was done **outside CI** — the ann-benchmarks table in the description (all 5 non-cosine sets, L2 + NegDot, at 1M, Sq16Adaptive vs the Sq8Residual baseline), with a 100M BIGANN run in flight. And once this merges, the leg runs on **every future PR** with a valid vs-`main` A/B (from then on `main` has both the 5-leg matrix *and* the knob). The standard 10M-Cohere gate is cosine, so it wouldn't attest the new codec regardless. `1b95111`

**3. Rebase over #550 + metering.** Rebased onto 0.5.3. #550's kernel-CPU metering is threaded through the generic scorer (`score_candidates_with_kernel` now returns `(scored, kernel_ns)`) and all three Sq16Adaptive arms (eager, lazy-with-meta early return, lazy fallback). New op-stats test `a_scoped_sq16_adaptive_vector_query_reports_kernel_cpu` asserts the new default's rerank leg reports non-zero kernel CPU. `4085c29`, `8ee166f`

One thing surfaced doing this: the kernel-CPU metering is **Linux-only** — `thread_cpu_ns()` reads `/proc/thread-self/schedstat`, so `kernel_cpu_ns` is always 0 off Linux (and any non-Linux infino deployment silently under-reports CPU, not just the tests). I gated the three kernel-CPU tests (mine + the pre-existing BM25/vector) `#[cfg_attr(not(target_os = "linux"), ignore)]` so they run on Linux CI but don't red a macOS `make ci`; making the metering platform-independent (`clock_gettime`) is tracked in #565.

### Non-blocking

- **inline `super::` path** → imported from `cell_posting`. `1e0ea10`
- **const placement + duplicated `65535`** → moved `SQ16_CLAMP_DETECT_SLACK_CODES` up; added a shared `SQ16_CODE_MAX` that `SQ16_FIXED_SCALE`, `code_max()`, and the encode/dequant math all reference, so the ruler and plane width agree structurally. `884acc1`
- **zero-span doc** → reworded: the build fit assigns `scale = 1.0` (not 0) for a constant dim, so the `scale == 0` branch is a defensive floor. `884acc1`
- **stale strings** → `builder.rs` mismatch message now says every row must share the table's rerank codec; `reader.rs` unknown-id and not-implemented lists now include `sq16` / `sq16_adaptive`. `884acc1`
- **merge-transcode clamp test** → added `sq16_adaptive_transcode_counts_saturated_components` (two disjoint-range inputs → destination clamp counted), serialized with the Sq8 counter test to avoid a shared-atomic race. `1e0ea10`, `884acc1`
- **Cosine + Sq16Adaptive override** → **not addressed.** Cosine's default is the fixed-grid `Sq16`, so this only arises via an explicit per-column override, which no path selects; I left it untested rather than add coverage for a config nothing produces. Happy to add a cosine self-query test or gate cosine out of `supports_metric` for Sq16Adaptive if you'd prefer one over the other.

### On the merge clamp (your earlier note)

To be explicit: union-sizing (the actual clamp-free fix) is deliberately **not** in this PR — it's a separate follow-up. What this PR does is make the clamp **observable**: the Sq16Adaptive merge transcode feeds the same process-wide clamp counter that compaction reads to emit `[supertable compaction] BUG: … recall silently degrades`, so it's at diagnostic parity with the residual family rather than silent. The enum/merge doc comments are reconciled to describe the real first-input-ruler + clamp behaviour (the earlier "union … by construction" wording is gone).

ekechinwokah (approved):

PR #566, opened 2026-08-08
superfile/vector: batch the warm Sq16 rerank path

ekechinwokah:
Three implementation-only changes to the deferred rerank, together 26.1 -> 19.3 ms warm k=100 on Cohere-1M engine defaults (phase-profile harness, scores equivalent within FMA/add-order rounding, recall machinery untouched):

- Tier the Sq16 cross-dot (AVX-512F / AVX2+FMA / safe `wide` fallback). The old inner loop converted u16 codes through per-byte slice indexing — measured as 24.1 ms of a 26.1 ms query; the intrinsic tiers convert via `cvtepu16` straight from the unaligned code bytes (9.9 ms, now DRAM-bound on the scattered rows). Tiers agree to add-order/FMA rounding, pinned by a parity test.

- Resolve warm survivor rows as ONE zero-copy span per probed cluster and slice each row out of it, instead of one source lookup per survivor (25.3K lookups ~= 8.5 ms -> 1.1 ms). Resident spans are mmap-backed Bytes views, so pages fault in only for rows actually read; any missing span falls back to the coalesced remote plan unchanged.

- Read per-candidate norms as single little-endian f32 loads from the raw norm-table bytes (NormLookup) instead of parsing the whole table on every rerank call (2.6 ms -> ~0).

Also adds the ignored vector_phase_profile diagnostic (per-phase decomposition of warm queries against a prebuilt local table) that produced these measurements.

### DIFF
--- a/src/superfile/vector/distance.rs
+++ b/src/superfile/vector/distance.rs
@@ -1153,7 +1153,11 @@ impl Sq16Kernel {
 /// [`BitQuantizer::estimate_dot_rotated_with_total`]).
 #[inline]
 fn sq16_cross(q_prime: &[f32], code_bytes: &[u8]) -> f32 {
-    debug_assert_eq!(code_bytes.len(), q_prime.len() * 2);
+    // Release-enforced: the intrinsic tiers below read `2 * dim` code
+    // bytes through unaligned pointer loads, so a short slice would be
+    // an out-of-bounds read, not a panic. One predictable branch per
+    // candidate is free next to the ~dim FMAs that follow.
+    assert_eq!(code_bytes.len(), q_prime.len() * 2);
     #[cfg(target_arch = "x86_64")]
     {
         if avx512_enabled() {
@@ -1166,10 +1170,9 @@ fn sq16_cross(q_prime: &[f32], code_bytes: &[u8]) -> f32 {
             return unsafe { sq16_cross_avx512(q_prime, code_bytes) };
         }
         if avx2_enabled() {
-            // SAFETY: gated on `avx2_enabled()` (AVX2; FMA is checked
-            // by the same gate's platform baseline — see the explicit
-            // `fma` enable on the function). Bounds as above with an
-            // 8-code stride.
+            // SAFETY: gated on `avx2_enabled()`, which detects both
+            // `avx2` and `fma` — the two features the function enables.
+            // Bounds as above with an 8-code stride.
             return unsafe { sq16_cross_avx2(q_prime, code_bytes) };
         }
     }
--- a/src/superfile/vector/distance.rs
+++ b/src/superfile/vector/distance.rs
@@ -1120,29 +1120,8 @@ impl Sq16Kernel {
     #[inline]
     pub fn distance_with_norm(&self, code_bytes: &[u8], norm: Option<f32>) -> f32 {
         debug_assert_eq!(code_bytes.len(), self.dim * 2);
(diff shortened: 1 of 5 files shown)


muralikpbhat (approved):
Approving. I verified the correctness core of all three changes and it's result-identical at the byte level:

- **Zero-copy per-cluster span slicing is byte-identical to the old per-row fetch** — `base = span.start` (per-cluster min over its candidates' ranges), the slice `range.start-base .. range.end-base` returns the same bytes the old `try_get_range_sync(range)` did, `ranges` and `rerank_cands` are built in one lockstep loop so they stay 1:1 and identically ordered, `range ⊆ span` rules out underflow/OOB, and the span/fallback is all-or-nothing (any non-resident span sends every row through the coalesced async plan in the same order). Stride math unchanged.
- **`NormLookup::Raw` returns the same value as the old parse-whole-table path** — same LE `pos*4..+4` offset, same source bytes; OOB panics exactly as before (not a regression).
- **The SIMD Sq16 cross-dot is memory-safe** — unaligned loads throughout, loop bounds strictly within the code-byte buffer with a scalar remainder tail (no OOB on non-lane-multiple dims), unsigned `cvtepu16` zero-extend (codes ≥ 32768 correct), runtime-gated with the safe `wide` fallback on non-x86/disabled. Nice win.

Three non-blocking notes for follow-up:

1. **Wording:** "identical rows scored / behavior-unchanged" isn't bit-exact — the kernel now uses fused `mul_add` where the old path did separate mul+add, so per-candidate scores shift by ULPs and can swap a near-tie at the top-k boundary (negligible recall impact, same tolerance contract as the existing tiered kernels). Suggest "equivalent within FMA/add-order rounding."
2. **AVX2 gate:** `sq16_cross_avx2` is `#[target_feature(enable="avx2,fma")]` and issues `vfmadd`, but the dispatch gate only runtime-checks `avx2`, not `fma`. No mainstream CPU has AVX2 without FMA and the prod `x86-64-v3` target bakes in FMA (and this matches the existing Sq8 kernels' gating, so no new risk) — but a `is_x86_feature_detected!("fma")` in the gate, or a corrected SAFETY comment, would make it airtight.
3. **Test gap:** the SIMD is well covered by `sq16_cross_tiers_agree` (tiers vs `wide` vs an f64 reference, sweeping tail/768 dims), but there's no targeted unit test that span-slice == the old per-lookup, or that `NormLookup::Raw(pos)` == `Arr`/`parse[pos]` — those two data-access rewrites lean on end-to-end recall (and the Lazy `Raw` arm's coverage is unverified). A direct equality test for both would lock them in.

None blocks; the core is verified and CI is green. (Branch is currently behind `main` — needs a rebase before merge.)

ekechinwokah:
thanks — all three addressed on the branch:

1. body wording fixed to "equivalent within fma/add-order rounding" — right call, the wide tier also moved to mul_add so even the fallback shifted ulps.
2. the avx2 gate now detects fma alongside avx2 (one cpuid read, and it retroactively covers the sq8 kernels with the same exposure); the safety comment now says what the gate actually checks. also promoted sq16_cross's length precondition to a release assert — a short slice is a panic now, not an oob read.
3. both equality tests added: span-slice is byte-identical to the per-range fetch (span map pulled out into survivor_cluster_spans so the contract is directly testable, including duplicate and overlapping ranges), and NormLookup::Raw is bit-exact against the eager arm over a parsed copy of the same table.

PR #569, opened 2026-08-09
bench/ci: block PR merge on significant benchmark regressions

ekechinwokah:
The A/B summary's 5% findings were advisory prose: bench_summary.py always exited 0 and nothing in the workflow failed, so a benchmark regression could not actually block a merge. And the transitions table (drain / optimize walls) rendered its numbers as text, so those never reached the report JSON at all — an optimize regression was invisible to the delta layer entirely.

- Transition walls are emitted as numeric lower-is-better metrics, so drain and optimize times diff against the main baseline like every latency row.
- bench_summary.py gains the blocking pair: a time-valued metric worse than the same-host main baseline by BOTH >50% and >5 ms blocks the merge; either alone does not, so sub-millisecond percent spikes and long-wall noise wiggles never block. The 5%/30% tier thresholds stay exactly as they were, as report-only advisory findings. The verdict (with reasons) goes to GATE_FILE.
- The azure bench workflow enforces the verdict in a separate step after the summary posts, so a FAIL still leaves the PR comment in place; runs with no baseline pass vacuously.

Verified on synthetic fixtures: (786 s -> 5,332 s optimize) blocks; +60%/+6ms blocks; +40%/+4ms and +300%-of-0.05ms stay advisory (each fails one leg of the AND); PASS and no-baseline paths write PASS.

### DIFF
--- a/.github/scripts/bench_summary.py
+++ b/.github/scripts/bench_summary.py
@@ -9,6 +9,7 @@ Inputs (env):
   BASELINE_DIR               dir holding <report>.json from the base-ref baseline
   CURRENT_DIR                dir holding <report>.json from this run
   BENCH_NOISE_THRESHOLD_PCT  threshold in percent (default 5)
+  GATE_FILE                  merge-gate verdict destination (default /tmp/bench-gate.status)
   OUT_FILE                   markdown destination (default /tmp/ai-summary.md)
   BENCH_LABEL                human label for the run (the `bench` input)
   BENCH_VM_SIZE              VM size label for context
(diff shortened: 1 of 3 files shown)


muralikpbhat (approved):
Approving. The gate is well-shaped: a merge blocks only when a lower-is-better time metric is worse than the same-host `main` baseline by BOTH >50% and >5 ms — the AND is a sound noise guard (a big % of a sub-ms metric never blocks, nor does a small % of a long wall). The 5% advisory tiers correctly stay report-only, the enforce step is separate so the summary still posts to the PR on FAIL, and a missing gate file (no baseline / first run) passes. I confirmed the drain/optimize **walls are actually covered**: the `supertable.rs` change emits them as `metric(…, Better::Lower)`, and `is_latency()` (lower-is-better, not rss/stored) returns true for them, so they reach the JSON and the gate — previously they were `text()` and invisible to the A/B.

Two notes, neither blocking:

1. **No waiver for *intended* regressions.** The gate is a hard `exit 1`, so a deliberate latency trade — e.g. a `target_recall` operating-point change, or a #574-style "wide probes pay their true price" — that exceeds the pair would block merge with no escape hatch. Consider a label-based waiver (an `accept-bench-regression` PR label the enforce step honors, logged in the summary) so intended trades aren't stuck behind a red gate.

2. **Cliff, not creep.** Because the base arm is `main` re-benched each run (rolling baseline), a merged regression is absorbed into the baseline and doesn't tax future PRs — good — but the flip side is the gate can't see cumulative drift: N PRs each regressing just under the pair, each measured against the prior drifted `main`, walk the absolute number arbitrarily far while every PR passes. Fine for the stated single-cliff purpose; if you later want to bound absolute latency, add a pinned/golden anchor (or a hard ceiling) alongside the rolling A/B.

PR #570, opened 2026-08-09
compaction: codec-agnostic saturation warning (stop hardcoding "Sq8")

### LINKED ISSUE #568: Compaction saturation BUG message hardcodes "Sq8" — mislabels the Sq16Adaptive clamp path
### Summary

The compaction saturation `BUG` line hardcodes **"destination Sq8 quantizer"** even when the clamp happens on the **Sq16Adaptive** (single-plane) path — so on a Sq16Adaptive table (the L2/NegDot default) it wrongly reads as if the codec were Sq8.

### Detail

The counter `TRANSCODE_CLAMPED_COMPONENTS` (`superfile/vector/cell_posting.rs`) is bumped by **two** paths:
- the `Sq8Residual` residual transcode, inline (`residual_family_materialize_into_cluster_quant`), and
- **single-plane codecs via `note_transcode_clamped_components`** — `Sq16Adaptive` calls it from its merge re-encode (`superfile/vector/rerank_codec.rs`: `encode_sq16_adaptive_row(...)` → `note_transcode_clamped_components(clamped)`).

But the emitted message (`supertable/compaction/mod.rs`) is:

```
[supertable compaction] BUG: {n} component(s) saturated their destination Sq8 quantizer during this pass's merges/splits (#512 failure mode) ...
```

So a Sq16Adaptive table that clamps on merge prints a "Sq8" `BUG`. This is purely a **logging/labeling defect** — the codec selection is correct (Sq16Adaptive), and the clamp itself is the separate union-sizing work in #554.

### Impact

Actively misleading: it produced a false "wrong-codec" diagnosis during LAION/T2I 100M investigation (looked like the table was Sq8 when it was Sq16Adaptive).

### Fix

Emit the **actual destination codec** in the message (or make it codec-agnostic, e.g. "destination quantizer"), so the `BUG` line names what really clamped instead of hardcoding "Sq8".

muralikpbhat:
## What

The compaction saturation warning hardcoded **"destination Sq8 quantizer"**, but the counter it reports (`TRANSCODE_CLAMPED_COMPONENTS`) now also tallies clamps from **single-plane codecs** — `Sq16Adaptive` bumps it from its merge re-encode via `note_transcode_clamped_components`. So a `Sq16Adaptive` table (the L2/NegDot default) that clamps on merge printed a `BUG` line saying "Sq8", reading as if the wrong codec were selected.

This makes the message **codec-agnostic** ("destination quantizer") and drops "silently" (the line is emitted, so it isn't silent).

```
- their destination Sq8 quantizer during this pass's merges/splits ... rows' recall silently degrades.
+ their destination quantizer during this pass's merges/splits ... rows' recall degrades.
```

Logging-only change; no behavior change. The underlying clamp (destination grid sized from a subset, not the union of inputs) is the separate union-sizing work in #554.

Fixes #568.

### DIFF
--- a/src/supertable/compaction/mod.rs
+++ b/src/supertable/compaction/mod.rs
@@ -458,9 +458,9 @@ impl Supertable {
         if clamped_components > 0 {
             eprintln!(
                 "[supertable compaction] BUG: {clamped_components} component(s) saturated \
-                 their destination Sq8 quantizer during this pass's merges/splits (#512 \
-                 failure mode) — a destination grid failed to cover its inputs; affected \
-                 rows' recall silently degrades.",
+                 their destination quantizer during this pass's merges/splits — a \
+                 destination grid failed to cover its inputs; affected rows' recall \
+                 degrades.",
             );
         }
         Ok(())
--- a/src/supertable/writer.rs
+++ b/src/supertable/writer.rs
@@ -4262,10 +4262,10 @@ pub(in crate::supertable) async fn drain_user_superfiles_to_hidden_cells(
     if clamped_components > 0 {
         eprintln!(
             "[supertable drain] BUG: {clamped_components} component(s) saturated their \
-             destination Sq8 quantizer during this drain's re-encodes (#512 failure \
-             mode). Cosine: an ingest path bypassed normalization; L2/NegDot: a \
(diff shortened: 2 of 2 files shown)


savannahar68 (approved):
Lgtm

PR #573, opened 2026-08-09
fts: faster COUNT + phrase search — skip-based negation, staged phrase verify

vinaykakade:
Two exact optimizations to the FTS count path; the phrase one also speeds scored phrase top-k. Counts and results are unchanged (brute-force oracle + corpus-truth tests pin them).

## 1. Negation count without materializing the negated union

The count path handled `+term -term` by materializing the **entire** negated union into a bitmap and subtracting it — e.g. `+jaguar -car -football` built a `car ∪ football` set of millions of docs to exclude from ~1.7k positive matches. Now the negated terms are fed as a **skip-based exclusion gate** (`AtomExcludeFilter`) to the existing count walk, so a common negated list is only `skip_to`-probed, never decoded whole — the same mechanism the scored search path already used for negation.

The delete path is unchanged (a tombstone bitmap still needs an explicit id set); only the delete-free case takes the skip path.

## 2. Phrase verification: rarest-first, staged, lazy

Phrase verification anchored on the query-first member and eagerly decoded every member's per-doc positions. The first word of a phrase is often a common word (`the`, `of`), and a phrase containing one produces a huge co-occurrence candidate set — e.g. `the book of life` aligns ~92k docs to find 95 matches. Decoding the common members' positions (a per-block run-offset walk) on every candidate was the dominant cost.

Now verification seeds candidate start positions from the **rarest** member and filters the rest **rarest-first with lazy decode**: a common member's positions are decoded only once a candidate has survived every rarer member. On these big co-occurrence sets almost every doc is rejected by a rare member first, so the common members are never decoded there. Doc alignment probes rarest-first too. This is the phrase cursor's shared doc-advance, so **scored phrase top-k benefits as well as count**.

## Results (single compacted superfile, single-threaded, min-of-10)

infino before (main) → after (this branch):

| shape | main | branch |
|---|---:|---:|
| phrase COUNT | 2210 ms | **1338 ms** |
| phrase TOP_100 | 1719 ms | **1002 ms** |
| phrase TOP_1000 | 2036 ms | **1231 ms** |
| negation COUNT (bucket) | 434 ms | **286 ms** |
| **total COUNT** | 3467 ms | **2500 ms** |

Worst-case per-query (COUNT): `+jaguar -car -football` ~24×, `the book of life` ~5.4×, `the news journal` ~5.6×. Plain-term, union, and intersection paths are untouched (their arm-to-arm differences are run-to-run noise).

## Tests

- Brute-force BM25 top-K oracle and phrase count-vs-oracle agreement stay green — verification changes cannot change which docs/counts return.
- Added `count_with_negation_matches_corpus_truth` (OR/AND positives, single/multiple negatives, and the no-negation identity) pinning the skip-based count against corpus truth.
- Full FTS integration + supertable query suites pass.

No on-disk format change; no public API change (the extended `atoms_match_count` is internal).

### DIFF
--- a/src/superfile/fts/reader.rs
+++ b/src/superfile/fts/reader.rs
@@ -1798,7 +1798,14 @@ impl FtsReader {
                 .build_atom_cursors(column_id, neg_terms, neg_phrases, None)
                 .await?;
             let neg_atoms: Vec<AnyCursor> = neg_built.into_iter().flatten().collect();
-            work.planned_ranges += neg_dict_ranges;
+            // Count the negated clause's posting work the same way the
+            // positive atoms above (and the scored path's `ExcludeFilter`)
+            // are counted — planned posting bytes + ranges from cursor
+            // metadata — so op_stats prices a negated count consistently.
+            // (Like every skip/leapfrog path, this is a planned figure, not
+            // the partial bytes the skip probe actually decodes.)
+            work.postings_bytes += atom_cursor_bytes(&neg_atoms);
+            work.planned_ranges += atom_planned_ranges(&neg_atoms) + neg_dict_ranges;
             if !neg_atoms.is_empty() {
                 filter = Some(AtomExcludeFilter::new(neg_atoms));
             }
--- a/src/superfile/fts/reader.rs
+++ b/src/superfile/fts/reader.rs
@@ -1772,6 +1772,8 @@ impl FtsReader {
         terms: &[&str],
         phrases: &[Vec<String>],
         mode: BoolMode,
+        neg_terms: &[&str],
+        neg_phrases: &[Vec<String>],
     ) -> Result<(u64, MatchWork), FtsError> {
         let column_id = self.resolve_column_id(column)?;
         // Unranked: idf is irrelevant to the match set, so build local.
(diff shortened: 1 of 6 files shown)


src/superfile/fts/reader.rs
```diff
@@ -1785,9 +1787,25 @@ impl FtsReader {
...
+        // against the negated cursors, so a common negated term's long list
+        // is only partially decoded. Empty ⇒ `None`, the same walk as an
+        // unnegated count.
+        let mut filter = None;
+        if !neg_terms.is_empty() || !neg_phrases.is_empty() {
+            let (neg_built, neg_dict_ranges) = self
+                .build_atom_cursors(column_id, neg_terms, neg_phrases, None)
+                .await?;
+            let neg_atoms: Vec<AnyCursor> = neg_built.into_iter().flatten().collect();
+            work.planned_ranges += neg_dict_ranges;
```
naman-modi: From what I (limitedly) understand about the `MatchWork`, it needs to accurately calculate the metrics, and here the the negated-count path only adds `neg_dict_ranges` to work. It never accounts the negatives' posting bytes or posting-fetch ranges (atom_cursor_bytes/atom_planned_ranges).

So, op_stats under-reports fetched cost for a negated count. This is more of a metric correctness issue rather than the actual query issue.

Possible fix is to add the negatives' atom_cursor_bytes + atom_planned_ranges to work, but that itself would be partial (which is exactly what this PR tends to do).

tests/superfile/fts/negation.rs
```diff
@@ -85,6 +85,51 @@ async fn or_single_positive_minus_negative() {
     assert_eq!(got, want, "rust -async (OR)");
 }
 
+/// Count with negation goes through the skip-based exclusion path
+/// (`atoms_match_count` with negated atoms) rather than materializing the
+/// negated union. Pin it against corpus truth for OR and AND positives
+/// with single and multiple negatives.
+#[tokio::test]
+async fn count_with_negation_matches_corpus_truth() {
```
naman-modi, line 110: This count test only passes term args (&[] for both phrase slots), so nothing exercises phrases in the count path. Can we have a test with:

1. phrase as a positive, combined with a negation, eg:`+"new york" -city`  
2. negated phrase, eg: `york -"new york"`

naman-modi (approved):
LGTM, few comments.

tests/superfile/fts/negation.rs
vinaykakade: Thanks, done

src/superfile/fts/reader.rs
vinaykakade: Good catch - fixed

PR #574, opened 2026-08-09
supertable/vector: residual codes adaptive rerank

ekechinwokah:
## What

Two serving-cost features for the hidden vector index, plus the kernel
integration that makes them pay at serve time:

1. **Residual 1-bit codes (subsection v3).** Hidden cells store RaBitQ
   codes as signs of `rot(v − c)` against each row's own subsection
   centroid, with a per-doc `‖r‖` region; the scan estimate becomes
   `q·c + κ·‖r‖·sign_dot`. Sharper estimates on real embeddings mean
   the drain-time calibration certifies narrower widths and smaller
   survivor budgets for the same coverage target.
2. **Adaptive rerank stopping.** Calibration also stamps a stop margin
   (the stage-coverage quantile of `exact − estimate` over pooled
   pairs, cosine only). The deferred rerank walks the shortlist in
   estimate bands and stops once no unprocessed candidate can reach
   the running kth exact score even granted its estimate plus the
   margin. The stamped budget stays the hard cap; easy queries stop
   early, hard queries run to the cap.

## Format

- Subsection version 3: residual flag in the reclaimed reserved slot
  `[24..28]`, per-doc `‖r‖` f32 region between codec_meta and the
  stable-id region. Readers accept {2, 3}; flag-off files stay
  byte-identical version-2 (every existing superfile unchanged).
- κ is NOT in the header — the estimator uses the analytic
  `sqrt(π/2)/sqrt(dim)` (within ~5% of the measured fit);
  `CellRoutingParams::residual_kappa` is a reserved override slot.
- Splice and parsed merges carry the norm region (cluster-order index
  space) and enforce all-or-none residual provenance, like the codec
  check; mixed v2+v3 cells rebuild through the pack core, which
  re-derives every code from the payload.

## Calibration semantics (reviewers: this is the part to weigh)

- Drain spills are user-table rows (raw codes), so the drain scores
  them raw — the blunter estimator's budget over-covers residual
  serving (recall-safe). The optimize recalibration rescores packed
  cells with the serving-exact estimator per cell provenance.
- **Recalibration now REPLACES rerank knots (and the stop margin)
  under current evidence** — the same gate width already replaces
  under — and max-merges otherwise. Without this, the sharper
  estimator's smaller certified budget can never enter the stamp
  (max-merge keeps the drain's blunter value forever). This also
  means an evidence-current recalibration can shrink a raw table's
  rerank stamp.

## fully warm — verified by tail percentiles; max within 2.2x of p50)

**Cohere-10M, 768d cosine, k=10, warm serial p50 (100 queries):**

| config  | main @637fb24     | this branch       |
|---------|-------------------|-------------------|
| default | 0.9970 @ 25.2 ms  | 0.9980 @ 21.9 ms  |
| rm=256  | 0.9950 @ 22.2     | 0.9950 @ 19.0     |
| rm=128  | 0.9850 @ 21.1     | 0.9900 @ 17.4     |
| rm=32   | 0.9590 @ 19.7     | 0.9440 @ 16.7     |

Stamped laws on the same data: width 30 vs main's 34 cells at k=10;
k=100 survivor budget 18,883 vs 25,835 rows (−27%); stop margin
0.037. Default k=100: 0.9973 @ 46.1 ms vs the rm=256 constant's
0.9980 @ 62.7 ms.

**Cohere-1M (same protocol): parity.** 0.9930 @ 5.2 ms vs main's
0.9940 @ 4.8 ms — within 100-query sampling noise. Expected: at 1M
the law serves ~1–2 cells and the shortlist sits below the banding
floor, so neither feature has room to engage. The win is a scale
effect.

**Standard synthetic gates:** 1M lifecycle recall@10 0.996 post-drain
/ 0.998 post-compact (bar: 0.99), both arms; latency parity.

## Serve-time integration

The first cut lost the calibration win to estimator arithmetic in the
scan loop (measured as a flat +12 ms across every budget at 10M). Two
fixes, both arithmetic-identical: fold the residual affine over each
LUT block's score vector in one streamed pass (norms are cluster-order
contiguous), and resolve `q·c` as a raw dot against the resident
centroid bytes instead of rotating every probed fine centroid per
query (the rotation is orthonormal — `q_rot·rot(c) ≡ q·c`).

## Tests

- End-to-end on-disk tripwire: hidden superfiles assert v3 across
  drain / delta-commit / optimize generations (the class of test whose
  absence let an earlier flag-only enablement ship as a no-op).
- Byte-level roundtrips: layout offsets + header; every row's code and
  `‖r‖` re-derived independently; splice + parsed-merge norm carry
  with per-cluster slice equality; mixed-provenance rejection.
- Search parity vs raw codes and brute force, with a full-budget
  exactness tripwire (corpus-independent wiring proof).
- Margin quantile math (bulk vs outlier tails, conservative edges);
  banded serving walked at unit scale via a test-helpers band floor
  (planted top-k intact, determinism pinned).
- Full gates: lib 2,204 / supertable 283 / superfile 137, clippy
  clean.

## Trade-offs

- Residual tables spend `n_docs × 4` extra bytes per cell (the norm
  region) and one extra streamed norm read per scanned row.
- The drain's rerank stamp on a fresh residual table is measured with
  the raw estimator (conservative until the first optimize
  recalibration replaces it with the serving-exact value).
- At 1M-scale tables the features are inert by design (band floor,
  tiny widths): measured parity, no regression.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -55,8 +55,11 @@ jobs:
           - { bench: supertable_vector, docs: "1000000", suffix: st-vec-l2, metric: l2sq }
           - { bench: supertable_sql,    docs: "1000000", suffix: st-sql }
     # A new push to the PR cancels the in-flight run for that bench.
+    # Keyed on the SUFFIX, not the bench name: the two vector cells
+    # (cosine st-vec and l2sq st-vec-l2) share `matrix.bench`, and a
+    # bench-keyed group made them cancel each other on every run.
     concurrency:
-      group: bench-${{ github.event.pull_request.number }}-${{ matrix.bench }}
+      group: bench-${{ github.event.pull_request.number }}-${{ matrix.suffix }}
       cancel-in-progress: true
     permissions:
       id-token: write
--- a/benches/utils/concurrent.rs
+++ b/benches/utils/concurrent.rs
@@ -252,6 +252,7 @@ fn build_supertable_options(storage: Arc<dyn StorageProvider>) -> SupertableOpti
             metric: Metric::Cosine,
             rerank_codec: RerankCodec::Fp32,
             provided_centroids: None,
+            residual_codes: false,
         }],
         Some(default_tokenizer()),
     )
--- a/benches/utils/corpus.rs
+++ b/benches/utils/corpus.rs
@@ -1510,6 +1510,7 @@ pub fn build_vector_index(vectors: &[f32], n_docs: usize, metric: Metric) -> Vec
     let mut b = VectorBuilder::new();
     b.register_column(VectorConfig {
         provided_centroids: None,
+        residual_codes: false,
         column: "v".into(),
         dim: DIM,
         rot_seed: ROT_SEED,
@@ -1561,6 +1562,7 @@ pub fn build_superfile(docs: &[String], vectors: &[f32]) -> Vec<u8> {
         }],
         vec![SfVectorConfig {
             provided_centroids: None,
+            residual_codes: false,
             column: "emb".into(),
             dim: DIM,
             rot_seed: ROT_SEED,
@@ -1603,6 +1605,7 @@ pub fn build_superfile_with_metric(docs: &[String], vectors: &[f32], metric: Met
         }],
         vec![SfVectorConfig {
             provided_centroids: None,
+            residual_codes: false,
             column: "emb".into(),
             dim: DIM,
             rot_seed: ROT_SEED,
--- a/benches/utils/harness/infino_sql_engine.rs
+++ b/benches/utils/harness/infino_sql_engine.rs
@@ -138,6 +138,7 @@ pub fn sql_options() -> SupertableOptions {
         ],
         vec![VectorConfig {
             provided_centroids: None,
+            residual_codes: false,
             column: VECTOR_COLUMN.into(),
             dim: SQL_DIM,
             rot_seed: ROT_SEED,
--- a/benches/utils/harness/infino_vector_engine.rs
+++ b/benches/utils/harness/infino_vector_engine.rs
@@ -56,6 +56,7 @@ fn build_superfile(
         vec![],
         vec![VectorConfig {
             provided_centroids: None,
+            residual_codes: false,
             column: column.into(),
             dim,
             rot_seed: ROT_SEED,
--- a/benches/utils/ingest/supertable.rs
+++ b/benches/utils/ingest/supertable.rs
@@ -314,6 +314,7 @@ pub fn options_for(
     let vector = if modality.has_vector() {
         vec![VectorConfig {
             provided_centroids: None,
+            residual_codes: false,
             column: VEC_COLUMN.into(),
             dim: DIM,
             rot_seed: ROT_SEED,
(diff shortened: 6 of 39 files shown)


muralikpbhat (approved):
Approving. This is a large, format-changing PR, and I reviewed it across three axes — the v3 format + merge/splice, the adaptive-stop soundness, and the calibration/reopen path. The correctness core holds up and the scale recall validates the design.

**Verified correct**
- **Format / v2 compat:** residual-off files are byte-identical to v2 (norm region zero-width, no offset shifts, reserved `[24..28]` reads back off — tested). Reader dispatches on subsection version with a version⇄flag coherence guard (rejects v3-flag-off / v2-flag-on; unknown version fails loud). The `‖r‖` gap-split sizing is exact with `checked_sub` fail-loud.
- **Merge/splice:** all-or-none residual provenance fires on both orderings (tested), and the `‖r‖` region is carried index-aligned with the codes/doc-ids (tested).
- **Encode ⇄ estimator:** `code = signs(rot(v−c))`, `‖r‖ = ‖v−c‖`, analytic κ, same centroid at encode and scan.
- **Adaptive stop:** monotone, off-by-one-free (descending-estimate walk, correct running k-th), gated to the law-served cosine default only, recall-safe missing-margin fallback (None/0 → full budget). It's a calibrated (0.997-coverage) probabilistic bound by design — and the 10M/k=100 measurement confirms the real cost is ~0.0007 (~0.07%), an order of magnitude inside the target.
- **Calibration:** recalibration REPLACE gated by `evidence_current`; drain-raw over-covers residual serving; `residual_codes` correctly kept out of the identity hash (data-determined by the on-disk v3 flag, mirroring the `rerank_codec` precedent).
- **Scale recall:** Cohere-10M default 0.9970→0.9980 (faster), 1M parity (features inert by design), synthetic gate green. Residual ≥ baseline where it engages.

**Follow-up notes (none blocking):**
1. **Parsed-merge centroid anchor.** The parsed merge writes each output cluster's centroid as the weighted mean of inputs' centroids but copies residual codes/norms verbatim (anchored to each input's own centroid). Correct only under the current single-global-fine-grid invariant (all inputs share centroids) — which splice structurally guarantees but the parsed merge leaves unchecked. A `debug_assert` comparing input centroids on residual merges would make that dependency fail loud instead of silently costing recall if fine-grid drift is ever introduced.
2. **κ doc + fallback.** The `SUBSECTION_VERSION_RESIDUAL` doc-comment says κ lives in the reserved header slot, but serving uses the analytic κ and the `residual_kappa` field is reserved/unconsumed — worth correcting the comment, and when the field is first wired into serving it must carry a `None → analytic` fallback (a 0.0 stamp would rank on centroid-only estimates → recall collapse).
3. **Test gaps.** No end-to-end test opens a genuine pre-v3 (v2) on-disk fixture (the v2 parse path is only exercised by freshly-built v3 tables), and the reopen+re-optimize survival test asserts `rerank_margin` + pool provenance but not `width_margin`.
4. **Behavior note.** On the adaptive path the same query can return a slightly different top-k warm vs cold (recall-safe direction — warm stops strictly later than a full-k-th stop; two warm runs identical). Worth documenting as accepted.

Also a low-likelihood edge: the `width_margin` REPLACE draws width-gap evidence only from multi-cell entries, so a mixed v1 + multi-cell re-optimize could narrow it below the full-corpus quantile — `evidence_current` proves the set is unchanged, not that every entry contributed width-gap evidence. Unreachable for all-multi-cell adaptive-width tables today.

ekechinwokah:
closing for now. we are considering a different approach.

PR #577, opened 2026-08-10
supertable query sql: withhold whole-domain bounds from statistics

naman-modi:
### What does this PR do?
A range filter (<, >, BETWEEN) on a number column that stores both the smallest and largest value its type can hold used to fail instead of returning rows.

Before running a filter, the planner estimates how many rows will survive by comparing the width of the range you asked for against the width of the column's whole range. It measures width as biggest minus smallest plus one. Once the column's range covers every possible value, that count is one larger than the counter can hold, so it wraps to zero, the estimate divides by zero, and the resulting infinity fails a sanity check before the query touches any data. 

### How do we do this change?

- `statistics_for` withholds `min_value` / `max_value` when the folded range spans the column type's whole domain, so the planner falls back to its range-unknown estimate.
- Floats are excluded: the planner counts them by bit pattern and already guards that path, while the value distance used here saturates on a wide range.

### Notes
- The overflow is a DataFusion bug, fixed upstream by PR #22309, "Return None for cardinality overflow", which will be included in DataFusion 55, so we need to revert this change then.

### DIFF
--- a/src/supertable/query/provider.rs
+++ b/src/supertable/query/provider.rs
@@ -126,6 +126,12 @@ use crate::{
 /// resolving filter columns to a physical pruning predicate.
 pub(crate) const TABLE_NAME: &str = "supertable";
 
+/// Distance between the endpoints of a min/max range that covers a
+/// 64-bit column type's entire domain (`0 ..= u64::MAX` for `UInt64`,
+/// `i64::MIN ..= i64::MAX` for `Int64`, and the same span for `Date64`
+/// / `Timestamp`). See [`spans_full_domain`].
+const FULL_DOMAIN_ENDPOINT_DISTANCE: u64 = u64::MAX;
+
 /// Object-store URL *prefix* the surviving superfiles are registered under
 /// for a scan. The authority is arbitrary — only a key into the session's
 /// object-store registry — but it must be **unique per provider**: a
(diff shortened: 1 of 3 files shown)


savannahar68:
> The overflow is a DataFusion bug, fixed upstream by PR #22309, "Return None for cardinality overflow", which will be included in DataFusion 55, so we need to revert this change then

Should we add this comment in cargo.toml above datafusion deps so that we take action while bumping

src/supertable/query/provider.rs
```diff
@@ -711,6 +721,68 @@ fn scalar_min_max(
     acc
 }
 
+/// Whether `[min, max]` covers the column type's entire domain — the one
```
savannahar68: can we shorten this comment and link #22309

src/supertable/query/provider.rs
```diff
@@ -711,6 +721,68 @@ fn scalar_min_max(
...
+///
+/// TODO: delete this once the engine moves to DataFusion 55. The overflow
+/// is fixed upstream by DataFusion PR #22309, "Return None for cardinality
+/// overflow", which makes the count a checked add and returns no count on
+/// overflow — the same fallback this guard forces by hand. That fix landed
+/// after the 54 release branch was cut and was not backported, so 54.x
+/// still needs this.
+fn spans_full_domain(min: &ScalarValue, max: &ScalarValue) -> bool {
+    !min.data_type().is_floating()
+        && max.distance(min).and_then(|d| u64::try_from(d).ok())
```
savannahar68: is there a possibility for try_from failing? in that case query will fail
imo we shouldnt fail query if pruning fails as this is part of pruning
wdyt

savannahar68:
anything left in this PR ?

src/supertable/query/provider.rs
naman-modi: Done.

src/supertable/query/provider.rs
naman-modi: No. Also, this is not pruning, this is sharing the stats with the planner. So, I think we are good.

savannahar68 (approved):
lgtm

PR #579, opened 2026-08-10
benches: optionally use external corpora

ekechinwokah:
## Problem

Every vector bench measures the synthetic planted-cluster corpus, whose queries are perturbed corpus members. That corpus is matched to the grid by construction, so it systematically flatters routing: the drain calibrates on corpus rows and the queries look like corpus rows. Real embeddings have neither property, and the gap only becomes visible with a real dataset and its own query set.

## What

Opt-in real corpora for the benches, selected **by argument** — not env — so a corpus can never change from the environment and silently make two runs incomparable:

```
cargo bench -- supertable vector corpus=annb:glove-100-angular corpus-dir=/data/annb
```

- `corpus=synthetic` (default) — the seeded planted-cluster generator, byte-identical to today; every existing baseline, delta, and CI A/B stays comparable.
- `corpus=annb:<slug>` — a published ann-benchmarks dataset (glove-100-angular, sift-128-euclidean, deep-image-96-angular, …), shipping its **official queries and top-k neighbours** — ground truth measured against the field's published answers, not our own oracle.
- `corpus=hf:<owner/repo>` — a Hugging Face parquet dataset: real embeddings at scale, ground truth computed from the ingested rows; shards are reconciled against the hub's file list and a query tail is reserved out of the ingest set.
- `corpus=parquet:<dir>` — a local VDBBench-layout directory, no download.
- `corpus-dir=<path>` stages downloaded corpora; present files are used as-is, and only the file actually opened is fetched.

The source sits behind the existing `MmapVectorCorpus` abstraction, so ingest, ground truth, and the recall machinery are untouched. Dimension is `corpus::dim()` — the dataset's own width from its schema; synthetic keeps 1024. The FTS/SQL tiers read a parquet dataset's text column, so real corpora are no longer vector-only. Doc counts clamp loudly with the delta headroom budgeted as one commit (1/16); the cold-GET ceiling scales with corpus row width; the prepared-dataset sidecar gains a serde-defaulted `corpus` field so a dataset built from one corpus refuses to grade another.

## Per-tier recall floors

The two tiers legitimately sit at different recall levels on real data, so the single tripwire became `RecallFloors`:

- **supertable** (calibrated: the drain stamps serving laws) keeps the existing floors — the 0.99 acceptance bar is unchanged and expected to hold here.
- **superfile** (uncalibrated, fixed probe width tuned on synthetic) drops to 0.60/0.60 — loose enough that a real dataset's legitimately lower default-config recall is *reported* rather than aborting the run, still tight enough that a broken index (which collapses toward chance) trips it. Motivating measurement: real Cohere at the superfile tier's default probe width served 0.660 where synthetic reads ~0.99.

A realistic-synthetic-shape default (Zipf vocabulary + anisotropic vectors, 0.70 floors) was tried on this branch and **reverted** — the synthetic default stays exactly as it was.

## Engine change riding along: `target_recall`

The branch carries one engine change (originally #581): `vector.target_recall` — a single per-table knob for the recall/latency operating point. The drain calibration targets it when stamping the table's probe laws: width crosses at the target and the staged depth pads off it, so lowering the value narrows width, fine depth, and the rerank budget together. Config default is **0.99 — the existing acceptance bar — so behavior is preserved unless a deployment sets it**; `SupertableOptions::with_target_recall` overrides per table, taking effect on the next drain/recalibration. No public-API surface change (`public-api.txt` unchanged). Measured shape (in the config comment): a 0.93 target serves 0.969 at 8.4 ms — serving lands above target because crossings are discrete.

## Build / CI

`infino-bench-utils` gains `hdf5` (system-linked `hdf5-metno`) + `ndarray` for the ann-benchmarks reader — **`libhdf5-dev` is now required to compile dev targets**; CI installs it on every runner that compiles the workspace (including the locomo example job). The bench workflow threads `corpus` / `corpus_dir` through its matrix so real-corpus runs are CI-selectable.

## Verified

- Wiring certificate on real data: at full probe width the engine is exhaustive, so all-cells recall@10 = 1.000 certifies corpus bytes, query loading, ground truth, and id mapping end-to-end (measured 1.000 on Cohere 100K on the original wiring; the annb path additionally grades against published neighbours).
- Unit tests: parquet fixture round-trip (row order/values, normalization, shortfall error, strided queries); an ignored network test asserts the multi-GB train file is *not* fetched for a queries-only pull; the #512 parity test now ranks ids instead of subtracting them.
- Synthetic path byte-identical when no `corpus=` is given; `make check` clean.


muralikpbhat (commented):
The mechanics here are clean — I verified the default synthetic path is unchanged (so the CI-gated baselines don't shift), the Cohere ground truth is recomputed exact over the same normalized vectors that get ingested, and the change is confined to `benches/` + the one debug test. No correctness concern with what it measures.

**Holding approval on a scope/dependency question, not the code.** This adds a **third-party vendor dependency into `benches/`**: a hardcoded external dataset bucket (`assets.zilliz.com/benchmark/…`) fetched at bench time, plus vendor/tool names in the source. Our bench tier is otherwise deliberately vendor-neutral and self-contained — synthetic corpora, infino-only timings, no external asset host in the loop — and this introduces a network dependency on a vendor bucket for a bench run to work.

**Ideally this lives in the VectorDBBench harness rather than in-tree `benches/`.** That harness already has the infino client and already manages exactly these real-world datasets (download, normalization, official query/GT handling) — so a real-corpus (Cohere) recall path is largely duplicating what it already does, but here it also drags the vendor bucket + dataset-plumbing into the core repo's bench tier. Keeping the synthetic-vs-real split as "in-tree benches = synthetic/self-contained; real-corpus recall = the dedicated harness" keeps `benches/` free of an external-vendor runtime dependency and keeps the datasets managed in the one place built for them.

If the decision is to keep it in-tree anyway, then before merge I'd want, at minimum:
- the URL behind an env indirection (e.g. `INFINO_BENCH_CORPUS_URL`) with no hardcoded vendor bucket in source, and the vendor names softened in comments;
- an integrity pin (checksum/size) on the download — today only HTTP-200 + `.part`+rename guard against bad bytes, so a truncated-but-200 or wrong-content response is accepted silently;
- a request/stall timeout (the client is built with `timeout(None)`, so a half-open connection to a multi-GB file hangs the run);
- and note the `from_parquet` write/normalize path isn't exercised by an offline test (the round-trip test resolves the synthetic dim in-process), so it's only validated by a live run.

None of that blocks the *default* bench — it's all on the opt-in `INFINO_BENCH_CORPUS=cohere` path — but the vendor-dependency-in-`benches/` question is worth settling before this lands.

ekechinwokah:
the vendor dependency wasn't intentional — fixed. pushed.

datasets now come from ann-benchmarks.com, the academic host the published ANN benchmarks are run against. `INFINO_BENCH_CORPUS` takes a slug (`glove-25-angular`, `nytimes-256-angular`, `deep-image-96-angular`) instead of a vendor-prepared parquet dir, and `INFINO_BENCH_CORPUS_URL` points at a mirror if you want one. no zilliz bucket, no vdbbench-specific layout, no vendor names in source.

bonus from the switch: those files ship the official query set and the official top-k neighbour ids, so ground truth travels with the data and our recall numbers are directly comparable to published ann-benchmarks results.

on scope though — this shouldn't live in the vdbbench harness. vdbbench is a vector benchmark. the point of this is real datasets for the bench tier generally, fts and sql included, and a vector-only harness structurally can't carry that. in-tree is where it has to be.

your other four are all in:
- download base is env-indirected, no host hardcoded in source
- length-verified against Content-Length, refuses partial data (a truncated-but-200 was silently accepted before)
- connect + whole-download timeouts, so a half-open connection can't hang a run
- `from_hdf5` takes the row width explicitly instead of reading the process-wide `dim()` — that's what makes the write/normalize path testable offline, and there's now a test that writes its own fixture and covers row order, normalization, shortfall + dim-mismatch errors, query order, and GT filtering. no network.

one thing i caught wiring up the official GT: those neighbours are true top-k over every row, so on a truncated corpus they're just the wrong answer — filtering the out-of-range ids would leave a short list that overstates recall. it asserts full-corpus ingest now rather than quietly grading against the wrong truth.

cost is one bench-crate dep (hdf5-metno, static — libhdf5 built from source so nothing to install on a bench box or in CI) plus ndarray. both infino-bench-utils only, never the shipped crate.

also split the recall tripwires per tier: supertable stays 0.80, superfile goes to 0.60. the superfile tier has no drain and no stamped laws — its default probe width is tuned against the synthetic corpus's planted clusters, so on real data it under-serves by construction (glove-25 0.740, cohere 0.660 at 100K; both ~0.99 synthetic). one constant can't serve both tiers without either failing honest real-corpus runs or blinding the calibrated one.

verified: glove-25 loads and reports 0.740 where it used to abort, synthetic unchanged at 1.000, bench-utils 61 tests, make check clean.

muralikpbhat (commented):
Good progress on the rework — the runtime vendor dependency is genuinely gone: downloads now target `ann-benchmarks.com` (neutral academic host) + HuggingFace, no `assets.zilliz.com` bucket anywhere in the tree, the corpus comes from ann-benchmarks with its official held-out queries, GT is recomputed locally under the engine's own metric, and the download path is now hardened (connect + transfer timeouts, Content-Length check, atomic rename, non-synthetic-only). I also confirmed the **default synthetic path is byte-identical** (`dim()` = `SYNTHETIC_DIM` = 1024, `corpus_source()` defaults to Synthetic, CI passes `corpus=""`), so the gated baselines don't shift. The #512 parity-test change is a genuine improvement.

**Holding on one thing — a CI recall tripwire is weakened on the default synthetic path.** `executors.rs:1420` drops the single-superfile recall floor `0.80 → 0.60`. The justification is sound *for real corpora* (they under-serve at 0.66–0.74 on the uncalibrated single-superfile path), but the floor isn't conditioned on corpus source, so it also halves the guard on the **synthetic** run — a future synthetic regression from ~0.99 down to 0.65 (a codec/routing bug) would now pass green where 0.80 caught it, and that's exactly the class of guard the new bench gate is meant to complement. Please keep `0.80` for `Synthetic` and apply `0.60` only to the real-corpus sources.

Also worth fixing (not a hold): the `--help` text at `benches/bench/main.rs:327` still reads `parquet:<dir>  local VDBBench-layout directory` — that's a third-party vendor tool named in the public help output, and it's also inaccurate (the loader reads any `*.parquet` in name order, not a vendor-specific layout). Rephrase to e.g. `local directory of *.parquet shards (no download)`. The vendor name also survives in a few commit messages (with the old bucket URL) — a nit, but a reword before merge would keep it out of public history.

Minor: the help lists a `sift-128-euclidean` example, but the pipeline always normalizes and grades cosine — internally consistent, but a euclidean example produces a misleading "sift" number; drop the euclidean examples or route `-euclidean` slugs to L2.

muralikpbhat:
Re-checked at the current head — the HDF5 build sorting looks good and the merge-syncs are clean. Just two small items left from the earlier round, then this is good to approve:

1. **Condition the superfile recall floor on corpus source** (`benches/utils/executors.rs`, `RecallFloors::SUPERFILE`). The 0.60 floor is right for real corpora (they legitimately under-serve on the uncalibrated single-superfile path), but as written it also applies to the synthetic run — halving that tripwire's sensitivity for no benefit. Keeping 0.80 for `Synthetic` and 0.60 for the real sources preserves the guard where it was doing its job.

2. **One help-text line** (`benches/bench/main.rs:327`): `parquet:<dir>  local VDBBench-layout directory` still names the vendor tool in the public `--help`, and it's also inaccurate — the loader reads any `*.parquet` in name order, no vendor-specific layout. Something like `parquet:<dir>  local directory of *.parquet shards (no download)` covers it.

That's everything — the rest (ann-benchmarks sourcing, default-path preservation, GT, download hardening) all checked out earlier.

ekechinwokah:
> Re-checked at the current head — the HDF5 build sorting looks good and the merge-syncs are clean. Just two small items left from the earlier round, then this is good to approve:
> 
> 1. **Condition the superfile recall floor on corpus source** (`benches/utils/executors.rs`, `RecallFloors::SUPERFILE`). The 0.60 floor is right for real corpora (they legitimately under-serve on the uncalibrated single-superfile path), but as written it also applies to the synthetic run — halving that tripwire's sensitivity for no benefit. Keeping 0.80 for `Synthetic` and 0.60 for the real sources preserves the guard where it was doing its job.
> 2. **One help-text line** (`benches/bench/main.rs:327`): `parquet:<dir>  local VDBBench-layout directory` still names the vendor tool in the public `--help`, and it's also inaccurate — the loader reads any `*.parquet` in name order, no vendor-specific layout. Something like `parquet:<dir>  local directory of *.parquet shards (no download)` covers it.
> 
> That's everything — the rest (ann-benchmarks sourcing, default-path preservation, GT, download hardening) all checked out earlier.

both taken.

recall floors are corpus-conditional now: the superfile tier keeps the full 0.80 tripwires on synthetic (where the fixed probe width is tuned to the planted clusters and ~0.99 is normal) and drops to 0.60 only for real sources, where the uncalibrated path legitimately under-serves. selection is a function reading the active corpus source, so the guard follows the run rather than the tier.

help text fixed — vendor name gone and the description now says what the loader actually does: any *.parquet in the directory, read in name order, no layout assumptions. that was the last vendor reference in the bench tree (swept rs+toml to be sure).

muralikpbhat (approved):
Approving. Both remaining items landed in `16cbee1` and I verified them end-to-end:

- **Corpus-conditional superfile floors** — `RecallFloors::superfile()` dispatches on `corpus_source()`: `Synthetic` keeps the full `0.80/0.80` tripwire sensitivity, real corpora get the `0.60/0.60` floor their legitimately-lower uncalibrated recall needs, and both call sites use the selector. The synthetic guard is back where it was doing its job.
- **Help text scrubbed** — `parquet:<dir>` now reads "local directory of *.parquet shards, read in name order (no download)", which is both vendor-neutral and, unlike the old line, actually accurate to what the loader does. No vendor tooling named anywhere in `benches/` now.

With the earlier rounds this closes out everything: ann-benchmarks/HF sourcing (no vendor bucket), default synthetic path byte-identical (CI baselines unmoved), locally-recomputed GT consistent with the engine's metric, hardened downloads (timeouts, length check, atomic rename, opt-in only), and the rank-based #512 parity test. Nice, thorough set of fixes.

(The `license/cla` check is pending due to earlier cla-assistant service outage — unrelated to this PR; it'll clear once your re-trigger.)

pratyushlokhande:
Was looking at what it'd take to run a schema-driven SQL corpus through this, and
hit one bug plus two scope questions.

**1. `superfile fts` and `superfile sql` accept `corpus=` and silently ignore it.**

`mod fts` (`superfile.rs:665`) and `mod sql` (`superfile.rs:1990`) call
`MmapTextCorpus::generate(...)` unconditionally — only `mod vector`
(`superfile.rs:1074`) matches on `corpus_source()`.

Reproduced on `16cbee1` with a 40K-row parquet fixture (`text` + `emb`, 167
bytes/doc vs synthetic's ~2 KB/doc), `INFINO_BENCH_SUPERFILE_DOCS=20000`:

| Cell | `corpus=parquet:<dir>` | synthetic |
|---|---|---|
| `superfile vector` | `64×20K fixture-corpus` | `1024×20K synthetic` |
| `superfile fts` | Corpus 38.34 MiB / Stored 13.22 MiB | **identical** |
| `superfile sql` | Corpus 116.46 MiB / Stored 69.68 MiB | **identical** |

Exit 0, no warning, and the corpus label never appears in the FTS or SQL output.
Had the fixture been read, FTS would have shown ~3.2 MiB.

Also worth noting: `clamp_docs_to_corpus` is on the shared `superfile_docs()`
path, so a corpus smaller than the requested count would clamp FTS/SQL's doc
count while still feeding them generated text — sized to the real dataset,
measuring synthetic data.

The supertable guard at `ingest/supertable.rs:422` is exactly the right shape.
Happy to send a small PR mirroring it on the superfile tier.

**2. Help text says "vector corpus source", but supertable FTS and SQL do consume
the corpus today.** Worth widening the wording — or narrowing the behavior,
depending on which you meant.

**3. SQL takes a corpus's text but not its schema — intended ceiling?**
`sql_schema()` (`harness/infino_sql_engine.rs:77`) is a fixed six-column fixture
with `emb` pinned to `SQL_DIM = SYNTHETIC_DIM`, so a dataset's own columns have
nowhere to go. Coherent as a line to draw; flagging it because the obvious next
ask — benchmarking ClickBench `hits` through the SQL cell — needs schema-driven
Parquet ingest, a much larger change than this PR.

Separately: any plan to key the baseline JSON by corpus? `target/infino-bench/<bench>.json`
is corpus-agnostic today, so a synthetic baseline diffed against a real-corpus run
produces deltas that read as regressions. Checking before I open anything separately.

ekechinwokah:
@pratyushlokhande all three verified, and your clickbench work is merged into the branch — answers against head 2528990d.

1. real bug, and the two halves ended up in different places. your schema-driven sql cell fixes the sql half properly (honoring the corpus beats refusing it), so i took that whole and kept a guard only where the gap is still live: superfile fts still calls MmapTextCorpus::generate unconditionally, so it now fails loudly via corpus::require_synthetic instead of reporting generated text under the real corpus's label. the clamp point was the sharpest part of your report — clamp_docs_to_corpus sits on the shared doc-count path, so the run was sized to a dataset it never read. that's in the guard's rationale.

2. widened the wording rather than narrowing behavior: supertable fts/sql consuming the dataset's text is intended (it's how the dbpedia lifecycle runs), and the help now states the split per cell — vector always, supertable fts/sql the text column, superfile sql the schema, superfile fts synthetic-only and refusing.

3. was the intended ceiling for this pr; you removed it, which is the better outcome. nothing left to answer here.

on the baseline json — real gap, and worse than a delta-reading problem, so i fixed it here rather than filing it. `save` MERGES over the existing file (so a partial run updates only what it measured), which means a real-corpus run over a synthetic file leaves every key it didn't measure holding synthetic numbers — and that mixed file is what the next delta reads, and what ci copies as an arm's result. the file now records a fingerprint of the corpus label and starts fresh when it differs, with a line saying so.

kept the fingerprint INSIDE the file rather than suffixing the filename, because the bench workflow scp's target/infino-bench/<bench>.json by exact name for its a/b baseline — renaming breaks that plumbing for no gain. two tests: one pins that a corpus switch drops the previous corpus's untouched keys instead of leaving them stale, the other that the fingerprint round-trips through json's f64 exactly (masked to 53 bits — without that the gate fires every run and no partial run ever merges).

gates at head cef7e17f: clippy -D warnings clean, bench-utils 85/85, lib 2229, license audit 3/3.

PR #582, opened 2026-08-11
vector: normalize cosine rows at the fp32 cell-build seam (#512)

### LINKED ISSUE #512: Cosine rerank codec silently clamps non-unit input (−9.6 pts recall on Cohere-1M) — decide: normalize vs. reject

ekechinwokah:
Fixes #512 

The fixed cosine rerank grid spans [-1, 1], so a non-unit component saturates at ENCODE — with no error and no clamp counter, since the drain's transcode tripwire only observes re-encodes. Reported cost: recall@10 capped at ~0.898 on Cohere, unchanged even at all-cells with full rerank (degraded data, not routing), recovering to ~0.998 when the identical corpus is re-ingested pre-normalized.

#520 normalized at VectorBuilder::add, described there as "the single seam every downstream consumer reads through". That holds for that builder and not for the engine: SupertableWriter::append buffers the caller's Arrow vector buffers zero-copy, and the commit-time hidden-index pack reads them directly — VectorColumnView -> PackRow::Fp32 -> drain_pack_assigned_cell ->
build_merged_subsection_from_fp32 — never passing through that seam. So the user table stored unit rows while the hidden index, which is what serves vector search, stored clamped ones.

Normalize in stream_fp32_rows_to_buckets, the choke point every fp32 cell build funnels through, so the guarantee holds regardless of caller. The scratch is chunk-bounded (same sizing as the existing chunk_rotated buffer), deliberately not at append(): normalizing there would reintroduce the commit-wide vector copy VectorColumnView exists to avoid (12.8 GiB at a 3.125M-row x dim-1024 commit). Cosine declares magnitude irrelevant, so this is semantics-preserving and unit input is an fp-noise no-op.

Also adds a debug-only assertion at the cell-posting encoder that cosine rows arrive unit, so the next path that bypasses a normalize seam fails in tests instead of silently degrading data.

Regression test drives the public append() path end to end with rows at magnitude 12 and asserts served recall matches the same corpus submitted pre-normalized, mapping engine _ids to dense corpus rows the way the bench harness does (one ordered SELECT _id scan) rather than comparing per-table ids. Without the fix: 0.6969 vs 1.0000, test fails. With it: parity, test passes. Suites green — lib 2192, supertable 284, superfile 138, make check clean.

### DIFF
--- a/src/superfile/vector/builder.rs
+++ b/src/superfile/vector/builder.rs
@@ -1759,6 +1759,33 @@ fn stream_fp32_rows_to_buckets(
     });
     let chunk_rows = materialized_chunk_rows_for_dim(dim);
     let mut chunk_rotated = vec![0.0f32; chunk_rows * dim];
+    // Cosine rows must be unit before ANY consumer below sees them: the
+    // fixed cosine grid spans [-1, 1], so a non-unit component saturates at
+    // encode — silently, with no error and no clamp counter (the drain's
+    // transcode tripwire only observes RE-encodes). Issue #512 measured
+    // about -10 points of recall@10 from exactly this.
+    //
+    // #520 normalized at `VectorBuilder::add`, described there as "the
+    // single seam every downstream consumer reads through". That holds for
+    // that builder and not for the engine: the supertable buffers the
+    // caller's Arrow vector buffers zero-copy (`BufferedBatch`) and the
+    // commit-time hidden-index pack views them directly
+    // (`VectorColumnView` -> `PackRow::Fp32` -> `build_merged_subsection_
+    // from_fp32`), never passing through that seam. So the user table
+    // stored unit rows while the hidden index — what actually serves vector
+    // search — stored clamped ones.
+    //
+    // Normalizing here covers every fp32 cell build regardless of caller,
+    // and the scratch is chunk-bounded (same sizing as `chunk_rotated`),
+    // not corpus-bounded: normalizing at `append` instead would reintroduce
+    // the commit-wide vector copy `VectorColumnView` exists to avoid
+    // (12.8 GiB at a 3.125M-row x dim-1024 commit).
+    let cosine = cfg.metric == Metric::Cosine;
+    let mut chunk_unit = if cosine {
+        vec![0.0f32; chunk_rows * dim]
+    } else {
+        Vec::new()
+    };
     let mut chunk_codes = vec![0u8; chunk_rows * code_bytes];
     let mut chunk_payload = if fixed {
         vec![0u8; chunk_rows * dim * 2]
@@ -1768,7 +1795,19 @@ fn stream_fp32_rows_to_buckets(
     let mut row_base = 0usize;
     while row_base < n_docs {
         let take = (n_docs - row_base).min(chunk_rows);
-        let chunk = &vectors[row_base * dim..(row_base + take) * dim];
+        let raw = &vectors[row_base * dim..(row_base + take) * dim];
+        let chunk: &[f32] = if cosine {
+            chunk_unit[..take * dim]
+                .par_chunks_mut(dim)
+                .zip(raw.par_chunks(dim))
+                .for_each(|(dst, src)| {
+                    dst.copy_from_slice(src);
+                    normalize(dst);
+                });
+            &chunk_unit[..take * dim]
+        } else {
+            raw
+        };
         let mut assignments = vec![0u32; take];
         assign_to_centroids(chunk, centroids, dim, n_cent, &mut assignments);
         chunk_rotated[..take * dim]
--- a/src/superfile/vector/cell_posting.rs
+++ b/src/superfile/vector/cell_posting.rs
@@ -162,6 +162,40 @@ impl CellPostingBuilder {
     }
 }
 
+/// Tolerance on `‖x‖` for the cosine unit-input tripwire below. Wide
+/// enough that fp32 accumulation over thousands of dims never trips it,
+/// tight enough that genuinely un-normalized input always does (raw
+/// embedding norms are O(1..100), not 1 ± 0.01).
+#[cfg(debug_assertions)]
+const COSINE_UNIT_NORM_TOLERANCE: f32 = 1e-2;
+
+/// Debug-only guard that cosine rows reach the encoder already unit
+/// (issue #512). Both ingest seams normalize — `VectorBuilder::add` and
+/// `CellPostingBuilder::add` — but the failure this guards is a NEW path
+/// bypassing them: the fixed cosine grid then clamps out-of-range
+/// components at encode with no error and no counter, which measured
+/// −10 pts recall@10 the last time it shipped. A `debug_assert` means
+/// the next such path fails in tests instead of silently degrading data,
+/// and costs release builds nothing.
+#[cfg_attr(not(debug_assertions), allow(unused_variables))]
+fn debug_assert_cosine_rows_unit(metric: Metric, dim: usize, vectors: &[f32]) {
+    #[cfg(debug_assertions)]
+    if metric == Metric::Cosine {
+        for (row, chunk) in vectors.chunks_exact(dim).enumerate() {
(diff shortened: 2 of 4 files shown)


muralikpbhat (approved):
Approving. I verified the two load-bearing claims against the code:

- **The seam really is the choke point.** `stream_fp32_rows_to_buckets` has exactly one caller — the fp32-source arm of `build_merged_subsection_from_fp32` — and that function is the funnel for every fp32 cell build, including the commit-time hidden-index pack that bypassed #520's `VectorBuilder::add` seam (`VectorColumnView` → `PackRow::Fp32`, zero-copy, never through the builder). The diagnosis is exactly right: the user table stored unit rows while the hidden index, which serves, stored clamped ones.
- **The tripwire has no bypass.** `encode_rows` has exactly one caller (`encode_blob`), which carries the debug assert — so the next ingest path that skips a normalize seam fails in tests instead of silently costing ~10 recall points.

Details that check out: the normalized chunk feeds everything downstream (assignment, rotation, 1-bit codes, residual payload), so routing and stored bytes are consistent on unit rows — and for cosine, per-row scaling doesn't change centroid argmax, so assignment semantics are unchanged; the normalize is cosine-gated (L2/NegDot untouched); zero-row policy is consistent between the normalizer and the assert; the chunk-bounded scratch placement (not at `append()`) is the right call given the commit-wide copy it avoids; the assert tolerance is well-chosen; and the regression test drives the public `append()` path end to end with magnitude-12 rows and asserts served-recall parity against the pre-normalized corpus.

One doc suggestion, non-blocking: since both seams now silently unit-normalize, a cosine column's stored fp32 differs from the caller's input on read-back (magnitudes are not preserved). Worth one line in the docs/binding docstrings — "cosine columns store unit-normalized vectors" — so it's a documented contract rather than a discovery; callers who need magnitudes keep them in a scalar column or choose an unbounded metric.

PR #583, opened 2026-08-11
supertable/vector: time the 1-bit scan on the deferred path

ekechinwokah:
`vec.shortlist` was recorded only in `probe_clusters_async`, the immediate probe path. Every warm hidden-index query is served by `search_clusters_scan_async` instead, where the scan went untimed and surfaced as unattributed `vec.fanout_wall` — so the engine could not see its own largest phase, and that phase's share ended up being argued from perf samples rather than measured.

Also make the phase-profile diagnostic's cache budget env-tunable (INFINO_DIAG_CACHE_GIB, default unchanged at 20 GiB). Against a 33 GiB table the hard-coded default profiles partly-cold I/O: 45 ms wall where the warm sweeps on the same table and code read 20 ms.

Measured with both, Cohere-10M k=10 warm (200 queries after warmup, width 35, rerank budget 3683): wall p50 23.1 ms; `vec.admit` 3.0 ms single-threaded ahead of the fan-out; `vec.fanout_wall` 22.4 ms, filled by the scan; `vec.rerank` 1.7 ms and `vec.survivor_fetch` 0.3 ms. The per-cell phases are summed across concurrently scanned cells and are inflated by queueing on the shared reader pool (464 ms of `vec.shortlist` against 22.4 ms of wall), so they bound wall time, not CPU.

### DIFF
--- a/src/superfile/vector/reader.rs
+++ b/src/superfile/vector/reader.rs
@@ -3567,9 +3567,20 @@ impl VectorReader {
                         budget,
                     };
                     let coarse_limit = k.saturating_mul(rerank_mult);
+                    // Record `vec.shortlist` here too. It was only recorded
+                    // on the immediate probe path, but every warm
+                    // hidden-index query is served by THIS arm, so the 1-bit
+                    // scan went untimed on the path that dominates the
+                    // query — it surfaced as unattributed `vec.fanout_wall`,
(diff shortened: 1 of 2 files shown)


ekechinwokah:
not needed in main.

PR #586, opened 2026-08-11
fts: df-anchor the disjunction count when one term dominates

vinaykakade:
## What

Speeds up unranked `OR` counts dominated by one very common term. `or_count_unranked` walked every posting of every term — including the entire list of a term appearing in a large fraction of the corpus — just to produce a cardinality. When one term's df dwarfs the rest, that full walk is the whole cost.

## How

For a disjunction whose size is dominated by one long list:

```
|A ∪ B ∪ …| = df(D) + |(the other terms' union) \ D|
```

where `D` is the highest-df term. `df(D)` is read from `TermCursor.df` (no decode); the rarer cursors are frontier-merged, and each distinct doc they produce is membership-tested against `D` by advancing its skip table monotonically. `D`'s skip table is traversed once instead of its postings decoded end to end — cost drops from `O(Σ df)` to `O(Σ df_rest + skip-traversal(D))`.

Gated behind a conservative dominance threshold (`8×`): the anchored path is taken only when the dominant term's df is at least 8× the others' combined, so balanced unions keep the windowed-bitset walk, which is already faster when no term dominates. The threshold was calibrated from measurement — a skip-probe costs several× a windowed `next()`, so the crossover sits well above a naive ratio (an earlier `4×` cut regressed a few unions whose dominance was only ~4–6×).

## Correctness

Result-identical: docs in `D` are counted once via its df, docs in the rarer union are added once each iff `D` lacks them, and a doc in both counts only through the df. A new test drives the anchored path on a multi-window corpus with gaps where the dominant term is absent (exercising the "doc not in the anchor" branch) and asserts equality with the independent naive merge. Full-corpus check: **0 count mismatches across all 962 queries** of the standard bench set.

## Measurement

Local A/B over the full 962-query count set (min-of-10, same machine, before = `main`, index unchanged — this is a read-path-only change):

| slice | AVG before→after | P90 before→after |
| ----- | ---------------- | ---------------- |
| union | 1612 → 1284 µs (**−20%**) | 7202 → 3313 µs (**−54%**) |
| ALL count | 1560 → 1467 µs | 3229 → 2658 µs |

Dominant-common-term unions collapse: `the incredibles` 7360 → **69 µs**, `the preakness` 7325 → 127 µs, `the progressive` 7660 → 1430 µs, `helen of troy` 6722 → 1273 µs. Non-union slices (phrase, intersection) are unchanged — those paths aren't touched.

## Verification

- `cargo test --lib superfile::fts` + `--test superfile fts::` + `--test supertable query::` green.
- `cargo fmt --check` and `clippy --all-targets --features test-helpers -- -D warnings` clean.

### DIFF
--- a/src/superfile/fts/reader/core.rs
+++ b/src/superfile/fts/reader/core.rs
@@ -1579,19 +1579,26 @@ pub(super) fn or_count_unranked(mut cursors: Vec<TermCursor>) -> u64 {
 /// df read + skip-probe wins. Requires ≥ 2 cursors (a single cursor's
 /// count is just its df, and the windowed walk handles it trivially).
 fn dominant_anchor_index(cursors: &[TermCursor]) -> Option<usize> {
-    if cursors.len() < 2 {
+    dominant_anchor_of_dfs(cursors.iter().map(|c| c.df))
+}
+
+/// The routing decision behind [`dominant_anchor_index`], over raw dfs so
+/// the boundary behaviour is unit-testable without building cursors.
+/// Returns the index of the term whose df is at least
+/// [`OR_COUNT_ANCHOR_DOMINANCE`]× the sum of all the others' (a `>=`
+/// boundary — exactly `N×` still routes to the anchor), or `None` when no
+/// term dominates or there are fewer than two terms.
+fn dominant_anchor_of_dfs(dfs: impl IntoIterator<Item = u64>) -> Option<usize> {
+    let dfs: Vec<u64> = dfs.into_iter().collect();
+    if dfs.len() < 2 {
         return None;
     }
-    let (max_idx, max_df) = cursors
-        .iter()
-        .enumerate()
-        .map(|(i, c)| (i, c.df))
-        .max_by_key(|&(_, df)| df)?;
-    let others_df: u64 = cursors
+    let (max_idx, &max_df) = dfs.iter().enumerate().max_by_key(|&(_, df)| *df)?;
+    let others_df: u64 = dfs
         .iter()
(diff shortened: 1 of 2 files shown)


src/superfile/fts/reader/core.rs
```diff
@@ -1546,6 +1571,72 @@ pub(super) fn or_count_unranked(mut cursors: Vec<TermCursor>) -> u64 {
...
 }
 
+/// Pick the cursor whose df dominates the union, or `None` if no term is
+/// dominant enough for the anchored count to pay off. Dominant means the
+/// largest df is at least [`OR_COUNT_ANCHOR_DOMINANCE`]× the sum of all
+/// the other dfs — so the dominant list is longer than everything else
+/// combined by a wide margin, exactly when replacing its full walk with a
+/// df read + skip-probe wins. Requires ≥ 2 cursors (a single cursor's
+/// count is just its df, and the windowed walk handles it trivially).
+fn dominant_anchor_index(cursors: &[TermCursor]) -> Option<usize> {
```
naman-modi, line 1581: hey Vinay, can we please add more tests for this to check the routing at the boundary levels:
- 8x exactly (800 vs 100) → dominant
- just below (799 vs 100) → not
- single term → not
- two-way tie (500 vs 500) → not

naman-modi (approved):
LGTM, minor comment.

src/superfile/fts/reader/core.rs
vinaykakade: Done

PR #588, opened 2026-08-11
vector: centroid-HNSW ivf router for the hidden vector index

muralikpbhat:
### Problem / motivation
Real-dataset scale runs — **LAION 100M**, tracked in #490 — showed the engine is **multi-second per query at 100M**, in *both* regimes:

- **warm** (cache-resident, 0 GETs): **~3–5 s**. The stamped router scans whole IVF cells, and #490 shows it probes ~29 of 256 cells for `k=10` — at 100M that's ~11.6M vectors scored per query, *even with the data cached*. This is compute, not I/O.
- **cold** (first touch): ~4–9 s (#490), Blob-read-bound as each serial query cold-fetches its cells.

Single-digit-ms latency needs a router that scores **far fewer candidates**. This PR adds one.

### What this adds
`ivf_router = centroid_graph`: instead of scanning whole stamped cells, it walks an HNSW over the resident fp32 fine centroids and scores only the top-`fanout` **fine clusters**. **Rebased onto current `main` (post-#603).**

```yaml
search_mode: ivf | hnsw_ivf
ivf_router:  stamped | centroid_graph   # ivf only; default stamped
```
Folds the experimental `global_fine_centroid` search mode into `ivf_router = centroid_graph` and drops the intermediate brute-force scan (superseded by the graph). Default stays `stamped`.

### stamped vs centroid_graph — LAION 100M, 768-d, co-located Azure Blob

| metric | `stamped` | `centroid_graph` |
|---|---|---|
| **warm** p50 (cache-resident, 0 GETs) | ~4.0 s (p95 6.4 s) | **~30 ms** |
| **cold** p50 | ~5.6 s | **~5.4 s** |
| cold q0 (fully cold, first query) | 15.4 s | 19.4 s |
| cold bytes / query (q0) | 12.8 GiB | **9.7 GiB** |
| recall@10 (fanout-dependent, see below) | 0.988 | 0.990–0.993 |

**The headline is warm: ~135× (4.0 s → 30 ms).** Scoring `fanout` targeted fine clusters instead of whole cells removes the per-query compute wall. Cold lands at **parity** with `stamped` (on fewer bytes) — but only after the fan-out fix below; without it the router's cold p50 was ~56 s (its per-cell reads were issued serially).

### Cold path: parallel per-cell fan-out

The router's phase-2 scan over the selected fine clusters ran one superfile at a time, awaiting each cell's cold read before the next — RTT-serializing the object-store fetches. The stamped whole-cell path already fans its per-superfile reads out concurrently; this brings the graph path to the same discipline: the per-superfile scans run through `buffer_unordered` bounded to the reader-pool width, sharing the same connection-memory budget and reader pool (so concurrent cold fetches still gate on `OverBudget`). Survivors are pooled and reranked exactly as before. Reads are unchanged — same clusters, same bytes — so recall is identical.

LAION 100M, `centroid_graph`, fanout 2048 (cold p50 over 20 cold queries):

| | serial | parallel |
|---|---|---|
| cold **p50** | 55.9 s | **5.4 s** |
| cold q0 (fully cold) | 60.5 s | 19.4 s |

~10× lower cold p50, pure read concurrency. The gain scales with how much cold I/O is serialized: **~10× at 100M**, **~2× at 10M** (2.05 s → 0.97 s). Recall verified unchanged (Cohere 10M `centroid_graph`, identical before and after).

The remaining q0 gap (19.4 s vs `stamped`'s 15.4 s) is the one-time per-handle centroid/graph load on the first query. **Building the graph at drain and persisting it (instead of per-handle at first query) is the next follow-up PR** — it removes that first-query build and the q0 tail entirely.

### Recall by fanout — Cohere 10M (cache-independent)

| router / fanout | recall@10 |
|---|---|
| `stamped` (ivf) | 0.988 |
| `centroid_graph`, 512 | 0.990 |
| `centroid_graph`, 1024 | 0.993 |

Ranking centroids by normalized cosine also beats the dropped scan's raw-dot ranking: 0.956 → 0.983 at fanout 128 on Cohere-1M.

### Centroid-HNSW router
- Built once per handle from the resident centroid section at the first query (no drain change); reuses #603's HNSW machinery.
- A graph node maps back to the exact `flat` cluster id the stamped read plan expects, so phase 2 is **byte-identical**.
- Knobs: `global_fine_fanout`, `global_fine_graph_ef` (0 = auto), `global_fine_rerank_mult`.
- **Next follow-up PR:** build + persist the graph at drain time (removes the first-query build and the q0 tail). Later: per-table fanout/ef calibration.

### Enabling fix (reader-cache)
Included because the router is unusable at scale without it: block-cached entries for sparse vector blobs reserved their whole superfile size up front, so a wide fanout over-reserved the cache budget and evicted live entries — vector reads never warmed. Now they account their live filled bytes (source-owned), gated on the promotion signal so FTS/SQL mmap promotion is unaffected.

### Validation
Full lib suite green (2246); clippy clean; `reader_cache` accounting suite 67/67. Parallel fan-out: reads byte-identical (recall unchanged), cold p50 ~10× at 100M / ~2× at 10M, memory-safe under concurrency (shared budget + pool).

### DIFF
--- a/src/supertable/options.rs
+++ b/src/supertable/options.rs
@@ -415,6 +415,14 @@ pub struct SupertableOptions {
     /// Warm queries never touch it — they resolve on the cache mutex's fast
     /// path — so the download never serializes steady-state serving.
     pub(crate) graph_hydration_lock: Arc<TokioMutex<()>>,
+    /// Build-once cache for the in-memory centroid-router HNSW (an HNSW over
+    /// the resident fp32 fine centroids that replaces the brute-force
+    /// `global_fine_cluster_scores` scan on the global-fine path). Built from
+    /// the resident centroid section at first global-fine query, so testing it
+    /// needs no re-drain; the persisted centroid graph is a follow-on.
+    /// Experimental: only populated when the global-fine graph path is enabled.
+    pub(crate) centroid_router_cache:
+        Arc<tokio::sync::OnceCell<Arc<crate::supertable::query::vector::CentroidRouterGraph>>>,
     /// Read-time reverse (`stable_id -> local`) lookup backing scalar
     /// projection over gapped user superfiles, so a hit resolves in O(k) after
     /// a one-time per-superfile build instead of the per-query O(corpus) `_id`
@@ -742,6 +750,7 @@ impl SupertableOptions {
             centroid_section_cache: Arc::new(TokioMutex::new(None)),
             graph_sections_cache: Arc::new(TokioMutex::new(None)),
             graph_hydration_lock: Arc::new(TokioMutex::new(())),
+            centroid_router_cache: Arc::new(tokio::sync::OnceCell::new()),
             gapped_id_placement_cache: Arc::new(TokioMutex::new(GappedIdPlacementCache::default())),
             user_centroid_cache: Arc::new(TokioMutex::new(None)),
             prepopulate_cache_on_commit: true,
--- a/src/supertable/options.rs
+++ b/src/supertable/options.rs
@@ -416,11 +416,11 @@ pub struct SupertableOptions {
     /// path — so the download never serializes steady-state serving.
     pub(crate) graph_hydration_lock: Arc<TokioMutex<()>>,
     /// Build-once cache for the in-memory centroid-router HNSW (an HNSW over
-    /// the resident fp32 fine centroids that replaces the brute-force
-    /// `global_fine_cluster_scores` scan on the global-fine path). Built from
-    /// the resident centroid section at first global-fine query, so testing it
-    /// needs no re-drain; the persisted centroid graph is a follow-on.
-    /// Experimental: only populated when the global-fine graph path is enabled.
+    /// the resident fp32 fine centroids, used by `ivf_router = centroid_graph`
+    /// to select clusters). Built from the resident centroid section at the
+    /// first such query, so testing it needs no re-drain; the persisted
+    /// centroid graph is a follow-on. Only populated when the centroid-graph
+    /// router is enabled.
     pub(crate) centroid_router_cache:
         Arc<tokio::sync::OnceCell<Arc<crate::supertable::query::vector::CentroidRouterGraph>>>,
     /// Read-time reverse (`stable_id -> local`) lookup backing scalar
--- a/src/supertable/query/vector.rs
+++ b/src/supertable/query/vector.rs
@@ -785,7 +785,12 @@ fn build_centroid_router(
     let mut node_map: Vec<(usize, u32)> = Vec::new();
     for (si, reader) in readers.iter().enumerate() {
         let Some(vr) = reader.vec() else { continue };
-        let sfid = superfiles[si].superfile_id;
+        // Checked access: a concurrent optimize can shift the reader/superfile
+        // set while the router builds; skip a missing entry rather than panic.
+        let Some(sf) = superfiles.get(si) else {
+            continue;
+        };
+        let sfid = sf.superfile_id;
         for (flat, mut vec) in vr
             .global_fine_cluster_vectors(column, section, sfid)
             .map_err(|e| QueryError::Execute(e.to_string()))?
@@ -2788,11 +2793,28 @@ impl SupertableReader {
             warn_hnsw_no_resident_graph(column, manifest.slow_vector_state_graphs_blob().is_some());
             // fall through to the ivf scan below
         }
+        // The centroid router ranks by cosine (unit-normalized centroids, a
+        // -dot scorer), so it is only correct for a Cosine column. A NegDot or
+        // L2Sq table falls through to the stamped router rather than being
+        // mis-ranked; per-metric centroid scoring is a router-productionization
+        // follow-on.
+        let centroid_graph_metric_ok = manifest
+            .options
+            .vector_columns
+            .iter()
+            .find(|vc| vc.column == column)
+            .is_some_and(|vc| {
+                matches!(
+                    vc.metric,
+                    crate::superfile::vector::distance::Metric::Cosine
+                )
+            });
         if !filtered
             && hidden_vector_index
             && vcfg.search_mode == config::VectorSearchMode::Ivf
             && vcfg.ivf_router == config::IvfRouter::CentroidGraph
             && vcfg.global_fine_fanout > 0
+            && centroid_graph_metric_ok
         {
             return self
                 .global_fine_fanout(
--- a/src/supertable/query/vector.rs
+++ b/src/supertable/query/vector.rs
@@ -746,6 +746,65 @@ fn union_cell_selection(grid: &[u32], fine: &[u32]) -> Vec<u32> {
     selected
 }
 
+/// In-memory centroid router: an HNSW over the pooled fp32 fine centroids,
+/// used on the global-fine path to select the top-`fanout` clusters in place
+/// of the brute-force `global_fine_cluster_scores` scan. Experimental,
+/// validation-grade — built once per handle from the resident centroid section
+/// (no drain change). A graph node maps back to the `(superfile index, flat
+/// cluster id)` the scan path would have produced, so the downstream per-cell
+/// read plan is byte-identical.
+pub(crate) struct CentroidRouterGraph {
+    scorer: crate::superfile::vector::hnsw::Fp32Scorer,
+    graph: crate::superfile::vector::hnsw::Hnsw,
+    node_map: Vec<(usize, u32)>,
+}
+
+/// Unit-normalize in place so the centroid graph's `−dot` scorer ranks by
+/// cosine (the fine centroids are means of unit vectors, not themselves unit).
+fn gfc_unit_normalize(v: &mut [f32]) {
+    let n: f32 = v.iter().map(|x| x * x).sum::<f32>().sqrt();
+    if n > 0.0 {
+        let inv = 1.0 / n;
+        for x in v.iter_mut() {
+            *x *= inv;
+        }
+    }
+}
+
+/// Build the in-memory centroid router from the resident centroid section.
+/// `readers[si]` corresponds to `superfiles[si]`; centroids are pulled via
+/// `global_fine_cluster_vectors` so the flat ids match the scan path exactly.
+fn build_centroid_router(
+    superfiles: &[Arc<SuperfileEntry>],
+    readers: &[Arc<SuperfileReader>],
+    column: &str,
+    section: &crate::supertable::slow_vector_state::CentroidSection,
+    dim: usize,
+) -> Result<CentroidRouterGraph, QueryError> {
+    use crate::superfile::vector::hnsw::{Fp32Scorer, Hnsw, HnswParams};
+    let mut vecs: Vec<Vec<f32>> = Vec::new();
+    let mut node_map: Vec<(usize, u32)> = Vec::new();
+    for (si, reader) in readers.iter().enumerate() {
+        let Some(vr) = reader.vec() else { continue };
+        let sfid = superfiles[si].superfile_id;
+        for (flat, mut vec) in vr
+            .global_fine_cluster_vectors(column, section, sfid)
+            .map_err(|e| QueryError::Execute(e.to_string()))?
+        {
+            gfc_unit_normalize(&mut vec);
+            vecs.push(vec);
+            node_map.push((si, flat));
+        }
+    }
+    let scorer = Fp32Scorer::from_vectors(&vecs, dim);
+    let graph = Hnsw::build(&scorer, HnswParams::default());
+    Ok(CentroidRouterGraph {
+        scorer,
+        graph,
+        node_map,
+    })
+}
+
 /// Widen a grid cell cutoff so the probed cells' indexed row counts cover at
 /// least `k`. `grid_cell_cutoff`'s slack window stops at the nearest near-tie
 /// cluster — a single cell for a query blended between two clusters — so a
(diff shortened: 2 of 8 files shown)


muralikpbhat:
**10M + 100M A/B: centroid-graph router vs brute-force centroid scan**

Ran `gfc-graph` (HNSW over the fp32 fine centroids) against `gfc-scan` (brute-force centroid scan) and the `ivf` baseline at 10M (Cohere, 768-d) and 100M (LAION, 768-d). Headline: **the graph is at parity with the scan it replaces — no clear win yet at current cluster sizes.**

**100M — warm (full cache)**

| metric | ivf | scan@2048 | graph@2048 | graph@4096 |
|---|---|---|---|---|
| recall@10 | 0.986 | 0.988 | 0.988 | 0.990 |
| p50 (ms) | 3064 | 1018 | 1048 | 1874 |
| p95 (ms) | 4607 | 1560 | 1550 | 3011 |

**100M — cold (low cache)**

| metric | ivf | graph@2048 | graph@4096 |
|---|---|---|---|
| recall@10 | 0.986 | 0.986 | 0.991 |
| bytes/query | 11.1 GiB | 5.8 GiB | 11.1 GiB |
| GETs/query | 284 | 920 | 1535 |

**10M — 12 GB cache**

| metric | ivf | scan@1024 | graph@512 | graph@1024 |
|---|---|---|---|---|
| recall@10 | 0.988 | 0.9875 | 0.990 | 0.993 |
| bytes/query | 4.0 GiB | 2.6 GiB | 1.4 GiB | 2.6 GiB |
| GETs/query | 51 | 235 | 167 | 219 |
| p50 (ms) | 5462 | 4424 | 3013 | 4672 |

**Takeaways**

- **Warm, graph ≈ scan** (100M: 1048 vs 1018 ms p50 at equal recall). The HNSW walk doesn't beat the brute-force centroid scan yet — at ~880-row fine clusters the centroid count is small enough that selection is already cheap relative to scoring, so replacing it saves little.
- **Graph edges scan on recall** (10M: 0.990 vs 0.9875) — it ranks centroids by true cosine (normalized), where the scan uses raw dot.
- **The cold cache-fit tradeoff vs `ivf` is unchanged**: `gfc` reads fewer bytes but issues more (scattered) GETs, so cold cost/latency rise. The graph selects the same clusters as the scan, so it doesn't move this.

**Next lever: fine-cluster size.** Shrinking clusters raises the centroid count (where the graph router should start to earn its keep) and cuts the per-query candidate set. Building smaller-cluster tables (256 KiB fine-run target, ~100-row) at 10M and 100M now to measure recall / bytes / latency against these baselines.

muralikpbhat:
**Follow-up: fine-cluster size at 10M — doesn't move warm latency at matched recall**

Tested the "next lever" from above — shrinking fine clusters. Built a **~100-row** table (256 KiB fine-run target) alongside the **~880-row** baseline (2 MiB target), 10M Cohere, 12 GB cache.

| arm | ~880-row recall | ~880-row p50 / cold | ~100-row recall | ~100-row p50 / cold |
|---|---|---|---|---|
| ivf (stamped) | 0.988 | 5462 / 4.00 | 0.9845 | 7759 / 4.09 |
| graph@512 | 0.990 | 3013 / 1.44 | 0.9685 | 2496 / 0.74 |
| graph@1024 | 0.993 | 4672 / 2.65 | 0.9855 | 3357 / 1.29 |
| graph@2048 | — | — | 0.996 | 4068 / 2.09 |

*(p50 ms, cold GiB)*

**At matched fanout**, ~100-row reads ~half the bytes but loses recall (512: 0.9685 vs 0.990) — fewer true neighbors land in 512 small clusters than 512 big ones.

**At matched recall (~0.99)**:
- ~880-row: `graph@512` → 0.990 at 1.44 GiB, **p50 3013 ms**
- ~100-row: needs `~1450` fanout → ~0.99 at ~1.6 GiB, **p50 ~3700 ms**

→ **Smaller clusters don't improve warm latency at matched recall** — higher fanout offsets the smaller per-cluster scan, so candidate count (fanout × cluster_size) stays ~flat. The one real gain is **recall-per-byte at the high end**: ~100-row reaches **0.996 at 2.09 GiB**, which ~880-row can't hit even at 2.65 GiB — finer routing helps the recall ceiling and cold cost, not warm speed.

**Build cost:** the fine k-means (cell-build) runs **~4.3× slower** at 256 KiB (185 s → 804 s at 10M) — ~8× more centroids to train. CPU-bound, memory flat (RSS unchanged).

**Takeaway:** cluster size alone doesn't move the warm-latency floor, which means the floor scales with **fanout / cluster-count or fixed per-query overhead**, not raw candidate count. Profiling the warm path (selection vs scoring vs rerank split, thread occupancy) next to localize it; a ~100-row build at 100M is in progress for the cold/recall-ceiling side.

*(These are cache-pressured numbers — latency mixes some I/O; the warm profile will isolate compute.)*

muralikpbhat:
**Correction: the warm-latency (p50) columns in the two comments above were contaminated — real warm is ms-scale**

The `p50` numbers in both comments (the ~880-row *and* ~100-row tables) were measured on a **loaded box** — a concurrent job competed for CPU/memory during the timed pass, so those "seconds" are contention artifacts, not warm compute. Re-measured on an **idle** box, same tables/queries/fanout:

| 10M, fanout 2048 | posted p50 | clean warm p50 | p95 | recall |
|---|---|---|---|---|
| ~880-row | — (4672 @1024) | **15.1 ms** | 16.5 | 0.991 |
| ~100-row (256 KiB) | **4067.7** | **12.7 ms** | 14.9 | 0.992 |

Working set is ~70 MB (RAM-resident); the tight p95 with no fat tail confirms the clean run.

**What stands / what changes:**
- **recall and cold-bytes / GETs columns remain valid** (recall is cache-independent; cold bytes = first-touch). The cache-fit tradeoff and recall-per-byte findings hold.
- **The warm-latency conclusion is retracted.** Warm 10M is ~13–15 ms, not seconds — so "candidate scoring is a ~1 s floor" was wrong; that was I/O + contention. Scan is ~12 ms for ~1.8M candidates when RAM-resident.
- **Cluster size doesn't move warm latency** (15 → 12.7 ms) and costs ~4.3× build-time fine k-means, so it's not a warm lever — its only edge is the cold recall-per-byte ceiling.

**Net:** the real axis is **cold / cache-miss / working-set-vs-RAM**, not warm compute. One number still to verify cleanly: 100M warm (posted 1048 ms at 0 GETs) — re-measuring on an idle box next.

muralikpbhat:
**Rebased + restructured (current state).** This PR was rebased onto `main` (post-#603) and reworked: routing is now an **ivf router choice** (`ivf_router = stamped | centroid_graph`, default `stamped`), the intermediate brute-force scan is **dropped**, and the description above is updated to match.

The comments below pre-date this — they're the earlier `scan`-vs-`graph` and cluster-size exploration that led here. The p50-contamination note (that the multi-second latencies were box contention, not compute) still applies to those older tables; the numbers in the updated description are from clean idle-box runs.

muralikpbhat:
**Cold-path parallel fan-out — validation**

The `centroid_graph` router's phase-2 scanned its selected clusters one superfile at a time in a serial `await`, which RTT-serialized the cold Blob fetches — the dominant cost behind the multi-second cold latencies in #490. Fanning the per-superfile scans out concurrently (`buffer_unordered` at reader-pool width, sharing the connection-memory budget so cold fetches still gate on `OverBudget`) closes it. Same clusters read, so recall is untouched.

Measured cold p50 over 20 cold queries, fully-cold first query = q0, 100 GB cache, co-located Azure Blob.

**LAION 100M, fanout 2048:**

| | serial | parallel |
|---|---|---|
| cold p50 | 55.9 s | **5.4 s** |
| cold q0 | 60.5 s | 19.4 s |
| GETs / bytes (q0) | 6851 / 9.7 GiB | 6851 / 9.7 GiB |

**Cohere 10M, fanout 1500:**

| | serial | parallel |
|---|---|---|
| cold p50 | 2.05 s | **0.97 s** |
| cold q0 | 9.4 s | 6.5 s |
| GETs / bytes (q0) | 1266 / 4.9 GiB | 1266 / 4.9 GiB |
| recall@100 | 0.9925 | 0.9925 |

Identical GETs, bytes, and recall at both scales — pure read concurrency. The gain tracks how much cold I/O was serialized: ~10× at 100M, ~2× at 10M. Warm improves as a side effect (the concurrent scan futures overlap their scoring).

Cold p50 now matches `stamped` (~5.4 s vs ~5.6 s — the #490 LAION baseline) on fewer bytes. Residual: cold q0 (19.4 s vs `stamped` 15.4 s) is the one-time first-query centroid/graph load — the drain-persisted-graph follow-up removes it.

Hardening: the concurrent scans share the connection-memory budget and reader pool, so wide-fanout cold fetches gate on `OverBudget` exactly as the stamped fan-out does (no unbounded in-flight reads).

savannahar68:
did we measure how much was memory overhead for this new hnsw graph for 100mn and 10mn scale

src/supertable/options.rs
```diff
@@ -415,6 +415,14 @@ pub struct SupertableOptions {
...
     /// path — so the download never serializes steady-state serving.
     pub(crate) graph_hydration_lock: Arc<TokioMutex<()>>,
+    /// Build-once cache for the in-memory centroid-router HNSW (an HNSW over
+    /// the resident fp32 fine centroids, used by `ivf_router = centroid_graph`
+    /// to select clusters). Built from the resident centroid section at the
+    /// first such query, so testing it needs no re-drain; the persisted
+    /// centroid graph is a follow-on. Only populated when the centroid-graph
+    /// router is enabled.
+    pub(crate) centroid_router_cache:
+        Arc<tokio::sync::OnceCell<Arc<crate::supertable::query::vector::CentroidRouterGraph>>>,
```
savannahar68, line 425: Can you elaborate a bit on how do we plan to share the hnsw graph created if infino engine is running in multiple nodes sharing the underlying data - i don't see it as part of the PR - not sure given this is expiremental if we want to think about it or not

Also if we rebuild this in every node which is running the engine, we take overhead across, not sure if we should do it - thoughts ?

src/supertable/query/vector.rs
```diff
@@ -2544,38 +2576,60 @@ impl SupertableReader {
...
             readers.push(reader);
         }
-        if cands.is_empty() {
+
+        // Cluster selection: the centroid-router HNSW walks a graph over the
+        // resident fp32 fine centroids to pick the top-`fanout` clusters; phase
+        // 2 then reads only those clusters. The graph is built once (cached) at
+        // the first query from the same fine centroids, so a graph node maps
+        // back to the exact `flat` cluster id and the read plan is identical.
+        let by_sf: HashMap<usize, Vec<u32>> = {
```
savannahar68, line 2599: in case of drain / optimize the graph becomes stale and needs to be dropped / rebuild. This logic is missing imo in this PR

src/supertable/query/vector.rs
```diff
@@ -761,6 +746,64 @@ fn union_cell_selection(grid: &[u32], fine: &[u32]) -> Vec<u32> {
...
+    column: &str,
+    section: &crate::supertable::slow_vector_state::CentroidSection,
+    dim: usize,
+) -> Result<CentroidRouterGraph, QueryError> {
+    use crate::superfile::vector::hnsw::{Fp32Scorer, Hnsw, HnswParams};
+    let mut vecs: Vec<Vec<f32>> = Vec::new();
+    let mut node_map: Vec<(usize, u32)> = Vec::new();
+    for (si, reader) in readers.iter().enumerate() {
+        let Some(vr) = reader.vec() else { continue };
+        let sfid = superfiles[si].superfile_id;
```
savannahar68: better to use a getter instead of refering to index which might panic if it doesn't exist in case optimize runs when we are building centroid router

src/supertable/options.rs
muralikpbhat: Yes, we have that item open even for the HNSW vectors. For HNSW of centroids here, it is much more smaller graph and we can reload the graph every time after drain. The plan is not to build the graph in every node. See https://github.com/infino-ai/infino/issues/630. One of the idea is to load the graph from S3 when we detect more drains compared to last graph build point. We will write about the exact design in that GH issue.

src/supertable/query/vector.rs
```diff
@@ -761,6 +746,64 @@ fn union_cell_selection(grid: &[u32], fine: &[u32]) -> Vec<u32> {
...
+/// per-cell read plan expects, so the downstream read is byte-identical.
+pub(crate) struct CentroidRouterGraph {
+    scorer: crate::superfile::vector::hnsw::Fp32Scorer,
+    graph: crate::superfile::vector::hnsw::Hnsw,
+    node_map: Vec<(usize, u32)>,
+}
+
+/// Unit-normalize in place so the centroid graph's `−dot` scorer ranks by
+/// cosine (the fine centroids are means of unit vectors, not themselves unit).
+fn gfc_unit_normalize(v: &mut [f32]) {
```
savannahar68, line 763: Should we pass the column's configured metric in here and branch the calculation accordingly? Right now gfc_unit_normalize always normalizes, and Fp32Scorer always ranks by plain dot product so this always routes by cosine similarity regardless of what metric the table is actually configured with. That's correct for Cosine, but for NegDot it means real magnitude information gets thrown away before ranking, and for L2Sq it's further off as straight-line distance depends on each centroid's own length too, so just skipping the normalize step for L2Sq isn't quite enough it likely needs an actual distance calculation

savannahar68 (commented):
i think we have not handled centroid_router_cache with optimize/ drain

Do you want to take it as part of separate PR given the default routing strategy is still the older one.

src/supertable/query/vector.rs
muralikpbhat: Agreed — this is cosine-only today: `gfc_unit_normalize` + the `-dot` scorer rank by cosine regardless of the column's configured metric, which drops magnitude for NegDot and isn't straight-line distance for L2Sq. Guarded in 273dcce: the `centroid_graph` router now only engages for a Cosine column and falls back to the stamped router for NegDot/L2Sq, so a non-cosine table is never mis-ranked. Full per-metric centroid scoring (so non-cosine tables can use the graph too) is tracked in the router-productionization follow-up (#630).

src/supertable/query/vector.rs
muralikpbhat: Fixed in 273dcce — switched `superfiles[si]` to a checked getter (`superfiles.get(si)`), so a concurrent optimize that shifts the reader/superfile set mid-build skips the entry instead of panicking.

src/supertable/query/vector.rs
muralikpbhat: Good catch — you're right the build-once cache goes stale after a drain, and it's a correctness issue, not just recall: the graph's nodes map to `(superfile, flat cluster id)` valid only for the drain generation they were built against, so after a drain/optimize (split/merge/supersede) those mappings can point at rewritten/superseded clusters or misalign with the reader set. The refresh is generation-keyed, every drain (no threshold): single-box, the drain pushes an in-memory invalidation so steady-state queries pay nothing and only the first post-drain query rebuilds; multi-node, the reload rides the manifest refresh serving nodes already do to pick up new drains, so there's no new per-query poll. Since this path is opt-in and the default router is `stamped`, I'd rather not land cache-invalidation machinery in this experimental PR — I've written the full refresh design into #630 and it's tracked there.

muralikpbhat:
> i think we have not handled centroid_router_cache with optimize/ drain
> 
> Do you want to take it as part of separate PR given the default routing strategy is still the older one.

Thanks. Yes, it will be handled in #630 as part of making this prod ready at scale(right now goes stale after next drain).

savannahar68 (approved):
lgtm
rest fixes will be part of separate PR

src/supertable/query/vector.rs
savannahar68: worth taking a look at other places in PR if similar pattern exist. I think i saw `readers[si].as_ref().vec()` somewhere

muralikpbhat: Good eye — the same `readers[si]` / `superfiles[si]` accesses recur in the fanout scan (~L2668/2705/2728). Those are only reachable out-of-bounds when the cached router's `node_map` (built at an earlier generation) is walked against the current readers — i.e. the stale-graph condition from the other thread. So the defensive getters there belong with the drain-invalidation that fixes the root cause, and I've grouped both in the follow-up (#630) rather than splitting the hardening from its cause. This PR keeps the one getter on the build path.

PR #589, opened 2026-08-12
bench/ci: exclude cold and footprint metrics from the merge gate

vinaykakade:
## Problem

The benchmark merge gate blocked a PR when any lower-is-better nanosecond metric was worse than the same-host `main` baseline by **both >50% and >5 ms**. That scope was too wide and swept in two classes of metric that vary far too much to gate on:

- **Cold-cache latencies** (`cold 1st query (median)`, `cold open (median)`, `+fetch cold`) are object-store first-touch timings. They swing hundreds to thousands of percent run-to-run purely from network and VM page-cache state. On *identical* code they trip both legs of the gate, so PRs that changed no engine code failed — and the failure also fires on `main` post-merge.
- **Peak memory-footprint columns** (`Peak file`, `Peak anon`) are byte sizes, not latencies. `is_latency()` only excluded the literal `"rss"` token, so these slipped through, got misformatted as `"ms"` by `human()`, and were gated as if a `338 MiB -> 524 MiB` move were a `+62%` latency regression.

### Observed

Recent no-code / post-merge runs failed the gate with blocking lists that were **100% cold or footprint** metrics — no warm latency moved:

```
# release version bump (no engine code):
ten_term_or / cold 1st query (median): 50.28 ms -> 1.47 s (+2823%)
three_wide_or / cold 1st query (median): 58.77 ms -> 685.55 ms (+1066%)
... (12 blocking, all cold)

# post-merge run on main:
two_term_and / cold 1st query (median): 18.72 ms -> 1.26 s (+6640%)
bm25_search / cold open (median): 227.44 ms -> 368.72 ms (+62%)
vector-only / Peak file: 322.59 ms -> 523.80 ms (+62%)   # a memory column, mislabeled as ms
```

## Fix

- Add `gates(header)`: the blocking gate now fires only on controllable, low-variance time metrics — **warm latency, `count()`, ingest `Time`, and the drain/optimize walls**. Cold (`cold`/`+fetch`) timings are excluded.
- Widen the footprint token set (`rss`, `stored`, `peak`, `anon`) so peak/anon/file columns are never treated as latencies — fixing both the gate and the MiB/GiB unit formatting in `human()`.
- The 5%/30% advisory tiers and the `>50% AND >5 ms` blocking pair are unchanged; only the **set of metrics eligible to block** is narrowed. Cold and footprint moves remain visible as advisory findings.

## Tests

Adds `.github/scripts/test_bench_summary.py` (stdlib `unittest`, run with `python3 .github/scripts/test_bench_summary.py`). It pins the boundary using the exact signatures above:

- cold and `+fetch cold` regressions → **do not block**
- `Peak file` footprint move → **does not block**, renders as MiB
- real warm p90 regression (+100%, +10 ms) → **blocks**
- ingest `Time` wall regression → **blocks**
- sub-5 ms warm move (+60%, +2.4 ms) → **does not block** (noise guard intact)

Verified the cold/footprint cases **fail against the pre-fix script and pass after**; all 11 tests green.

### DIFF
--- a/.github/scripts/bench_summary.py
+++ b/.github/scripts/bench_summary.py
@@ -40,6 +40,22 @@ COST_TOKENS = ("$", "cost", "measured", "per-unit")
 SECONDARY_HEADERS = ("cold search", "peak rss")
 SECONDARY_THRESHOLD_PCT = 30.0
 
+# Memory-footprint columns are byte sizes (MiB/GiB), not nanosecond latencies:
+# peak/median RSS, peak anon/file mappings, and total bytes stored. They are
+# run-order biased, so they are never treated as latencies and never gate.
+# is_latency() and human() both key off this set - a token here keeps a
+# footprint column from being mistaken for a nanosecond latency (which would
+# both misformat it as "ms" and sweep it into the merge gate).
+FOOTPRINT_TOKENS = ("rss", "stored", "peak", "anon")
+
+# Cold-cache latencies - object-store first-touch ("cold ...") and the
+# fetch-augmented variants ("+fetch ...") - swing widely run-to-run with
+# network and VM page-cache state, so they are advisory-only and never block a
+# merge. Only warm, CPU-bound latencies and the ingest/optimize/drain walls
+# gate. Identical code routinely moves a cold metric by hundreds of percent;
+# gating on that fails PRs that changed no engine code at all.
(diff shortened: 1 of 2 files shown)


ekechinwokah:
why are cold latencies growing? perhaps we need different thresholds but we should not be excluding any of the metrics IMO.

PR #591, opened 2026-08-12
fts: faster COUNT — position sub-index for phrases + dedup repeated tokens

vinaykakade:
## What

Two independent speedups for unranked FTS `count` (and the count half of `TOP_k_COUNT`), the slowest of which was phrase counting:

1. **Position run-offset sub-index (new blob version V3)** — the big one. Phrase verification reached a doc's position runs by walking *every* run in its 128-doc block from the block start; for a phrase filtering a common member (docs scattered one-per-block) that walk dominated, often skipping far more position bytes than it decoded. V3 stores, per positional term, one run offset every 16 pairs between the skip table and the blocks, so the decode jumps to the nearest checkpoint and skips `< 16` runs. A cached last-pair keeps high-reuse blocks (all-common phrases) from re-skipping.

2. **Dedup repeated query tokens on the count path** — an unranked AND/OR/exclude over a term repeated in the query (e.g. `+to +be +or +not +to +be`) is idempotent, yet each occurrence built a cursor that intersected/unioned a list with itself. Drop duplicates per clause role (phrase members and the scored path are left untouched).

## Backwards-compatible — no reindex

The sub-index is **additive**: readers accept V1/V2/V3, and existing V1/V2 indices read unchanged through the block-start walk. Validated on a real V2 index — the new reader returned identical counts to `main` across all 962 bench queries × {COUNT, TOP_100_COUNT}, **0 mismatches**. A unit test also downgrades a V3 blob to V2 to exercise the fallback. Old superfiles migrate to V3 as compaction rewrites them. (Not forward-compatible: an old reader rejects a V3 blob cleanly as an unknown version — reader and writer ship together.)

## Correctness

Result-identical throughout — **0 count mismatches** on the full bench set. The brute-force phrase oracle and the unranked id/count oracle both pass; positions decoded via the sub-index match the block-walk exactly. New tests: the V2-fallback equality test, a dedup-equivalence test, and (unchanged) the oracles.

## Measurement

Local A/B, full 962-query set, min-of-10, 0 mismatches.

**Sub-index, isolated** (same binary, V2 index / block-walk vs V3 index / sub-index):

| slice | COUNT AVG | TOP_100_COUNT AVG |
| ----- | --------- | ----------------- |
| phrase | 2542 → 1433 µs (**−44%**) | 4353 → 2571 µs (**−41%**) |

`the movement` −57%, `secretary of state` −59%. Union/intersection/term unchanged (untouched paths).

**Cumulative** (both changes vs `main`):

| command | ALL AVG | P90 |
| ------- | ------- | --- |
| COUNT | 1556 → 1119 µs (**−28%**) | 3227 → 1790 µs (−45%) |
| TOP_100_COUNT | 2858 → 2195 µs (**−23%**) | 5651 → 3644 µs (−35%) |

Storage: ~0.25 B/posting of added index on positional columns only; positionless columns are byte-identical to V2.

## Verification

- `cargo test` fts + supertable-query + format suites green; brute-force oracles green.
- `cargo fmt --check` + `clippy --all-targets --features test-helpers -- -D warnings` clean.
- `make ci` passed — check + doctest + coverage + test-remote all green; coverage 93.1% lines / 89.6% functions (above the ≥90%/≥89% gate).

### DIFF
--- a/src/superfile/fts/reader/cursor.rs
+++ b/src/superfile/fts/reader/cursor.rs
@@ -54,6 +54,12 @@ pub(super) struct TermMeta {
     /// Byte length of this term's position runs (positional columns;
     /// zero otherwise).
     pub(super) positions_length: u32,
+    /// Absolute offset (within the postings region) of this term's
+    /// position run-offset sub-index — the block of
+    /// `num_blocks × ENTRIES_PER_BLOCK` `u32`s sitting right after the
+    /// skip table on a `VERSION_V3` positional term. `None` on
+    /// `V1`/`V2` (no sub-index) and on positionless terms.
+    pub(super) subindex_start: Option<usize>,
 }
 
 impl TermMeta {
@@ -65,6 +71,7 @@ impl TermMeta {
         postings: &[u8],
         metadata_offset: usize,
         positional: bool,
+        has_subindex: bool,
     ) -> Result<Self, FtsError> {
         // Positional columns carry the extended 32-byte header (the
         // term's positions offset + length after `num_blocks`); the
@@ -122,6 +129,23 @@ impl TermMeta {
                 "skip table runs past postings region".into(),
             )));
         }
+        // v3 positional terms store a run-offset sub-index right after the
+        // skip table: `num_blocks × ENTRIES_PER_BLOCK` u32s. Bound it now;
+        // the blocks follow it (their offsets are read from the skip
+        // table, which the writer already shifted past the sub-index).
+        let subindex_start = match has_subindex {
+            true => {
+                const ENTRIES_PER_BLOCK: usize = BLOCK_LEN / format::fts::POSITION_SUBINDEX_STRIDE;
+                let subindex_end = skip_end + num_blocks * ENTRIES_PER_BLOCK * U32_BYTES;
+                if subindex_end > postings.len() {
+                    return Err(FtsError::Read(ReadError::MalformedVersion(
+                        "position sub-index runs past postings region".into(),
+                    )));
+                }
+                Some(skip_end)
+            }
+            false => None,
+        };
         Ok(Self {
             df,
             postings_length,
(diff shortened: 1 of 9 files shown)


tests/supertable/query/match_search.rs
```diff
@@ -195,6 +195,30 @@ fn count_agrees_with_token_match_cardinality() {
     }
 }
 
+/// A token repeated in the query is idempotent for an unranked count —
+/// AND/OR/exclude over the same term twice is the same set — so the
+/// count path drops duplicate tokens, and the count must match the
+/// query with the repeat removed. (Phrase members are exempt: position
+/// matters there.)
+#[test]
+fn count_dedups_repeated_query_tokens() {
```
naman-modi, line 204: This only covers AND/OR positives, can we also add tests like:
-  `+rust -rust` → count stays 0
- `-rust -rust` → same as -rust

tests/superfile/format/crc_corruption.rs
naman-modi, line 1: Do we need any specific crc test for the subindex inclusion in ver3?

naman-modi: Can we change all the comments from `v2` to `v3`?

src/superfile/fts/reader/cursor.rs
```diff
@@ -122,16 +129,58 @@ impl TermMeta {
...
+    /// sub-index (`V1`/`V2`) — the caller falls back to the block-start
+    /// walk. The skip is always `< POSITION_SUBINDEX_STRIDE`.
+    #[inline]
+    pub(super) fn positions_subindex_offset(
+        &self,
+        postings: &[u8],
+        block: usize,
+        pair_in_block: usize,
+    ) -> Option<(u32, usize)> {
+        const ENTRIES_PER_BLOCK: usize = BLOCK_LEN / format::fts::POSITION_SUBINDEX_STRIDE;
```
naman-modi: This is used at multiple places, can we create a const next to `POSITION_SUBINDEX_STRIDE` and use it?

src/superfile/fts/reader/cursor.rs
```diff
@@ -122,16 +129,58 @@ impl TermMeta {
...
+        let start = self.subindex_start?;
+        let slot = pair_in_block / format::fts::POSITION_SUBINDEX_STRIDE;
```
naman-modi:
```suggestion
        let slot = pair_in_block / POSITION_SUBINDEX_STRIDE;
```

Can we import this at the top of the file, rather than inline?

naman-modi (approved):
Few comments, else the overall change lgtm.

tests/supertable/query/match_search.rs
vinaykakade: Done

tests/superfile/format/crc_corruption.rs
vinaykakade: Done

src/superfile/fts/reader/cursor.rs
vinaykakade: Done

src/superfile/fts/reader/cursor.rs
vinaykakade: Done - added const POSITION_SUBINDEX_ENTRIES_PER_BLOCK

tests/superfile/format/crc_corruption.rs
vinaykakade: Done

PR #592, opened 2026-08-12
fix deletes: acquire lease on WAL during delete

Pranav2612000:
### What does this PR do?
- Acquires the lease on the WAL file created during deletes, to allow it to complete without any CAS issues

- During an in-progress delete, opening another table starts its sweep flow, which looks at the WAL files and starts "work" on them, even though a delete is already in progress
- This is because, the "delete" process does not acquire a lease. ( `lease: None` before this change )
- This may cause the delete to fail, due to CAS errors, and return an error to the user. But the data may be deleted by the background delete job

### How do we do this change?
- Create the WAL file with a lease
- Periodically extend this lease, while the delete is in progress
- In case of errors, relesae the lease before erroring out

### How has this been tested?
- Unit tests added

### DIFF
--- a/src/supertable/writer.rs
+++ b/src/supertable/writer.rs
@@ -73,7 +73,7 @@ use arrow_array::{
 };
 use blake3::Hasher as Blake3Hasher;
 use bytes::Bytes;
-use chrono::Utc;
+use chrono::{DateTime, Duration as ChronoDuration, Utc};
 use datafusion::prelude::Expr;
 use futures::{
     future::try_join_all,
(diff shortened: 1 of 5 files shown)


src/supertable/writer.rs
```diff
@@ -1441,24 +1444,38 @@ impl SupertableWriter {
...
+    ///
+    /// The lease stays advisory: the etag CAS chain in the tombstone phase
+    /// is what keeps concurrent drivers correct. This only keeps a peer
+    /// from duplicating the work and knocking us off our etag.
+    ///
+    /// One `now` stamps `created_at` and both lease timestamps so the
+    /// doc's creation time and its lease window come from a single clock
+    /// reading.
+    fn delete_wal_doc(&self, entry: &PendingDeleteEntry, now: DateTime<Utc>) -> WalStateDoc {
+        let lease_span = ChronoDuration::from_std(DEFAULT_LEASE_DURATION).unwrap_or_default();
```
vinaykakade: Instead of defaulting to 0, let us fail hard if this is not found? (default will be 0 - which would basically mean no lease).

src/supertable/writer.rs
vinaykakade, line 1383: You'll need to add the update path as well - and please add a test for the update path too

vinaykakade (approved):
LGTM after making a similar change for the update path + a minor comment regarding unwrap_or_default for the default lease duration

Pranav2612000:
Fixed the default to 0 issue. Separate PR for update comint up.

PR #597, opened 2026-08-13
fts: dense bitset blocks + count kernels — union/intersection/phrase COUNT to parity

vinaykakade:
## Summary

Closes the unranked **COUNT** performance gap on a dense corpus. Four changes,
all on the same posting-list machinery — a new dense block encoding plus three
count kernels that exploit it:

1. **Dense posting blocks stored as bitsets** — a block may be a presence bitset
   instead of PFOR deltas when it is dense; membership is answered directly
   against it, often with no decode.
2. **Two-phase phrase atoms in the AND walk** — a phrase AND'd with a rarer term
   verifies positions only where the rare term lands.
3. **Bit-test membership in the df-anchored union count** — a dominant common
   term is membership-tested, not decoded.
4. **Bitset-AND intersection count** — dense intersections AND presence bitsets
   word-at-a-time instead of walking the rarest term doc by doc.

Net: **all-COUNT −44%** on a dense 5M-doc corpus vs the current release, every
query shape faster, **0 count mismatches**, and **no reindex**.

## 1. Dense blocks as bitsets

A posting block is now encoded either as today's PACKED (PFOR deltas, unchanged)
or, when dense, as a BITSET of doc presence. The header's previously reserved
byte becomes a self-describing `encoding` byte; the encoder emits BITSET only
when it does not grow the block. Three count paths exploit it:

- **Union COUNT** word-ORs BITSET blocks into the result — a common term is
  merged at memory bandwidth, never PFOR-decoded.
- **Intersection / phrase membership** answers "does this doc contain the term?"
  with a single word-load + bit-test, no decode.

### On-disk format & compatibility

The blob version is bumped and a block encoding variant added. Existing indices
**need no reindex**: older blobs contain only PACKED blocks and read unchanged;
the new reader branches per-block on the encoding byte; compaction migrates
blocks organically. An older reader cleanly refuses a newer-version blob.
Storage is neutral-or-smaller (BITSET is chosen only when no larger than the
PACKED deltas); tfs are stored identically, so scoring is byte-for-byte
unchanged.

## 2. Two-phase phrase atoms

A phrase AND'd with a rarer term (e.g. an exact phrase of two common words next
to a rare term) was pathologically slow: the AND walk advanced the phrase atom
with `skip_to`, which position-verifies to produce its next match, so it decoded
positions across the phrase's whole match set instead of only where the rare
term also lands. The phrase atom is split into an **approximation** (`approx_seek`
— next doc holding all members, by bit-test, no positions) and **verification**
(decode positions, check adjacency). The AND walk aligns every atom by its
approximation first, so a rare term prunes candidates, then verifies phrase
adjacency only on survivors. Pure-phrase and OR behavior unchanged; ranked search
untouched.

## 3. Bit-test membership in the df-anchored union count

The df-anchored disjunction count (one dominant common term plus rarer ones)
membership-tested the anchor with `skip_to`, which decodes a bitset block (~128
doc ids expanded) — and the frontier merge drags the anchor across its whole
list, so a `the ∪ x ∪ y` count decoded all of `the`. Use `contains` (bit-test,
no decode) instead — the same primitive the intersection count uses.

## 4. Bitset-AND intersection count

The intersection count drove off the rarest term and probed the rest by bit-test.
Optimal when the rarest term is sparse — but when every term is common the
rarest is still millions of docs, so the driver iterates a long list. Added a
word-parallel path for that case: build each term's presence into a doc-space
bitset (dense blocks word-copied, never expanded), AND the bitsets, popcount.
Cost is `n_terms × max_doc/64`, independent of the terms' lengths. Gated on the
rarest term being dense; sparser intersections keep the membership probe. The
byte-level fill is shared with the union count (`or_cursor_into_bitset`).

## Performance (COUNT, dense 5M-doc single superfile, min-of-runs, 0 mismatches)

Cumulative vs the current release, on the same corpus rebuilt in the new format:

| slice        | before → after            |
| ------------ | ------------------------- |
| union        | 1319 → 426 µs (**−67.7%**), P90 −67% |
| intersection | 504 → 335 µs (**−33.4%**)  |
| phrase       | 1494 → 1283 µs (**−14.1%**) |
| phrase AND rare term (worst case) | 142 → 22 ms (**−84%**) |
| all-common intersection (worst case) | 18.5 → 2.3 ms (**8×**) |
| **all COUNT**| 1196 → 670 µs (**−44.0%**), P90 −33% |

Cross-engine, this brings all-COUNT from ~1.56× the reference engine to roughly
parity. The scored top-k paths are deliberately untouched (scoring needs tfs,
unchanged) and measured flat.

## Correctness

- Brute-force FTS oracles (union / intersection / count / phrase, ranked and
  unranked) pass; added a `phrase AND rare-term` case and a dense-corpus
  `AND count` case that exercises both the bitset-AND and membership kernels
  against an independent flat-merge reference.
- Block codec round-trips both encodings over dense and sparse corpora.
- A new-reader-on-old-index A/B returns identical counts (0 mismatches).
- `make ci` (fmt, clippy, api-parity, doctest, coverage, remote) green.

### DIFF
--- a/src/superfile/fts/posting.rs
+++ b/src/superfile/fts/posting.rs
@@ -60,9 +60,30 @@ use bitpacking::{BitPacker, BitPacker4x};
 /// match `BitPacker4x::BLOCK_LEN`.
 pub const BLOCK_LEN: usize = BitPacker4x::BLOCK_LEN;
 
-/// Header size in bytes (doc_count + delta_bits + tf_bits + reserved +
+/// Header size in bytes (doc_count + delta_bits + tf_bits + encoding +
 /// base_doc_id).
 pub const HEADER_SIZE: usize = 8;
+
+/// Header byte offset of the block `encoding` field (formerly reserved).
+pub const ENCODING_OFF: usize = 3;
+
+/// Block `encoding` (header byte 3): doc ids stored as PFOR-delta packing
+/// (today's layout). Every `V1`–`V3` block is this.
+pub const ENCODING_PACKED: u8 = 0;
+/// Block `encoding`: doc ids stored as a **presence bitset** over
+/// `[base_doc_id, last_doc_id]`, `base_doc_id` aligned down to a 64-bit
+/// word so the union count can OR it in word-aligned. Chosen only when it
+/// does not grow the block (dense blocks). Tfs follow, packed identically
+/// to PACKED. Only `VERSION_V4` blobs contain these.
+pub const ENCODING_BITSET: u8 = 1;
+
+/// Align a doc id down to the 64-bit word that contains it — the origin of
+/// a [`ENCODING_BITSET`] block's presence bitset, so its words line up
+/// with the union bitset and the OR needs no per-word bit shift.
+#[inline]
+pub fn bitset_block_base(doc_id: u32) -> u32 {
+    doc_id & !63
+}
 
 /// One block of postings — sorted-ascending `doc_ids` plus per-doc
 /// `tfs`. Both vectors must have the same length, ≤ [`BLOCK_LEN`].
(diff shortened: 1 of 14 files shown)


src/superfile/fts/posting.rs
```diff
@@ -181,59 +236,105 @@ pub fn encode_block(b: &Block) -> EncodedBlock {
...
+        let words = &bytes[HEADER_SIZE..bytes.len() - tfs_size];
+        let mut j = 0usize;
+        for (wi, chunk) in words.chunks_exact(8).enumerate() {
+            let mut word = u64::from_le_bytes(chunk.try_into().expect("8 bytes"));
+            while word != 0 {
+                dest_doc_ids[j] = base_doc_id + (wi as u32 * 64 + word.trailing_zeros());
+                j += 1;
+                word &= word - 1;
+            }
+        }
```
naman-modi, line 333: Nothing checks over here if the bitmap size it more than 128, so a corruption can happen. A simple check like `j <= dest_doc_ids.len()` should help.

tests/superfile/format/crc_corruption.rs
naman-modi, line 1: Do we want to add a test around  flip inside a BITSET block, caught via CRC?

naman-modi (approved):
Few comments Vinay, else lgtm.

naman-modi:
Missed mentioning in the review, can we have more tests around the negated terms & phrases?

tests/superfile/format/crc_corruption.rs
vinaykakade: Done

src/superfile/fts/posting.rs
vinaykakade: Since this is provably unreachable, added a comment that why this is safe. The checks in hot path do increase latency, so trying to optimize.

vinaykakade:
> Do we want to add a test around flip inside a BITSET block, caught via CRC?

Done

PR #599, opened 2026-08-13
feat writer: shard commit fanout by buffered bytes, and pool width

### LINKED ISSUE #594: writer: commit shard fanout tracks core count, fragmenting tables on big machines
## Problem

A commit splits its buffer into one shard per writer-pool thread (`n_shards = n_threads.min(total_rows)`), and each shard becomes its own superfile. So the larger the machine (more cores), the more superfiles created per commit. This works well when the ingested data size is in 10s of GBs for a single commit, but when the size is smaller, it causes a lot of unnecessary fragmentation. Even though this gets handled once we call `optimize`, it is still a lot of unnecessary overhead.

For context, ingesting a 100M-row load (c7a.metal-48xl vs c6a.4xlarge) produced 184 live superfiles on 192 cores vs 54 on 16 cores for the same data, with ~9,600 tiny intermediate fragments accumulating in the disk cache on the big box vs ~840 on the small one. Every extra file is paid for downstream: a footer parse and an object-store GET on cold open, a disk-cache entry, a compaction input, and a PUT at write.

As a matter of fact, the initial thesis behind the n_cores fanout (parallelize across as many cores as possible to increase ingestion throughput) does not hold at real buffer sizes: 192 shards of ~11 MB build slower than ~33 shards of ~64 MiB (271K vs 335.6K docs/s), because per-shard setup dominates the work.

## Possible Approach

Fan out by buffered bytes instead: `ceil(buffered_bytes / target)` shards, capped by the pool width and the row count, so data volume sets the file count and the pool only caps build parallelism. Can possible add a variable related to this to config.

## Acceptance

- The same commit (of a smaller data size) produces the same superfile count on a 4-core and a 192-core host.
- The 10M FTS bench cell runs on a 192-core host under the default fd limit.
- No ingest-throughput or query regression on the bench suite; the fts build cell at 10M should improve.

naman-modi:
### What does this PR do?
This PR changes how the number of superfiles a commit produces is decided: it is now based on the amount of data being committed instead of only the machine's number of cores. Closes #594.

A commit splits its buffer into one shard per writer-pool thread, and each shard becomes its own superfile. So the larger the machine, the more superfiles a commit creates, even when the data is small. At real buffer sizes this fragments the table without buying any ingestion throughput. #594 has detailed observation.

### How do we do this change?
- The number of shards for a commit is now `ceil(buffered_bytes / target)`, still capped by the pool width. So the shard count grows with the data, each shard stays around the target size, and the cores only decide how many build in parallel.
- The target defaults to 64 MiB and is configurable as `shard_target_size_mb` in the config, next to the existing commit threshold. Setting it to `0` restores the old per-thread fanout.
- Vector commits already shard by cell and are untouched.

### Performance
- Measured on c7a.metal-48xl (192 cores), `supertable fts build` at 10M docs against main: ingest throughput went from 271K to 335.6K docs/s (+24%), superfiles from 3072 to 528, stored index from 13.94 to 8.88 GiB, object-store PUTs from 3122 to 578, and peak ingest RSS from 10.5 to 6.8 GiB.
- Small machines are unaffected: a 4-core host produces the same shards per flush before and after. The table content is identical either way, only the file boundaries change.

### Notes
- Existing configs pick up the 64 MiB default on upgrade; that is the fix taking effect, and `0` is the opt-out.

### DIFF
--- a/src/config/config.yaml
+++ b/src/config/config.yaml
@@ -48,6 +48,11 @@ supertable:
   # superfiles.
   commit_threshold_size_mb: 1024
 
+  # Target buffered bytes per commit shard, in mebibytes. Each shard becomes one
+  # superfile, so this decides how many files a commit produces from data volume
+  # rather than core count. Set 0 to fall back to one shard per writer-pool thread.
+  shard_target_size_mb: 64
+
   # Verify the trailing whole-blob CRC and per-subsection CRCs
   # on every SuperfileReader::open. Defaults to true. Set to
   # false only when the underlying storage already validates
--- a/src/config/config.yaml
+++ b/src/config/config.yaml
@@ -17,7 +17,7 @@
 # readable from files instead of reconstructed from process env.
 
 # Supertable runtime knobs. Reader fan-out (skip + per-superfile
-# search + top-k merge) and writer commit-time rayon-shard run on
+# search + top-k merge) and the writer's parallel commit builds run on
 # separate pools so a long-running commit can't spike reader p99 or
 # vice versa.
 #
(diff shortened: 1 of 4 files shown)


src/config/config.yaml
```diff
@@ -48,6 +48,11 @@ supertable:
   # superfiles.
   commit_threshold_size_mb: 1024
 
+  # Target buffered bytes per commit shard, in mebibytes. Each shard becomes one
+  # superfile, so this decides how many files a commit produces from data volume
+  # rather than core count. Set 0 to fall back to one shard per writer-pool thread.
+  shard_target_size_mb: 64
```
vinaykakade: Let us avoid the term "shard" - to avoid confusion with the other tools? Say something like superfile fragment / superfile target size at commit?

vinaykakade (approved):
LGTM after the terminology change

src/config/config.yaml
naman-modi: Done.

PR #601, opened 2026-08-13
vector: probe enough cells to fill top-k across cluster boundaries

muralikpbhat:
## Problem

An unfiltered vector `top-k` search can return fewer than `k` rows when the query vector sits between two clusters.

After the drain distributes rows across hidden cells, the coarse cell selector chooses how many cells to probe by a **score-slack window**: start at the nearest cell and keep adding cells only while the next cell's score stays within a slack margin of the best. A query blended between two clusters (primary weight 1.0, secondary 0.5) has its secondary cell fall outside that margin, so only the nearest cell is probed. If that cell holds fewer than `k` rows, the query returns short — e.g. **12 of a requested 24** — silently, under the default config with no caller `nprobe`.

The slack window decides coverage by score proximity, never by whether `k` is satisfied.

## Fix

`cover_k_cell_cutoff`, applied right after the slack cutoff: keep extending the probed cells down the ranking — summing each cell's indexed row count — until the probed cells can hold `k`, or the ranking is exhausted.

- A **k-sufficiency floor** on top of the existing slack window: slack still handles the common single-cluster early stop; this guarantees a top-k request can't come back short.
- **No-op whenever the nearest cell(s) already hold `k`**, so single-cluster queries and every query at scale (top cell ≫ k) are unaffected — no extra cells probed, scored, or fetched.
- The widened cutoff flows into the must-include set, so the extra cells are scored and their rows returned.

## Testing

New unit test `blended_vector_search_fills_topk_across_hidden_cells`: 60 one-hot rows across 5 axis clusters, drained into separate hidden cells; a `k=24` query blended between two axes must return 24. Returns 12 without the fix, 24 with it. Existing vector/query integration suite stays green.

### DIFF
--- a/src/supertable/query/vector.rs
+++ b/src/supertable/query/vector.rs
@@ -742,6 +742,31 @@ fn union_cell_selection(grid: &[u32], fine: &[u32]) -> Vec<u32> {
     selected
 }
 
+/// Widen a grid cell cutoff so the probed cells' indexed row counts cover at
+/// least `k`. `grid_cell_cutoff`'s slack window stops at the nearest near-tie
+/// cluster — a single cell for a query blended between two clusters — so a
+/// top-k request that spans cells could otherwise probe one undersized cell and
+/// return fewer than `k` rows. Walks the cell ranking past `cutoff`, adding each
+/// cell's indexed row count (from `postings_by_cell`) until coverage reaches `k`
+/// or the ranking is exhausted. A no-op once the already-probed cells hold `k`
+/// (the common single-cluster case, and every cell large relative to `k` at
+/// scale), so a sufficient query is never widened.
+fn cover_k_cell_cutoff(
+    cutoff: usize,
+    ranked: &[(u32, f32)],
+    postings_by_cell: &HashMap<u32, u64>,
+    k: usize,
+) -> usize {
+    let cell_rows = |cell: u32| postings_by_cell.get(&cell).copied().unwrap_or(0);
+    let mut covered: u64 = ranked[..cutoff].iter().map(|(c, _)| cell_rows(*c)).sum();
+    let mut cut = cutoff;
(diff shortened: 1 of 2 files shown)


src/supertable/query/vector.rs
```diff
@@ -1786,6 +1811,19 @@ impl SupertableReader {
...
+            // blended between two clusters selects one undersized cell and
+            // returns fewer than `k`; widen along the grid ranking until the
+            // probed cells cover `k`. No-op once the nearest cell(s) already
+            // hold `k`. An explicit caller `nprobe` is a deliberate width
+            // constraint and is left untouched even when it under-fills `k`.
+            let cutoff = if options.nprobe.is_none() {
+                cover_k_cell_cutoff(cutoff, &ranked_for_beam, &postings_by_cell, k)
+            } else {
+                cutoff
+            };
```
Pranav2612000, line 1826: Nit: would it be better to merge it with the looping in grid_cell_cutoff?

Pranav2612000 (approved):

src/supertable/query/vector.rs
muralikpbhat: Kept them separate deliberately:

1. **Different bounds** — the slack loop is capped at `nprobe_max` (a routing width limit), whereas the k-floor intentionally widens *past* that cap (up to all cells with postings) to guarantee `k` is fillable. Folding them in would blur that bound.
2. **Different applicability** — the k-floor runs only on the default path (`nprobe.is_none()`); the slack loop runs for every path. Merging means threading `k` / `postings_by_cell` / the default-path flag into the closure and guarding inside it.
3. **Separation of concerns** — "near-tie slack coverage" vs "correctness k-floor" read more clearly as two named steps.

No perf difference either way (both are O(cells) over a tiny cutoff), so I've left it as two steps for clarity. Happy to revisit if you feel strongly.

PR #602, opened 2026-08-14
fts: borrow block bytes in the membership probe instead of cloning

vinaykakade:
## What

The unranked intersection-count path probes the common term with `TermCursor::contains` once per driver doc. Each probe called `Bytes::slice` on the block's byte range, cloning an owned `Bytes` — an atomic refcount round-trip — even though on the PACKED path the sliced bytes are only read for the encoding byte before the probe falls to the cached decode. `decode_current_block` did the same clone once per block.

This borrows the block bytes in place (`&self.bytes[range]`) instead: the buffer is held for the cursor's life, so a subslice needs no owned clone. Behavior is identical; the only change is dropping the per-probe/per-block allocation.

## Why

A sampling profile of a large-cardinality intersection-count run attributed **~11% of the walk's time to the `Bytes` refcount churn** (`shared_drop`) from these per-probe clones — pure overhead.

## Result

Intersection COUNT is **~15% faster** overall and on the large-cardinality bucket, with the tail improving most (p99 −18%, max −20%). Counts are unchanged — verified against the flat-merge oracle across the standard intersection query set, 0 mismatches. Covered by the existing `count::tests` intersection/merge oracles; no behavior change to test beyond those.

No format change, no public-API change.

### DIFF
--- a/src/superfile/fts/reader/cursor.rs
+++ b/src/superfile/fts/reader/cursor.rs
@@ -494,14 +494,14 @@ impl TermCursor {
 
     pub(super) fn decode_current_block(&mut self) {
         let block = self.blocks[self.current_block];
-        let bytes = self
-            .bytes
-            .slice(block.block_byte_offset..block.block_byte_end);
+        // Borrow in place rather than clone an owned `Bytes` (disjoint from the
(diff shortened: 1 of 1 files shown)


naman-modi (approved):
LGTM!

Issue #590, opened 2026-08-12
Cohere-1M vector query: reach p95 ~2 ms at recall@10 ≥ 0.95


muralikpbhat:
## Target

Drive warm hidden-vector-index query latency on **Cohere-1M** (768-d, cosine) to **~2 ms p99 at recall@10 ≥ 0.95** (k=10). Today the warm p99 at that recall is **~4.9 ms** — decomposed below (the phase split is p50, to show where the time goes; the target metric is total p99).

## Baseline (warm, resident cache, Cohere-1M, k=10)

Measured with `vector.routing = global_fine_centroid`, sweeping `global_fine_fanout` to the 0.95 point. Per-phase split over 1000 warm queries (fanout 114, recall@10 = 0.9525):

| phase | p50 (ms) | p99 (ms) | share (p50) |
|---|---|---|---|
| route (global fine-centroid scan) | 0.44 | 0.92 | 12% |
| **scan (1-bit cluster scan)** | **2.70** | **3.56** | **72%** |
| rerank (Sq16 exact) | 0.59 | 0.77 | 16% |
| overhead (merge + binding) | ~0.05 | — | ~1% |
| **total** | **3.75** | **4.71** | |

Per-phase percentiles are measured independently, so per-phase p99 does not sum to the total; the **scan tail (p99 3.56 ms) drives the total tail**. A separate uninstrumented 1500-query run gives total p50 3.80 / p95 4.47 / **p99 4.87 ms**. The metric this issue targets is **p99 ≤ 2 ms**.

## Read

- **The 1-bit cluster scan dominates (72%) and is linear in fanout** (~0.018 ms/cluster, ~0.58 ms fixed floor): it is row-bound — ~100K rows streamed at 1-bit to surface the top-10. Fanout is pinned by the recall target and the per-row kernel is already at its floor, so there is no input left to trim on this path.
- **Routing is cheap (0.38 ms) and fanout-independent** — the centroid scan is not the bottleneck at this scale.
- **Rerank is small** and grows slowly with fanout.

## Direction

At 0.95 the scan cost is intrinsic to *scanning whole clusters*. The lever is to **touch fewer rows at equal recall** — navigate rather than scan. Structural options (a resident navigation layer over the warm rows) are tracked separately; this issue owns the **target + this baseline**, and every candidate reports back against the table above.

## Reproduce

[...]

PR #603, opened 2026-08-14
vector: self-driving hnsw graph search mode with per-table (m0, ef) calibration

muralikpbhat:
## What

Adds a config-driven **`search_mode = hnsw`** for the hidden vector index: a resident per-row Sq16 HNSW graph, built at drain and walked in place of the per-cell scan, with a **drain-time calibrator that picks per-table graph parameters `(m0, ef)` to the table's recall bar** — and serves `ivf` automatically wherever the graph can't compete. The graph is self-driving: no hand-tuning and no per-table tuning state to manage.

Default `search_mode` is `hnsw`; `ivf` remains the fallback and the serving path for every case the graph declines.

## Why

**ivf's latency floor is the *number* of distance calculations.** ivf answers a query by scoring every row in the probed clusters — on a high-recall query that's ~100K Sq16 distances. Those kernels are already optimized to the metal (SIMD AVX-512/AVX2 paths, the Sq16 codec), so the per-comparison cost has no slack left; the floor is set by *how many* comparisons, not how fast each one is. Holding the cluster in memory doesn't move it either — a warm ivf scan is already served from cache and still computes every one of those ~100K distances.

**hnsw wins by *avoiding* most of those calculations.** The graph walk navigates straight to the neighborhood, touching only ~1–2K nodes — so it simply never computes the other ~98% of the distances ivf does (~50–80× fewer). That avoided computation *is* the latency win, and it's why the win holds even fully warm: it's an algorithmic cut in comparison count, the only lever left once the kernels are floored.

**The trade is memory.** The graph is resident — the Sq16 code plane plus ~m0 adjacency links per node — and that memory is what buys the avoided comparisons. Caching the ivf cluster in memory instead would spend memory without avoiding a single distance. The extra `m0` memory is bounded: layer-0 only, ∝ docs × m0, capped by the scale ceiling.

**Why it's calibrated.** The avoided-computation win only materializes if the base-layer degree `m0` is provisioned for the data's dimension — too sparse and the walk mis-routes (recall collapses); too dense and it wastes memory and touches more nodes than it needs. So the engine calibrates `m0` (and the query beam `ef`) per table, to a recall target, at drain.

## Results — calibrated `hnsw` vs `ivf`, real cosine datasets

k=10, serial, recall@10 vs exact-NN ground truth; p95 is served query latency. `(m0, ef)` is what the calibrator stamped for that table. Within every row, the `hnsw` and `ivf` columns are measured on the **same harness and the same drained table** (only the serving `search_mode` differs) — that same-target, same-table pair is the comparison that matters; latency is not meant to be compared *across* rows of different dimension/scale. Measured on x86 (Intel Xeon, AVX-512) — absolute latency is silicon-dependent (the score-heavy walk leans on AVX-512), but the same-box `hnsw`-vs-`ivf` ratio is the robust quantity.

| dataset | dim | n | stamped `(m0, ef)` | hnsw recall / p95 | ivf recall / p95 | speedup |
|---|---:|---:|---|---|---|---:|
| glove-25 | 25 | 1.2M | (64, 128) | 0.9947 / 0.6 ms | 0.9975 / 14.2 ms | 24× |
| glove-50 | 50 | 1.2M | (128, 256) | 0.9893 / 1.2 ms | 0.9960 / 17.8 ms | 15× |
| glove-100 | 100 | 1.2M | (128, 1024) | 0.9903 / 5.7 ms | 0.9975 / 27.7 ms | 4.9× |
| glove-200 | 200 | 1.2M | (256, 2048) | 0.9873 / 18.7 ms | 0.9982 / 27.7 ms | 1.5× |
| nytimes-256 | 256 | 290K | (256, 512) | 0.9714 / 3.4 ms | 0.9813 / 4.7 ms | 1.4× |
| deep-image-96 | 96 | 10M | (64, 256) | 0.9906 / 2.4 ms | 0.9971 / 29.6 ms | 12× |
| Cohere-1M | 768 | 1M | (64, 256) | 0.9922 / 2.4 ms | 0.996 / 5.3 ms | 2.2× |

> *Recall carries real run-to-run noise on **both** `hnsw` and `ivf` — the same settings don't reproduce an identical recall each run (non-seeded graph build, calibration query sampling, drain ordering). The swing is material at sparse `m0`: identical m0=32/ef=128 served 0.948–0.9746 across runs (~±0.02), with ~±0.3 ms of latency spread. Treat these as representative points, not exact constants; the stamped `(m0, ef)` can likewise shift a rung between runs. A fixed build seed (follow-up 8) would tighten this.*

The recall gap to `ivf` is ≤ 0.014 on every set (the calibrator finds the base-layer degree the corpus needs — `m0` ranges 64→256 by difficulty), at a 1.4–24× latency win. glove-200 is the hardest cosine set (needs `ef=2048`), where the win narrows but recall stays near `ivf`.

This table is a **living set** — Cohere-1M lands with this PR; **Cohere-10M is appended when it finishes** (a longer run). The mechanism is dataset-independent; the table just accumulates evidence. Only cosine sets appear here: L2/negdot metrics have no graph and fall back to `ivf` (follow-up 1), so there's nothing to compare; and beyond the graph scale ceiling (e.g. 100M) the table serves `ivf` with centroid routing (follow-up 7).

## How it works

- **Calibrate at drain.** Build the graph once at the max `m0` candidate on the full corpus, then *prune* the base layer down to each smaller `m0` (cheap — truncate layer-0 adjacency) and sweep `ef` by re-search (free). Stamp the **latency-first** pair (smallest `ef`, then smallest `m0`) that clears `vector.target_recall`.
- **Register or step aside.** If nothing clears `target_recall − hnsw_recall_slack`, the graph is not registered and the table serves `ivf` — reusing the existing `None → fallback`. `target_recall` is the same per-table knob the `ivf` stamping law uses (one operating point per table, engine-agnostic).
- **Self-describing calibration.** `m0` lives in the graph structure; `ef` rides in the bundle header (the former reserved slot, back-compatible with older bundles) — so the per-table `(m0, ef)` needs no manifest state. The graph blob itself is referenced from the existing per-table slow-vector-state, alongside the routing and centroid refs already there (one new `slow_vector_state_graphs` field; absent ⇒ decodes to `None` ⇒ serves `ivf`). Format documented on #590.
- **Probe-gate large builds.** On a corpus larger than `hnsw_probe_max_docs` (100K), calibrate on a bounded subsample first. Subsample recall is *optimistic* (the `m0` requirement grows with N), so a probe that can't register is a hard "graph-hostile distribution" signal → skip the expensive full build, serve `ivf`. This gates on *distribution, not size*: a large but graph-friendly corpus (deep-image at 10M) passes the probe and keeps its graph. Re-derived every drain, so no table is permanently locked out.
- **Incremental drains** inherit the prior graph's stamped `(m0, ef)` and extend it in place — no rebuild.

## The `target_recall` lever — Cohere-1M (768d)

`target_recall` is the one knob that moves the whole recall/latency operating point — no per-engine tuning. The calibrator is latency-first (smallest `ef`, then smallest `m0` that clears the bar), so **lowering the target automatically stamps a cheaper graph and serves faster**, no code change. Same Cohere-1M (768d) table — five calibrated targets plus one hand-configured point below the grid floor — both engines at each target, latency shown as **p95 / p99** ms:

| `target_recall` | (m0, ef) | hnsw recall | hnsw p95/p99 (ms) | ivf recall | ivf p95/p99 (ms) | speedup (p99) |
|---|---|--:|--:|--:|--:|--:|
| 0.99 | (64, 256) | 0.9922 | 2.4 / 2.6 | 0.996 | 5.3 / 5.9 | 2.3× |
| 0.97 | (128, 128) | 0.9888 | 1.9 / 2.2 | 0.9914 | 4.8 / 5.4 | 2.5× |
| 0.95 | (64, 128) | 0.9813 | 1.7 / 1.9 | 0.9809 | 4.3 / 4.8 | 2.5× |
| 0.93 | (32, 128) | 0.9502 | 1.3 / 1.4 | 0.9725 | 4.2 / 4.7 | 3.4× |
| 0.91 | (32, 128) | 0.948 | 1.2 / 1.4 | 0.9728 | 4.0 / 4.5 | 3.2× |
| — (pinned) † | (16, 128) † | 0.9233 | 0.9 / 1.0 | — | — | — |

> † **Hand-configured, not calibrator-stamped:** `m0`/`ef` pinned via config; m0=16 is *below* the candidate-grid floor (32), so the calibrator can't reach it today. It's a genuinely leaner rung — 0.9233 at 0.9/1.0 ms — showing the dial continues below the current floor. (Going *too* sparse eventually degrades connectivity and gets slower again; that lower bound is characterized separately.) No `ivf` counterpart (an hnsw-only pinned point).

Reading the curve:

- **hnsw tracks the target cleanly, down to the grid floor.** Served recall and latency step down together from 0.9922 (2.4/2.6 ms) to 0.9502 (1.3/1.4 ms) as the target drops to 0.93, because the graph's cost is per-node along the `ef` beam. **Don't be surprised that 0.93 and 0.91 read the same** — both stamp the smallest rung the grid offers, (32, 128), so once the target falls below what that rung already achieves, lowering it further changes nothing (the 0.9502→0.948 difference is just build noise on one stamp). This is a **config limit, not a cost limit**: (32, 128) is simply the smallest candidate today, and the hand-configured m0=16 row shows a leaner rung still below it (0.9/1.0 ms). If a deployment needs that lower band, extending the grid downward opens it up (opt-in follow-up 9) — down to a connectivity-bounded floor, not to zero.
- **ivf floors for a different reason — architecturally.** It tracks down to target ~0.93 and then *stops* at ~0.972 / ~4.5 ms (p99). ivf reads whole clusters, so once width bottoms out it sheds neither recall nor latency; its cost (cold especially) tracks cluster/block reads, not the recall bar. The overshoot above target is free bonus recall from cells it reads anyway — benign — but ivf simply cannot serve a low-recall/low-latency point.
- **Both floor; hnsw's floor is lower, faster, and liftable.** At the grid floor hnsw serves 0.948 / 1.4 ms p99 vs ivf's 0.972 / ~4.5 ms, so the graph reaches operating points ivf can't (2.3× → 3.2× as the target drops). And because hnsw's floor is the candidate grid, extending it downward (a smaller `m0` rung) pushes the dial further still — the hand-configured `m0=16` row (last in the table) is exactly that leaner point below the current floor.

One number dials a deployment from max-recall to latency-first, and on the graph the latency genuinely follows — as far as the candidate grid reaches. (The 0.97→0.95 non-monotonicity in `m0` — 128 then 64 — is calibration-sample noise, not a real inversion; the served curve stays monotonic.)

## Config

All under `vector:` (YAML only; env never overrides):

- `search_mode: hnsw | ivf` (default `hnsw`)
- `target_recall` — recall/latency operating point, shared with the `ivf` stamping law. **Lowering it lowers latency for free**: the calibrator is latency-first, so a lower bar stamps a smaller `ef`/`m0` automatically.
- `hnsw_recall_slack` — give-up band; register floor is `target_recall − slack` (absolute)
- `hnsw_m0` — `0` = calibrator picks (sweeps candidates); non-zero pins it
- `hnsw_ef_ceil` — caps the `ef` sweep (default 2048)
- `hnsw_probe_max_docs` — probe subsample cap (default 100K)
- `hnsw_max_docs` — scale ceiling for building the resident graph at all

## Why this is ready to merge

- **Strictly additive and self-gating.** `hnsw` only ever *replaces* the scan when the calibrator proves it clears the recall bar on that table's real data; every other case — declined graph, filtered query, non-cosine metric, cold/restart, over-scale — falls through to the unchanged `ivf` path. There is no table the change can make *worse*: the worst case is "serve `ivf`, exactly as today."
- **No migration.** The `(m0, ef)` params are self-describing (`m0` in the graph, `ef` in the bundle header, back-compatible with existing bundles); the only manifest addition is the optional `slow_vector_state_graphs` blob ref, which is strictly additive — an absent field decodes to `None` and serves `ivf`, so older manifests need no migration. The probe gate is re-derived every drain, so there's no stored flag to migrate and no one-time bootstrap to coordinate.
- **Validated at scale on real data**, not micro-benchmarks — up to 10M, across dimensions 25→768, with the calibrator picking correctly per table (and correctly *declining* graph-hostile distributions → `ivf`).
- **The deferrals below are optimizations, not correctness gaps.** Each one's absence just means "serve `ivf`" or "leave latency on the table," never a wrong answer. They're enumerated with reasons so the follow-up work is legible.

## Scope & follow-ups (not in this PR)

1. **Cosine only.** L2/negdot use a per-cluster adaptive Sq16 codec (no single grid), so the graph isn't built for them and those tables serve `ivf`. A real graph there needs per-cluster decode plumbing.
2. **Register on recall, not yet on graph-vs-`ivf` latency.** nytimes registers on recall but is a small set where the win is marginal; the register decision should also require the graph to be *faster* than `ivf` (needs `ivf`'s predicted latency at calibration time).
3. **`ef_construction` as a 3rd lever.** Today it's fixed; it's a real build-beam knob (should be ≥ `m0`) and belongs in the calibration sweep.
4. **Streaming / spill graph build.** The `m0_max` build is full-in-memory; bounding its peak (spill-to-disk streaming merge) is what lets the full build scale past the current ceiling.
5. **Calibration fidelity.** Calibration queries are synthesized near corpus nodes and scored against Sq16 GT, both optimistic vs real out-of-distribution queries against fp32 GT (nytimes shows ~0.015 of gap). Harden the synthesized queries / score against fp32 so the internal bar tracks production recall.
6. **Recalibration on growth / migration first-build.** Incremental drains keep the stamped `(m0, ef)` indefinitely; since the `m0` requirement grows with N, served recall can drift as a table grows. A growth-threshold recalibration (and the one-time first-build allowance for a pre-existing table's first drain) is deferred.
7. **Above the graph scale ceiling** (`> hnsw_max_docs`, e.g. 100M), the table doesn't build a per-row graph — it serves `ivf` routed through a centroid graph over the cluster centroids. That centroid-routing path is the subject of the ongoing https://github.com/infino-ai/infino/pull/588, so it's out of scope here.
8. **Deterministic build.** The graph build is non-seeded, so recall/latency carry run-to-run variance (~±0.02 recall at sparse `m0`, ~±0.3 ms). A fixed build seed would make points reproducible and de-noise the calibrator's own measurements.
9. **Lower the `m0` candidate floor (opt-in).** The grid starts at m0=32 (the conventional HNSW base degree) because the original pressure was high-dim recall, which needs *more* density. The hand-configured m0=16 row shows a leaner rung beneath the current floor; letting the calibrator *stamp* it is future opt-in work — bounded, since going *too* sparse degrades connectivity (search takes more hops and gets slower again), so the useful floor is a small degree, not zero. Best paired with follow-up 5, since a sparse-`m0` graph's calibrate estimate runs optimistic and could under-serve at the low target boundary.

## Testing

Unit tests cover the calibrator's latency-first pick + register/graceful decision, base-layer pruning, the `ef` bundle round-trip, and config validation (including the new `target_recall`/`hnsw_recall_slack` range checks). Real-dataset behavior (the table above) and the probe-gate — every dataset probes at 100K, registers, and proceeds to the full-corpus build — were validated at scale up to 10M; details in https://github.com/infino-ai/infino/issues/590.


src/supertable/query/vector.rs
```diff
@@ -1503,7 +1506,424 @@ fn score_cell_fp32(
...
+        };
+        let hits: Vec<SuperfileHit> = data
+            .graph
+            .search(&data.scorer, query, k, ef)
+            .into_iter()
+            .filter_map(|(node, dist)| {
+                Some(SuperfileHit {
+                    superfile: SuperfileUri(Uuid::nil()),
+                    local_doc_id: 0,
+                    score: dist,
```
ekechinwokah: confirmed, merge-blocking: this stamps the graph's raw -dot as SuperfileHit.score (Sq16Scorer hardcodes NegDot, norm-uncorrected), while the ivf arm returns 1 - dot/norm for the same cosine column (reader.rs Sq16Kernel::new(col.metric, ...)). the two arms then merge on raw score in top_k_ascending at ~4281. on unit rows the shift is exactly +1.0, so an appended exact match (ivf score ~0) loses to every drained graph hit with dot > 0, and hits_id_score_batch emits -1.0 for a perfect match where the documented contract is distance-with-0-perfect. fix at this boundary: return 1 + dist (or rescore via the column metric) so both arms share one scale — and add the cross-arm merge test that fails today.

src/supertable/query/vector.rs
```diff
@@ -1503,7 +1506,424 @@ fn score_cell_fp32(
...
+    /// time.
+    ///
+    /// Each hit's `superfile`/`local_doc_id` are left `nil`/`0`: the
+    /// `_id`+score projection answers straight from `stable_id`, and any
+    /// wider projection resolves the live `(superfile, local)` from that id
+    /// through [`user_placement_for_scalar_resolve`] on the shared
+    /// `vector_search` path — compaction-correct without baking physical
+    /// rows into the graph. The graph distance is `−dot` on the Sq16 grid;
+    /// smaller is nearer, matching the `SuperfileHit.score` convention.
+    async fn hnsw_search(
```
ekechinwokah, line 2175: confirmed, merge-blocking: hnsw_search has no column argument and the bundle records none, but build_hnsw_graph_ref builds only vector_columns.first() — and the hidden manifest carries ALL the user's vector columns (handle.rs builds hidden_vector_columns from the full list). a query on a same-dim second column passes the only guard (dim == query.len()) and returns the first column's neighbors, silently. the router at 2263 also returns before the unknown-vector-column check at ~2340, so undeclared names return hits on a fully drained table (the existing guard test never drains, so it still passes). stamp the column name in the bundle and check it here, and hoist the column validation above the hnsw branch.

src/supertable/manifest/mod.rs
```diff
@@ -1875,16 +1890,26 @@ impl ManifestSnapshot {
...
             slow_vector_state_content_hash: None,
             slow_vector_state_centroids: None,
+            // The `hnsw` graph ref, by contrast, IS carried forward:
+            // the graph is a function of which stable doc ids exist, not how
+            // they are packed, so it survives a repack. The post-drain /
+            // post-compaction settle keys on the doc-id watermark to decide
+            // whether the carried graph still covers the population (reuse)
+            // or the rows changed (rebuild). Carrying it forward keeps it
+            // referenced (GC-live) and lets a packing-only settle reuse it.
+            slow_vector_state_graphs: self
```
ekechinwokah, line 1933: confirmed: carrying the graph ref forward is deliberate, but the batch commits that ride it also advance drained_ranges in the same commit — and the rebuild only happens at the final settle. from the second drain on, every query in that window serves a graph that does not contain the just-drained rows while get_undrained_superfiles_loaded(&drained) excludes those same rows from the user arm: invisible to both, for the whole drain, and durable across a crash between batch commit and settle. hnsw_search has no population/high-water staleness guard to catch it. either advance the watermark only in the settle commit that also swaps the graph ref, or make hnsw_search compare a stamped high-water against the manifest and fall back to ivf when stale.

src/supertable/query/vector.rs
```diff
@@ -1503,7 +1506,424 @@ fn score_cell_fp32(
...
+    if prior.doc_ids.len() + new_doc_ids.len() != current_total {
+        return Ok(None);
+    }
+
+    let inserted = new_doc_ids.len();
+    // Incremental drains INHERIT the prior graph's calibrated `(m0, ef)`: the
+    // new nodes must match the existing base-layer degree (a mixed-degree
+    // graph would be inconsistent), and the stamped query beam carries
+    // forward. No recalibration here — that only happens on a full build.
+    let inherited_m0 = prior.graph.base_degree();
```
ekechinwokah, line 2062: confirmed, this breaks the self-driving claim: the incremental path inherits (m0, ef) and never recalibrates — build_hnsw_graph_ref tries incremental FIRST regardless of the population key, the append guard is satisfied by any pure append, and nothing re-measures recall on the extended graph. your own comment at ~1681 states the m0 requirement is scale-dependent (50k looks 0.99 while the full corpus serves 0.86); an append-only table calibrated at 200k and grown to 5m is exactly that regime, still registered, still served. cheapest fix: re-run graph_recall on the extended graph and de-register below the floor; better: full recalibration once total exceeds some multiple of the calibrated population.

src/superfile/vector/hnsw.rs
```diff
@@ -0,0 +1,2101 @@
...
+        let neighbors = self
+            .neighbors
+            .iter()
+            .map(|levels| {
+                levels
+                    .iter()
+                    .enumerate()
+                    .map(|(lvl, lst)| {
+                        if lvl == 0 && lst.len() > m0 {
+                            lst[..m0].to_vec()
```
ekechinwokah: confirmed: lst[..m0] prunes positionally, but a list that never overflowed its cap is ordered [distance-sorted own selection | reverse links in arrival order] — and with m0_max=256 over ef_construction=200 the cap rarely overflows, so unsorted tails are the norm. the prune therefore preferentially deletes reverse links regardless of distance: the calibration proxy is unsound (line 656's 'approximates a native m0 build' claim doesn't hold), small m0 rungs measure worse than a native build would, and the calibrator over-provisions m0 (up to 8x base-layer memory) or declines graph-friendly corpora — plus a run-varying set of in-degree-zero nodes is permanently unreachable in the persisted graph at any ef. the fix is local and you already have the machinery: re-run select_neighbors_heuristic per node at m0 (the same call link_into makes on overflow) instead of slicing.

src/config/mod.rs
```diff
@@ -788,7 +949,46 @@ impl Config {
...
+        if v.hnsw_ef_mult < 1 {
+            return Err(ConfigError::Invalid(format!(
+                "vector.hnsw_ef_mult must be >= 1, got {}",
+                v.hnsw_ef_mult
+            )));
+        }
+        // A recall target outside (0, 1] is meaningless and, shared with the
+        // ivf stamping law, would silently mis-stamp both engines; an
+        // unreachable one (e.g. > 1) would decline the graph on every table.
+        if !(v.target_recall > 0.0 && v.target_recall <= 1.0) {
```
ekechinwokah, line 984: two problems with the new validate. first, its only production caller is global()'s Config::load().unwrap_or_default() — so one invalid hnsw knob now silently discards the operator's ENTIRE config file, reverting storage, commit thresholds, budgets, and drain tuning to embedded defaults with no log line; before this pr that file loaded. fail loud (log the ConfigError at minimum) or validate per-section. second, the cross-field checks that actually matter are missing: hnsw_recall_slack >= target_recall collapses the register floor to 0 so any graph with recall > 0 registers; hnsw_ef_ceil below 128 empties the calibration grid and the drain then logs 'graph-hostile' when no sweep ran; and hnsw_m0 is never related to hnsw_ef_construction, so m0 above ~ef_construction is pure sentinel padding (hnsw_m0: 1024 = 4kb/node of sentinels, 40gb at 10m, for a graph that is by construction degree <= 200).

src/supertable/query/vector.rs
```diff
@@ -1895,6 +2492,23 @@ impl SupertableReader {
...
+                    "[WIDTH] k={} law_width={:?} nprobe_min={} nprobe_max={} fine_nprobe={} prepin_fine_depth={}",
+                    k,
+                    law_width,
+                    cell_routing.nprobe_min,
+                    cell_routing.nprobe_max,
+                    cell_routing.fine_nprobe,
+                    prepin_fine_depth
+                );
+            }
+            if let Some(cap) = std::env::var("INFINO_FINE_CAP")
```
ekechinwokah: these two env reads have to come out before merge. INFINO_FINE_CAP silently clamps fine_nprobe AFTER the stamped width pin, on every cell-routed query — hidden, filtered, and default — so a stray export serves below the stamped law and the 0.99 bar with no config audit able to see it. INFINO_LOG_WIDTH is a per-query eprintln plus a per-query getenv. the yaml-only contract is pinned in config/mod.rs (env_vars_do_not_override_config) and CLAUDE.md, and this pr's own f0d6a87 removed exactly this pattern (INFINO_GLOBAL_FINE_FC/_COALESCE promoted to config). nothing else in the repo references either name — debug leftovers. promote the cap to a vector.* knob if it's wanted; tracing::debug for the width line.

src/supertable/query/vector.rs
```diff
@@ -1503,7 +1506,424 @@ fn score_cell_fp32(
...
+            );
+            return Ok(None);
+        }
+        eprintln!(
+            "[supertable hnsw] probe col={column} dim={dim} n={n} (sampled {pn}): \
+             registrable (m0={} ef={} recall {:.3}) — proceeding to full-corpus build",
+            pchoice.m0, pchoice.ef, pchoice.recall
+        );
+    }
+    let scorer = Sq16Scorer::from_codes(codes.clone(), dim, n);
```
ekechinwokah: drain peak rss is 3.3-5x the code plane at the DEFAULT 10m ceiling: codes plus this codes.clone() both live through encode_hnsw(&codes, ...) at 1719; calibrate_graph holds the m0_max base graph plus a byte-for-byte pruned copy (the 256 rung truncates nothing, so it's a full deep copy); encode_hnsw then allocates graph.to_bytes() plus the reserved output buffer. that's ~50-78gb for a table the config's hnsw_max_docs doc advertises as fitting in ~15gib — the drain ooms at the supported scale. the incremental path holds three planes and two graphs. arc the plane between scorer and encoder, prune via an m0 cap consulted at neighbor iteration instead of copying, and stream the encode.

src/supertable/query/vector.rs
```diff
@@ -1503,7 +1506,424 @@ fn score_cell_fp32(
...
+        // The calibrated per-table `ef` stamped in the bundle is the default;
+        // fall back to the ef law only for older bundles with no stamp (0).
+        let ef = if data.ef_search > 0 {
+            data.ef_search.max(k)
+        } else {
+            hnsw_ef(k)
+        };
+        let hits: Vec<SuperfileHit> = data
+            .graph
+            .search(&data.scorer, query, k, ef)
```
ekechinwokah: the concurrency contract (rayon for cpu on options.reader_pool, tokio for i/o, bridge with a oneshot) is violated at four new sites: this serving walk runs inline on a tokio worker (up to ef=2048 x m0=256 sq16 scores, ~19ms p95 on glove-200 per the pr body); calibrate_graph at drain runs inline in assemble_hnsw_sections AND its rayon build lands on the GLOBAL pool, not reader_pool; fetch_graph_sections blake3-hashes and triple-copies a multi-gib bundle inline INSIDE the graph_sections_cache mutex (both call sites), so every concurrent query on the table blocks behind the first post-drain hydration instead of falling back to ivf. the correct pattern is already in these files with comments explaining why — run_on_pool at 1165, load_full_state's spawn_blocking. same fix here.

src/superfile/vector/hnsw.rs
```diff
@@ -0,0 +1,2101 @@
...
+        }
+        let prepared = scorer.prepare(query);
+        let mut ep = self.entry;
+        let top = self.node_level[self.entry as usize];
+        let mut l = top;
+        while l >= 1 {
+            ep = self.greedy_nearest(scorer, &prepared, ep, l);
+            l -= 1;
+        }
+        let mut visited = VisitedSet::new(self.len);
```
ekechinwokah: VisitedSet::new(self.len) on every search call — a zeroed vec![0u32; n], 40mb at 10m nodes, plus a soft page fault for most of the ~ef x m0 scattered node touches — wrapped around a walk of the same order or less. the type's own doc two screens up says 'O(1) reset ... no per-search allocation', and clear() exists and is unused on this path (the build side does it right with per-worker scratch). calibration multiplies this by up to 4000 searches per drain. take caller-owned scratch (search_layer already accepts &mut VisitedSet) or pool it. relatedly the for_each_init comment at 415 says once-per-thread; rayon documents init as once per JOB — harmless at build cadence, but fix the comment.

src/supertable/query/vector.rs
```diff
@@ -1503,7 +1506,424 @@ fn score_cell_fp32(
...
+            .filter_map(|(node, dist)| {
+                Some(SuperfileHit {
+                    superfile: SuperfileUri(Uuid::nil()),
+                    local_doc_id: 0,
+                    score: dist,
+                    stable_id: Some(*data.doc_ids.get(node as usize)?),
+                })
+            })
+            .collect();
+        Ok(Some(top_k_ascending(vec![hits], k)))
```
ekechinwokah, line 2252: plausible today, live the day someone sets vector.drain_replica_target_factor > 1.0: assemble pushes every materialized row as its own node with no stable-id dedup (boundary replicas share the user _id), graph.search returns exactly k nodes, top_k_ascending collapses same-id hits, and the refill loop at ~4303 only grows the request when deletes exist — so a delete-free table returns fewer than k rows. the ivf arm handles this exact case explicitly (k_fetch = k + replica_overhead at ~3068); the graph arm is the only path without the term. over-fetch by the replica factor here, or dedup by stable id at assemble time.

src/superfile/vector/hnsw.rs
```diff
@@ -0,0 +1,2101 @@
...
+        let m = c.u32()? as usize;
+        let m0 = c.u32()? as usize;
+        let ef_construction = c.u32()? as usize;
+        let entry = c.u32()?;
+        if n == 0 || entry as usize >= n || m0 == 0 {
+            return None;
+        }
+        let mut node_level = Vec::with_capacity(n);
+        for _ in 0..n {
+            node_level.push(c.u32()?);
```
ekechinwokah: from_bytes underr-validates in two ways that both defeat the 'corrupt section degrades to a fallback rather than a panic' contract at 953: (1) neighbor ids are checked < n but never node_level[id] >= level, and greedy_nearest indexes self.neighbors[best][level] unguarded — a level edge to a shorter tower is an oob panic at query time (ParBuild::snapshot guards this exact pattern on the build side); (2) allocations are driven by wire values before any length cross-check — a node_level word of u32::MAX requests ~103gb (abort via handle_alloc_error), n off the wire feeds with_capacity unchecked. blake3 gates the production path so the trigger is a writer bug or same-magic skew, not bit rot — but that's exactly when you want None-and-ivf, not a panic inside a query or a drain worker. reject lvl > MAX_LEVEL, check tower coverage per edge, cross-check n/m0/len against bytes.len().

src/supertable/query/vector.rs
```diff
@@ -1503,7 +1506,424 @@ fn score_cell_fp32(
...
+        } else {
+            hnsw_ef(k)
+        };
+        let hits: Vec<SuperfileHit> = data
+            .graph
+            .search(&data.scorer, query, k, ef)
+            .into_iter()
+            .filter_map(|(node, dist)| {
+                Some(SuperfileHit {
+                    superfile: SuperfileUri(Uuid::nil()),
```
ekechinwokah: every graph hit carries (Uuid::nil(), 0), which collapses hit_order's documented tie-break to a constant for graph hits — and on the merge/refill paths (pool larger than the request) equal-distance hits at the k boundary are then kept or evicted by std HashMap iteration order in best_by_id. two identical queries in one process can return different _id sets among ties, which the comment above top_k_ascending explicitly promises cannot happen. hnsw_search's own single-arm set is deterministic (Scored breaks ties on node id), so the one-line fix is a stable_id leg in hit_order.

src/supertable/query/vector.rs
```diff
@@ -1503,7 +1506,424 @@ fn score_cell_fp32(
...
+        };
+        let Some(data) = sections.data.as_ref() else {
+            return Ok(None);
+        };
+        if data.dim != query.len() || data.doc_ids.is_empty() {
+            return Ok(None);
+        }
+        // The calibrated per-table `ef` stamped in the bundle is the default;
+        // fall back to the ef law only for older bundles with no stamp (0).
+        let ef = if data.ef_search > 0 {
```
ekechinwokah: this ef_search == 0 branch is dead: the bundle format is new in this pr, both encode_hnsw call sites always stamp a calibrated non-zero ef (the grid floor is 128 and unregistered choices return None before encode), and the 'older bundles' in the comment exist only in mid-branch commits that never shipped. consequently vector.hnsw_ef_floor and vector.hnsw_ef_mult — documented with tuning guidance in config.yaml, validated, unit-tested — cannot influence any served query. drop the branch and both knobs (keep hnsw_ef_ceil, which genuinely caps the calibration grid), or a future format bump has to disambiguate ef=0-means-old from ef=0-means-unset.

src/supertable/manifest/list.rs
```diff
@@ -1216,6 +1224,10 @@ struct ManifestDto {
     slow_vector_state_centroids_uri: Option<String>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     slow_vector_state_centroids_content_hash: Option<String>, // "blake3:<64hex>"
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    slow_vector_state_graphs_uri: Option<String>,
```
ekechinwokah, line 1228: process + hygiene, none of it blocking on its own: new on-disk blob layout (INFVGB01 fixed-offset header) and these new manifest wire fields plus the stamp-semantics change in writer.rs are ask-first items per CLAUDE.md — retroactive issue please (for the record: the never-propose hnsw-inside-ivf item does NOT apply, this is a whole-table graph and the pr body carries the new evidence that list demands). style sweep to fold in with the fixes: fnv and calibration constants inlined mid-expression (writer.rs 7636/7640, hnsw.rs 780/784/789), inline crate:: paths where the file already imports the module (writer.rs 7712 vs its own line 1796), function-local test imports duplicating file-level ones (hnsw.rs 1834/1949/2047), 6 of 13 commit subjects over ~70 chars, 22 eprintln sites in modules that already use tracing, and the pr-body perf tables belong in benches/README.md per the repo convention.

ekechinwokah (changes requested):
not mergeable yet. the graph work itself is strong — the epoch visited-set, the seeded order-independent towers, the diversity heuristic with its rationale written down — and the calibrate/prune/register design is the right shape. the bugs are at the seams to the scan path it replaces, and three of them bite on the shipped default config. inline comments carry each finding with the trigger; the headliners:

- graph hits score -dot while ivf hits score 1-dot, and the two arms merge on raw score. any freshly appended (undrained) row loses to any drained hit with positive similarity, and the public score column goes negative. this makes the 'worst case is serve ivf, exact same results' claim false the moment a table has undrained rows — which is every table between drains.
- the graph is built for vector_columns.first() only and hnsw_search carries no column identity, so a same-dim second column is served from the wrong column's graph, silently. unknown column names also stop erroring on drained tables.
- from the second drain on, each batch commit advances the drained watermark while the graph only rebuilds at the final settle: the just-drained rows are invisible to BOTH arms for the drain's duration, and a crash mid-drain makes that state durable.
- incremental drains inherit (m0, ef) forever. an append-only table never recalibrates, there is no growth bound, and your own comment says why that's wrong: 50k looks 0.99 while the full corpus serves 0.86. the acceptance bar is 0.99 served; this sinks below it with the graph still registered.
- pruned_base_layer truncates layer-0 positionally, which preferentially drops reverse links (they sit unsorted at the tail unless the cap overflowed). the calibration proxy is therefore unsound — it over-provisions m0 or spuriously declines the graph — and a run-varying minority of nodes is permanently unreachable in the persisted graph.

five sharp-looking candidates were checked hard and CLEARED, so nobody chases them: sq16 is cosine-locked at three build sites so the missing metric check is unreachable; updates cannot stale the population key (fresh minted ids move count and max); the calibration self-hit bias nets out under the slack once you use the right counterfactual; probe-subsample stride aliasing does not transfer (the sample is its own universe); and the registered/graph expect is derived from one boolean and cannot panic.

two asks beyond the fixes: search_mode should land defaulted to ivf until the seams above are closed — flipping the default and fixing the merge scale can be its own follow-up with the bench evidence attached. and the on-disk bundle (INFVGB01), the new manifest wire fields, and the stamp-semantics change are ask-first territory per CLAUDE.md — retroactive issue please, the pr body already has the evidence it needs.

one adjacent bug surfaced during verification that is NOT this pr's fault (updates never record their old copy into the hidden deleted-id set; pre-existing, byte-identical on main) — filed separately, don't take it on here.

gates: 10 finder angles, 18 adversarial verifications — 12 confirmed, 3 plausible, 5 refuted. no test in the tree exercises VectorSearchMode::Hnsw; the five headliners each need a regression test that fails today.

src/supertable/query/vector.rs
muralikpbhat: Fixed in `cd86131`. The graph arm now returns `1 + dist`, so it merges on the same scale as the ivf arm's `1 − dot` (Sq16 is cosine-only, so no other metric reaches the graph). Added the cross-arm merge test that failed before the fix: `hnsw_graph_arm_emits_cosine_scale_scores`.

src/supertable/query/vector.rs
muralikpbhat: Fixed in `cd86131`. The bundle now stamps the column name and `hnsw_search` checks it; the unknown-column validation is hoisted above the hnsw branch; and the magic is bumped so un-stamped older bundles decode as absent → ivf. A same-dim second column now falls back to ivf on the mismatch instead of silently returning the first column's neighbours. Test: `hnsw_unknown_column_errors_on_drained_table`.

src/supertable/manifest/mod.rs
muralikpbhat: Fixed in `cd86131`, via your first option: the graph is built against the prospective manifest and stamped in the **same membership commit that advances `drained_ranges`**, so the watermark never leads the graph and the just-drained rows are covered the moment they become visible. `6f1199b` goes further and removes the settle's rebuild entirely — the membership commit is now the sole builder; the slow-state settle only carries the ref forward. Test: `drain_window_keeps_just_drained_rows_visible` (asserts visibility across a crash between the membership commit and the settle).

src/supertable/query/vector.rs
muralikpbhat: Fixed in `cd86131`: the incremental path now re-measures recall on the extended graph and de-registers (→ ivf) when it drops below the floor, so an append-only table grown out of its calibrated regime stops being served a stale graph. `6f1199b` runs that recheck on the reader pool and bounds it to a strided subsample above the probe cap. Test: `measure_recall_reflects_graph_quality`.

src/superfile/vector/hnsw.rs
muralikpbhat: Fixed in `cd86131`: `pruned_base_layer` now re-runs `select_neighbors_heuristic` per node at the target m0 — the same call `link_into` makes on overflow — instead of slicing `lst[..m0]`. The proxy is distance-sound again and no node is left with in-degree zero. `6f1199b` parallelizes it across the reader pool. Test: `pruned_base_layer_tracks_native_small_m0`.

src/config/mod.rs
muralikpbhat: Fixed in `cd86131`, both parts. (1) A config-load failure now surfaces (the `ConfigError` is logged) instead of silently reverting the operator's whole file to defaults. (2) Added the cross-field checks: `hnsw_recall_slack >= target_recall`, `hnsw_ef_ceil < 128` (would empty the grid and mislabel it "graph-hostile"), and `hnsw_m0` vs `hnsw_ef_construction` (sentinel padding). `6f1199b` also rejects a zero `hnsw_probe_max_docs`. Test: `hnsw_calibration_config_validates`.

src/supertable/query/vector.rs
muralikpbhat: Fixed in `cd86131`: both `INFINO_FINE_CAP` and `INFINO_LOG_WIDTH` are removed — the yaml-only contract holds again — and the width line is now `tracing::debug`.

src/supertable/query/vector.rs
muralikpbhat: Fixed in `cd86131`: the code plane is shared (Arc) between the scorer and the encoder rather than cloned, and collection is bounded to the probe subsample instead of the whole plane, so drain peak RSS no longer runs multiples of the code plane at the ceiling. `6f1199b` additionally frees the resident graph when the manifest drops its ref (was pinned for process life).

src/supertable/query/vector.rs
muralikpbhat: Fixed in `cd86131`: the serving walk and the drain-time calibrate/build run on `options.reader_pool` via `run_on_pool` (not inline on a tokio worker, not on the global rayon pool), and the graph-sections hydration no longer holds the cache mutex across the multi-GiB fetch. `6f1199b` extends the same treatment to the incremental recheck and single-flights the hydration so a post-drain burst can't launch duplicate multi-GiB fetches.

src/superfile/vector/hnsw.rs
muralikpbhat: Fixed in `cd86131`: the search path takes caller-owned `VisitedSet` scratch and resets it via `clear()` instead of allocating a zeroed vec per search; the `for_each_init` comment is corrected to "once per job."

src/supertable/query/vector.rs
muralikpbhat: Fixed in `cd86131`: the graph arm now over-fetches by the replica factor, matching the ivf arm's `k_fetch = k + replica_overhead`, so dedup by stable id never returns fewer than k on a delete-free replicated table. `6f1199b` adds a comment noting this is a build-time property (don't lower the factor below the value used at build).

src/superfile/vector/hnsw.rs
muralikpbhat: Fixed in `cd86131`: `from_bytes` now rejects `lvl > MAX_LEVEL`, checks tower coverage per edge, and cross-checks `n`/`m0`/len against `bytes.len()` before allocating — so a corrupt or same-magic-skewed bundle degrades to `None` → ivf instead of panicking in a query or drain worker. `6f1199b` adds the matching pre-alloc guard to `decode_hnsw`. Tests: `from_bytes_rejects_tower_violations` and a corrupt-`n` case in `hnsw_bundle_roundtrip`.

src/supertable/query/vector.rs
muralikpbhat: Fixed in `cd86131`: graph hits now carry their real stable id, and `hit_order` gains a `stable_id` leg, so the merge/refill paths break equal-distance ties deterministically instead of by HashMap iteration order. Two identical queries now return the same set among ties.

src/supertable/query/vector.rs
muralikpbhat: Fixed in `cd86131`: the dead `ef_search == 0` branch is removed, along with `vector.hnsw_ef_floor` and `hnsw_ef_mult`. `hnsw_ef_ceil` stays — it genuinely caps the calibration grid.

src/supertable/manifest/list.rs
muralikpbhat: On-disk format — the `INFVGB01` fixed-offset header, the `slow_vector_state_graphs` manifest wire field, and the drain-time stamp semantics — is documented on #590 per the ask-first convention, reusing the tracking issue rather than opening a new one. Agreed the never-propose "hnsw-inside-ivf" item doesn't apply (this is a whole-table graph). Style: the graph/drain `eprintln` diagnostics are migrated to `tracing` (`cd86131`/`6f1199b`), and the 13 commits are squashed to 2 clean subjects. The remaining micro-nits (inlined constants, redundant `crate::` paths, function-local test imports) and moving the infino-only hnsw-vs-ivf tables to `benches/README.md` will come as a follow-up.

muralikpbhat:
Thanks for the thorough pass — all confirmed findings are addressed.

**Round 1 (`cd86131`)** covers the 15 threads above: cosine score scale, column identity, atomic drain-window stamping, incremental recalibration, distance-aware base prune, config validation + load-failure logging, env-var removal, bounded drain memory, the reader-pool concurrency contract, `VisitedSet` reuse, replica over-fetch, decode hardening, deterministic tie-break, and dead-`ef` removal.

**A follow-up self-review pass (`6f1199b`)** hardened eight further items: reject a zero probe cap (divide-by-zero), exclude split-superseded parent cells from graph assembly (was double-ingesting after a split), stamp the graph only in the membership commit, free + single-flight the resident graph cache, bound the incremental recheck, parallelize the prune, and guard the decode allocation.

On-disk format is documented on #590. CI is green (fmt/clippy, 2226 lib + integration tests, coverage 92.9 / 89.6 / 92.5), and it's validated at scale on Cohere-1M (recall@10 0.9944, p95 2.3 ms), including a forced cell-split stress run.

muralikpbhat:
## Drain cost on object storage — cold-read fixed + cost tables (Azure Blob)

The earlier revision of this branch regressed the drain wall on the object-store bench (**drain Wall +102%** vs the same-host main arm). Root cause: the resident graph is built by re-reading the just-drained hidden cells, but the reader-tier warm-through ran **after** the build — so on an object store the build re-fetched the whole Sq16 plane **cold** (one full-plane GET of data written seconds earlier). Local-fs never showed it (the re-read hits the page cache for free).

**Fix (`ba6a0f8`):** warm the disk cache before the graph build, so the re-read is served locally. Drain object I/O drops back to the no-graph baseline.

Also folded in:

- **`ffed1e1`** — assemble the drain graph in one plane decode instead of two/three.
- **Review fixes** — size the built scorer from the **decoded** plane (not the metadata pre-count), so a transient read fault can't slice out of bounds; warm **only** the disk cache (the in-memory reader tier has no eviction, so pre-committing to it would pin bytes on a failed drain); small-corpus reuse of the already-decoded plane.

Both tables below are measured on **Azure Blob**, main (`637fb24`, no graph) vs this branch (latest).

### Synthetic 1M — Azure Blob

| metric | main (ivf, no graph) | this branch | delta |
|---|---|---|---|
| drain wall | 27.9 s | 37.0 s | +9.1 s (+32%; graph CPU, under the 50% gate) |
| drain object I/O (down) | 139 GET / 2.14 GiB | 139 GET / 2.14 GiB | parity — cold re-read gone |
| optimize (drain+compact) | 112.8 s | 95.0 s | ≈ |
| ingest wall | 67.6 s | 61.6 s | ≈ |
| post-drain recall@10 | 0.990 | 0.992 | ≈ (both serve ivf) |
| warm p50 | 613 µs | 611 µs | ≈ (both serve ivf) |
| stored / Storage$ | 4.55 GiB / $0.11 | 4.55 GiB / $0.11 | ≈ |
| drain $/pass | $0.0046 | $0.0054 | +$0.0008 |
| monthly total | ~$0.28 | ~$0.28 | ≈ |

Synthetic is graph-hostile, so both arms serve `ivf` and the fix is drain-only. The decisive row: drain object I/O is restored to no-graph parity — the cold re-read is gone. The residual +9 s drain wall is graph CPU (probe build + warm decode), **+32% — under the merge gate's 50% threshold** (was +102%, blocked).

### Cohere-1M — Azure Blob

| metric | main (ivf, no graph) | this branch (hnsw) | delta |
|---|---|---|---|
| recall@10 | 0.998 | 0.9889 | −0.009 (parity) |
| warm p50 | 7.05 ms | 1.32 ms | 5.3× faster |
| qps (serial) | 129 | 746 | 5.8× |
| cpu-s / query | 0.037 | 0.00144 | 26× lower |
| peak RSS (search) | 9.66 GiB | 12.91 GiB | +3.25 GiB (the graph) |
| serving $/1M queries | $1.16 | $0.136 | 8.5× cheaper |
| queries per $ | 0.86 M | 7.36 M | 8.5× |
| ingest wall | 73.1 s | 72.9 s | ≈ |
| drain+index wall | 34.5 s | 234.3 s | 6.8× — one-time graph build cost |
| drain+index $ (amortized 1×/mo) | ~$0.003 | ~$0.07 | small |
| monthly total | $1.37 | $0.46 | 3× cheaper |

The graph serves **5.3× faster / 8.5× cheaper reads → 3× lower monthly total**, despite the build tax. The cache-reorder holds on the register path too: the Blob drain+index (234 s) ≈ its local-fs value (~215 s) plus the inherent Blob write, with **no cold re-read penalty**.

The **6.8× drain+index wall is the one-time graph build cost** — the register path is build-bound (the build runs at the top of the `m0` grid regardless of the stamped degree). Reducing it is tracked as a **follow-up**, not part of this change.

**Pricing notes:** Cohere serving priced on compute/RAM only (`get_per_query=0`, matching the prior ledger); stored bytes are backend-independent; drain+index $ is compute-only (the drain/optimize PUTs ≈ $0.005 are not added).

muralikpbhat:
## Why `hnsw_ivf` is opt-in (default `ivf`) — and still worth landing

`search_mode = hnsw_ivf` builds a resident in-memory HNSW graph over the Sq16 data vectors. It's **default off** deliberately:

- **The graph is RAM-pinned; the scan cache isn't.** `ivf` serves from the NVMe-backed block cache, which grows and shrinks with activity (down to zero when idle). The graph has to be RAM-resident to deliver its latency — a walk hits scattered nodes, so it can't run usefully off disk. Forcing every table to pin ~2 GB (1M) / ~20 GB (10M) of RAM isn't the right default.
- **Refresh doesn't fit a multi-node deployment yet.** The graph is built on the draining (primary) node; any other serving node — even a single read replica in a primary+replica pair — must reload the *whole* graph blob every time the primary publishes a new generation (there's no delta today). That ∝-corpus reload per drain can't keep up, so the graph is effectively single-box until cheap ∝-delta propagation lands — an LSM-style overlay (append-only delta blobs merged at read), real non-trivial follow-up work.

**Why it's still valuable as an opt-in:**

- **Genuine single-node latency win** — for a RAM-provisioned, single-box table the walk touches ~1–2K nodes and scores Sq16 inline (no rerank tail) instead of scanning ~100K rows: 5.3× faster / 8.5× cheaper reads at recall parity on Cohere-1M (tables in the comment above).
- **Correctness never depends on it** — the `_ivf` suffix is the contract: a missing graph (pre-drain, above `hnsw_max_docs`, hostile probe, or a non-matching column) serves `ivf` automatically. Enabling it can only add a fast path, never break one.
- **The machinery generalizes** — the same build/search will drive a **centroid router**, a far smaller resident set that refreshes cheaply, which is where the large-scale cost win lives (separate change).

Net: default `ivf` keeps the reclaimable, cold-safe economics for everyone; `hnsw_ivf` is there for single-box deployments sized for it.

muralikpbhat:
## Cohere-10M — Azure Blob: `ivf` vs `hnsw_ivf` (same binding)

With the drain-time gate in place, `search_mode = ivf` no longer builds the graph, so both arms run on the **same branch binding** — a true same-code comparison (the 1M table had to use `main` for the ivf column because the pre-gate build ran the graph unconditionally). Cohere-10M (768d cosine), n_cent 256, both arms measured by reopening the built table on a fresh 64 GiB cache; build metrics from the drain/optimize run.

| metric | `ivf` | `hnsw_ivf` | delta |
|---|---|---|---|
| recall@10 | 0.9968 | 0.9940 | −0.003 (parity, both ≥0.99) |
| warm p50 | 14.34 ms | 5.83 ms | **2.46× faster** |
| qps (serial) | 65.0 | 170.9 | **2.63×** |
| cpu-s / query | 0.1233 | 0.00595 | **20.7× lower** |
| peak RSS (search) | 20.14 GiB | 20.64 GiB | +0.5 GiB (see note) |
| serving $/1M queries | $3.87 | $0.948 | **4.08× cheaper** |
| queries per $ | 258 K | 1.055 M | 4.09× |
| ingest wall | 813.7 s | 863.7 s | ≈ |
| drain+index wall | 427.0 s | 3077.7 s | 7.2× — one-time graph build |
| drain+index $ (amortized 1×/mo) | $0.053 | $1.11 | 21× (one-time) |
| stored | 42.40 GiB | 64.46 GiB | +22.1 GiB (the graph) |
| **monthly total** | **$5.03** | **$3.71** | **1.36× cheaper** |

The graph wins reads at scale (2.5× faster / 4× cheaper at recall parity) and still lowers the monthly total, but the story is more muted than at 1M (which was 5.3× faster / 3× cheaper): at 10M the build tax is heavier (7.2× drain wall, +22 GiB storage), so the net monthly win narrows to 1.36×.

### RAM: the reclaimability distinction, measured

The near-equal serving RSS (+0.5 GiB) is worth unpacking, because it sharpens the "why not default" argument rather than weakening it. Reopening each table and sampling RSS through the query stream:

| stage | `ivf` | `hnsw_ivf` |
|---|---|---|
| after open | 0.26 GiB | 0.26 GiB |
| after 1st query | 9.50 GiB (17.6 s) | 19.91 GiB (48.9 s) |
| after 100 q | 19.31 GiB | 19.91 GiB (flat) |
| after 1000 q | 19.44 GiB | 19.91 GiB |

Both settle at ~20 GiB — that ~20 GiB is the **Sq16 vector plane, which `ivf` holds too**. So the graph's *marginal* RAM over a fully-warmed `ivf` is ~0; the difference is *when* and *whether it's reclaimable*: `hnsw_ivf` hydrates the whole plane into a pinned resident graph on the first query (the 48.9 s), while `ivf` accumulates the same plane in its block cache as queries touch cells.

The reclaimability shows directly under a tight cache budget:

| cache budget | `ivf` RSS | `hnsw_ivf` RSS |
|---|---|---|
| 64 GiB | 19.44 GiB | 19.91 GiB |
| 4 GiB | 9.95 GiB (bounded — served, colder) | 19.86 GiB (unchanged) |

`ivf`'s footprint tracks the cache budget and shrinks with it (and toward zero when idle); the graph holds ~20 GiB flat regardless, because it's resident heap, not the reclaimable block cache. That is the real reason `hnsw_ivf` stays opt-in: not that it costs more RAM under load, but that its RAM is pinned where `ivf`'s is reclaimable.

A natural follow-up (tracked separately) is to make the resident graph itself a reclaimable cache — background-hydrate without blocking the first query (serve `ivf` meanwhile), then evict under memory pressure with hysteresis, falling back to `ivf`. The `_ivf` fallback contract already makes every graph-absent state correct, so this is a lifecycle change, not a correctness one — and it would remove both the cold-load stall and the permanent pin. Complementary to the incremental-delta build follow-up already noted above.

### Other notes

- **Cold first query: 48.9 s (`hnsw_ivf`) vs 17.6 s (`ivf`)** — the +31 s is the 22 GiB graph blob loading from object storage on first touch (the per-generation load the background-hydrate follow-up would hide).
- **Build peak: `hnsw_ivf` optimize hit 49.6 GiB RSS over 51 min** (vs `ivf` 31.8 GiB / 7 min) — the build, not serving, is where the graph's memory and time cost concentrates.

Net: the 10M numbers hold the read-latency/cost win at recall parity and keep `hnsw_ivf` a monthly-cheaper option for a single box sized for it, while confirming — with numbers — why `ivf` remains the right default.

ekechinwokah (approved):
approving. i verified each fix by reading c00c506:

- graph hits now stamp 1.0 + dist onto the scan arm's cosine scale (monotonic, non-negative, cross-arm merge compares like with like)
- the bundle stamps the column (v02; a v01 bundle decodes to None and serves ivf) and hnsw_search rejects a column mismatch before the dim check
- the graph ref lands IN the membership commit, with a crash-after-commit test pinning the window shut
- incremental drains run a bounded recheck (~O(probe_cap)) on the reader pool and fall back to a full rebuild that recalibrates or de-registers
- pruned_base_layer re-runs the diversity heuristic per node, parallel, with the native-tracking test — the positional-slice hazard is documented in its comment
- env vars gone, ef law and both dead knobs deleted, slack >= target rejected at validate with a real message, hit_order carries the stable_id leg, decode rejects lvl > MAX_LEVEL and tower-short edges, and the walk rides the reader pool
- replica over-fetch on the graph arm matches the ivf arm's

the default flip to ivf with hnsw_ivf opt-in is the right call, and the ram table is what makes it stick: both arms settle at the same ~20 gib plane, but the block cache reclaims to zero while the graph pins — "when and whether it's reclaimable" is the actual distinction, not the footprint. the 10m table narrowing to 1.36x monthly (from 5.3x at 1m) stated plainly is exactly how i want results reported; the 7.2x drain wall prices correctly now that only tables that opted in pay it.

one non-blocking nit for a follow-up: the serving path calls the allocating `search` rather than `search_scratch`, whose own doc says to prefer it — at 10m that is a ~40 mb zeroed alloc per query, inside the p95 you measured, so no urgency; pool a per-worker scratch whenever convenient.

format doc on #590 noted — thanks for closing the ask-first loop. gates: all 15 review threads verified fixed at head, ci green, cohere-10m recall 0.9940 vs bar 0.99.

ekechinwokah (approved):

PR #604, opened 2026-08-14
fix gc: erase the disk-cache copy also of each deleted superfile

### LINKED ISSUE #595: gc: deleting a superfile from storage leaves its disk-cache copy behind
## Problem

When a superfile gets superseded by compaction (via`optimize()`), GC deletes it from storage, but the disk cache never hears about it. The cache only ever removes entries under byte-budget pressure, so the copies of deleted superfiles sit there until the budget forces them out. This is an issue for larger disk-cache sizes, where they never get deleted.

For context, after a 100M-row load on c7a.metal-48xl, the disk cache held ~9,600 files of which ~39 were live; the 24 GiB budget never filled, so nothing was ever evicted.

## Possible Approach

GC already computes the verified set of dead objects (keep-set plus reclaim grace, so snapshot-pinned readers stay safe). When it deletes an object from storage, also drop that URI's cache entry and file - ideally one lifecycle, one truth.

## Acceptance

- After a GC sweep, the deleting process's cache holds no entry or file for the objects that were deleted from storage.
- A churn-then-gc sequence converges the cache dir toward the live set.
- Snapshot-pinned readers holding superseded superfiles are unaffected (same grace as storage reclaim).

naman-modi:
### What does this PR do?
GC now also erases the disk-cache copy of every superfile it deletes. Closes #595.

The disk cache holds copies of storage objects, and superfiles are immutable, so a copy whose source object is gone can never be wanted again. The sweep deletes storage's copy, but the cache only ever removes entries under byte-budget pressure, so the two sides of the same object have a lifecycle on one side only. On a small cache the budget does eventually force the dead copies out, but it evicts by recency rather than by liveness, so a live entry can go while dead ones stay.

### How do we do this change?
- The sweep now erases a superfile's local cache copy right before it deletes the object from storage, so the cache never holds a superfile whose source is gone.
- Erasing covers all of it: the in-memory entry, the promoted file on disk, the partial block file, and the pending-fetch slot. A query already reading that superfile is unaffected, since unlink removes the name and not the inode.
- Only superfile keys are touched. Manifests, tombstone sidecars, and in-flight temp objects fall through untouched.

### Notes
- `drop_table(purge = true)` deletes objects directly rather than through the sweep, so a purged table's cache directory is still left behind. Each table has its own cache directory, so that one is a directory removal rather than per-URI work, tracked separately.
- The hidden vector index shares the user table's cache and sweeps through its own prefixed storage provider, so the same wiring covers it.

### DIFF
--- a/src/supertable/reader_cache/disk.rs
+++ b/src/supertable/reader_cache/disk.rs
@@ -228,6 +228,8 @@ pub struct CacheStats {
     /// by the idle-threshold sweep thread. Includes individual
     /// `sweep_once()` invocations.
     pub n_madvise_calls: u64,
+    /// Total count of entries dropped because GC deleted from objectstore.
+    pub n_gc_drops: u64,
 }
 
 /// Pulls superfile bytes through a [`StorageProvider`] and
@@ -262,6 +264,7 @@ pub struct DiskCacheStore {
     budget_warned: AtomicBool,
     n_cold_fetches: AtomicU64,
     n_evictions: AtomicU64,
+    n_gc_drops: AtomicU64,
     n_madvise_calls: AtomicU64,
     /// Number of callers explicitly waiting for lazy background
     /// promotion. A waiter means promotion is now latency-critical,
@@ -334,6 +337,7 @@ impl DiskCacheStore {
             budget_warned: AtomicBool::new(false),
             n_cold_fetches: AtomicU64::new(0),
             n_evictions: AtomicU64::new(0),
+            n_gc_drops: AtomicU64::new(0),
             n_madvise_calls: AtomicU64::new(0),
             n_promotion_waiters: AtomicU64::new(0),
             pinned_fn: std::sync::Mutex::new(pinned_fn),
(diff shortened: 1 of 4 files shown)


src/supertable/reader_cache/disk.rs
```diff
@@ -1806,6 +1811,30 @@ impl DiskCacheStore {
...
+    pub(crate) fn erase_superfile_local_copy(&self, uri: &SuperfileUri) -> bool {
+        let present = if let Some((_, entry)) = self.cached.remove(uri) {
+            self.release_entry_accounting(&entry);
+            true
+        } else {
+            false
+        };
+
+        self.coordinators.remove(uri);
+        let _ = fs::remove_file(self.cache_path(uri));
```
savannahar68, line 1829: curious why are we not using tokio::fs anywhere in this file

savannahar68 (approved):
looks good. Only concern is we are using std::fs which blocks tokio thread from the task pool. I see that for entire reader cache. Not sure why.

src/supertable/reader_cache/disk.rs
naman-modi: Fair question, followed the convection in this PR. Going to take a separate pass to check and fix this. Did not feel appropriate to have it in this PR.

naman-modi:
>  I see that for entire reader cache. Not sure why.

Yes, going to take a separate pass at this, refactor, if required.

PR #607, opened 2026-08-14
fts: prune ranked multi-term OR instead of full-scanning it

vinaykakade:
## What

A common-heavy multi-term OR (uniform term upper bounds, no dominant term) was always routed to the windowed union scanner, which SIMD-scores every posting of every term in every non-empty window and only consults the top-k threshold at heap admission — so it does **no block pruning** and its cost is independent of `k`.

This makes the pruning MaxScore path the default for multi-term OR, **k-gated**:

- At small/mid `k`, common-heavy OR now runs MaxScore. Once its heap fills, the competitive threshold rises above the common terms' per-block maxima and MaxScore skips those blocks — pruning the windowed scan ignored.
- At deep `k` (> `OR_WINDOWED_UNIFORM_MAX_PRUNING_K`), with no dominant term the threshold is too low for any block to skip, so MaxScore would degrade to a scalar full scan; the OR falls back to the windowed SIMD scan, which does that scan faster.

Both the single-shot and ranged (fan-out) OR entries now route through one `route_or_to_windowed` seam, so a query runs the same kernel whether or not the fan-out sliced it — closing a latent inconsistency where the ranged path ignored the pruning-dead test.

## Result

Ranked multi-term OR: a large latency reduction at page-sized `k` (worst-case tail cut by more than half), neutral at deep `k` — no regression. Verified with a warm A/B on the ranked-union set. Correctness is unchanged: the windowed / MaxScore / WAND kernels already agree with each other and with the brute-force BM25 oracle, and a new oracle checks the rerouted default against ground-truth BM25 at k ∈ {10, 100, 1000}.

No format change, no public-API change.

### DIFF
--- a/src/superfile/fts/reader/core.rs
+++ b/src/superfile/fts/reader/core.rs
@@ -325,16 +325,6 @@ pub(super) const OR_COUNT_BITSET_DENSITY_DIVISOR: u64 = 16;
 /// this same boundary to decide when a ranged kernel is heavy enough to
 /// ship to the reader pool (see `RANGED_KERNEL_POOL_MIN_TERMS`).
 pub(crate) const OR_WINDOW_MIN_TERMS: usize = 3;
-/// Route a multi-term OR to the windowed union scorer only when the top
-/// term's score upper bound is at most this multiple of the *average*
-/// term upper bound — i.e. no single term dominates. Uniform terms sit at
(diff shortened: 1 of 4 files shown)


vinaykakade:
Consolidating the ranked top-k work onto a single branch; this phase will land as part of the combined change once it reaches parity.

PR #611, opened 2026-08-15
fts: prune ranked multi-term OR and tighten the ranked-AND bound

vinaykakade:
Two independent improvements to ranked (scored) multi-term top-k, both routing/bound changes with **unchanged top-k** (the windowed / MaxScore / WAND kernels agree with each other and with the brute-force BM25 oracle; a new common-heavy OR oracle checks the rerouted default against ground truth at k ∈ {10, 100, 1000}).

## 1. Prune ranked multi-term OR instead of full-scanning it

A common-heavy multi-term OR (uniform term upper bounds, no dominant term) was always sent to the windowed union scanner, which SIMD-scores every posting of every term in every non-empty window and consults the top-k threshold only at heap admission — so it never skips a block and its cost is independent of `k`. Make the pruning MaxScore path the default, **gated on `k`**: at small `k` the competitive threshold rises above the common terms' per-block maxima once the heap fills, so MaxScore skips their blocks; above the cutoff (`OR_WINDOWED_UNIFORM_MAX_PRUNING_K`) the threshold is too low to skip and it falls back to the windowed SIMD scan. Both the single-shot and ranged (fan-out) OR entries route through one `route_or_to_windowed` seam, so a query runs the same kernel whether or not the fan-out sliced it. Removes the now-unused dominance heuristic.

## 2. Tighten the ranked-AND block-max bound to a true Block-Max-AND

The AND flat-merge's block-max skip bounded each non-leader term by the max block-max over *every* block overlapping the leader's whole current block. On a rare∧common query the rare leader's one block spans thousands of the common term's docs, so that collapsed to the common term's global block-max — a bound so loose the skip almost never fired. Bound each non-leader by the single block that actually contains the leader's current doc (a per-block-pair sum) and skip only to the nearest block boundary across all cursors. Shared by the 2-term and general flat-merge (`block_max_and_bound`), so both the pure-AND and must+should paths benefit.

## Measured — infino before → after

Warm benchmark, min-of-runs, average per-query latency (µs) over each ranked query shape; `k` is the top-k size:

| shape | k | before | after | Δ |
| ----- | - | -----: | ----: | -: |
| ranked AND | 10 | 555 | 478 | **−14%** |
| ranked AND | 100 | 684 | 626 | −8% |
| ranked AND | 1000 | 784 | 745 | −5% |
| ranked OR (common-heavy) | 10 | 770 | 651 | **−15%** |
| ranked OR (common-heavy) | 100 / 1000 | — | — | neutral (routes to the unchanged windowed scan) |
| boolean (must+should)¹ | 100 | 838 | 506 | **−40%** |
| boolean (must+should)¹ | 1000 | 1165 | 749 | **−36%** |

¹ boolean uses the median — its mean is skewed by one heavy outlier query present in both arms.

No COUNT / phrase / single-term regression. No on-disk format change, no public-API change.

### DIFF
--- a/src/superfile/fts/reader/core.rs
+++ b/src/superfile/fts/reader/core.rs
@@ -325,16 +325,6 @@ pub(super) const OR_COUNT_BITSET_DENSITY_DIVISOR: u64 = 16;
 /// this same boundary to decide when a ranged kernel is heavy enough to
 /// ship to the reader pool (see `RANGED_KERNEL_POOL_MIN_TERMS`).
 pub(crate) const OR_WINDOW_MIN_TERMS: usize = 3;
-/// Route a multi-term OR to the windowed union scorer only when the top
-/// term's score upper bound is at most this multiple of the *average*
-/// term upper bound — i.e. no single term dominates. Uniform terms sit at
-/// ~1.0× the average (MaxScore can't prune them → windowed wins); a
-/// dominant rare term sits well above it (MaxScore prunes hard → it stays
-/// on MaxScore). Calibrated on the 1M tier; re-measured on every bench
-/// run by the superfile tier's per-algorithm probes
-/// (`benches/utils/superfile.rs`), whose shapes sit on both sides of
-/// this threshold.
-pub(super) const OR_WINDOW_DOMINANCE_MULT: f32 = 1.5;
 
 /// Largest `k` for which a 2-term OR routes to WAND+BMW instead of
 /// MaxScore. WAND's pivot pruning needs a high top-k threshold to skip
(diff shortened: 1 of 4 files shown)


src/superfile/fts/reader/core.rs
```diff
@@ -356,13 +346,36 @@ pub(super) const WAND_BMW_2TERM_MAX_K: usize = 128;
...
 
+/// Score upper-bound spread below which a multi-term OR counts as
+/// "common-heavy" (no single term dominates): `max_ub <=
+/// OR_WINDOW_DOMINANCE_MULT * avg_ub`. Uniform terms sit near the average;
+/// a dominant (rarer) term sits well above it. See [`no_dominant_term_ub`].
+pub(super) const OR_WINDOW_DOMINANCE_MULT: f32 = 1.5;
+
+/// For a common-heavy (no-dominant-term) multi-term OR, the `k` at or below
+/// which MaxScore still out-prunes the non-pruning windowed scan, so the OR
+/// stays on MaxScore; above it the windowed SIMD scan wins.
```
naman-modi: nit: all the comments are too long, and many a times loose the essence of what it is trying to convey. Can we please try to keep them simple, precise, and to the point?

naman-modi (approved):
LGTM.

src/superfile/fts/reader/core.rs
vinaykakade: Done

PR #615, opened 2026-08-16
fts: prune phrase-in-boolean and tighten the MaxScore non-essential bound

vinaykakade:
Two independent ranked-search pruning fixes. Both are routing/bound changes with **unchanged top-k** — gated by the phrase, must+should, and brute-force BM25 oracles plus the all-algo-agreement tests (and 0 COUNT mismatches over the measured query set). No on-disk format change, no public-API change.

## 1. Two-phase must-walk for phrase-bearing boolean queries

The ranked must-walk advanced each phrase must to its next *verified* occurrence during alignment, decoding member positions across the whole co-occurrence set while searching for adjacency. For a phrase of common words that co-occur widely but are rarely adjacent, adding a *selective* must term (which should prune) instead made the query several times slower. Split the walk into two phases — approx-align every must (a phrase advances only to a member-co-occurrence doc, no positions), bar-skip on block-max upper bounds, then verify adjacency + score only on the survivors — so the selective co-clause prunes the position work. `verify_at_aligned` now stores the occurrence count in `current_tf` so two-phase scoring matches the single-phase path.

## 2. Per-block bound for MaxScore non-essentials

The f=1 MaxScore fast path already skips the non-essential `skip_to` + score for a candidate the essential alone can't lift over the threshold, but it bounded the non-essentials by their *global* term-max, so the filter rarely fired. Bound each non-essential by its block-max for the block that *contains* the candidate instead (via the `inspect_block` hint — amortized O(1), no decode). Far tighter, so the existing filter skips the per-doc cost on many more docs.

## Measured (warm, min-of-runs, average per-query latency at k=10; before = current `main`, after = this branch)

| shape | before | after | Δ |
| ----- | -----: | ----: | -: |
| phrase-in-boolean, worst case (common-word phrase + selective term) | 153 ms | 23 ms | **−85%** |
| boolean bucket, avg | 3436 µs | 1263 µs | **−63%** (entirely the case above; other boolean queries unchanged) |
| ranked multi-term OR (union), avg | 647 µs | 536 µs | **−17%** (worst case 34 ms → 15 ms) |

Phrase-only, uncommon-phrase, single-term, and pure-AND shapes are unchanged (each fix touches only its own path). COUNT unchanged (0 mismatches).

### DIFF
--- a/tests/superfile/fts/brute_force_oracle.rs
+++ b/tests/superfile/fts/brute_force_oracle.rs
@@ -332,6 +332,60 @@ async fn oracle_common_heavy_or_matches_brute_force_at_depth() {
     }
 }
 
+/// Corpus where a common non-essential ("hot") has a high block-max only in a
+/// *mid-range* block. Five anchors in block 10 (ids 1281..1289) carry `lead` +
+/// `hot`×{10,9,8,7,6} — strictly-decreasing scores far above the bulk (bulk hot
+/// tf=1), so the tie-free top-5 lives entirely in that block. It gates the
+/// per-block bound's failure mode: the filter must advance each non-essential's
+/// `inspect_block` hint *into* block 10 and read its high max there; a stale
+/// hint would under-bound and wrongly drop a top-5 anchor — something the
+/// uniform / block-0-hot corpora can't catch.
+fn mid_hot_block_corpus() -> Vec<(u64, String)> {
+    const N: u64 = 2560; // 20 × 128-doc blocks
+    let anchors: [(u64, usize); 5] = [(1281, 10), (1283, 9), (1285, 8), (1287, 7), (1289, 6)];
+    let mut docs = Vec::with_capacity(N as usize);
+    for i in 0..N {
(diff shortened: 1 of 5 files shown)


tests/superfile/fts/brute_force_oracle.rs
naman-modi, line 1: Can we add some tests in `tests/superfile/fts/phrase.rs` for the change 1. i.e. skipping co-occur-but-not-adjacent docs? Like in the `truncated_top_k_pruning_agrees_with_oracle` test, we can add the alpha-gamma shapes, and test `"alpha gamma"`, `"alpha gamma" delta` type queries?

naman-modi (approved):
LGTM.

tests/superfile/fts/brute_force_oracle.rs
vinaykakade: Done - added must_phrase_skips_cooccurring_nonadjacent in tests/superfile/fts/phrase.rs

PR #621, opened 2026-08-17
fts: rarest-driven membership walk for ranked AND with a common term

vinaykakade:
## Summary

Ranked AND (`BoolMode::And`) over a set of terms where one term is common and
the rarest is very sparse spends most of its time in the wrong place: for each
doc of the rarest term, the block flat-merge does a `skip_to` that **fully
decodes a common term's 128-doc block just to read one doc**. Profiling the
`n≥3`-term AND-with-a-common-term shape put that per-candidate block decode at
the top.

This adds a **rarest-driven membership walk** for that shape: iterate the rarest
term, and for each of its docs probe the other terms with `TermCursor::contains`
— an **O(1) bitset bit-test, no block decode** — short-circuiting on the first
miss. Only a doc present in *every* term is then materialized (decoded +
positioned) and scored. It reuses the same idea the unranked count path already
uses (`count_and_intersect_membership`), but scores the matches; the BM25 sum is
identical to the flat-merge, emitted through the same `ScoreSink`.

## Gating (deliberately narrow)

Routing is behind `and_prefer_membership`, true only when all hold:

- **v4 bitset blocks present** (`has_bitset_blocks`) — the common term's blocks
  are bitset-encoded, so `contains` is a real bit-test rather than a
  binary-search + decode.
- **≥3 terms** — a 2-term AND keeps the specialized `and_flat_merge_2term`
  two-pointer merge, which the walk was measurably losing to.
- **rarest term df < max_doc / 64** — the walk gives up the flat-merge's
  block-max heap-bar skip, so it only pays when the rarest list is genuinely
  short. A looser bound regressed the ranked tail (p99) where the bar-skip was
  doing real work.

Everything outside that shape — 2-term AND, non-sparse rarest, the all-dense /
pre-v4 case, and must+should (which keeps the flat-merge so its should-clause
scoring stays in one place) — is unchanged.

## Performance

Internal same-box A/B against the current main baseline (min-of-runs per query,
warmup + 15 iterations, same index and machine for both binaries), infino-only:

| bucket | avg | p50 | p90 |
|---|---|---|---|
| **≥3-term intersection (the target)** | **−13%** | **−16 to −18%** | **−18%** |
| 2-term intersection (gated out) | ~0% (neutral) | ~0% | ~0% |
| union (untouched path) | ~0% (neutral) | ~0% | ~0% |

The win is confined to the targeted shape; the paths the change does not touch
(2-term intersection, union) measure neutral on the same box, confirming no
collateral. No latency, throughput, or cost regression elsewhere.

## Testing

Two brute-force BM25 oracle tests (scores compared against the textbook formula,
no shared code with the optimized walk):

- `oracle_and_membership_path_matches_brute_force` — forces the walk (dense
  common term ⇒ bitset blocks; very-sparse rarest ⇒ sparsity gate) and checks
  the intersection set and every per-doc score against brute force.
- `oracle_and_membership_rejects_partial_matches` — the companion that exercises
  the **reject branch**: the rarest driver term only partially overlaps a
  selective co-term, so most driver docs must be bit-tested, found missing, and
  excluded. Asserts the result is the true 3-way intersection (not the whole
  driver list), that rejects actually fired, and adds a `k < match-count`
  truncation check on the same route.

`make ci` (fmt, clippy, tests) green.

### DIFF
--- a/src/superfile/fts/reader/core.rs
+++ b/src/superfile/fts/reader/core.rs
@@ -319,6 +319,17 @@ pub(super) const OR_COUNT_ANCHOR_DOMINANCE: u64 = 8;
 /// `total_df >= max_doc / N`.
 pub(super) const OR_COUNT_BITSET_DENSITY_DIVISOR: u64 = 16;
 
+/// Rarest-term sparsity gate for the ranked-AND membership walk
+/// (`FtsReader::and_membership_scored`): route there only when the rarest term
+/// covers less than `1/N` of the doc-id space. The membership walk drives the
+/// rarest term's *entire* list (bit-testing the others) and gives up the
+/// flat-merge's block-max heap-bar skip, so it only pays when that list is
+/// genuinely short. A looser `1/16` (the count path's divisor) let moderately
+/// sparse rarest terms through and regressed the ranked tail (p99), where the
+/// bar-skip was doing real work; `1/64` restricts it to the clearly-rare∧common
+/// shape the walk targets. Bench-calibrated against the ranked-AND tail.
+pub(super) const AND_MEMBERSHIP_RAREST_SPARSE_DIVISOR: u64 = 64;
+
 /// Multi-term OR dispatch floor. A 2-term OR is already sub-millisecond
 /// on MaxScore, so the window's per-window bookkeeping isn't worth it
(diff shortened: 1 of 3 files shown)


naman-modi (approved):
LGTM

PR #623, opened 2026-08-18
Expand FTS BM25 correctness oracle coverage

vinaykakade:
Expands the FTS BM25 correctness coverage so the optimized search
kernels can't silently drift from correct BM25. All changes are
test-only — no engine code is touched.

## What's added

- **Property-based oracle** (`fuzz_oracle.rs`): generates random corpora
  and random clause/phrase/mode/k query strings and diffs the reader
  against the textbook brute-force BM25 reference on every case. Both
  sides interpret the same generated query through the same parser, so
  clause semantics can't diverge by construction. Documents are capped
  below the length-norm table's exact region, so the comparison needs no
  quantization slack. A second variant uses a skewed vocabulary to drive
  the rare-anchor and dominant-term routing branches that a uniform
  vocabulary rarely produces. Corpus and case sizes are env-tunable and
  capped by default to keep CI fast.

- **Multi-column weighted search** (`multi_column.rs`): pins the per-doc
  score of `bm25_search_multi` against a weighted two-column reference
  (previously only membership was checked).

- **Prefix and score floor** (`prefix_and_floor.rs`): prefix expansion
  vs the brute-force OR of the same expansion; the score floor verified
  by top-k preservation across the single-term, multi-term and
  intersection kernels.

- **Edge cases and the unranked spine** (`edge_and_unranked.rs`): k=0,
  single-doc, all-equal-length (constant length norm), and duplicate
  query terms; `token_match` / `token_match_count` against corpus truth
  with the ordering and `count == ids.len()` invariants, plus the
  cross-check that a ranked search covering all matches has exactly
  `token_match_count` hits.

- **Standard (UAX#29) tokenizer** (`standard_tokenizer.rs`): builds and
  references under the standard analyzer over text with non-ASCII
  letters, apostrophes and digits, and confirms non-ASCII terms are
  indexed and searchable.

## Verification

`cargo fmt` clean, `clippy` clean, and the full FTS integration suite
(103 tests) passes, including fuzz stress runs of several thousand cases.

### DIFF
--- a/tests/superfile/fts/edge_and_unranked.rs
+++ b/tests/superfile/fts/edge_and_unranked.rs
@@ -1,0 +1,311 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The Infino Authors
+
+//! Edge-case matrix and the unranked match spine.
+//!
+//! * **Edge cases** the fuzz oracle doesn't reach by construction: `k =
+//!   0`, a single-doc corpus, a degenerate all-equal-length corpus
+//!   (constant dl-norm), and duplicate query terms (the fuzz generator
+//!   dedups atoms, so this pins whatever the parser + scorer actually do
+//!   with a repeated term, against the same parsed clauses).
+//! * **The unranked spine** (`token_match` / `token_match_count`): a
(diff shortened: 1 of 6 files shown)


naman-modi (approved):
LGTM.

PR #626, opened 2026-08-22
improv error: add error variant for permission error

Pranav2612000:
### What does this PR do?
- Exposes a new error variant for "Permission" related errors. Previously, these errors were treated as Io errors

- Permission errors are different from Io errors, in the sense that retries will not fix them. The user needs to explicitly update the credentials and try again.
- Exposing the error as a separate variant makes it easy for consumers to catch these kind of errors, update their credentials and try again.

### How do we do this change?
- All error variants ( which may fail due to permission errors ) now have a way to throw a Permission Error enum variant. ( For certain errors, we need to iterate over the error chain )
- This error is propagated to the consumers


naman-modi (approved):
LGTM, thanks.

PR #629, opened 2026-08-24
superfile/vector: SQ8 int8-VNNI hnsw walk with Sq16 refine

muralikpbhat:
## What

For `search_mode = hnsw_ivf`, navigate the resident HNSW graph on a cheap int8 (SQ8) plane derived at load from the Sq16 codes' high byte — scored with an AVX-512-VNNI `vpdpbusd` kernel where present, scalar fallback otherwise — then re-rank the final beam on the exact Sq16 codes before returning the top `k`. The graph walk dominates query cost (tens of thousands of distance evals, far more than `ef`), so making the per-node kernel ~3× cheaper roughly halves warm latency while the Sq16 refine restores exact ranking.

Two new knobs under `vector:` (both `#[serde(default)]`, so existing `config.yaml` files keep working):

- `hnsw_sq8_walk` (default `true`) — walk on the SQ8 plane + Sq16 refine. `false` walks entirely on Sq16 (the prior behavior), no extra plane.
- `hnsw_refine_k` (default `256`) — how many of the SQ8 walk's nearest candidates to re-rank on Sq16 before returning `k`. Clamped to `[k, ef]`.

Reader-side only: the SQ8 plane is derived when a graph bundle is hydrated, never persisted — no writer, bundle-format, or on-disk change. The plane is built only when `hnsw_sq8_walk` is on, so the feature costs no memory when disabled.

## Why / results

Warm p95 on a 1M-row, 768-dim cosine corpus, comparing the Sq16 walk against SQ8 walk + Sq16 refine **at matched recall** (same graph, swept `ef`):

**k=100**

| recall | Sq16 walk p95 | SQ8 + refine p95 | speedup |
| ------ | ------------- | ---------------- | ------- |
| ~0.982 | 3.2 ms        | 1.8 ms           | 1.78×   |
| ~0.993 | 5.5 ms        | 2.9 ms           | 1.90×   |
| ~0.997 | 9.4 ms        | 4.9 ms           | 1.92×   |

**k=10**

| recall | Sq16 walk p95 | SQ8 + refine p95 | speedup |
| ------ | ------------- | ---------------- | ------- |
| ~0.990 | 1.9 ms        | 0.9 ms           | 2.11×   |
| ~0.996 | 3.0 ms        | 1.5 ms           | 2.00×   |

Recall matches the Sq16 walk to within ~0.001 at every `ef`; the win is 1.8–2.0× lower warm p95. The `refine_k` default of 256 is the knee: k=10 saturates by refine_k≈32, k=100 by refine_k≈256; at the higher `ef` used for 0.99 recall, 256 sits well below `ef` and fully saturates.

### 10M scale

Same corpus family at 10M rows (same-graph flip, stamped ef=512):

| k | Sq16 walk recall / p95 | SQ8 + refine recall / p95 | speedup |
| - | ---------------------- | ------------------------- | ------- |
| 100 | 0.9826 / 8.8 ms | 0.9823 / 5.9 ms | 1.49× |
| 10 | 0.9887 / 5.5 ms | 0.9880 / 3.8 ms | 1.45× |

Recall stays within 0.0007 of the Sq16 walk at 10M. The speedup is smaller than at 1M because the larger resident graph is more DRAM-bandwidth-bound (the int8 kernel's compute advantage is partly masked by memory-stall time), but it remains a recall-neutral 1.45–1.49× on warm p95.

## Cost / trade-off

The SQ8 plane is an extra resident array of `~1 byte/dim/row` — about 50% on top of the Sq16 plane — for each hydrated data graph, built only while `hnsw_sq8_walk` is on. This is the deliberate trade for the ~2× latency win; `hnsw_sq8_walk: false` reclaims the memory and walks on Sq16.

## Correctness

- The SQ8 walk is a **navigation proxy** only; the exact top-`k` always comes from the Sq16 refine. The walk kernel scores the *centered* dot `Σ(code−128)·q` (a per-query constant shift off the raw `u8·i8` dot), which keeps the value inside f32's exact-integer range at high dim.
- New unit tests: the VNNI kernel is asserted bit-exact against its scalar reference across 64-aligned and tail widths; `search_sq8_refine` is asserted to track the Sq16 walk's recall (both measured against brute-force exact Sq16) within a small margin.
- `cargo fmt`, `clippy`, and the vector unit + integration tests pass.

### DIFF
--- a/src/superfile/vector/hnsw.rs
+++ b/src/superfile/vector/hnsw.rs
@@ -46,7 +46,7 @@ use std::{
 use rayon::prelude::*;
 
 use crate::superfile::vector::distance::{
-    Metric, Sq16Kernel, dequantize_sq16_into, dot, encode_sq16_row,
+    Metric, Sq16Kernel, dequantize_sq16_into, dot, encode_sq16_row, quantize_query_i8, sq8_walk_dot,
 };
 
 /// Per-node distance the graph is generic over. Lower = nearer.
@@ -1176,6 +1176,11 @@ pub(crate) struct HnswIndex {
     /// different column (→ ivf) rather than silently answer it from this
     /// column's neighbors.
     pub column: String,
+    /// Resident contiguous SQ8 walk plane (`n × dim` bytes) — the high byte of
+    /// each Sq16 code, derived here at load, never persisted (no writer/bundle
+    /// change). Used by [`HnswIndex::search_sq8_refine`] for a cheap int8-VNNI
+    /// walk; the exact ranking comes from the Sq16 refine over `scorer`.
+    pub sq8_plane: Vec<u8>,
 }
 
 /// Serialize a `hnsw` index to a persistable byte bundle: header,
@@ -1216,7 +1221,7 @@ pub(crate) fn encode_hnsw(
 /// Rebuild a resident [`HnswIndex`] from [`encode_hnsw`].
 /// Returns `None` on any malformation so the caller falls back to the lazy
 /// build or scan path rather than failing the query.
-pub(crate) fn decode_hnsw(bytes: &[u8]) -> Option<HnswIndex> {
+pub(crate) fn decode_hnsw(bytes: &[u8], sq8_walk: bool) -> Option<HnswIndex> {
     let mut c = Cursor::new(bytes);
     if c.take(HNSW_DATA_MAGIC.len())? != HNSW_DATA_MAGIC {
         return None;
@@ -1251,6 +1256,19 @@ pub(crate) fn decode_hnsw(bytes: &[u8]) -> Option<HnswIndex> {
     if graph.len() != n {
         return None;
     }
+    // Resident SQ8 walk plane: the high byte of each little-endian u16 code,
+    // laid out contiguously (n × dim). Reader-side only — derived from the
+    // Sq16 plane, never persisted. Built only when SQ8-walk serving is on;
+    // empty otherwise, so the plane costs no memory when disabled and serving
+    // falls back to the Sq16 walk.
+    let sq8_plane: Vec<u8> = if sq8_walk {
+        // High byte of each little-endian u16 code — the second byte of every
+        // 2-byte pair. `chunks_exact(2)` elides the per-element bounds check on
+        // this `n × dim` hydration loop.
+        plane.chunks_exact(2).map(|w| w[1]).collect()
+    } else {
+        Vec::new()
+    };
     let scorer = Sq16Scorer::from_codes(plane, dim, n);
     Some(HnswIndex {
         scorer,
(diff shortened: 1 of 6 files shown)


src/supertable/slow_vector_state.rs
```diff
@@ -596,7 +597,14 @@ pub(crate) async fn fetch_graph_sections(
     let raw = fetch_graph_section(storage, reference).await?;
     let bundle = hnsw::decode_graph_bundle(&raw)
         .ok_or_else(|| SlowVectorStateError::Parse("graph bundle frame".into()))?;
-    let data = bundle.data_bundle.as_deref().and_then(hnsw::decode_hnsw);
+    // Build the resident SQ8 walk plane only when SQ8-walk serving is enabled
```
savannahar68, line 600: this function is also called from drain/incremental append path which doesn't require sq8_plane. So we are wasting compute here in background path.

src/superfile/vector/hnsw.rs
```diff
@@ -1259,9 +1277,106 @@ pub(crate) fn decode_hnsw(bytes: &[u8]) -> Option<HnswIndex> {
...
+            .map(|&c| (c as i16 - 128) as i8)
+            .collect();
+        let baseline = 128 * q.iter().map(|&x| x as i32).sum::<i32>();
+        Sq8Query { q, baseline }
+    }
+    #[inline]
+    fn score(&self, q: &Sq8Query, node: u32) -> f32 {
+        // Centered dot: raw `Σ code·q` minus the `128·Σq` baseline — see
+        // `Sq8Query::baseline` for why the centering is needed for f32 exactness.
+        -((sq8_walk_dot(self.row(node), &q.q) - q.baseline) as f32)
```
savannahar68, line 1344: so here we are casting to fp32
for fp32 we can only hold upto 2^24 whole numbers after which the precision starts to degrade. 

in our tests we have tried vectors of 768 dimensions which is within 2^24 limits
but for higher dims which we support upto 4096 the casting won't be precise

768 dimensions:  768 × 16,256 = 12,484,608 (less than 16mill)
4096 dimensions: 4096 × 16,256 = 66,584,576 (3-4x more than 16mil)

this might lead to loss in recall

savannahar68 (approved):
lgtm
i've posted 2 comments. Not sure if those needs to be tested / addressed immediately or create a gh issue and pick it later when we do scale testing

src/supertable/slow_vector_state.rs
muralikpbhat: Good catch. The follow-up mmap change (#636) already makes the SQ8 plane a zero-copy bundle slice rather than a derived plane, so the expensive derivation goes away — and I'll add a serving-only gate there so the drain/incremental-append hydration builds no SQ8 at all. Will land in the #636 PR.

src/superfile/vector/hnsw.rs
muralikpbhat: Right. The score is centered (Σ(code−128)·q), which keeps it exact in f32 up to ~1032 dims — so 768 is fine — but 4096 overflows 2^24 as you note. Filed #643 to measure SQ8-walk recall at high dims and switch the beam comparison to integer-precise (i64/scaled) if it degrades; the Sq16 refine should recover the ranking, but that's unverified at 4096.

PR #634, opened 2026-08-24
catalog/remote: carry the FTS analyzer in create_table requests

vinaykakade:
## Problem

The remote transport encoded an `IndexSpec`'s FTS indexes as bare column names
(`"fts": ["body"]`), using only `fts_columns()` and never `fts_analyzers()`. So
a non-default analyzer chosen at create time —
`IndexSpec::new().fts_with_analyzer("body", "standard")` — was dropped on the
wire, and the server built the column with the default `ascii_lower` analyzer.
Non-ASCII query terms against a remote table then returned no rows, while the
same table over a local connection (which passes the `IndexSpec` straight to
the engine) honored the analyzer.

## Fix

Encode each FTS entry as a `{column, analyzer}` object when it uses a
non-default analyzer, and keep the bare column-name form for the default
(`ascii_lower`):

```json
"fts": ["title", {"column": "body", "analyzer": "standard"}]
```

Keeping the bare form for the default means a default-analyzer request is
byte-identical to before, so a server that predates the object form is
unaffected for the common case; only a request that actually asks for a
non-default analyzer uses the new shape.

## Testing

- `index_spec_json_carries_a_non_default_analyzer` — a `standard` column
  serializes as `{"column","analyzer"}` while a sibling default column stays a
  bare string.
- `index_spec_json_bare_form_for_the_default_analyzer` — an explicit
  `ascii_lower` serializes identically to a bare column name.
- The existing `index_spec_json_shape` (default analyzer → `["body"]`) is
  unchanged.
- `cargo fmt --check` clean; `cargo test --features remote catalog::remote::wire`
  green.

### DIFF
--- a/src/catalog/index_spec.rs
+++ b/src/catalog/index_spec.rs
@@ -68,9 +68,9 @@ impl IndexSpec {
 
     /// Mark `column` as full-text (BM25) indexed with a named analyzer
     /// (`"ascii_lower"` or `"standard"` — the Unicode-aware UAX #29
-    /// tokenizer that keeps non-ASCII text). All FTS columns in a table
-    /// must use the same analyzer; a table mixing analyzers is rejected
-    /// at [`create_table`](crate::Connection::create_table).
+    /// tokenizer that keeps non-ASCII text). The analyzer is per column:
+    /// each FTS column is tokenized with its own, so columns in one table
+    /// may use different analyzers.
     pub fn fts_with_analyzer(
         mut self,
         column: impl Into<String>,
--- a/src/catalog/remote/wire.rs
+++ b/src/catalog/remote/wire.rs
@@ -22,8 +22,7 @@ use arrow_schema::Schema;
 use base64::{Engine, engine::general_purpose::STANDARD as BASE64};
 use serde_json::{Value, json};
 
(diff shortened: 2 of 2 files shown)


pratyushlokhande (approved):

pratyushlokhande:
nit: the doc on `fts_with_analyzer` (`src/catalog/index_spec.rs:72`) still says "All FTS columns in a table must use the same analyzer; a table mixing analyzers is rejected at `create_table`". Not enforced — `table_tokenizers` (`src/catalog/mod.rs:911`) resolves one tokenizer per column, which is why this PR needs a per-entry `{column, analyzer}` on the wire. Worth dropping those two sentences.

vinaykakade:
> nit: the doc on `fts_with_analyzer` (`src/catalog/index_spec.rs:72`) still says "All FTS columns in a table must use the same analyzer; a table mixing analyzers is rejected at `create_table`". Not enforced — `table_tokenizers` (`src/catalog/mod.rs:911`) resolves one tokenizer per column, which is why this PR needs a per-entry `{column, analyzer}` on the wire. Worth dropping those two sentences.

Thanks, done

PR #638, opened 2026-08-24
vector: serve the stamped width and depth laws on the filtered path

ekechinwokah:
Filtered vector search consulted none of the laws the drain stamps. They were gated behind `!filtered`, so an allow-set query fell back to fixed constants -- sweep every cell (FILTERED_HIDDEN_CELL_NPROBE = 256) at a fixed depth of 16 fine runs -- while the manifest carried a measured width of ~21 and a measured fine depth of ~62 for the same k.

Wide-and-shallow is the wrong trade for an allow-set. Only a fraction of what a probed run yields is eligible, so the matching neighbours sit deeper in each cell's fine ranking rather than in farther cells, and the deficit grows with cell size. That is the measured shape: post-drain filtered recall@10 on real Cohere was 0.793 at 200K and 0.630 at 9.4M, against 0.995 unfiltered on the same table with the same queries.

Serve the stamped width and fine-depth laws on this path. The depth law is consumed as a floor (`max`), so it can only deepen a read.

Measured, 9.4M Cohere, post-drain, ~10% selectivity:

  recall@10  0.630 -> 0.993
  p50        4.01 ms -> 12.31 ms

Two laws stay gated off, both deliberately:

- The rerank law, because its measured budget (1066 rows at k=10) is SMALLER than the k x rerank_mult constant it would replace, and an allow-set needs that shortlist widened rather than narrowed.
- The #515 near-tie serve window, which #525 scoped out of filtered search on purpose. Measured here for completeness: admitting it as well moves recall 0.993 -> 0.998 and p50 12.31 -> 16.95 ms. Whether that trade is worth reopening the exclusion is a separate decision and is not taken here.

Known and unexplained: this serves ~21 cells read in full (~534K rows) where the constants swept ~256 cells at 16 of ~36 fine runs (~2.9M rows) -- five times less data for three times the latency. The cost is therefore not the volume read, and the mechanism behind it is worth finding on its own; it is not specific to filtered search.

Cell and fine selection remain allow-blind -- routing ranks cells by vector proximity and filters rows only once inside a probed run -- so a selectivity-aware admit is still the deeper fix. `|allow|` is already on hand before routing for whoever writes it.

### DIFF
--- a/src/supertable/query/vector.rs
+++ b/src/supertable/query/vector.rs
@@ -2878,20 +2878,38 @@ impl SupertableReader {
             // exactly where a flat config floor was measured scale-fragile:
             // 10M post-drain 0.982 at floor 4 vs 0.996 at 8, identical
             // latency).
+            // The stamped fine-depth law applies to FILTERED queries too.
+            // It is consumed as a floor (`max`), so it can only deepen a
+            // read, never narrow one — and an allow-set needs at least the
+            // unfiltered depth, not less: only a fraction of what a cell
+            // yields is eligible, so the matching neighbours sit deeper in
+            // the fine ranking than the unfiltered top-k ever has to reach.
+            // Filtered was pinned to the fixed FILTERED_HIDDEN_FINE_NPROBE
+            // floor while the drain's own measurement sat unused, which is
+            // the shape of the loss on real corpora (measured post-drain on
+            // Cohere: 0.793 at 200K falling to 0.630 at 9.4M as cells grow,
+            // against 0.997 unfiltered on the same table).
             if hidden_vector_index
-                && !filtered
                 && let Some(fine) = hidden_routing.and_then(|r| r.fine_for_k_at(k))
             {
                 cell_routing.fine_nprobe = cell_routing.fine_nprobe.max(fine);
             }
-            let law_width: Option<usize> =
-                if hidden_vector_index && !filtered && options.nprobe.is_none() {
-                    hidden_routing
-                        .and_then(|r| r.width_for_k_at(k))
-                        .filter(|w| *w > LAW_WIDTH_WITHIN_DEFAULT)
-                } else {
-                    None
-                };
+            // The stamped width law serves FILTERED queries too. Sweeping
+            // every cell (`FILTERED_HIDDEN_CELL_NPROBE`) was a fallback for
+            // not consulting it: wide-and-shallow costs more reads than the
+            // law's cell set AND misses, because an allow-set's eligible
+            // neighbours sit deeper in the fine ranking rather than in
+            // farther cells. With the fine-depth law above now applying,
+            // this serves filtered in the same shape as the unfiltered
+            // default — law floor at whole-cell depth, extended by the
+            // near-tie window below.
+            let law_width: Option<usize> = if hidden_vector_index && options.nprobe.is_none() {
+                hidden_routing
+                    .and_then(|r| r.width_for_k_at(k))
+                    .filter(|w| *w > LAW_WIDTH_WITHIN_DEFAULT)
+            } else {
+                None
+            };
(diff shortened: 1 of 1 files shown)


muralikpbhat (approved):
Approving. Verified against the code:

- The fine-depth law is consumed as a **floor** (`max`) on the filtered path — it can only deepen a read, never narrow one, so the change is recall-safe by construction on that axis.
- The width law now serves filtered, while an explicit caller `nprobe` still bypasses (`options.nprobe.is_none()` retained).
- The two deliberate exclusions actually hold: `law_default` still requires `!filtered`, so the #515 near-tie serve window and the rerank-law budget genuinely stay off for filtered — respecting #525's scoping, with the trade measured and deferred rather than silently taken. That discipline (and the honestly-flagged 5x-less-data / 3x-latency anomaly, plus the allow-blind-routing note with `|allow|` already on hand for the deeper fix) is exactly how these should be written.

The measured result speaks for itself: filtered recall@10 on Cohere-9.4M 0.630 -> 0.993 at ~10% selectivity.

**One fix before merge — a misleading in-code comment:** the new comment on the width-law block says the filtered path is served "law floor at whole-cell depth, **extended by the near-tie window below**" — but the window is gated off for filtered (`law_default = !filtered && ...`), as this PR's own description states. The next reader of that comment will believe the window applies to filtered. Please drop or correct that clause.

(The `bench (superfile, sf)` red is unrelated — the BMW-vs-brute-force BM25 divergence panic inherited from main.)

ekechinwokah:
comment fixed on the branch, and i have a phase breakdown of the latency anomaly that landed after the body was written.

**the misleading clause is gone.** you're right that it would have taught the next reader the opposite of what the code does. the block now reads:

```
// With the fine-depth law above now applying, filtered reads the
// law's cell set at whole-cell depth. It stops there: the near-tie
// serve window below is gated on `law_default`, which keeps its
// `!filtered` condition, so no extension cells are added for filtered.
```

naming `law_default` rather than describing the behaviour, so the comment breaks visibly if that gate is ever relaxed instead of quietly going false again.

**the 3x latency isn't more data — it's the fan-out critical path.** phase instrumentation (`INFINO_TRACE_VECTOR_WARM_PHASES=1`), cohere-9.4m, same box, filtered at ~10% selectivity, laws off vs on:

| phase | laws off | laws on |
|---|---|---|
| `vec.admit` | 2745 µs | 1168 µs |
| `vec.survivor_fetch` | 89 µs | 185 µs |
| `vec.rerank` | 439 µs | 862 µs |
| `vec.fanout_wall` | 1318 µs | 10405 µs |
| recall@10 / p50 | 0.605 / 3.99 ms | 0.993 / 11.47 ms |

admit gets *cheaper* — it's looking at a law-sized cell set instead of sweeping the grid, which is the point of the change — and the three phases that actually touch data move by under 2 ms combined. the fan-out wall moves 9 ms. so the cost is in orchestrating the per-cell reads, not scoring more of them: with fewer, larger units of work the wall converges on the slowest single cell instead of amortising across a wide sweep.

that's a scheduling shape to fix on its own terms, and it doesn't argue against serving the laws — 0.630 -> 0.993 isn't a trade i'd give back for 8 ms. not chasing it in this pr; it wants its own change to the fan-out, and it's worth doing because the same shape shows up wherever a pin narrows the cell set.

**on the two exclusions staying off:** deliberate, and both still measurable from here. the rerank-law budget and the #515 near-tie window are the next two to try on filtered, each with its own measurement, rather than folded in behind a recall win that already lands.

agreed on `bench (superfile, sf)` — same BMW-vs-brute divergence, present with none of these changes.

gates: 64 supertable vector unit tests, clippy clean, fmt clean; filtered recall@10 measured 0.630 -> 0.993 on cohere-9.4m and 0.793 -> 1.000 at 200k.

Issue #636, opened 2026-08-24
vector: mmap-serve the resident HNSW graph (zero-copy, shared across processes)


muralikpbhat:
Builds on #629 (the SQ8 walk + Sq16 refine adds a second resident plane).

## Problem

`hnsw_ivf` serving (#603) decodes the persisted graph blob into **private per-process heap** on first query — the Sq16 plane, the SQ8 walk plane (#629), the fixed-stride adjacency, and the doc-id map all become owned heap structures. That has three costs:

1. **Memory scales with reader count.** Every process/handle that opens the table loads its *own* full copy — so N concurrent readers cost N × the graph (multi-GB each). A single serving process with a thread pool is fine (one copy, shared across threads), but any multi-process topology multiplies RAM and OOMs on a memory-constrained host.
2. **The RAM is pinned, not reclaimable.** The decoded graph lives on the heap for the life of the handle; it can't be evicted under memory pressure the way a block cache can.
3. **Open pays a multi-GB decode.** First query hydrates the whole graph before it can serve.

## Proposal

Serve the graph structures **directly from an mmap of the persisted blob, zero-copy**, instead of decoding into owned heap. The on-disk layout is already mmap-friendly: the Sq16 and SQ8 planes are contiguous byte arrays, and the adjacency is fixed-stride `[u32; M0] × N` — both are sliceable in place from the mapped region, and the shipped local-read path is already mmap-based.

Payoffs:
- **One physical copy shared across processes** via the OS page cache — memory is bounded regardless of how many readers open the table (each just maps the same file).
- **Reclaimable memory** — page-cache pages evict under pressure and re-fault from NVMe, instead of pinning heap for the handle's lifetime.
- **Fast open** — a ~seconds mmap instead of a multi-GB decode-on-first-query.

## Perf note (needs measurement)

Warm mmap reads are ordinary RAM loads (page-cache pages are physical RAM), so warm serving should be heap-class. The HNSW walk is random-access pointer-chasing, which can pay a TLB penalty under 4 KB pages — mitigable with `madvise(MADV_HUGEPAGE)` / `MADV_RANDOM` + a warmup pass. The exact mmap-vs-heap warm-latency gap and the multi-reader scaling curve should be measured before this replaces the heap path (or it can land as an opt-in serving mode with heap as the default).

## Relation
- Builds on #629 (the SQ8/Sq16 planes this would mmap).
- Mechanism for #632's two default-on blockers — reclaimable graph memory, and shared serving across processes.
[...]

Issue #632, opened 2026-08-24
vector: productionize hnsw_ivf serving toward default-on


muralikpbhat:
`search_mode = hnsw_ivf` shipped in #603 as an opt-in mode (default stays `ivf`). It serves from a resident, self-calibrated HNSW graph over the Sq16 codes. This tracks the work to make it production-grade and eventually the default.

## Blockers to making it the default

1. **Reclaimable graph cache.** The resident graph decodes into a heap structure built on first query and pinned for the life of the table handle — never evicted (~20 GiB at 10M rows). The `ivf` block cache, by contrast, is reclaimable. So `ivf` is the safe default not because the graph costs more RAM under load, but because its RAM is pinned where `ivf`'s is reclaimable. Treat the resident graph as a background-hydrated, evictable cache entry over the always-correct scan fallback (graph absent/evicted ⇒ fall back to the scan arm), so its marginal RAM is reclaimable under pressure.

2. **Multi-node graph refresh.** The graph is a per-process resident artifact keyed by manifest generation. In a multi-process / multi-node deployment, a graph rebuilt at one writer's commit must refresh on the other serving processes to the newly committed generation — readers pick up the new graph blob on a generation change and evict the stale one. Default-on needs this refresh path.

## Data-graph maintenance optimizations (core done at 1M; optimize for scale)

Incremental persist is fused into the drain (copy → insert → PUT → flip at the generation commit, MVCC/crash-safe), with interim shortcuts to replace:

- **Clone the resident graph at the settle instead of re-fetching + decoding the prior blob.** The prior-generation graph is already resident (co-located cache); clone it, fetch only on a cold miss. Removes the read half of the ∝-corpus settle I/O.
- **Capture `(Sq16, id)` in the drain's encode loop.** The drain already encodes each new row's Sq16 and assigns its stable id; capture there instead of re-reading delta rows above the high-water id.
- **Bound insert RAM at scale.** Streaming / mmap'd-disk insert (page in only the working set the walk touches), or a disk-resident merge-based build, so construction RAM is the working set rather than the whole graph.
- **Deletes.** Currently a full-rebuild fallback (pure-append guard); add tombstone-skip-at-query + purge-at-periodic-rebuild, and guard the delete-all-rows case (it leaves a stale graph ref until the next drain).

## Related
- #603 — shipped the mode.
- #630 — centroid-graph router persisted at drain (the `> N` scale path).
- #590 — the Cohere-1M latency target this mode serves.

[...]

PR #644, opened 2026-08-25
superfile/vector: mmap the graph bundle; SQ8 as a bundle section

muralikpbhat:
Implements #636. Stacks on #629 (merged).

## What

For `search_mode = hnsw_ivf`, the resident graph is now served from an **mmap of the graph bundle** instead of being decoded into private per-process heap:

- **mmap is the default** (no config flag). Local backends map the bundle file directly (via the disk cache's `mmap_readonly_bytes`); object-store backends keep the existing striped fetch; heap fallback on any mmap I/O error. The Sq16 plane and doc-ids are **zero-copy slices** of the one mapping, and the resident index holds the mapping alive.
- **The SQ8 walk plane becomes a persisted data-bundle section**, not a per-open derivation — data-bundle magic **`INFDDG02` → `INFDDG03`**. On `INFDDG03` the SQ8 plane is a zero-copy slice of the same mapping; on older `INFDDG02` bundles it is derived on read (the prior behavior) — **backward-compatible, no forced rebuild**.
- Hydration (hash-verify + decode over the mapping) runs on the **blocking pool**, off the tokio workers (per the runtime concurrency contract).
- 2nd commit (review follow-up): the **drain / incremental-append** maintenance path hydrates the graph but never walks SQ8, so it now passes `need_sq8 = false` and **skips the SQ8 plane entirely** — no wasted derivation/slice in the background path. (Serving-neutral: serving passes `true`, i.e. the prior behavior.)

## Why

The resident graph decoded into **private per-process heap** — so every process/reader that opened a table loaded its own full copy (multi-GB), which OOMs on memory-constrained or multi-process serving. Mapping the bundle shares **one physical copy across all readers via the page cache** (reclaimable, not pinned heap), and open becomes a `~seconds` map instead of a multi-GB decode. And even single-process it is not a trade-off — warm serving is marginally *faster* than the heap decode (see below), so there is no latency cost to pay for the memory win.

This also lays the groundwork for two follow-ups tracked in #632:
- **Multi-node / multi-replica serving** — several serving processes or replicas on one host share a single physical bundle copy via the page cache, instead of N× the graph in private heap.
- **Unpinning the resident graph from RAM** — page-cache pages are reclaimable under memory pressure (they re-fault from NVMe on demand), unlike the decoded graph which is pinned on the heap for the handle's lifetime. That is exactly the reclaimable-graph-cache goal, and mapping the bundle is the enabler.

## Verification (1M Cohere, 768-dim, az2 x86_64)

**Correctness — recall bit-identical to the pre-mmap (heap) path**: same result-id `sha256` (`e43090b2…`) on a shared `INFDDG02` table. Rebuilt `INFDDG03` table serves healthy recall:

| metric | value |
| ------ | ----- |
| recall@10 | 0.9928 |
| recall@100 | 0.9765 |

**Throughput — mmap unlocks concurrent serving the heap path couldn't reach.** With a private per-process graph copy, concurrent multi-process serving OOM'd before it got useful — it hung/crashed past a handful of readers (heap dies at N≈6, see the memory table). The shared mapping removes that wall, so concurrent throughput scales and holds through N=40 (direct probe, `hnsw_ivf`, 8 cores):

| readers (N) | 8 | 16 | 24 | 40 |
| ----------- | -- | -- | -- | -- |
| heap | ❌ OOM | ❌ OOM | ❌ OOM | ❌ OOM |
| mmap QPS | 6102 | 6209 | 6360 | 6217 |

So this goes from *"single/low-concurrency only, crashes when scaled"* to **sustained ~6.2K QPS across the concurrency an 8-core host saturates at (~N=16)**. (The section-in-bundle layout itself is perf-neutral — QPS matches a plain mmap of the planes; it's the shared mapping that buys the scaling.)

**Warm latency, single process (N=1, 3000 warmed queries, K=100)** — mmap is not just heap-class, it is **consistently a touch faster** across every percentile (a warm mmap read is a plain RAM load, and the single shared copy has tighter cache residency than a private heap copy):

| percentile | heap | mmap |
| ---------- | ---- | ---- |
| p50 | 0.620 ms | 0.590 ms |
| p95 | 0.775 ms | 0.738 ms |
| p99 | 0.873 ms | 0.837 ms |

<sub>(heap-vs-mmap A/B captured via a development-time toggle; the shipped build maps unconditionally on local backends.)</sub>

**Multi-process memory (8-core / 32 GB budget, pure multi-process ladder)** — mmap shares one physical bundle copy, so many readers fit where the heap path OOMs:

| readers (N) | heap | mmap (#636) |
| ----------- | ---- | ----------- |
| 4 | ✅ | ✅ |
| 6 | ✅ (~25 GB) | ✅ |
| 8 | ❌ OOM | ✅ |
| 40 | ❌ | ✅ (~20 GB peak) |

Max readers survived: **heap = 6, mmap = 40** (~6.7×), ~0.5 GB/process incremental (the ~2.8 GB Sq16+SQ8 planes are mapped once, `smaps` `Pss_File` ≈ shared/N).

**Backward-compat**: the new build serving a pre-existing `INFDDG02` table returns identical results via derive-on-read. **`INFDDG03` bundle** produced by a drop-old rebuild (no temp files).

## Notes

- On-disk **format bump `INFDDG02` → `INFDDG03`** — additive, backward-compatible via derive-on-read; no forced rebuild.
- The HNSW **adjacency stays decoded in heap** (small — ~128 MB — and not the bottleneck at the concurrency an 8-core host saturates at); flattening it is deliberately out of scope.
- `cargo fmt` / `clippy --all-targets` / `check -D warnings` clean; 305 vector + 9 slow_vector_state + 67 writer lib tests pass, incl. new `INFDDG03` round-trip + `INFDDG02` backward-compat + heap-path-doesn't-alias tests.

### DIFF
--- a/src/supertable/slow_vector_state.rs
+++ b/src/supertable/slow_vector_state.rs
@@ -35,11 +35,14 @@ use crate::{
     config,
     storage::{StorageError, StorageProvider},
     superfile::vector::hnsw,
-    supertable::manifest::{
-        SuperfileEntry, VectorSummary,
-        encoding::SummaryWireMode,
-        list::RoutingRef,
-        part::{self, ContentHash, ManifestPart, PartId},
+    supertable::{
+        manifest::{
+            SuperfileEntry, VectorSummary,
+            encoding::SummaryWireMode,
+            list::RoutingRef,
+            part::{self, ContentHash, ManifestPart, PartId},
+        },
+        reader_cache::disk,
     },
 };
 
@@ -554,20 +557,67 @@ pub(crate) async fn write_graph_section(
 }
 
 /// Fetch a graph section written by [`write_graph_section`], verifying its
-/// bytes hash to `reference.content_hash`. Returns the raw section bytes
-/// (the caller decodes them with `hnsw::decode_hnsw` /
-/// `Hnsw::from_bytes`). Striped like every other slow-state fetch. Callers
-/// fall back to the lazy build / scan path on any error — a bad or missing
-/// graph blob must never fail an open or a query.
+/// bytes hash to `reference.content_hash`. Returns the verified section bytes as
+/// a [`Bytes`] handed straight to the decoder (no `to_vec` copy) plus whether
+/// the backing is an `mmap` (the caller decodes with `hnsw::decode_hnsw` /
+/// `Hnsw::from_bytes`).
+///
+/// A local bundle file is loaded as a zero-copy `mmap`-backed `Bytes` (one
+/// faulted read for the hash check, then the mapping is what the resident
+/// graph holds), so the multi-GiB open-time transient never touches heap and N
+/// processes share one physical page-cache copy — the default serving path.
+/// Every non-local backend (`local_path` is `None`, i.e. S3/Azure/GCS) keeps
+/// the striped heap fetch, so remote correctness is unchanged. Callers fall
+/// back to the lazy build / scan path on any error — a bad or missing graph
+/// blob must never fail an open or a query.
+///
+/// The hash verification faults the whole (possibly multi-GiB, mmap-backed)
+/// bundle in synchronously — a CPU/IO wave that must not run on a tokio worker
+/// (the runtime anti-pattern). It runs on the blocking pool, mirroring
+/// [`load_full_state`].
 pub(crate) async fn fetch_graph_section(
     storage: &dyn StorageProvider,
     reference: &RoutingRef,
-) -> Result<Vec<u8>, SlowVectorStateError> {
-    let bytes = fetch_blob_striped(storage, &reference.uri, STRIPED_FETCH_CHUNK_BYTES).await?;
-    if ContentHash::of(bytes.as_ref()) != reference.content_hash {
-        return Err(SlowVectorStateError::HashMismatch);
+) -> Result<(Bytes, bool), SlowVectorStateError> {
+    let (bytes, is_mmap) = fetch_graph_bytes(storage, reference).await?;
+    let expected = reference.content_hash;
+    let bytes = spawn_blocking(move || {
+        if ContentHash::of(bytes.as_ref()) != expected {
+            return Err(SlowVectorStateError::HashMismatch);
+        }
+        Ok(bytes)
+    })
+    .await
+    .map_err(|join_error| {
+        SlowVectorStateError::Parse(format!("graph section hash task failed: {join_error}"))
+    })??;
+    Ok((bytes, is_mmap))
+}
+
+/// The bundle bytes for [`fetch_graph_section`] plus whether the backing is an
+/// `mmap`: `mmap`-backed for a local file (the default), else striped from
+/// storage into a heap buffer for remote backends. A local mmap failure is
+/// non-fatal — it falls through to the striped fetch so an open never fails on
+/// it. The mmap acquisition itself is cheap (no fault); only the hash + decode
+/// over the map are CPU/IO-heavy, and those run on the blocking pool.
+async fn fetch_graph_bytes(
+    storage: &dyn StorageProvider,
+    reference: &RoutingRef,
+) -> Result<(Bytes, bool), SlowVectorStateError> {
+    if let Some(path) = storage.local_path(&reference.uri)
+        && path.exists()
+    {
+        match disk::mmap_readonly_bytes(&path) {
+            Ok(bytes) => return Ok((bytes, true)),
+            Err(error) => tracing::warn!(
+                uri = reference.uri,
+                %error,
+                "graph bundle mmap failed; falling back to striped heap fetch"
+            ),
+        }
     }
-    Ok(bytes.to_vec())
+    let bytes = fetch_blob_striped(storage, &reference.uri, STRIPED_FETCH_CHUNK_BYTES).await?;
+    Ok((bytes, false))
 }
 
 /// The graph sections of one published generation, decoded resident. Held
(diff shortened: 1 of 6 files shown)


src/supertable/slow_vector_state.rs
```diff
@@ -554,20 +557,67 @@ pub(crate) async fn write_graph_section(
...
+/// it. The mmap acquisition itself is cheap (no fault); only the hash + decode
+/// over the map are CPU/IO-heavy, and those run on the blocking pool.
+async fn fetch_graph_bytes(
+    storage: &dyn StorageProvider,
+    reference: &RoutingRef,
+) -> Result<(Bytes, bool), SlowVectorStateError> {
+    if let Some(path) = storage.local_path(&reference.uri)
+        && path.exists()
+    {
+        match disk::mmap_readonly_bytes(&path) {
```
savannahar68: instead of falling back its better to actually catch mmap error and based on that take action
eg: if its setrlimit or ENOMEM then we actually need to return back error to the user

silently falling back might be risky resource consumption wise

src/supertable/slow_vector_state.rs
```diff
@@ -554,20 +557,67 @@ pub(crate) async fn write_graph_section(
...
     reference: &RoutingRef,
-) -> Result<Vec<u8>, SlowVectorStateError> {
-    let bytes = fetch_blob_striped(storage, &reference.uri, STRIPED_FETCH_CHUNK_BYTES).await?;
-    if ContentHash::of(bytes.as_ref()) != reference.content_hash {
-        return Err(SlowVectorStateError::HashMismatch);
+) -> Result<(Bytes, bool), SlowVectorStateError> {
+    let (bytes, is_mmap) = fetch_graph_bytes(storage, reference).await?;
+    let expected = reference.content_hash;
+    let bytes = spawn_blocking(move || {
+        if ContentHash::of(bytes.as_ref()) != expected {
```
savannahar68, line 585: if we mmap the file then the below line
`ContentHash::of(bytes.as_ref())`

will load everything in memory right?

savannahar68: so across readers / process this helps, but i think i read somewhere that the graph is only loaded on query is not true

src/supertable/slow_vector_state.rs
```diff
@@ -554,20 +557,67 @@ pub(crate) async fn write_graph_section(
...
+/// [`load_full_state`].
 pub(crate) async fn fetch_graph_section(
     storage: &dyn StorageProvider,
     reference: &RoutingRef,
-) -> Result<Vec<u8>, SlowVectorStateError> {
-    let bytes = fetch_blob_striped(storage, &reference.uri, STRIPED_FETCH_CHUNK_BYTES).await?;
-    if ContentHash::of(bytes.as_ref()) != reference.content_hash {
-        return Err(SlowVectorStateError::HashMismatch);
+) -> Result<(Bytes, bool), SlowVectorStateError> {
+    let (bytes, is_mmap) = fetch_graph_bytes(storage, reference).await?;
```
savannahar68, line 582: so fetch_graph_bytes i touching files / doing IO
Should we make this entire function within spawn_blocking to not block the worker threads ?

savannahar68 (approved):
lgtm
some comments please check

src/supertable/slow_vector_state.rs
muralikpbhat: Good call — done in e6b994a. A resource-exhaustion mmap failure (ENOMEM: address space / vm.max_map_count / an mmap rlimit) now returns Err instead of falling back, because the striped fallback would pull the whole bundle into *heap* — strictly more memory than the mapping that just failed, i.e. exactly the wrong move under memory pressure. The caller then degrades safely: the serving path falls back to the ivf scan (the query still succeeds), and the drain path to a full rebuild. Benign per-file failures (missing/moved file, permissions) still fall through to the striped fetch so an open never fails on them.

src/supertable/slow_vector_state.rs
muralikpbhat: Yes — the hash pass walks every byte, so it does fault the whole bundle in at hydration. The key distinction is *where*: those pages land in the OS page cache (shared across all readers/processes, reclaimable under pressure), not in the private per-process heap the old decode path allocated. So it is one sequential faulting read into shared, reclaimable memory instead of N× multi-GiB private copies — that is the memory win, not a claim that we only touch the pages a query needs.

muralikpbhat: Right, thanks for catching that — "only loaded on query" would be inaccurate. Hydration is lazy until the first serve, but at that point the hash-verify + decode fault the entire bundle in; it is not per-query, only-what-is-needed paging. The two real wins are (1) shared across readers/processes via the page cache and (2) reclaimable — pages re-fault from NVMe under pressure — versus the pinned heap decode. I have kept the doc-comment wording aligned with that so it does not imply lazy-per-query loading.

src/supertable/slow_vector_state.rs
muralikpbhat: Done in e6b994a. The heavy wave (hash-fault + decode) was already on the blocking pool; I have now also moved the exists()/mmap syscalls in fetch_graph_bytes onto spawn_blocking, so no local-open I/O sits on a tokio worker. The remote branch (fetch_blob_striped) stays as-is since it is already async and yields — so the whole function is now either blocking-pool or async, nothing blocking on a worker thread.

PR #646, opened 2026-08-25
feat reader cache: open cache files on first read, not at table open

### LINKED ISSUE #596: cache: opening a table pays for every cached file, not the ones queries touch
## Problem

`DiskCacheStore::new` eagerly restores the whole cache directory: every file gets mmapped and its footer parsed before the first query runs. That is what makes cross-process cache reuse work at all, and it is fine while the directory is roughly the live set, but per #595 the directory also accumulates a lot of dead files, and every restart pays for them.

As observed on a c7a.metal-48xl, the first cold query took 15.9s against a cache dir of ~9,600 files (~39 live) and 1.33s against a clean one, a 12x difference where the query itself did the same work.

## Possible Approach

One possible approach could be to index on startup and open on first touch, though this shifts the open cost onto the first query that touches each file. Another way could be to restore only what the latest manifest references.

naman-modi:
### What does this PR do?
Opening a table no longer opens every file in the cache directory; each cache file is opened by the first read that needs it. Closes #596.

The open-time restore mmaps, parses, and checksum-verifies the whole cache directory, so its cost grows with the directory instead of with what queries touch. A large cache pays it on every restart, and a directory that has collected dead copies pays for those too. #596 has the observation: 15.9s for the first query against a churned directory versus 1.33s against a clean one, for identical query work.

### How do we do this change?
- At open, one pass only stats the directory to seed the byte budget; nothing is mmapped, parsed, or verified. Leftovers that can never be used are deleted here: orphaned block sidecars and stale tempfiles (a fresh tempfile belongs to another process's in-flight fetch, so only ones old enough to be a crashed fetch's leftover are reclaimed).
- Each cold fetch first looks for the file on local disk and opens just that one; unusable files are deleted and the read falls back to the normal fetch. Since this checks the filesystem and not an open-time snapshot, files another process writes after open are now found too (before, those were re-fetched from storage for the life of the process).
- Eviction reclaims the never-opened files first: nothing holds them, so removing one is just an unlink. Without this, a directory already at its budget would fail the first read outright.

### Performance
- End to end on a 100M-row table (38 superfiles of ~256 MB, 16 cores, page cache dropped before every cold query): the cold sweep went from 131.6s to 81.3s (-38%). The opens now happen inside the scan, overlapping its parallel reads, instead of serially before the first query.
- Isolated open cost, fresh connection, first query with one live file among dead ones (local NVMe): 2,000 dead files 0.176s to 0.011s (16x), 8,000 dead 0.724s to 0.036s (20x). What remains is a stat per file instead of an open.
- Worst case for this change, a full scan needing every file over a clean small-file directory: 6.4ms to 7.0ms, within run noise.

### DIFF
--- a/src/supertable/reader_cache/disk.rs
+++ b/src/supertable/reader_cache/disk.rs
@@ -16,7 +16,7 @@ use std::{
         atomic::{AtomicBool, AtomicU64, Ordering},
     },
     thread,
-    time::{Duration, Instant},
+    time::{Duration, Instant, SystemTime, UNIX_EPOCH},
 };
 
 use async_trait::async_trait;
@@ -80,6 +80,11 @@ const STORE_UPGRADE_RETRY_INTERVAL: Duration = Duration::from_millis(10);
 /// Filename suffix for per-superfile sparse block-cache files.
 const BLOCKS_FILE_SUFFIX: &str = ".blocks";
 
+/// How old an in-flight tempfile must be before the open-time scan reclaims it. A live cold
+/// fetch's tempfile is seconds old, so one this stale can only be a crashed fetch's leftover;
+/// deleting a live one would fail the owner's rename and cost it a retried fetch.
+const TMP_RECLAIM_AGE: Duration = Duration::from_secs(15 * 60);
+
 /// Process-global count of in-flight foreground queries. Used with
 /// [`foreground_notify`] so a fill's `select!` wakes promptly when a query
 /// begins and can re-check its per-URI pause condition; it is **not** a
@@ -115,6 +120,15 @@ impl Drop for ForegroundQueryGuard {
         // query releases same-URI readers.
         foreground_notify().notify_waiters();
     }
+}
+
+/// File mtime in microseconds since the unix epoch, 0 if the filesystem has none.
+fn file_mtime_us(meta: &fs::Metadata) -> u64 {
+    meta.modified()
+        .ok()
+        .and_then(|t| t.duration_since(UNIX_EPOCH).ok())
+        .map(|d| d.as_micros() as u64)
+        .unwrap_or(0)
 }
 
 /// Pause this URI's background full-object fill while a caller besides the
@@ -223,6 +237,8 @@ pub struct CacheStats {
     pub current_bytes: u64,
     pub budget_bytes: u64,
     pub n_cold_fetches: u64,
+    /// Reads served by adopting a cache file that was on disk but not yet indexed.
+    pub n_disk_reuses: u64,
     pub n_evictions: u64,
     /// Cumulative count of entries `madvise(MADV_DONTNEED)`'d
     /// by the idle-threshold sweep thread. Includes individual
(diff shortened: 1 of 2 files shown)


src/supertable/reader_cache/disk.rs
```diff
@@ -232,6 +248,14 @@ pub struct CacheStats {
     pub n_gc_drops: u64,
 }
 
+/// A cache file found on disk at open but not opened yet. Its bytes already count against the budget.
+#[derive(Debug, Clone, Copy)]
+struct UnindexedFile {
+    size_bytes: u64,
+    /// File mtime in microseconds since the unix epoch; eviction drops the oldest first.
+    mtime_us: u64,
```
Pranav2612000, line 258: Is `atime` better than `mtime`? Since superfiles don't change, i guess the `mtime` of a file will never change after its added to the cache?

src/supertable/reader_cache/disk.rs
```diff
@@ -1098,28 +1127,160 @@ impl DiskCacheStore {
...
                 continue;
             }
+
+            if SuperfileUri::from_cache_tmp_filename(name).is_some() {
+                // A fresh tempfile belongs to a sibling's in-flight fetch; only a stale one is a
+                // crashed fetch's leftover. See [`TMP_RECLAIM_AGE`].
+                let stale = entry
+                    .metadata()
+                    .ok()
+                    .and_then(|m| m.modified().ok())
```
Pranav2612000, line 1139: Should we look at the `modified()` time or the `accessed` time?

Pranav2612000: Although this could get tricky. A read to get this `accessed` time should not cause the `accessed` time to change

src/supertable/reader_cache/disk.rs
```diff
@@ -1768,6 +1950,14 @@ impl DiskCacheStore {
     /// is freed or no eligible victims remain (→
     /// `BudgetExceeded`).
     async fn evict_at_least(&self, bytes_needed: u64) -> Result<(), DiskCacheError> {
+        // Never-opened files go first; freeing one cannot hurt a live reader.
```
Pranav2612000, line 1962: May not be a part of this PR, and this may not be the best place for the check, but I guess deleted/inaccessible files should be the first ones to go. Thoughts?

Pranav2612000 (approved):

src/supertable/reader_cache/disk.rs
naman-modi: Right, `mtime` is the file creation time for us, i.e. when we fetch it from the object store. The issue is `atime` its tracking is not reliable; many prod systems mount `noatime` (for perf reasons), and even when it is enables, it is updated once in 24 hours. Also `atime`'s behaviour with mmap is tricky - so for now its better to have `mtime` itself ([ref](https://lwn.net/Articles/621046/); good read). However, as future work, we should persist last access times (and other CacheStats) across restarts ourselves, and use those for eviction.

src/supertable/reader_cache/disk.rs
naman-modi: Here we are checking if the temp file is being written to or stale, so better to have the modified time, as that would keep getting updated, and we can make a clear decision.

src/supertable/reader_cache/disk.rs
naman-modi: Agreed.

PR #649, opened 2026-08-25
fts: locate ranked-OR non-essentials by bit-test, not block expansion

vinaykakade:
## What

The ranked-OR MaxScore completion decoded a non-essential term's entire
128-doc posting block on **every candidate**, just to locate the candidate
doc and read its term frequency. On a dense (bitset-encoded) block — exactly
what common non-essential terms use — that expansion is a scalar set-bit loop
over 128 doc ids, and it dominated the non-essential completion cost.

This locates the candidate by a presence **bit-test** and reads its tf by
**popcount-rank** into the block's tf array (decoded once per block via the
new `decode_block_tfs`, reused across a run of candidates in the same block),
never expanding the doc ids. PACKED blocks keep the decode-and-locate path —
there is no rank shortcut when doc ids are delta-packed.

## Why it's correct

Top-k is unchanged: in a bitset block a doc's position in doc order equals its
popcount-rank, so the tf read is bit-for-bit the same value the old
`block_tfs[pos]` returned. The brute-force BM25 top-k oracle and the proptest
fuzz oracle pass unchanged, and exercise the new tf-only decode through the
ranked-OR path.

No on-disk format change — this is a read-path-only optimization.

## Performance

Internal A/B, this branch vs `main`, same host, both builds on this base
commit so the delta isolates this change. SBG-style metric (min across
repeated runs per query, then averaged), on a multi-term FTS query workload
over a large single-node index. `COUNT` is an unchanged-code drift control.

Improvement concentrates on multi-term ranked unions — the class that pays the
non-essential completion cost:

| query class      | TOP_10 | TOP_100 | TOP_1000 |
|------------------|-------:|--------:|---------:|
| 3-term union     | −7.0%  | −11.8%  | −5.2%    |
| 4+-term union    | −6.9%  |  −9.5%  | −7.9%    |

No query class regresses beyond the ±noise band that untouched paths
(intersection, phrase) show across commands. The one directional cost is
two-term unions at large k (TOP_1000, ~+3%): with only two terms and a large
candidate set the per-candidate popcount-rank isn't fully amortized. It sits
within run-to-run noise and is outweighed by the ≥3-term wins. Single-term,
phrase, intersection, and count workloads are unchanged within noise.

## Coverage

Existing brute-force + fuzz BM25 top-k oracles gate the scoring path and pass
unchanged (top-k identical). No new public surface; no new dependency.

### DIFF
--- a/src/superfile/fts/posting.rs
+++ b/src/superfile/fts/posting.rs
@@ -265,6 +265,32 @@ pub fn decode_block(bytes: &[u8], dest_doc_ids: &mut [u32], dest_tfs: &mut [u32]
     count
 }
 
+/// Decode only the tf array of a block (the trailing tf-packed bytes) into
+/// `dest_tfs`, in doc order, skipping the doc-id half. The ranked-OR
+/// membership probe locates a bitset-block doc by bit-test + popcount-rank
+/// and needs only that doc's tf, never the expanded doc ids — so it decodes
+/// the tfs (which are BitPacker4x-packed and can't be single-value-indexed)
+/// once per block and reads the rank-th one, avoiding the doc-id expansion.
+pub fn decode_block_tfs(bytes: &[u8], dest_tfs: &mut [u32]) {
+    assert!(
+        dest_tfs.len() >= BLOCK_LEN,
+        "decode_block_tfs: dest_tfs must have at least {BLOCK_LEN} slots"
+    );
+    let tf_bits = bytes[2];
+    assert!(tf_bits <= 32, "decode_block_tfs: tf_bits {tf_bits} > 32");
+    let tfs_size = BLOCK_LEN * tf_bits as usize / 8;
+    assert!(
+        bytes.len() >= HEADER_SIZE + tfs_size,
+        "decode_block_tfs: bytes shorter than header+tfs"
+    );
+    let tfs_start = bytes.len() - tfs_size;
+    BitPacker4x::new().decompress(
+        &bytes[tfs_start..tfs_start + tfs_size],
+        &mut dest_tfs[..BLOCK_LEN],
+        tf_bits,
+    );
+}
+
 /// Decode only the doc ids of a posting block, skipping the term-frequency
 /// half entirely. Unranked counts (union, intersection) never look at
 /// `tf` — they tally doc ids — so decoding + reading the packed tfs is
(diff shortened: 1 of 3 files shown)


src/superfile/fts/posting.rs
naman-modi, line 1: The rank->tf read is only checked by value on a single-word block. `exhaustive_and_bmm_agree_on_top_k` (scorers.rs:1721) is 20 docs, so every doc's bit is < 64, word_idx is always 0, and the `for w in presence[..word_idx*8]` loop never actually runs.

Can we add one deterministic test that forces word_idx >= 1 with a tf that differs, e.g.:
- common: present in ~half of 200 docs (dense -> bitset, block spans >64 docs), tf=1 everywhere except tf=3 at doc 100 and tf=2 at doc 165
- rare: present only at docs 100 and 165 (high idf -> it's the essential, common is the bitset non-essential probed by bitset_probe_tf)
- query "rare common", mode OR, k=10
- assert the SCORE of doc 100 and 165 by value vs the oracle

naman-modi (approved):
LGTM after comments.

vinaykakade:
@naman-modi good catch — you're right, `exhaustive_and_bmm_agree_on_top_k` is a 20-doc block so every bit is < 64, `word_idx` is always 0, and the cross-word popcount loop never ran.

Added `bitset_probe_tf_reads_correct_tf_across_presence_words` (in `scorers.rs`) that closes exactly this gap. I went with a **direct probe** rather than the end-to-end score-vs-oracle route, for a reason worth noting: with a 2-term `rare common` query, MaxScore keeps `common` **essential** (its `max_score` stays above the k-th threshold, since the common-only docs are what fill the tail of the top-k), so `bitset_probe_tf` isn't reliably reached — and if I lower `common`'s idf enough to make it non-essential, its tf stops affecting the score measurably. A direct probe deterministically exercises the multi-word rank and asserts the exact value the score is built from.

The plant:
- `common` present at doc 0 and docs 100..=165. First-doc 0 word-aligns the bitset base to 0, so **doc 100 → bit 100 (presence word 1)** and **doc 165 → bit 165 (presence word 2)**; the 0→100 gap widens the deltas enough to force the bitset encoding (asserted in the test, so it fails loudly if that ever changes).
- tf = 1 everywhere except **3 at doc 100** and **2 at doc 165**.
- `bitset_probe_tf(100)` must return `Some(3)` (rank counts all of word 0), `bitset_probe_tf(165)` must return `Some(2)` (rank sums words 0 and 1 → rank 66), plus `bitset_probe_tf(50)` → `None`.

Verified it has teeth: disabling the `for w in presence[..word_idx*8]` accumulation makes doc 100 read `Some(1)` (rank collapses to 0 → doc 0's tf) and the test fails.

src/superfile/fts/posting.rs
vinaykakade: thanks, done

Issue #633, opened 2026-08-24
vector: query-adaptive ef via a stamped k→ef calibration curve


muralikpbhat:
## Problem

`hnsw_ivf` calibration (#603) stamps a **single** `ef`, sized for the calibration recall-k (recall@10 today). A query asking for a larger `k` is then under-served by that stamp. Measured on Cohere-1M:

- A graph calibrated for k=100 (wide `ef`) **over-serves** k=10 — k=10 costs ~4.2 ms on the k=100 graph vs ~1.5–1.8 ms on its own narrower `ef`.
- The k=10-calibrated `ef` **under-serves** k=100 — recall@100 falls short until `ef` is widened (k=100 needs roughly 2× the `ef` that k=10 needs).

A single stamped `ef` cannot serve all `k`, and a linear `ef = C·k` does not work either: the measured minimal `ef` is **sublinear** in `k` (~2× `ef` for 10× `k`), so `C·k` under-serves small `k` and tanks recall.

## Proposal — stamp a compact k→ef curve, serve `ef` per requested `k`

**Build (near-free).** The calibrator already computes an exhaustive top-Kmax ground truth per probe query. Extend the `recall[m0][ef]` sweep to `recall[m0][ef][k]` across a few k-anchors (e.g. 1, 10, 50, 100) using **prefixes** of that same GT — no extra graph walks. For the chosen `m0`, stamp the minimal `ef` that clears the target at each k-anchor: a compact curve of a handful of `(k, ef)` pairs.

**Serve.** `ef = ef_for_k(requested_k)` (round the requested `k` up to the next anchor), replacing `ef = data.ef_search.max(k_fetch)` (too weak — the over-fetch width never reaches the `ef` a large `k` needs). `k` above the top anchor clamps to the `ef` ceiling (no measured-recall promise beyond the anchors).

**Backward compatible (required).** The curve is a **new optional bundle section behind a format-version bump**. Existing bundles carry a single stamped `ef` and no curve; decode treats that as a degenerate 1-point curve (`ef_for_k(k) = stamped_ef` for every `k`) — i.e. exactly today's behavior. No rebuild is forced; only tables rebuilt after this change gain per-`k` `ef`.

## Notes

- This is an **on-disk format change**, so per the format-change policy it gets an issue first (this one) before implementation.
[...]

PR #651, opened 2026-08-25
vector: a 4-bit walk plane as a first-class codec

ekechinwokah:
## Why

The codec ladder has a gap where the embedded profile needs one. Every shipping
rerank codec is 2 bytes/dim — `Sq16`, `Sq16Adaptive`, and both `Sq8*` variants,
whose residual leg makes them 2 B/dim despite the name — except `RabitqOnly` at
0.125, whose recall sits outside the bar. There is nothing at 1.0, 0.5, or 0.25.

That gap makes resident memory the binding constraint for single-box serving:
the graph is RAM-pinned, and at 10M × 768 the Sq16 plane alone is ~15 GiB, which
is why `hnsw_ivf` ships opt-in. This adds the 1.0 and 0.5 B/dim rungs, the
AVX-512 kernels to score them, and a bundle format that treats a walk plane as a
named part of the layout rather than a special case.

## What changed

**`v04` bundle.** The walk codec is a `u8` in the header and only that codec's
sections are written. `v03` always wrote the SQ8 plane, which is affordable at
1 byte/dim and stops being so once denser codecs exist — a bundle should not
carry a plane it never walks. The Sq16 plane stays unconditional because every
codec re-ranks its beam on it. `v03` and `v02` still decode: their layouts imply
what `v04` states.

**`Sq4Scorer` is `Plane`-backed** (owned or zero-copy `Bytes`), so a 4-bit
section serves straight out of the mapped bundle like the Sq16 and SQ8 planes.
Owning a `Vec` would copy the whole plane on every open and undo #644.

**One config knob.** `vector.hnsw_plane` is now `{sq16, sq8, sq4, sq4_residual}`,
absorbing `hnsw_sq8_walk`. Both answered the same question — what the walk scores
on — and could contradict each other. Default is `sq8`, so shipped behaviour is
unchanged.

**Unpadded rotation.** The transform padded `dim` to the next power of two, so at
1536 a third of every stored plane encoded zeros. For a graph walk that is wasted
residency; for a terminal scan it is paid twice, since per-query cost there is
bytes-read ÷ bandwidth. A block-decomposed form (1536 → 1024+512, 200 →
128+64+8) with a seeded permutation between stages keeps the transform an exact
isometry with no padding.

**Kernels.** A 512-bit nibble arm (32 coords/iteration against the AVX2 arm's 16)
and a `vpdpbusd` integer arm (64 integer MACs per register against 16 float
ones). The query is quantized to `i8` once per query and laid out in the order
nibbles fall out of a mask/shift unpack, so no per-row shuffle is needed:
reordering the query is free, reordering codes would cost a shuffle on every
stored row. Partial final blocks are masked rather than handled scalar, in both
the new nibble kernel and the existing SQ8 walk kernel.

**Calibration walks the serving plane and grades against Sq16.** Grading a walk
against an exhaustive scan of its *own* representation measures routing only: if
the plane misranks, the ground truth misranks identically, so codec error is
invisible and a plane too coarse to serve still calibrates as passing. Grading
against a Sq16 walk instead measures a search that never runs. Walking the
serving plane and grading against Sq16 measures what will actually serve, so a
plane below the table's `target_recall` de-registers itself to ivf through the
existing `None` → fallback. Probes are derived from Sq16 for the same reason —
through a lossy plane they would carry that plane's error.

**Ruler loaded over mean ± 2.7σ rather than min/max.** A min/max ruler is set by
the single most extreme value per coordinate, so most of the 16 levels cover
range almost no row occupies. On dbpedia-1536 this moved the bare plane's
recall@10 from 0.902 to 0.934 and recall@1 from 0.895 to 0.955. The residual
construction keeps min/max: with 8 effective bits a wide span is cheap while
clamping a coarse code is not — its residual nibble then refines an interval the
value does not lie in — and σ is unsafe on multimodal data, where it reflects the
spread between modes rather than within them.

## Measured

dbpedia-1536 and glove-200 at 100K rows, 200 held-out queries, one query per
call (not batch-amortized), graded against the exact oracle.
`test_helpers::sq4_flat` is the measurement seam — feature-gated, absent from
`public-api.txt`, driving the shipping encoder and kernels rather than a parallel
implementation.

Terminal ranking, which isolates codec error from routing error:

| dbpedia-1536 @100K | resident | B/vec | R@1 | R@10 | R@100 | p50 |
| --- | --- | --- | --- | --- | --- | --- |
| 4-bit | 73.25 MiB | 768 | 0.950 | 0.933 | 0.954 | 1.47 ms |
| 4-bit + residual | 146.50 MiB | 1536 | 0.995 | 0.991 | 0.992 | 3.52 ms |

| glove-200 @100K | resident | B/vec | R@1 | R@10 | R@100 | p50 |
| --- | --- | --- | --- | --- | --- | --- |
| 4-bit | 9.54 MiB | 100 | 0.795 | 0.865 | 0.885 | 0.178 ms |
| 4-bit + residual | 19.08 MiB | 200 | 0.965 | 0.982 | 0.985 | 0.311 ms |

Recall is reported per `k` because a coarse codec loses the tail of the
neighbourhood long before it loses the top-1: the bare 4-bit rung is usable at
k=1 on dbpedia (0.950) and not at k=10 (0.933), while the residual rung holds
across all three. A single `k` would hide that.

The SQ8 walk on glove-200 (`hnsw_ivf`, dim 200 = 3·64 + 8) serves recall@1 1.000
/ @10 0.994 / @100 0.971 at 0.85 ms, against the ivf arm's 0.976@10 at 2.26 ms on
the same corpus.

The SQ8 kernel change is a uniformity fix rather than a speedup: on the same
cell, 0.992 → 0.994 recall@10 and 0.856 → 0.850 ms, both inside run-to-run noise.
A scalar remainder is expensive when the kernel *is* the query — 100K sequential
rows on a terminal scan — and disappears in a walk over ~`ef · m0` scattered
candidates where memory latency dominates.

## Scope and follow-ups

Not covered by this PR:

- The standard 10M vector bench, where `recall@10 ≥ 0.99` is the acceptance bar.
  Everything measured here is 100K.
- A full `make bench` comparison against `main`; the measurements above are
  targeted cells rather than the suite.
- `hnsw_plane: sq4` as an exercised serving mode. It is wired, unit-tested and
  round-trips through the bundle, but no drain has served on it. Its memory
  arithmetic is also unfavourable in that role, since a 4-bit walk keeps the Sq16
  refine plane and therefore *adds* 0.5 B/dim rather than replacing 2 — the
  memory win belongs to a terminal scan, which this PR measures but does not ship
  as a serving mode.

## Tests

2274 lib tests pass; `fmt` and `clippy --all-features` clean.

- Blocked rotation is an exact isometry and round-trips at 1536 and 200 — the
  dims that pay padding. An unpadded transform that merely approximated the dot
  product would trade recall for the bytes it saves, so the test pins exactness
  rather than closeness.
- Both new kernels are compared against the portable tier. The VNNI arm quantizes
  the query and so is not exact; what is pinned is that the error stays inside
  the `sqrt(padded)` quantization bound, since a wrong nibble/query interleave
  would exceed it while still producing plausible numbers.
- The blocked scan matches per-node scoring on both sides of the shape switch. A
  row/query misalignment there would score the wrong vectors while returning a
  plausible top-k, which an aggregate recall number does not reliably catch.
- The SQ8 kernel equivalence test covers nine width shapes at the mask edges
  (1, 63, 64, 65, 100, 128, 200, 768, 1536).
- The SQ8 compatibility test covers all three bundle versions — `v04` persisted,
  `v03` persisted, `v02` derived must agree — since a version bump is where an old
  bundle can begin decoding into a subtly different plane.
- Bundle decode declines a corrupt node count rather than attempting an
  oversized allocation.


src/supertable/query/vector.rs
```diff
@@ -2139,14 +2193,84 @@ pub(crate) async fn assemble_hnsw_incremental(
...
+    // delta must land on the same one. A change to `vector.hnsw_plane` takes
+    // effect at the next FULL rebuild, not mid-extend.
+    let inherited_walk = if prior.sq4.is_some() {
+        if prior.sq4.as_ref().is_some_and(Sq4Scorer::has_residual) {
+            hnsw::WalkCodec::Sq4Residual
+        } else {
+            hnsw::WalkCodec::Sq4
+        }
+    } else if prior.sq8_plane.is_empty() {
+        hnsw::WalkCodec::Sq16
```
muralikpbhat: **Blocker — this inference violates the contract stated two lines up, and it destroys data.** `inherited_walk` is derived from the *want-filtered* decoded planes, but the maintenance fetch feeding this path passes `need_sq8=false` (writer.rs:7918) → `fetch_graph_sections` decodes with `want=Sq16` → `prior.sq4` is `None` and `prior.sq8_plane` is empty **regardless of what the bundle stores**. So inheritance always resolves to `Sq16` here. Consequences: under the shipped default (`sq8`), the first incremental drain re-encodes the bundle without the persisted SQ8 section — every later open re-derives it into owned heap, permanently losing the #644 zero-copy property (pre-PR v03 always rewrote the section, so this is a regression). Under `sq4`/`sq4_residual` it's worse: the non-derivable plane + ruler are destroyed by one append-only drain, and the delta extends a graph whose neighbors were built on Sq4 geometry using the Sq16 scorer. Fix: surface the bundle's **stored** codec tag from `decode_hnsw` and inherit that (fetching the planes the tag names), rather than inferring from filtered state.

src/superfile/vector/distance.rs
```diff
@@ -1530,6 +1544,581 @@ pub(crate) fn sq16_adaptive_norm_sq(
...
+    /// Quantizing the QUERY rather than the codes is what keeps this
+    /// exact enough to use: the codes are already 4-bit, so the only new
+    /// error is on the query side, one shared scale per leg, and it is
+    /// spread over `padded` independent terms rather than compounding.
+    fn build(padded: usize, q_code: &[f32], q_residual: Option<&[f32]>) -> Option<Self> {
+        // No block-count requirement: a dim that is not a whole number of
+        // blocks (200, say) keeps its remainder on a scalar tail rather
+        // than losing the integer path entirely — gating on divisibility
+        // silently dropped every such corpus back to the float kernel.
+        if !has_avx512vnni() {
```
muralikpbhat, line 1664: **Blocker — aarch64 does not compile.** `has_avx512vnni` is imported under `#[cfg(target_arch = "x86_64")]` but this call site is unconditional → `E0425` on every ARM build (reproduced locally on an aarch64 Mac; the macos-14 Node/Python and ubuntu-arm lanes are red with it). The function itself is portable (returns false off-x86), so the fix is un-gating the import. Same class as the E0133 that bit #629's CI — worth an ARM `cargo check` alongside the x86 one on SIMD PRs.

src/superfile/vector/distance.rs
```diff
@@ -1530,6 +1544,581 @@ pub(crate) fn sq16_adaptive_norm_sq(
...
+/// together at the end instead of one per row on the critical path.
+///
+/// Rows are read at `stride` apart (the plane is row-major), so this
+/// amortizes instruction count but not locality; making the reads
+/// sequential too would require storing the plane block-major.
+///
+/// Callers must ensure the VNNI feature set and that `first + SQ4_ROW_BLOCK`
+/// rows exist in `plane`.
+#[cfg(target_arch = "x86_64")]
+#[target_feature(enable = "avx512f,avx512bw,avx512vnni")]
```
muralikpbhat: `sq4_dot_vnni_rows` (and `sq4_dot_vnni` below) omit `avx512vl` from `#[target_feature]`, but call `_mm256_maskz_loadu_epi8`, which requires BW+VL. Verified in codegen on the pinned toolchain: the intrinsic can't inline into the narrower envelope and is emitted as an **out-of-line call wrapping a single masked `vmovdqu8`** — a function call on the partial-block path of the hottest loop (once per row per block; at dim 200 that's 1 of 4 blocks on every row). Runtime soundness currently rests on `avx512_enabled()` happening to check VL, with no SAFETY note saying so. Adding `avx512vl` to both lists fixes the perf and the fragility in one line.

muralikpbhat (commented):
Reviewed across the SIMD kernels (down to the disassembly), the v04 bundle format and its v02/v03 compatibility, the rotation change, calibration/codec selection, config, and the relationship to the #629 → #644 → #633 chain. The architecture is sound and most of the high-stakes surfaces verify clean; **holding approval on two blockers plus the shorthand reword** (inline).

**Verified clean:**
- **Memory safety at the instruction level**: the masked final blocks (both the SQ8 tail-mask change and the SQ4 partial blocks) compile to genuinely fault-suppressing masked loads (`vmovdqu8 {k}{z}`), so a row ending at an mmap page boundary cannot fault; mask math is sound; nibble order is consistent encoder ↔ all four tiers; a swapped interleave would exceed the VNNI parity bound ~80x, so the test genuinely catches it.
- **v04 format**: section arithmetic is header-driven (never `want`-driven), the v02/v03 compat test decodes *hand-built* old-format bytes (not re-encodes), node counts are overflow-checked before allocation, an unknown codec tag maps to `None` → ivf, and a binary downgrade after a v04 drain falls back gracefully rather than mis-parsing.
- **Zero-copy**: the Sq4 sections serve as `Bytes` slices of the mapped bundle (kernels are unaligned-safe), preserving #644 — at open time. (The inheritance blocker un-does it after the first incremental drain; see inline.)
- **Rotation keying**: rotated Sq4 planes exist only in v04 bundles, decode reconstructs the transform from persisted `(dim, rot_seed)` with the same blocked apply the only encoder used, old bundles stay unrotated on the walk path, and the legacy RaBitQ sign codes reconstruct bit-identically. The mean±2.7σ ruler (with min/max deliberately kept for the residual construction) is well-reasoned.
- **Config**: default `sq8` preserves shipped behavior (roundtrip-pinned), and `hnsw_plane` is correctly *not* in the options hash — the codec rides in the bundle header, matching the `rerank_codec` precedent.

**Chain fit (#629/#644/#633):** this builds directly on #644 (merge-base) and generalizes #629's bolt-on into the first-class dial — the sq8 default is #629's behavior, and #633 (k→ef) is untouched. Two chain notes: issue #636 is implemented by #644 and can be closed; and downstream runners that set the old walk knob need the key rename when this lands.

**Measurement gap to close before any 4-bit rung is enabled:** the PR measures the new codecs as *terminal scans* (good per-k tables) plus one sq8 walk point, but there is **no walk-mode ladder** — `hnsw_plane` ∈ {sq16, sq8, sq4, sq4_residual} on the same corpus with recall@1/10/100 + p50 + resident — and the body itself notes no drain has ever served on sq4. Since a too-coarse plane silently de-registers to ivf, a sq4 deployment could be entirely inert and look healthy. The ladder table (1M+, ideally the 10M gate) is the natural gate for the follow-up that first serves on these planes — the analog of what #629 provided for the sq8 rung.

**Test asks (non-blocking):** the float-tier parity sweeps cover only lane-multiple or tail-only shapes — add mixed block+tail dims (40, 200) so a broken vector-loop→tail handoff can't pass; and consider one encode→decoded-reference ground-truth test (all tiers are currently anchored to `sq4_dot_wide`, so a *consistent* nibble-convention flip across tiers stays green).

Blockers inline: the incremental-drain codec inheritance (data-destroying) and the aarch64 compile break (4 CI lanes red). With those fixed, the shorthand reworded, and the doc sweep done, this is a strong addition to the ladder.

src/superfile/vector/hnsw.rs
```diff
@@ -1250,47 +1727,104 @@ pub(crate) struct HnswIndex {
...
+        }
+    };
     let mut out = Vec::with_capacity(
-        HNSW_DATA_FIXED_BYTES + col.len() + n * 16 + sq16_codes.len() + sq8_len + graph_bytes.len(),
+        HNSW_DATA_FIXED_BYTES
+            + 1
+            + col.len()
+            + n * 16
+            + sq16_codes.len()
+            + walk_len
```
muralikpbhat, line 1983: `encode_hnsw` writes `sq4.parts()` verbatim but never validates the section lengths against the header it stamps — the v04 reader sizes the ruler and nibble planes purely from header `dim`/`n`, so a caller whose `Sq4Scorer` dim disagrees with the header dim produces a bundle whose implied offsets mis-slice every following section (residual plane, graph length, graph) with no decode error. Cheap insurance: release-mode asserts that `offset.len() == dim` and `codes.len() == n * dim.div_ceil(2)` here.

src/superfile/vector/hnsw.rs
```diff
@@ -1364,27 +1910,65 @@ pub(crate) fn decode_hnsw(bundle: &Bytes, sq8_walk: bool) -> Option<HnswIndex> {
...
+            let step = read_f32_le(c.take(dim.checked_mul(4)?)?);
+            let stride = dim.div_ceil(2);
+            let plane_len = n.checked_mul(stride)?;
+            let codes = bundle.slice_ref(c.take(plane_len)?);
+            let residual = if stored.with_residual() {
+                Some(Plane::Shared(bundle.slice_ref(c.take(plane_len)?)))
+            } else {
+                None
+            };
+            if want.is_sq4() {
```
muralikpbhat, line 2193: When `want.is_sq4()` and `Sq4Scorer::from_parts` declines a malformed section (non-finite step, shape mismatch), decode returns `Some` with `sq4 = None` **and** an empty `sq8_plane` — the SQ8-derive fallback below is unreachable in this arm, and `from_parts`'s doc claim that failure "degrades to the ivf fallback" is wrong (decode still succeeds). Net effect: a bundle with a corrupt ruler (e.g. torn write) silently serves a full-width Sq16 walk — correct results at ~4x the configured walk bandwidth, indistinguishable from healthy. Suggest deriving the SQ8 fallback (or at least logging) when `from_parts` returns `None`, and fixing the doc.

benches/utils/supertable.rs
```diff
@@ -4090,6 +4117,58 @@ pub mod vector {
...
+                            &warm_reader,
+                            supertable::VEC_COLUMN,
+                            &q_correct,
+                            &truths,
+                            k,
+                            nprobe,
+                            rerank,
+                        );
+                        eprintln!(
+                            "[078-curve] infino/post-drain recall@{k} = {recall:.3} \
```
muralikpbhat: The `[078-curve]` tag is internal planning shorthand — please rename to something self-describing (e.g. `[codec-curve]`). The same shorthand appears in one commit message on this branch; a reword/squash before merge keeps it out of public history.

src/config/mod.rs
```diff
@@ -524,6 +550,13 @@ pub struct VectorSettings {
     /// Default `ivf`; `global_fine_centroid` is experimental (see
     /// [`VectorSearchMode`]).
     pub search_mode: VectorSearchMode,
+    /// For `search_mode = hnsw_ivf`: which resident plane the graph walk
+    /// scores candidates on. Every variant re-ranks its final beam on the
+    /// full Sq16 plane, so this decides which candidates reach the beam and
+    /// what each costs — never the returned order. See
+    /// [`VectorHnswPlane`] for each variant's bytes and rebuild semantics.
+    /// Ignored under any other search mode.
+    pub hnsw_plane: VectorHnswPlane,
```
muralikpbhat, line 576: The retired `hnsw_sq8_walk` key is **silently ignored** (no `deny_unknown_fields`, figment drops unknown keys): a user who deliberately set `hnsw_sq8_walk: false` to walk Sq16 and reclaim the +1 B/dim upgrades and silently flips back to the `sq8` default — plane, resident memory, and per-open cost all change with no warning. A validate-time rejection (or a deprecation alias) for the old key would make the migration loud. Also: two doc references to the removed knob survive — the `hnsw_refine_k` struct doc (lines ~603/607) and the comment at `query/vector.rs:2456` — both should name `hnsw_plane` now.

src/superfile/vector/hnsw.rs
```diff
@@ -280,6 +285,387 @@ impl NodeScorer for Fp32Scorer {
...
+    offset: Vec<f32>,
+    step: Vec<f32>,
+    /// The seeded structured rotation the plane lives in. Queries rotate
+    /// on `prepare`; nodes come back through the inverse on
+    /// [`NodeScorer::decode_node`]. Reconstructed from `(dim, rot_seed)`
+    /// at decode — deterministic, nothing about it is persisted beyond
+    /// the seed.
+    rot: RandomRotation,
+    rot_seed: u64,
+    /// Rotated component count: `dim` rounded up to a power of two. The
```
muralikpbhat: **Load-bearing stale doc.** This describes `padded` as "dim rounded up to a power of two" that "stores all padded components" — but with the blocked rotation `padded == rot.dim == dim` always, and the v04 wire format is only correct *because* of that equality (decode sizes the ruler as exactly `dim` f32s and the planes as `n·ceil(dim/2)`). A maintainer trusting this comment and reintroducing pow2 padding on the write side would mis-slice every non-pow2-dim bundle — the exact header-vs-layout divergence the format exists to prevent. Same stale family: `distance.rs` tail comments ("padded is a power of two, so this runs only when the plane is narrower than 32") and the Sq4 SAFETY comments still asserting a "whole number of VNNI blocks" precondition that `Sq4IntQuery::build` deliberately dropped. Please sweep these — stale SAFETY text is what invites someone to "fix" a masked path back into a real OOB.

ekechinwokah:
thanks — both blockers, the shorthand, the doc sweep and both test asks are on the branch. per-thread detail inline; the two things that belong up here:

**the codec inheritance was the real defect, and it was worse than one plane.** `need_sq8=false` → `want=Sq16` → `prior.sq4=None` and an empty `sq8_plane`, so the inference could only ever resolve to `Sq16` — on every codec, including the shipped default. `decode_hnsw` already had the stored tag in a local and dropped it. surfaced as `HnswIndex.stored_walk` now, with the maintenance hydration asking for the plane the tag names.

**measurement gap: agreed, and it's the follow-up's gate.** one thing worth writing down before that run, because it cost me a wasted 1M battery: the bench installs `EnvFilter::from_default_env()`, so with `RUST_LOG` unset the `search_mode=hnsw_ivf but no resident graph … serving via ivf scan` warning cannot print at all. a runner grepping for it proves nothing either way — absence is not evidence the graph served. the ladder run pins `RUST_LOG=infino=warn` so a silent de-registration shows up instead of being inferred from a recall number.

agreed on both chain notes — #636 is closed by #644, and the walk-knob rename lands wherever a runner sets it.

gates: 2290 lib + 539 integration tests, clippy `--all-features --all-targets` clean, fmt clean, and `cargo check --target aarch64-unknown-linux-gnu --lib` clean on default and `--all-features`, all under `-D warnings`.

src/supertable/query/vector.rs
ekechinwokah: fixed. traced exactly as you called it: `writer.rs:7918` and `:9183` were the two `need_sq8=false` sites, `slow_vector_state` mapped that to `WalkCodec::Sq16`, and `decode_hnsw` already had the stored tag in a local it then discarded.

**the tag is surfaced, not inferred.** `HnswIndex.stored_walk` comes off the header regardless of what the decode asked for. `decode_hnsw` takes `Option<WalkCodec>` where `None` means as-stored, and the bool on `fetch_graph_sections` becomes `GraphWalkRequest::{Configured, AsStored}` so the two intents are named rather than living in a comment. maintenance passes `AsStored` and hydrates the plane the tag names — which is what the extend path already wanted.

**plus a guard:** codec names a 4-bit plane, none resident → `Ok(None)`, full rebuild. a rebuild is expensive and re-fits the ruler; a lossy re-encode is unrecoverable.

both consequences you named are gone — the SQ8 section is rewritten from the concatenated Sq16 plane so #644 survives the first incremental drain, and the 4-bit delta extends onto the prior ruler with the graph built on the Sq4 scorer.

`an_as_stored_reencode_preserves_the_walk_plane` is the drain's write step with the manifest stripped out: hydrate as-stored, re-encode on the header's codec, assert byte-identical, all four codecs. `stored_walk_reports_the_header_not_the_decoded_planes` asserts the narrow view really does drop the planes, so why inference cannot work is pinned in place rather than in a comment.

src/superfile/vector/distance.rs
ekechinwokah: fixed at `09d3322b`, pushed ahead of the rest.

not by un-gating the import. that compiles, but it leaves the integer leg on targets with no `vpdpbusd` kernel to read it — a permanently-`None` `q_int` plus four never-read quantizer fields, two dead-code warnings on ARM. the gate went on the data instead: `Sq4IntQuery`, its impl, `VNNI_BLOCK_COORDS`, `I8_MAX`, the field and its construction, which is where the two kernels reading them already were. `has_avx512vnni` takes the `cfg_attr(not(target_arch = "x86_64"), allow(dead_code))` that `avx512_enabled` and `has_vbmi` in that same file already carry.

**your process ask is the durable half**, and it's local now — `gcc-aarch64-linux-gnu` installed, `cargo check --target aarch64-unknown-linux-gnu --lib` clean on default and `--all-features` under `-D warnings`. worth knowing for the next simd pr: the multi-os matrix jobs set `RUSTFLAGS: ""` while the global is `-D warnings`, so an arm-only dead-code warning never surfaces in ci even on a green lane. those two were invisible until I graded the cross-compile myself.

src/superfile/vector/distance.rs
ekechinwokah: fixed on both. checked the asymmetry first — `sq8_dot_vnni` masks with `_mm512_maskz_loadu_epi8` (512-bit, bw only) so its list was already right; only the two sq4 kernels use the 256-bit `_mm256_maskz_loadu_epi8`, and only they were short a feature.

the safety note states the distinction rather than burying it: omitting vl is not unsound — `has_avx512vnni` implies `avx512_enabled`, which checks vl — it puts the intrinsic outside the declared envelope so it cannot inline.

while in there, the safety block two lines down claimed the loop bound was `d + 64 <= padded`, which it is not (`while d < padded`), and a scalar tail that no longer exists. same stale-precondition class as your doc-sweep comment, so it now names the branch that actually keeps the read inside `packed` and says not to simplify the mask away.

src/superfile/vector/hnsw.rs
ekechinwokah: added, release-mode as you suggested — four integer compares once per drain against a silently corrupt bundle is not a trade worth thinking about. `offset.len() == dim`, `step.len() == dim`, `codes.len() == n * ceil(dim/2)`, and the residual plane's presence and length against the header's codec.

the comment states the failure mode rather than the rule, since the rule reads off the asserts: a mismatch mis-slices every *later* section — residual plane, graph length, graph — and decode reports nothing at all.

src/superfile/vector/hnsw.rs
ekechinwokah: fixed, both halves. `decode_hnsw` derives the SQ8 plane when `from_parts` declines, so a torn section degrades one rung instead of all the way to a full-width Sq16 walk, and warns naming the column — "indistinguishable from healthy" was the whole problem, and nothing in the served answers would have shown it.

`from_parts`'s doc no longer claims failure reaches the ivf fallback. it says what happens: the decode still succeeds, the walk runs on the derived SQ8 plane, and that is why the warning exists.

`a_rejected_4bit_section_falls_back_to_the_derived_sq8_plane` walks the header to the ruler's `step` vector and zeroes one entry — minimal poison, models a torn write rather than a structurally short section — then asserts `sq4` is `None`, `sq8_plane` is exactly `n * dim`, and `stored_walk` still reads `Sq4`.

benches/utils/supertable.rs
ekechinwokah: renamed to `[codec-curve]`.

src/config/mod.rs
ekechinwokah: fixed — rejected, not aliased. an alias keeps working and quietly stops meaning what it meant; failing the load is the loud version, and this knob shipped one release.

`Config::from_figment` checks a `RETIRED_CONFIG_KEYS` table via `figment::find_value` *before* extraction — which is exactly where the key would otherwise vanish — and the error names the replacement plus the value mapping (`sq8` = the old `true`, `sq16` = the old `false`). a bare "unknown key" leaves the reader guessing which knob took over.

both stale references fixed too: `hnsw_refine_k`'s doc keys off `hnsw_plane` and says it is ignored at `sq16` where the walk already scores Sq16, and the `query/vector.rs` comment describes the plane decode actually made resident.

`a_retired_config_key_is_rejected_and_names_its_replacement` covers `true` and `false`, asserts the message names both keys, and confirms the shipped default does not trip the check.

src/superfile/vector/hnsw.rs
ekechinwokah: fixed, and you were right that this one is load-bearing. checked the equality before rewriting anything: `from_sq16_plane` sets `padded = rot.dim` and `RandomRotation::new` does `dim.max(1)`, so `padded == dim` unconditionally.

the doc says that now, and says why it matters rather than just what it is — decode sizes the ruler as exactly `dim` f32s and each plane as `n·ceil(dim/2)`, so pow2 padding on the write side without widening those reads mis-slices every section after the ruler on any non-pow2 dim.

rest of the family swept: both float-arm tail comments (they claimed the tail runs only for planes under 32 coords on avx-512 and "0 or 8" on avx2 — under an unpadded rotation it is up to 31 and 15 on ordinary dims), and the sq4 safety texts still asserting the "whole number of vnni blocks" precondition `Sq4IntQuery::build` deliberately dropped, including the one in the vnni parity test where several sweep lengths (8, 100, 200) are not block multiples and exercising that is the point.

both test asks taken. the float-tier sweeps carry mixed block+tail dims now — 40, 100, 200 on both arms; your reasoning was the right one, a broken vector-loop→tail hand-off is invisible on lane multiples and invisible again tail-only, and only shows up when both run. and `sq4_nibble_convention_is_pinned_by_hand_computed_dots` is the encode-side anchor: a one-hot query, so the dot reduces to exactly the code at coordinate `d`, readable straight off the packed byte with no kernel involved, swept across block and tail boundaries. a convention flip applied consistently across every tier and the encoder now fails there.

muralikpbhat (approved):
Approving. Re-verified the round-2 head (`a375901`) item by item rather than from the descriptions:

- **Blocker 1 (codec inheritance) — fixed at the right layer.** `inherited_walk = prior.stored_walk` now reads the bundle's persisted header tag instead of inferring from want-filtered planes, with an explicit guard for a sq4 tag whose plane is missing. With this, the mmap zero-copy property (`Plane::Shared` slices of the one mapped bundle) survives incremental drains — the failure mode where the first append-only drain silently downgraded the bundle to Sq16 and pushed every later open onto the heap-derive path is gone.
- **Blocker 2 (aarch64) — empirically fixed**: `cargo check --lib` passes on an ARM Mac, the ARM CI lanes are green, and the integer-query path is `cfg`-gated to x86_64 with the float tiers serving elsewhere. The VNNI arm is intact on x86 (`has_avx512vnni()` dispatch, `vpdpbusd` kernels), and both `sq4_*_vnni` functions now carry `avx512vl`, so the masked partial-block loads inline again.
- **Test asks landed**: the parity sweeps gained mixed block+tail shapes (40, 200), and the `*_matches_decoded_reference` anchors ground the kernel family against the decoded reference.
- **Shorthand**: the bench tag is renamed. One residual: commit `7cc3409`'s body still carries the old shorthand — a squash-merge with a clean message resolves it; if this merges as a merge-commit, please reword that one first.
- **v05 note**: the rebase composed the walk-codec header with the stamped k→ef curve as a v05 bundle (v04 retained for decode). The composition is newer than the deep review pass; the version-matrix compat test claim is spot-checked rather than exhaustively traced — flagging for the record, not blocking.

Benches pending as of this review; understood the merge waits on them — the reverted/new ceilings make the bundle changes largely self-testing there.

PR #656, opened 2026-08-26
vector: the undrained tail inherits the table's stamped probe width

ekechinwokah:
Fixes regression introduced in #635.

The undrained user tail read at a blanket UNDRAINED_CELL_NPROBE_MAX regardless of what the table had already measured about itself. Once a drain stamps a width law, the delta rows are the same distribution the law measured -- they are assigned cells against the same grid -- so the tail now inherits the stamped width for this k. A table stamped 1 at k=10 reads its delta at one cell; a diffuse table reads its delta at its own measured width. Only a table with no stamp yet falls back to the blanket cap, and only on cosine, where the one-cell collapse on real embeddings was measured.

Measured, synthetic 1M on Azure, post-delta cold: user data 8 -> 1 GET on the warmup query and 7 -> 1 steady, warm p50 1.12 ms -> 962 us, at recall@10 0.995 unchanged. That lands on the invariant the ceiling table documents (post-delta = post-drain + 1, the tail being one extra coalesced GET), so the post-delta ceilings revert to their original calibration -- the 14/18 and 10/13 rebudget existed only to accommodate the blanket.

### DIFF
--- a/benches/utils/supertable.rs
+++ b/benches/utils/supertable.rs
@@ -981,14 +981,14 @@ fn open_consumer(modality: Modality, built: &supertable::IngestResult) -> (TempD
 /// whole-cell at <5M (cells ~6 MiB), 2–4 islands at 10M.
 const VECTOR_COLD_GET_CEILINGS_FIRST: &[(&str, u64, u64)] = &[
     ("post-drain", 4, 8),
-    // post-delta is the only state that serves an UNDRAINED user tail
-    // beside the drained bulk, and the user table never carries a probe
-    // law -- the laws describe the hidden table. That tail therefore
-    // probes up to `UNDRAINED_CELL_NPROBE_MAX` (8) cells rather than the
-    // single cell these ceilings were calibrated against, measured at 8
-    // user + 1 hidden GET at 1M. Budget the widened tail; the hidden
-    // side of the same query is unchanged.
-    ("post-delta", 14, 18),
+    // post-delta serves the undrained user tail beside the drained bulk,
+    // and the tail INHERITS the table's stamped width law (the delta is
+    // the same distribution the drain measured), so a stamped-1..1 table
+    // reads its delta at one cell and these ceilings hold at their
+    // original calibration. They were briefly rebudgeted to 14/18 while
+    // the tail read at a blanket 8-cell cap that ignored the stamp --
+    // that accommodation is gone with the blanket.
+    ("post-delta", 6, 10),
     ("post-compact", 4, 8),
 ];
 /// Per-state `(label, <5M ceiling, 5M–20M ceiling)` on the SECOND
@@ -1000,8 +1000,8 @@ const VECTOR_COLD_GET_CEILINGS_FIRST: &[(&str, u64, u64)] = &[
 /// shard generations after a budgeted optimize).
 const VECTOR_COLD_GET_CEILINGS_SECOND: &[(&str, u64, u64)] = &[
     ("post-drain", 1, 4),
-    // Same undrained-tail allowance as the first-query table above.
-    ("post-delta", 10, 13),
+    // Stamp-inherited tail, as in the first-query table above.
+    ("post-delta", 2, 5),
     ("post-compact", 1, 5),
 ];
 /// Doc-count cutoffs for the two ceiling columns (exclusive upper
--- a/src/supertable/query/vector.rs
+++ b/src/supertable/query/vector.rs
@@ -3070,32 +3070,31 @@ impl SupertableReader {
                     nprobe_max: FILTERED_USER_CELL_NPROBE,
                     ..CellRoutingParams::default()
                 }
-            } else if metric == Metric::Cosine {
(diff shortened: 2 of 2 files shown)


muralikpbhat (approved):
Approving. Verified against the code:

- **The width-1 edge is handled correctly** — the trap that caused #635's first-round regression. Inheritance reads the raw stamp (`width_for_k_at(k)`), not the `LAW_WIDTH_WITHIN_DEFAULT`-filtered `law_width`, so a table stamped 1 reads its delta at one cell instead of filtering to `None` and re-triggering the blanket cap.
- **Semantics are right**: the stamp wins on any metric (`Some(w) => w.max(1)` — the stamp is this table's own measured geometry under its own metric); with no stamp, the #635 behavior is preserved exactly (cosine → bounded blanket, other metrics → the one-cell default). Filtered and explicit-`nprobe` arms untouched; the manifest fetch threads `op_stats` correctly.
- **`undrained_nprobe_max` is a pure function with a unit test covering all six arms** — stamped on two metrics at two widths, unstamped cosine, unstamped L2/NegDot.
- **The ceilings genuinely revert rather than re-tune** — post-delta 14/18 → 6/10 and 10/13 → 2/5 match the pre-#635 calibration, and the measured result (user GETs 8 → 1 warmup / 7 → 1 steady, warm p50 1.12 ms → 962 µs, recall unchanged) lands exactly on the documented invariant: post-delta = post-drain + one coalesced tail GET.

Two notes, neither blocking:

1. The reverted post-delta ceilings make this change self-testing — the pending `st-vec` bench lane is the empirical proof, since a broken inheritance would trip its own reverted ceiling. Worth letting it land green before merge.
2. Design note, not a defect: a delta whose distribution *shifts* from what the drain measured inherits a possibly-stale narrow width — identically true of the drained bulk itself until the next `optimize()`, so it's consistent; just worth remembering when reasoning about tables with regime-changing ingest.

ekechinwokah:
> Approving. Verified against the code:
> 
> * **The width-1 edge is handled correctly** — the trap that caused #635's first-round regression. Inheritance reads the raw stamp (`width_for_k_at(k)`), not the `LAW_WIDTH_WITHIN_DEFAULT`-filtered `law_width`, so a table stamped 1 reads its delta at one cell instead of filtering to `None` and re-triggering the blanket cap.
> * **Semantics are right**: the stamp wins on any metric (`Some(w) => w.max(1)` — the stamp is this table's own measured geometry under its own metric); with no stamp, the #635 behavior is preserved exactly (cosine → bounded blanket, other metrics → the one-cell default). Filtered and explicit-`nprobe` arms untouched; the manifest fetch threads `op_stats` correctly.
> * **`undrained_nprobe_max` is a pure function with a unit test covering all six arms** — stamped on two metrics at two widths, unstamped cosine, unstamped L2/NegDot.
> * **The ceilings genuinely revert rather than re-tune** — post-delta 14/18 → 6/10 and 10/13 → 2/5 match the pre-#635 calibration, and the measured result (user GETs 8 → 1 warmup / 7 → 1 steady, warm p50 1.12 ms → 962 µs, recall unchanged) lands exactly on the documented invariant: post-delta = post-drain + one coalesced tail GET.
> 
> Two notes, neither blocking:
> 
> 1. The reverted post-delta ceilings make this change self-testing — the pending `st-vec` bench lane is the empirical proof, since a broken inheritance would trip its own reverted ceiling. Worth letting it land green before merge.
> 2. Design note, not a defect: a delta whose distribution _shifts_ from what the drain measured inherits a possibly-stale narrow width — identically true of the drained bulk itself until the next `optimize()`, so it's consistent; just worth remembering when reasoning about tables with regime-changing ingest.

thanks — and the bench lane has since landed green on `43b17f0` (18/18 checks), which is the proof your first note asks for: with the post-delta ceilings reverted to 6/10 and 2/5, a broken inheritance trips its own ceiling, so that lane passing is a positive result rather than an absence of failure.

**on note 2 — one degree sharper, and worth writing down.** the delta's staleness isn't quite identical to the drained bulk's. the bulk's law was measured *on those exact rows*, so it is only stale in the sense that the corpus around them grew. the delta's width was measured on *different* rows and extrapolated onto it — right whenever the delta is drawn from the same distribution, which is the overwhelmingly common case, but an extrapolation rather than an aging measurement.

what that changes is the direction of the failure. the blanket cap over-read a regime-shifted delta: cost GETs, recall safe. inheritance under-reads it: costs recall, silently, and cheaply — so there is no cost signal to notice it by either. and the bench cannot see it by construction, since its delta comes from the same generator as its bulk.

**not arguing for a change.** the exposure is bounded by drain cadence; a never-drained table still gets the blanket, so a table's first data is never under-read; and any floor added to hedge would reintroduce exactly the regression this PR removes. if we ever want a signal instead of a guess, the cheap one is comparing the width the next drain measures against the stamp it had been serving — a diagnostic, not a serving change. worth an issue only once regime-changing ingest shows up as a real workload.

PR #662, opened 2026-08-26
fts: single windowed-MaxScore kernel for multi-term union top-k

vinaykakade:
## Summary

Multi-term OR (disjunction) ranked search chose per query between a per-candidate MaxScore + block-max scorer and a windowed SIMD OR-sum scorer via a df / upper-bound routing heuristic — each kernel lost badly in the other's regime, so a heuristic picked between them. This replaces both with **one continuous-partition windowed MaxScore kernel** that covers both regimes in a single loop:

- The essential / non-essential partition is a continuous function of the running k-th threshold.
- When a term dominates, it degenerates to a per-candidate dominant-term leader scan (block-max pruning plus a tight per-block bound on the non-essentials).
- When no term dominates, it stays a windowed OR-sum across the essential block.

Every multi-term union routes through it, in both the single-shot and the ranged fan-out entries, so a query runs the identical kernel whether or not the fan-out sliced it. The 2-term rare+common WAND path is unchanged. The now-unused routing heuristic and its constants are removed.

## Correctness

Gated against the per-candidate scorer by the existing brute-force and randomized-fuzz oracles, extended to cover the new kernel across query shapes, `k` values, negation, and doc-id sub-ranges:

- `windowed_maxscore_agrees_with_bmm`, `..._randomized`, `..._negation_agrees_with_bmm`, `..._dense_common_leader_agrees_with_bmm`, `..._ranged_agrees_with_bmm_range`
- `filter_survivors_compacts_in_order`

Each asserts bit-for-bit identical top-k to the per-candidate scorer. `make ci` (fmt, clippy `-D warnings`, tests) is green.

## Performance

Measured on a multi-term union workload (Wikipedia corpus, 103 distinct three-or-more-term union queries, min-of-5 on a single fixed host). **Before** is the previous routing heuristic; **after** is the single kernel. Aggregate query time over the whole set:

| depth | before (router) | after (single kernel) | change |
|------:|----------------:|----------------------:|-------:|
| TOP_10   | 194,828 µs | 154,323 µs | **−20.8%** |
| TOP_100  | 399,030 µs | 260,716 µs | **−34.7%** |
| TOP_1000 | 668,777 µs | 454,407 µs | **−32.1%** |

The single kernel also matches the per-query best-of-both-old-kernels bound at every depth, so no query shape regresses at the aggregate. Representative per-query TOP_10 deltas:

| query | before | after | change |
|-------|-------:|------:|-------:|
| `george r brown convention center` | 11,939 µs | 4,503 µs | −62% |
| `civil war battlefields` | 480 µs | 399 µs | −17% |
| `to be or not to be` | 14,703 µs | 15,737 µs | +7% |

The big wins are queries the old heuristic routed to the losing kernel; the near-parity cases (like the dense-stopword `to be or not to be`) are ones it already routed well. A handful of very small unions see a sub-millisecond increase (fixed per-window setup on a union too small to amortize it), dwarfed by the aggregate improvement.

### DIFF
--- a/src/superfile/fts/reader/core.rs
+++ b/src/superfile/fts/reader/core.rs
@@ -256,33 +256,34 @@ impl PreparedClauses {
 }
 
 /// Multi-term OR algorithm selector for the bench harness's
-/// `search_with_algo_for_bench` entry point. Production code routes
-/// through `FtsReader::dispatch_or_algo`, which picks
-/// automatically; this enum exists so head-to-head bench runs can
-/// compare all three under identical inputs.
+/// `search_with_algo_for_bench` entry point, so head-to-head bench runs can
+/// compare the kernels under identical inputs. Production code routes through
+/// `FtsReader::dispatch_or_algo`, which sends every multi-term union to
+/// `WindowedMaxscore` and a 2-term rare+common OR to WAND+BMW.
 #[doc(hidden)]
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub enum OrAlgo {
-    /// Block-Max MaxScore: production default for dominant-term ORs.
+    /// Per-candidate Block-Max MaxScore. Superseded on the production path by
+    /// `WindowedMaxscore`; retained as a bench baseline and the correctness
+    /// oracle the windowed kernels are checked against.
     Bmm,
-    /// WAND + Block-Max-WAND: historical baseline; retained for
-    /// regression comparisons.
+    /// WAND + Block-Max-WAND. The production path for a 2-term rare+common OR
+    /// (it pivots on the rare term to skip the common term's long list); a
+    /// bench baseline otherwise.
     WandBmw,
-    /// Exhaustive union walk with SIMD scoring + top-K heap. Wins
-    /// when no term dominates (uniform `term_max_bm25` upper bounds)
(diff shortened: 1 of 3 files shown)


src/superfile/fts/reader/core.rs
savannahar68: this comment is not true anymore right with this PR?

src/superfile/fts/reader/scorers.rs
```diff
@@ -1443,6 +1575,363 @@ impl FtsReader {
...
+                }
+            }
+
+            // Complete the non-essentials over the compacted survivors. Once
+            // the heap is full, do it the reference way: strongest non-essential
+            // first, re-filtering the survivor list (progressively tighter
+            // budget) before each and batch-scoring matches — so weaker terms
+            // touch a shrinking survivor set. While filling, every doc must be
+            // admitted, so score all non-essentials over every candidate.
+            let _ = non_ess_ub;
```
savannahar68: why do we need this line?

savannahar68 (approved):
lgtm

src/superfile/fts/reader/scorers.rs

src/superfile/fts/reader/core.rs
vinaykakade: Done (also for other stale comments)

PR #672, opened 2026-08-27
vector: a flat 4-bit index as a peer of the hnsw graph

ekechinwokah:
# vector: a flat 4-bit index as a peer of the graph

## Why

The 4-bit codec landed in #651 as a **walk plane for the graph** — `vector.hnsw_plane = sq4`. In that role its memory arithmetic works against it: the walk still re-ranks its beam on the full Sq16 plane, so a 4-bit plane *adds* 0.5 bytes/dim on top of 2 rather than replacing them.

The memory win belongs to a terminal scan that ranks on the codes alone and keeps no Sq16 plane resident. #651 measured exactly that — 73.25 MiB and 1.47 ms at recall@10 0.933 on dbpedia-1536 at 100K — but through `test_helpers::sq4_flat`, a feature-gated type no table could be pointed at. Real numbers, for a configuration that did not exist.

This makes it exist: `vector.search_mode = flat_ivf`, built at drain, published, hydrated, and served.

## What changed

**A flat index, shaped as a peer of the graph rather than a mode of it.** Its own magic (`INFDFL01`), its own resident type, its own drain and serve paths — and deliberately **no Sq16 section and no graph section**. At 1536 dims the nibble plane is 768 bytes/row against Sq16's 3072, so holding Sq16 to refine a 4-bit scan lands at 3840 and gives back the entire advantage. `residency_is_the_nibble_plane_only` asserts the per-row rate against the codec's own rate, so a future change that retains the Sq16 codes fails there instead of passing every other test while quadrupling residency.

A format of its own rather than a graph bundle with sections omitted: encoding one as a degenerate other would mean every reader had to know which of its mandatory sections were secretly optional. A distinct magic makes an old reader decline the bytes rather than mis-slice them.

**Three config knobs, each stating one thing.**

The mode is spelled `flat_ivf`, not `flat`, for the reason `hnsw_ivf` is: the `_ivf` suffix names the fallback, so nothing about where a query actually lands is hidden in the name. Above `flat_max_docs`, pre-drain, on a filtered query, or on a column the plane was not built for, the query serves `ivf` — and a mode named `flat` would have concealed exactly that half of its behaviour.

| knob | default | what it decides |
| --- | --- | --- |
| `search_mode = flat_ivf` | `ivf` | serve by scanning the resident 4-bit plane |
| `flat_max_docs` | 1000000 | doc ceiling; above it, serve `ivf` |
| `flat_register_floor` | 0.80 | tripwire the fitted plane must clear to be published |

**The plane is bare `sq4`, and the residual rung is deliberately carved out.** An earlier revision shipped a `flat_plane` knob with a 1.0 B/dim `sq4_residual` default (measured 0.991 recall@10 at 146 MiB / 3.57 ms). Review caught what the in-repo precedent already proves at 2 B/dim (`sq16_vs_sq8residual_kernel_recall`: a single uniform plane beats coarse+residual at equal bytes on recall *and* latency): one rung down, a single fitted 8-bit plane — 256 uniform levels, one pass, kernels already shipped — should dominate `sq4_residual`'s two passes at the same bytes. Shipping a default that comparison may retire is a migration for nothing, so the knob is gone and the drain builds the bare plane. The 1.0 B/dim slot stays open until a matched-bytes A/B (`sq8` vs `sq4_residual`) fills it — the persisted form carries a codec tag, so a future rung is a config variant, not a format change.

`flat_register_floor` defaults to **0.80, a broken-plane tripwire, not a quality bar** — deliberately not the graph's 0.98. A 4-bit plane's recall ceiling is set by the codec (16 levels per coordinate measures ~0.93 on dbpedia-1536), so a 0.98 default would leave the mode structurally unable to register: dead on arrival, by configuration. A healthy plane sits well above 0.80; a degenerate ruler or mis-sliced section collapses toward random and lands far below it. Operators who want a bar raise it — that is what a per-index-type floor is for.

`flat_max_docs` is a **latency** ceiling, not the RAM ceiling `hnsw_max_docs` is. The plane is 0.5 B/dim, so 10M × 1536 would still fit one host at ~7.7 GiB — but an exhaustive scan is linear (~1.5 ms measured at 100K × 1536), so 1M lands near 15 ms and 10M near 150 ms.

**Cosine only, gated at build and at serve.** Both resident index types score `−dot` and map it onto the `1 − dot` cosine scale. That is the column's own ordering only under `Cosine`: under `L2Sq` it is a different ranking (`‖a−b‖² = ‖a‖² + ‖b‖² − 2a·b` agrees with `−dot` only when the norms match), and under `NegDot` the order is right but the published score carries a meaningless `+1`. The register gate cannot catch this by itself — `probe_recall` grades a `−dot` scan against a `−dot` exhaustive scan, so a mis-metriced index measures *perfect* recall against a ground truth wrong in the same direction, clears its floor, registers, and serves wrong neighbours at the right latency. `MetricUnsupported` now declines it at build for both index types, and again at serve, since a graph published by an earlier binary can still be resident. The centroid-graph router already gated on `Cosine` for the same reason.

**The scan over-fetches for boundary replicas.** Under `drain_replica_target_factor > 1` one user row occupies several nodes carrying one `stable_id`, and `top_k_ascending` collapses them — so a scan of width `k` returns fewer than `k` distinct rows. The graph arm widened for this and the flat arm did not; both now call one `replica_fetch_width(k)`, because two arms deciding this separately is how the divergence arose. Dormant at the shipped `1.0` default, which is why no test caught it.

**A register floor per index type, stated rather than derived.** One derived number, `target_recall - hnsw_recall_slack`, gated both index types. That coupling has a measured cost: accepting a deliberately coarse flat plane meant lowering `target_recall`, which also detunes the ivf width laws answering **filtered** queries — filtered recall@10 fell to 0.792 against the bench's 0.80 floor purely as a side effect of testing a coarse plane. `hnsw_register_floor` defaults to 0.98, the value the shipped `0.99 - 0.01` pair produced, so a graph table that changes nothing behaves as before; `flat_register_floor` gets its own number (above). `hnsw_recall_slack` is retired and rejected at load with its replacement named.

That covers the config that *set* the retired key. It does not cover the one that lowered `target_recall` and never set the slack key: nothing retired is present, the load succeeds, and the floor silently snaps from (say) 0.94 to 0.98 — after which an index calibrating at 0.96 de-registers at the next drain and the table quietly serves ivf. A floor above `target_recall` now warns at load, where both numbers are in hand, naming the fact that the floor is no longer derived. A warn rather than a rejection because holding the resident index to a higher bar than the ivf laws is a legitimate thing to want; it is only an unnoticed migration that looks the same.

The floors are separate because their jobs differ. The graph is a latency optimization over a correct scan, so its floor is a quality bar. A flat plane is a memory trade the operator chose, so its floor's job is narrower: catch a plane that is *broken* (degenerate ruler, mis-sliced section) rather than one that is merely coarse.

**Declines return their reason instead of a bare `None`.** Fourteen `Ok(None)` returns across the two paths, roughly half logging nothing, meant a table configured for one index served another with no way to tell — the only symptom being that the numbers looked like the mode you did not ask for. `IndexUnavailable` now carries the reason (no such column, unsupported codec, no rows, over the ceiling, below the floor, gather empty, row mismatch, not hydrated, wrong kind, column/dim mismatch, empty, not a pure append, no new rows, plane not resident) and `IndexOutcome<T>` keeps "declined, fall back" distinct from `Err`, which is a real failure. `IndexOutcome::or_warn` is the single place a decline is reported, so no path can fall back quietly by forgetting to. Falling back to `ivf` stays automatic — that part was never wrong.

Applied to the graph path, both halves, not just the new one. `hnsw_search` previously returned a bare `Option` and collapsed **wrong column**, **wrong dim** and **empty index** into a single `return`, diagnosed after the fact by a separate `warn_hnsw_no_resident_graph` that could only guess between them from whether a graph ref existed. Those are three different problems calling for three different reactions, so they are now three reasons, and that warner is deleted — one reporting path, not two.

`or_warn` carries the dedup the deleted warner had, keyed on the rendered message rather than on `(diagnosis, column)`. A serving decline happens per query, so a table configured for an index it will never have would otherwise warn on every one; keying on the message means two columns declining for the same reason each get said once.

**The shared plumbing is named for its job.** `publish_hnsw_blob`, `GraphBundle`, `ResidentGraphSections`, `fetch_graph_sections`, `graph_sections_cache`, `slow_vector_state_graphs_blob` and friends were always generic — a content-addressed blob, an envelope with a population key, a per-uri resident cache slot — but named after their only caller. That naming left a second index two options: publish itself through a function called `publish_hnsw_blob`, or grow a duplicate stack. Renamed in one commit with no behaviour change (68 occurrences, 7 files, reviewable on its own).

**The envelope states which index its payload holds** rather than a reader sniffing the payload's magic. The tag rides the payload section's existing presence byte — `0` was already "absent", `1` has only ever meant a graph — so an envelope written before the tag reads back as `Graph`, and a binary predating a kind sees a present section, fails to decode it as a graph, and serves `ivf`. A downgrade, not a mis-parse. `ResidentVectorIndex.data` is an enum rather than a second `Option`, so "both present" and "neither" are not states the serving path has to consider.

**One row gather, two callers.** The loop that reads every superfile's Sq16 rows moved out of `assemble_hnsw_sections` into `gather_sq16_rows`. The iteration order, superseded-cell exclusion, and local→stable id resolution together define what node index N *means*; two callers computing that separately would eventually disagree, and the failure would be silently wrong ids rather than an error. Its `probe_step` parameter is the one thing the callers differ on: the graph takes a strided sample so a hostile corpus never materializes the whole plane just to decline, while a flat build takes the whole plane because there is no calibration sweep to skip.

**The bench back door is gone.** `test_helpers::sq4_flat` re-exported the index so a harness could build one from an fp32 slab and scan it directly. Nothing needs that — the index is chosen by a config key, so measuring it means setting the YAML and running the engine normally. `Sq4FlatIndex` is `pub(crate)`; `build` (fp32 → plane) and `minimum_bytes` fall to `#[cfg(test)]`, since the serving path never holds fp32 and the byte floor exists to be compared against the real figure rather than reported instead of it.

## Measured

dbpedia-1536, Azure store, ingest → drain → publish → hydrate → query. No bench arm and no adapter: the standard `supertable vector` cell with the mode set in `./infino.yaml`, so the arms differ only by config. Both scales measured on the **shipped configuration** — `search_mode: flat_ivf` and nothing else; the bare plane registers at the default tripwire with no overrides.

**100K — the scale this mode is for:**

| arm | resident | R@10 | p50 |
| --- | --- | --- | --- |
| `flat_ivf` (bare `sq4`) | **73 MiB pinned** | 0.944 | 1.62 ms |
| `ivf` @ nprobe=8 | ~337 MiB hidden index, reclaimable cache | 0.942 | 0.83 ms |
| `ivf` @ nprobe=32 | 〃 | 0.988 | 1.59 ms |

**1M — the `flat_max_docs` ceiling:**

| arm | resident | R@10 | p50 |
| --- | --- | --- | --- |
| `flat_ivf` (bare `sq4`) | **726 MiB pinned** | 0.938 | 20.0 ms |
| `ivf` @ nprobe=8 | ~3.05 GiB hidden index, reclaimable cache | 0.959 | 1.99 ms |
| `ivf` @ all cells | 〃 | 0.996 | 17.8 ms |

Reading them honestly, in both directions:

- **Recall is scale-invariant** (0.944 → 0.938 across 10×), which is what a codec-set ceiling should do — the number is a property of 16 levels per coordinate, not of the corpus.
- **Latency is linear** (1.62 → 20.0 ms), and resident is exact at 0.5 B/dim at both scales (73 MiB at 100K, 726 MiB at 991.8K live rows). Both are the mode's claims, now measured rather than extrapolated.
- **At the ceiling, warm ivf dominates the scan on recall *and* latency.** The mode's case at 1M is not speed: it is the memory story below. The sweet spot is ≤ a few hundred K, and `flat_max_docs` is a ceiling, not a recommendation.
- The pre-carve run also measured the carved `sq4_residual` arm at **146 MiB / 0.991 / 3.57 ms** — kept as data, since it is the number the matched-bytes A/B has to beat.

### What serving actually holds

| | RAM to serve the quoted p50 | traffic, steady state |
| --- | --- | --- |
| `flat_ivf` @ 100K | **153 MiB pinned** (77 plane + 76 manifest), cache-independent | **0 GET/query** |
| `ivf` @ 100K | 86 MiB pinned + ~337 MiB hidden index kept warm in cache | 0 GET warm; degrades toward cold (61 GET/query class) under eviction |
| `flat_ivf` @ 1M | **841 MiB pinned** (726 plane + 115 manifest), cache-independent | **0 GET/query** |
| `ivf` @ 1M | 109 MiB pinned + ~3.05 GiB hidden index kept warm | 0 GET warm; cache-conditional |

Stating the memory multiple honestly, because it varies with what you compare against:

- **~2–4× floor-vs-floor** — the defensible number. Warm ivf's minimum working set is essentially its hidden index (queries spread across cells), so 337 MiB vs 153 MiB at 100K (2.2×) and 3.05 GiB vs 841 MiB at 1M (3.6×).
- **4× per row vs any other resident plane** — 0.5 B/dim against Sq16's 2.0; **~6–7× vs the hnsw graph bundle** (Sq16 plane + walk plane + edges + ids).
- **~9–10× vs the memory the bench run actually held** to produce the ivf rows (1.44 GiB / 8.0 GiB superfile cache as-run) — real, but flattering, since that cache was not tuned down to its minimum.
- Against the fp32-plus-HNSW shape embedded engines typically pin (~6.5 KiB/row at 1536d): **~8–9×**.

The property that doesn't fit a table: the flat figure is **pinned and sufficient**. ivf's warm latency is conditional on the cache staying resident — squeeze memory and it slides toward its cold numbers. The scan's latency is the worst case, period: nothing to evict, zero object-store traffic, and width-invariant (the sweep reads 20.05/20.00/19.97/20.00 ms across nprobe — the scan ignores widths, which is also the proof it is serving). Cold start is one striped blob fetch (2 GETs at 100K), then cold-steady equals warm: 1.87 ms / 0 GET, $0.33/1M queries against $28.81/1M for the pre-index state.

A table that needs 0.99-class recall keeps `ivf`.

## Tests

**Codec and format.**

- `residency_is_the_nibble_plane_only` — per-row rate against the codec's rate at 1536 dims, so retaining Sq16 to refine with fails here.
- `persisted_form_round_trips_and_serves_identically` — byte round-trip **and** identical scan results, on both codecs and both sides of the row-blocking switch. A ruler that survived encode but was mis-sliced on read would still produce a plausible top-k.
- `blocked_scan_matches_per_node_scoring` — the blocked kernel reorders when query loads and reductions happen and nothing else; a row/query misalignment would score the wrong vectors while returning a plausible top-k.
- `bounded_top_k_matches_an_unbounded_scan` — the heap's eviction arms, which the two scans above miss by asking for every row.
- `decode_declines_malformed_input` — truncation, a graph magic, a plane one byte short, and an implausible row count declining before it reserves.

**Build, publish, serve.** `search_mode` is read from the process-wide config, which a unit test cannot set, so these drive the same seams the graph tests drive: the drain's build step and the reader's serve arm, called directly on a drained table.

- `flat_index_assembles_off_the_drained_cells` — assembles, clears the register floor, and the published plane's per-row residency is asserted against the codec's own rate. Its top-1 is cross-checked against the **ivf arm's** answer through the public `_id`, so a misaligned node→id map fails rather than returning a plausible row.
- `flat_search_serves_the_resident_plane_and_names_its_declines` — the serve arm answers on the ivf arm's cosine scale (leaking the scan's internal `-dot` would rank a drained non-match above an exact match), and every refusal — not hydrated, wrong column, wrong dim — names itself.
- `flat_search_declines_a_generation_that_published_a_graph` and `hnsw_search_serves_the_graph_and_names_its_declines` — the envelope's kind tag doing its job in both directions: neither arm answers from the other's index, and the graph arm's three formerly-collapsed refusals are asserted apart.
- `envelope_kind_tag_round_trips_and_declines_the_unknown` — both kinds through `encode_resident_envelope` and back, plus an unknown tag declining the *section* while the centroid graph above it survives and the cursor still consumes the payload. The compat story was asserted only for `Graph`, which is the arm that cannot regress; `Flat` is the one a future edit to `put_opt_section` would break silently.
- `the_index_owns_nothing_beside_the_scorer_and_the_id_map` — `resident_bytes()` reports what the *scorer* exposes, so it is blind by construction to a buffer owned beside the scorer — which is exactly what "retain the Sq16 plane to refine with" would be. Pinning the struct's own width closes that: a retained buffer needs a field, and a field changes `size_of`.
- The metric gate declining `L2Sq` and `NegDot` at build, on both arms.
- `flat_drain_build_publishes_an_index_that_serves` — `build_flat_index_ref` → register gate → publish, hydrated back as `Flat` and serving its rows.
- `flat_build_declines_name_their_reason` and `every_index_decline_reason_renders_its_cause` — the decline reasons carry the value that decided them, over all fifteen variants.
- `a_retired_config_key_is_rejected_and_names_its_replacement` — now every entry in `RETIRED_CONFIG_KEYS`, not just the first. The generic scan covers both today, but a refactor of `reject_retired_keys` could break detection for an untested key with the suite green.
- Per-index-type floors, and the incremental extend re-checking against the same floor the full build cleared, so an append cannot quietly lower the standard its resident graph was registered at.

The serving fixtures use distinct pseudo-random directions rather than the one-hot rows most fixtures here use. The register gate grades recall against an exhaustive Sq16 scan, and one-hot rows repeat each direction many times over, so the head of every ground-truth list is a block of exact ties that two rankings break differently — the tie count, not the codec, would decide whether an index registers at all.

2346 lib + integration tests pass; `clippy --all-features --all-targets`, `fmt` (under CI's `imports_granularity`/`group_imports` config), and `cargo check --target aarch64-unknown-linux-gnu --lib` (default and `--all-features`) clean under `-D warnings`. No dead-code allowances added.

## Scope and follow-ups

Not covered here:

- **No incremental arm.** A flat build always rebuilds. Extending would mean encoding new rows against a ruler fitted over the previous population — which the graph must do, because its stored neighbour lists describe distances measured under that ruler. A scan stores nothing between rows, so a refit invalidates nothing and strictly improves the codes; a rebuild is simply correct and costs one pass. That trade is cheap under `flat_max_docs` and is what would need revisiting if the ceiling rises.
- **Unfiltered queries only.** A filtered query falls through to `ivf`, as it does for the graph.
- **The 10M gate.** Everything measured here is 100K, which is the scale this index type is for; the ceiling defaults below 10M.
- **A comparison arm in the benchmark harness.** Since the index is config-selected, that is an arm setting the YAML rather than an adapter.
- **The 1.0 B/dim rung.** The matched-bytes A/B — a single fitted `sq8` plane vs the carved `sq4_residual`, same corpus, recall@1/10/100 + p50 + resident at k=10 and k=100 — decides what fills the slot. Whichever wins lands as a config variant behind the format's codec tag; nothing here needs a migration for it.
- **A mode ladder (`flat_hnsw_ivf`).** Naming the escalation past `flat_max_docs` as an explicit operator choice, per review; needs `flat_max_docs <= hnsw_max_docs` load-time validation and its own issue.


muralikpbhat (commented):
Reviewed across format/lifecycle, the serve path, config/scope, and the rename — with the codec-ladder question below as a first-class ask. The architecture verifies clean; **holding on the two inline blockers** (the missing metric gate and the replica fetch width).

**Verified clean:**
- **Format**: the envelope kind-tag rides the presence byte correctly (pre-tag envelopes read as Graph by construction; unknown kind declines the section, keeps the centroid graph, consumes the cursor — downgrade, not mis-parse), and `INFDFL01` decode is fully hardened (magic/dim/codec checks, `checked_mul` before every allocation, ruler validated finite/positive-step, real malformed-input asserts including the 2^60 row count).
- **Lifecycle**: GC retains the manifest-referenced blob and reclaims superseded ones with no unservable window; rebuild-always supersedes atomically at the manifest stamp; scan ids provably follow `gather_sq16_rows`'s node order; the graph's incremental extend re-checks its floor.
- **Scope**: the default path is byte-identical for a config that changes nothing (mode gates verified at drain and serve; graph envelopes encode identically); the rename commit is **mechanically verified pure** (rename map applied to the parent tree and diffed — only rustfmt reflow remains); the bench back door is fully gone with `public-api.txt` untouched; the four knobs are global config, outside the options hash; and the floor decoupling is real (`flat_register_floor` read at exactly one gate, `target_recall`'s width-law uses untouched — the filtered-recall side-effect is genuinely severed). Hygiene clean. One merge-mechanics note: the branch carries #651's pre-squash history (including the commit whose message we already flagged there), so **this should squash-merge**.
- The reasoned-decline design (`IndexUnavailable` through one `or_warn` seam) and the load-time rejection of the retired key are exactly the right responses to the #651 round — the findings below are about finishing them, not the direction.

**The codec-ladder ask (pre-recommendation gate, same standard as #521):** `flat_plane` offers `{sq4, sq4_residual}` — the ladder jumps from 0.5 B/dim straight to a *composite* 1.0, skipping the single-plane 1.0 rung. We have direct precedent that this matters: at 2 B/dim, the single uniform plane (Sq16) beat the coarse+residual construction (Sq8+8) on recall *and* latency — finer effective step, one pass, no second plane (#521). The same algebra at 1.0 B/dim: a single fitted 8-bit plane is 256 uniform levels in one pass vs `sq4_residual`'s 16×16 two-stage combos (effective step ~1/240) in two — and the SQ8 plane + VNNI kernels already exist. Before `sq4_residual` becomes the recommended flat rung, please add `flat_plane = sq8` and run the matched-bytes A/B through the same config-only three-arm harness this PR built: same corpus, recall@1/10/100 + p50 + resident, **at k=10 and k=100** (scan cost shows at high k). If the single plane wins — as the precedent predicts — `sq4_residual` may not deserve to exist; if it doesn't, the numbers justify it. Either way the 1.0 B/dim recommendation shouldn't ship untested against its natural competitor.

**Blockers inline**: the missing non-cosine gate (silent wrong-metric serving the register floor structurally cannot catch) and the replica under-fetch (`< k` results whenever replication is on). Should-fixes inline: the per-query warn flood, the self-reporting residency tripwire, the Flat-kind envelope/e2e test gap, the retired-key test claim, the register-floor migration edge, and the publish log string. Minor, no anchor: the `config.yaml` flat-knob comments are interleaved (the scale-ceiling paragraph sits above `flat_plane`, leaving `flat_max_docs` bare), and the `// Slack >= target collapses…` comment above the floor-range validation cases describes the retired mechanism.

The measured table reproducing #651's direct-construction numbers through pure config selection is exactly the productization proof this needed — with the two blockers fixed, the sq8 arm measured, and the test gaps closed, this is a strong addition.

src/supertable/query/vector.rs
```diff
@@ -2458,6 +2838,95 @@ impl SupertableReader {
...
+                    .into_iter()
+                    .filter_map(|(node, dist)| {
+                        // Shift the scan's `−dot` onto the ivf arm's `1 − dot`
+                        // cosine scale so a cross-arm merge compares like with
+                        // like. Monotonic in `dist`, so intra-arm order is
+                        // unchanged and the public `score` stays non-negative.
+                        Some(SuperfileHit {
+                            superfile: SuperfileUri(Uuid::nil()),
+                            local_doc_id: 0,
+                            score: 1.0 + dist,
```
muralikpbhat, line 2988: **Blocker — the non-cosine decline didn't land, and the register gate can't catch it.** `assemble_flat_sections` gates on the rerank codec but never reads `vc.metric`, and this scoring maps `1.0 + dist` (the cosine shift) unconditionally. Because `probe_recall` grades −dot against −dot, an L2Sq/NegDot column **builds, clears the floor self-consistently, registers, and serves inner-product rankings** — wrong neighbors at the right latency, exactly the silent failure the gate exists to prevent. The centroid-graph router just below gates on `Metric::Cosine` for this reason; this arm needs the same gate with an `IndexUnavailable` reason (the graph arm inherits the same gap — worth fixing both while here).

src/supertable/query/vector.rs
```diff
@@ -2458,6 +2838,95 @@ impl SupertableReader {
...
+    /// scan.
+    ///
+    /// Peer of [`Self::hnsw_search`]. It has no beam, no `ef`, and no refine
+    /// width: the scan visits every row and the codes' own ranking IS the
+    /// answer. That is also why there is no over-fetch for boundary replicas —
+    /// a replicated row appears once per hidden cell in the graph's node
+    /// space, but this index is built from the same gathered rows and
+    /// `top_k_ascending` collapses duplicate stable ids either way, so the
+    /// fetch width is `k`.
+    async fn flat_search(
```
muralikpbhat, line 2911: **Blocker (dormant at the shipped default) — replica under-fetch returns fewer than k.** The flat plane is gathered from the same rows the graph is, where boundary replicas appear once per cell with the **same stable id**, and `top_k_ascending` collapses duplicates — so fetching exactly `k` nodes yields `< k` distinct hits whenever `drain_replica_target_factor > 1`. The graph arm over-fetches `k_fetch = ceil(k · factor)` for precisely this; this arm needs the same widening. The justifying comment ("collapses duplicate stable ids either way, so the fetch width is k") inverts its own conclusion.

src/supertable/query/vector.rs
```diff
@@ -3061,9 +3538,25 @@ impl SupertableReader {
...
+        // only, and only when a valid index for THIS column is resident. The
+        // drain declines to build one above `flat_max_docs` or below the
+        // register floor, and either way the query serves ivf.
+        if !filtered
+            && hidden_vector_index
+            && vcfg.search_mode == config::VectorSearchMode::Flat
+            && let Some(hits) = self
+                .flat_search(column, query, k)
+                .await?
+                .or_warn("flat search")
```
muralikpbhat, line 3642: This `or_warn` fires on **every query** in legitimate steady-state fallback — a `search_mode = flat` table over `flat_max_docs`, or pre-drain, warns per request at serving QPS. The graph arm's equivalent (`warn_hnsw_no_resident_graph`) dedupes per `(diagnosis, column)` via a `OnceLock<HashSet>` for exactly this reason; the reasoned-decline design is right, it just needs the same dedup.

src/superfile/vector/flat.rs
```diff
@@ -0,0 +1,714 @@
...
+    }
+
+    /// The resident footprint must be the nibble plane and its ruler, and
+    /// nothing else — in particular no Sq16 plane.
+    ///
+    /// This is the index's entire reason for existing, and it is the kind of
+    /// property that regresses silently: retaining the Sq16 codes to refine
+    /// with would leave every test above green while quadrupling residency.
+    #[test]
+    fn residency_is_the_nibble_plane_only() {
```
muralikpbhat, line 687: `residency_is_the_nibble_plane_only` measures the scorer's **self-report**: `resident_bytes()` sums what `Sq4Scorer::parts()` exposes, and `minimum_bytes()` recomputes the same expectation — both blind to a buffer retained **outside** the scorer's parts. The exact regression this test's doc-comment names (retaining the Sq16 plane to refine) would keep it green while residency quadruples. Make it measure the resident struct's total (or assert the index type holds nothing but the scorer + ids), not the scorer's own accounting.

src/superfile/vector/hnsw.rs
```diff
@@ -2553,18 +2591,26 @@ pub(crate) fn decode_graph_bundle(raw: &Bytes, mmap_backed: bool) -> Option<Grap
...
     // (the full striped blob) frees.
-    let data_bundle = if c.take(1)?[0] == 0 {
+    let tag = c.take(1)?[0];
+    let data_bundle = if tag == 0 {
         None
     } else {
+        // An unknown kind is a payload written by a newer build. Decline the
+        // SECTION rather than the whole envelope: the centroid graph above it
+        // is still valid and still worth serving.
+        let kind = PayloadKind::from_tag(tag);
```
muralikpbhat, line 2610: Test gap on the envelope kind machinery: nothing encodes `PayloadKind::Flat` through `encode_resident_envelope` and hydrates it back to `ResidentIndexKind::Flat`, and no test feeds an unknown tag (≥3) to pin "decline the section, keep the centroid graph, cursor consumed correctly." The compat story is asserted only for the Graph arm — and there's no end-to-end drain-and-serve test under `search_mode = flat` anywhere, so a future edit to `put_opt_section`/decode breaks flat hydration with every test green (the symptom being every flat table silently serving ivf).

src/config/mod.rs
```diff
@@ -62,10 +62,18 @@ const EMBEDDED_DEFAULT: &str = include_str!("config.yaml");
 /// would otherwise be handed that default back with nothing to indicate their
 /// setting had stopped applying — resident memory and per-open cost changing
 /// under them on an upgrade. Failing the load is the loud version.
-const RETIRED_CONFIG_KEYS: &[(&str, &str)] = &[(
-    "vector.hnsw_sq8_walk",
-    "vector.hnsw_plane — `sq8` is the old `true`, `sq16` the old `false`",
-)];
+const RETIRED_CONFIG_KEYS: &[(&str, &str)] = &[
+    (
+        "vector.hnsw_recall_slack",
```
muralikpbhat, line 67: The PR body's Tests section claims `a_retired_config_key_is_rejected_and_names_its_replacement` covers this newly retired key — it doesn't; the test still exercises only `hnsw_sq8_walk`. The generic scan does cover it today, but a refactor of `reject_retired_keys` could break detection with the suite green, handing an operator's `hnsw_recall_slack: 0.02` a silent 0.98 — the exact drift the retirement mechanism exists to prevent (and the same lesson as the walk knob). One more case in the existing test closes it.

src/config/mod.rs
```diff
@@ -329,12 +337,16 @@ const DEFAULT_VECTOR_HNSW_M0: usize = 0;
...
-/// graph (accepting the shortfall for the latency win) before giving up and
-/// serving ivf. The register floor is `target_recall - this`. Widen it to keep
-/// more hard high-dim tables on the graph; tighten it to demand near-target
-/// recall from the graph or step aside to ivf.
-const DEFAULT_VECTOR_HNSW_RECALL_SLACK: f64 = 0.01;
+/// Default register floor for the graph index: the value the shipped
+/// `target_recall - hnsw_recall_slack` pair produced (0.99 - 0.01), so
+/// behaviour is unchanged by stating it directly.
+const DEFAULT_VECTOR_HNSW_REGISTER_FLOOR: f64 = 0.98;
+
```
muralikpbhat, line 344: Migration edge worth stating (and ideally warning on): the 0.98 default equals the retired derivation only for **untouched** configs. A config that overrode `target_recall` (say 0.95 → old floor 0.94) but never set the slack key gets no retired-key rejection — nothing retired is present — and its floor silently snaps to 0.98: a graph calibrating at 0.96 de-registers at the next full drain and the table quietly serves ivf. A release note plus a drain-time warn when the floor exceeds `target_recall` would make the snap visible.

src/supertable/writer.rs
```diff
@@ -8211,20 +8218,21 @@ fn graph_population_key(manifest: &ManifestSnapshot) -> u64 {
 }
 
 /// Encode + PUT a data bundle as a graph section, logging the outcome.
-async fn publish_hnsw_blob(
+async fn publish_resident_index(
```
muralikpbhat, line 8451: `publish_resident_index` now takes a `PayloadKind` but its success log still says "hnsw: published graph section" — a flat publish logs as a graph publish, which contradicts this PR's own observability theme ("the numbers look like the mode you did not ask for," here in the log stream). Thread the kind into the message.

muralikpbhat:
One design suggestion on top of the review, best taken **before** this merges because it's a pure rename now and a retired-key migration forever after:

**Name modes as their fallback chains, and make the graduation an explicit mode.**

`hnsw_ivf` already reads as "graph while eligible, else ivf." The new mode behaves the same way — flat while eligible, else ivf — but is named `flat`, which hides its second half. And the discussion above about what should happen past `flat_max_docs` (auto-escalate to a graph vs fall to ivf) is really a question the *operator* should answer at config time, not a threshold behavior we pick for them. Both resolve with one naming scheme:

- `ivf` — the spine (unchanged)
- `flat_ivf` — this PR's mode, renamed: flat scan under `flat_max_docs`, else ivf
- `hnsw_ivf` — unchanged
- `flat_hnsw_ivf` — the explicit three-step graduation: flat under `flat_max_docs` → graph up to `hnsw_max_docs` → ivf beyond

The ladder mode makes the RAM consequence *chosen* rather than sprung: an operator selecting `flat_hnsw_ivf` has opted into the graph's residency when the corpus outgrows the scan, so the escalation that would be a surprise under `flat` is the contract under the ladder. Everything it needs already exists in this PR — the envelope kind enum, the single resident slot, per-index register floors (a graph failing its floor still drops to ivf), and rebuild-always on the flat side makes the rung transition at drain trivially clean. The only new requirement is load-time validation that `flat_max_docs <= hnsw_max_docs` when the ladder is selected.

`flat_hnsw_ivf` itself can be a follow-up — but `flat` → `flat_ivf` is worth doing in this PR while the name has never shipped.

src/supertable/query/vector.rs
ekechinwokah: blocker, agreed, fixed — and your diagnosis of *why the gate can't catch it* is the part that mattered. `probe_recall` grades the 4-bit scan against an Sq16 exhaustive scan and both score −dot, so a mis-metriced index measures perfect recall against a ground truth that is wrong in the same direction. the gate is self-consistent and therefore useless here.

`IndexUnavailable::MetricUnsupported { column, metric }`, declined in `assemble_flat_sections` and `assemble_hnsw_sections` — you were right that the graph arm inherits the same gap, so both. also re-checked at serve in `flat_search` and `hnsw_search`: a plane published by *this* binary can no longer reach the scan on a non-cosine column, but a graph published by an earlier one can, and it ranks by inner product regardless of what the column says.

restricted to Cosine rather than Cosine+NegDot, matching the centroid-graph router's existing gate. under NegDot the *order* is actually right — the metric is the inner product — but the published score carries a meaningless `+1`, and admitting a metric whose score scale is wrong to save a decline is not a trade worth making silently. if NegDot wants serving later it should be a deliberate extension with the scale fixed.

test: `flat_build_declines_name_their_reason` now walks `[L2Sq, NegDot]` x `[flat, hnsw]` and asserts `MetricUnsupported` naming the column and the metric. no drain needed — the decline lands before any row is read.

src/supertable/query/vector.rs
ekechinwokah: blocker, agreed, fixed — and the sentence you quoted is the tell. "collapses duplicate stable ids either way, so the fetch width is k" reads its own mechanism backwards: the collapse is why the width has to *grow*.

`replica_fetch_width(k)` is now one function both arms call, rather than the graph arm carrying `k_fetch = ceil(k · factor)` and the flat arm a bare `k`. the doc comment on it says what it is for and carries the build-time caveat that was previously buried in the graph arm — lowering `drain_replica_target_factor` below the value a resident index was built at under-counts its replicas until the next full rebuild.

what I could not do is test the behaviour. the factor is process-global config, same seam as `search_mode`, so the widened case is unreachable from a test process; `replica_fetch_width_is_identity_at_the_shipped_factor` pins the identity case and the fact that the width is a shared function of `k` rather than two arms each deciding. flagging that rather than claiming coverage I don't have — it is on the list of things per-table selection would make testable.

src/supertable/query/vector.rs
ekechinwokah: already fixed before this review landed, at `66ea19ac` — and fixed the way you'd want rather than by adding a second dedup beside the first.

`or_warn` now dedupes internally, keyed on the rendered message. that keeps one place deciding how a decline is reported, which is the whole point of the type, and it covers the graph arm too — `warn_hnsw_no_resident_graph` is deleted, because everything it diagnosed by guessing from `has_graph_ref` is now stated by the reason itself. keyed on the message rather than `(diagnosis, column)` so two columns declining for the same reason each get said once.

you spotted a regression I introduced, not a gap I left: the graph arm HAD the dedup and my conversion would have taken it away. the set stays small because a serving decline's reasons carry only column names and dimensions — a build decline can carry a measured recall, but that path runs once per drain.

src/superfile/vector/flat.rs
ekechinwokah: agreed, and the framing is exact: `resident_bytes()` sums `Sq4Scorer::parts()` and `minimum_bytes()` recomputes the same expectation, so the two assertions agree with each other and neither can see a buffer held beside the scorer — which is precisely what "retain the Sq16 plane to refine with" would be.

took your second option. `the_index_owns_nothing_beside_the_scorer_and_the_id_map` pins `size_of::<Sq4FlatIndex>()` against the sum of its declared fields, so the index type is asserted to hold nothing but the scorer, the id map and its two descriptors. any retained buffer has to be reachable from the struct and any owning handle to one costs inline bytes, so a field appears and the test fails naming the reason.

measuring the resident struct's true total was the other option and I didn't take it: it needs either a deep-size crate or an RSS sample, and an RSS assertion at 1536x1000 is a flaky test that fails for unrelated reasons. the struct-width pin catches the same regression at compile-time cost and cannot flake.

src/superfile/vector/hnsw.rs
ekechinwokah: agreed on both halves, and the second half was the real gap.

`envelope_kind_tag_round_trips_and_declines_the_unknown` covers both kinds through `encode_resident_envelope` and back on both the mmap and heap paths, then feeds tags `[3, 9, 255]` and asserts what the code claims but nothing checked: the SECTION declines, the centroid graph above it survives, the header still reads, and the cursor consumes the payload's length prefix and body so nothing after it mis-slices.

your point about the asymmetry is why this matters — `Graph` is the arm that cannot regress, since `1` is what the presence byte has always meant. `Flat` is the arm a future edit to `put_opt_section` breaks, and it breaks into "every flat table silently serves ivf", which no other test would notice.

on the end-to-end gap: `flat_drain_build_publishes_an_index_that_serves` now drives drain -> `build_flat_index_ref` -> register gate -> publish -> hydrate -> serve and asserts the plane answers, and the serve arms are driven directly on a drained table. it is not a test under `search_mode = flat_ivf`, and it cannot be — `config::global()` is a `OnceLock` and every `tests/` binary shares one config, so no test process can select the mode. that is the same limitation the graph tests live with and the reason per-table selection is worth doing.

src/config/mod.rs
ekechinwokah: you're right and the PR body was wrong — it claimed the test covered the new key and it covered only `hnsw_sq8_walk`. body corrected.

the test now loops `RETIRED_CONFIG_KEYS` itself rather than adding one hardcoded case, so an entry added to the table without a test case fails immediately instead of being covered only by the generic scan. it asserts the error names both the retired key and the first token of its replacement.

src/config/mod.rs
ekechinwokah: agreed — the config that *set* the retired key is rejected; the config that lowered `target_recall` and never set the slack key is exactly the one that slips through, because nothing retired is present to reject.

warned, at load rather than at drain. `Config::validate` has both numbers in hand and says it once per process instead of once per drain, and it fires before any table is touched: "flat_register_floor = 0.98 is above vector.target_recall = 0.95 ... set it explicitly - it is no longer derived from target_recall."

a warn and not a rejection deliberately: holding the resident index to a higher bar than the ivf laws is a legitimate thing to want, and an unnoticed migration looks identical to it from inside the config. the release note is yours to place - happy to draft it if you want it in this PR.

src/supertable/writer.rs
ekechinwokah: fixed. `PayloadKind::label()` -> `"hnsw"` / `"flat"`, threaded into both the success and the failure message, so a flat publish reads as one.

you named the reason better than I would have: this is the same class as the reasoned declines, one layer out. an operator reading "hnsw: published graph section" on a flat table is being told the mode they did not ask for, in the log stream instead of in the latency.

ekechinwokah:
agreed on the rename and it's in at `65e99c2b` as `flat_ivf`. your now-or-never framing is the argument that decides it — a rename today, a retired-key migration forever after — and the consistency point is right on its own: `hnsw_ivf` spells out the chain it serves and `flat` hid half of its own behaviour in the one place an operator looks first. the `search_mode` block in `config.yaml` gained the mode's entry while I was there, which it never had.

`flat_hnsw_ivf` I've deliberately not built. I think you're right that it belongs, and right about why — the escalation past `flat_max_docs` is a question about who owns the RAM consequence, and a ladder makes the operator own it explicitly instead of the engine springing it. but that is a design decision with a public-surface shape, and it should not ride a PR whose subject is test coverage. the pieces you listed are all in place (kind enum, single resident slot, per-index floors, rebuild-always on the flat rung), and the `flat_max_docs <= hnsw_max_docs` load-time validation is the one new requirement — noted so the follow-up issue starts from your comment rather than from scratch.

src/supertable/query/vector.rs
ekechinwokah: blocker, agreed, fixed at `65e99c2b` — and your diagnosis of *why the gate can't catch it* is the part that mattered. `probe_recall` grades the 4-bit scan against an Sq16 exhaustive scan and both score −dot, so a mis-metriced index measures perfect recall against a ground truth that is wrong in the same direction. the gate is self-consistent and therefore useless here.

`IndexUnavailable::MetricUnsupported { column, metric }`, declined in `assemble_flat_sections` and `assemble_hnsw_sections` — you were right that the graph arm inherits the same gap, so both. also re-checked at serve in `flat_search` and `hnsw_search`: a plane published by *this* binary can no longer reach the scan on a non-cosine column, but a graph published by an earlier one can, and it ranks by inner product regardless of what the column says.

restricted to Cosine rather than Cosine+NegDot, matching the centroid-graph router's existing gate. under NegDot the *order* is actually right — the metric is the inner product — but the published score carries a meaningless `+1`, and admitting a metric whose score scale is wrong to save a decline is not a trade worth making silently. if NegDot wants serving later it should be a deliberate extension with the scale fixed.

test: `flat_build_declines_name_their_reason` now walks `[L2Sq, NegDot]` x `[flat, hnsw]` and asserts `MetricUnsupported` naming the column and the metric. no drain needed — the decline lands before any row is read.

src/supertable/query/vector.rs
ekechinwokah: blocker, agreed, fixed at `65e99c2b` — and the sentence you quoted is the tell. "collapses duplicate stable ids either way, so the fetch width is k" reads its own mechanism backwards: the collapse is why the width has to *grow*.

`replica_fetch_width(k)` is now one function both arms call, rather than the graph arm carrying `k_fetch = ceil(k · factor)` and the flat arm a bare `k`. the doc comment on it says what it is for and carries the build-time caveat that was previously buried in the graph arm — lowering `drain_replica_target_factor` below the value a resident index was built at under-counts its replicas until the next full rebuild.

what I could not do is test the behaviour. the factor is process-global config, same seam as `search_mode`, so the widened case is unreachable from a test process; `replica_fetch_width_is_identity_at_the_shipped_factor` pins the identity case and the fact that the width is a shared function of `k` rather than two arms each deciding. flagging that rather than claiming coverage I don't have.

ekechinwokah: blocker, agreed, fixed at `65e99c2b` — and the sentence you quoted is the tell. "collapses duplicate stable ids either way, so the fetch width is k" reads its own mechanism backwards: the collapse is why the width has to *grow*.

`replica_fetch_width(k)` is now one function both arms call, rather than the graph arm carrying `k_fetch = ceil(k · factor)` and the flat arm a bare `k`. the doc comment on it says what it is for and carries the build-time caveat that was previously buried in the graph arm — lowering `drain_replica_target_factor` below the value a resident index was built at under-counts its replicas until the next full rebuild.

what I could not do is test the behaviour. the factor is process-global config, same seam as `search_mode`, so the widened case is unreachable from a test process; `replica_fetch_width_is_identity_at_the_shipped_factor` pins the identity case and the fact that the width is a shared function of `k` rather than two arms each deciding. flagging that rather than claiming coverage I do not have.

muralikpbhat:
Round-2 verification done on my side: both blockers are fixed and code-verified (`metric_supported` on both resident indexes; `replica_fetch_width` shared by both arms), all eight inline items are closed — the struct-width residency pin is the elegant version of the fix, and the retired-key test looping `RETIRED_CONFIG_KEYS` is better than what I asked for — `flat_ivf` landed, CI is green.

One item from the review summary went unanswered, and it hardens into a merge position: **`sq4_residual` should not ship unless it beats a single-plane `sq8` at the same bytes.**

The precedent is direct: at 2 B/dim, the single uniform plane (Sq16) beat the coarse+residual construction (Sq8+8) on recall *and* latency — finer effective step, one pass, no second plane. The same algebra at 1.0 B/dim says a single fitted 8-bit plane (256 uniform levels, one pass, kernels already shipped) should beat `sq4_residual`'s 16×16 two-stage combos in two passes — and this PR's own numbers hint at the pass cost (`sq4_residual` at 3.57 ms vs bare `sq4` at 1.49 ms is 2.4× for 2× the bytes).

Two acceptable resolutions, either unblocks approval:

1. **Run the matched-bytes A/B** through the three-arm config harness this PR already built — `flat_plane = sq8` (new arm) vs `sq4_residual`, same corpus, recall@1/10/100 + p50 + resident, at k=10 and k=100. If `sq4_residual` wins, the numbers justify it and it merges as-is. If it loses, drop it.
2. **Carve `sq4_residual` out of this PR** and land everything else now (the mode, `sq4`, the format, the tests). Honest consequence to state if so: bare `sq4` (0.937) doesn't clear the default 0.98 floor, so `flat_ivf` ships as a mode that registers only when the operator lowers `flat_register_floor` deliberately — which is arguably the right posture until a floor-clearing plane has proven itself.

Everything else in this PR is approve-ready — this is the one open question.

ekechinwokah:
took resolution 2 — `sq4_residual` is carved out at `60fa15f0`, and your direction call is accepted on its merits: `sq16_vs_sq8residual_kernel_recall` is a checked-in theorem that the single uniform plane beats coarse+residual at equal bytes, and shipping a default that the matched-bytes A/B may retire is a migration for nothing. we can come back to a 1.0 B/dim rung when a measured winner claims it.

what the carve looks like:

- the `flat_plane` knob is gone entirely rather than reduced to one variant — a knob that can only say `sq4` is dead config. the drain builds the bare plane, stated at the build site with the reasoning.
- the scorer, the persisted form's codec tag, and the round-trip tests keep the residual capability. the A/B follow-up re-adds a config variant, not a format; an `sq2` or `sq8` rung later is the same shape.
- the residual arm's measured row (146 MiB / 0.991 / 3.57 ms) stays in the PR body as data — it is the number the A/B has to beat.

one consequence stated honestly, where I land differently from your "arguably the right posture": `flat_register_floor` now defaults to **0.80**, not 0.98. with bare `sq4` as the only rung, a 0.98 default is not a posture — it is a mode that structurally cannot register, ever, because 16 levels per coordinate tops out around 0.93. the floor's own doc says its job is catching a BROKEN plane (degenerate ruler, mis-sliced section, recall collapsed toward random), not a coarse codec; 0.80 is a tripwire a healthy plane clears comfortably and a broken one cannot reach. an operator who wants a bar raises it — that is exactly what the per-index-type floor you reviewed is for. `hnsw_register_floor` stays at 0.98, where the graph can actually live.

the A/B itself and the `flat_hnsw_ivf` ladder are both named in the PR body's follow-ups now, with your `flat_max_docs <= hnsw_max_docs` validation noted so the ladder issue starts from your comment.

gates on the carve: 2355 lib + integration tests, clippy `--all-features --all-targets`, fmt, both aarch64 checks, all clean under `-D warnings`.

muralikpbhat (approved):
Approving. The carve is verified end to end, and it resolves the one question this was held on:

- **`flat_plane` is gone entirely** rather than reduced to a one-value knob — right call, dead config is worse than no config. The drain builds the bare plane with the reasoning stated at the build site (`with_residual = false`, citing the checked-in single-plane-beats-coarse+residual theorem).
- **The format keeps the residual capability** (codec tag, scorer, round-trip tests), so the matched-bytes A/B follow-up re-adds a config variant, not a format — and the residual arm's measured row stays in the body as the number that A/B has to beat. Exactly the right shape for "come back when a measured winner claims the 1.0 B/dim rung."
- **On the 0.80 floor: your pushback is correct and I'm updating my position.** I'd framed keeping 0.98 as a deliberate opt-in posture — but with bare `sq4` as the only rung topping out ~0.93, a 0.98 default isn't a posture, it's a mode that structurally cannot register. The floor's documented job is catching a *broken* plane, not a coarse codec; 0.80 does that job, and the per-index floors exist precisely so an operator who wants a bar sets one. `hnsw_register_floor` staying at 0.98 where the graph can actually live completes the picture.
- Round-2 items all previously verified on this branch: both blockers fixed (`metric_supported` on both resident indexes; shared `replica_fetch_width`), the struct-width residency pin, `or_warn` dedup, the Flat-kind envelope + unknown-tag tests, the e2e drain→publish→hydrate→serve test, the retired-key test looping the table, the load-time floor-migration warn, the labeled publish log, and the `flat_ivf` rename. CI green including both aarch64 checks under `-D warnings`.

Follow-ups named in the body are the right ones: the matched-bytes A/B for the 1.0 B/dim rung, and the `flat_hnsw_ivf` ladder with the `flat_max_docs <= hnsw_max_docs` load-time validation. One candidate worth adding to the A/B matrix when it runs: alongside `sq8` and `sq4_residual`, an optimal-scalar-plus-sign shape — (b−1) bits of per-coordinate optimal scalar quantization after rotation plus a 1-bit sign quantizer on the residual — which recent theory (TurboQuant, arXiv:2504.19874) argues is the inner-product-optimal way to spend a bit budget, and which is close to machinery this codebase already has.

ekechinwokah:
closing this out with the full measurement set before merge — the serving crossover, what each mode holds in RAM, the cold economics, and the drain cadence the mode wants. all on the branch head (carve at `60fa15f0`, sweep points at `1adea260`), dbpedia-1536 (OpenAI ada-002) over the Azure store, 8-thread EPYC 9V74 with full AVX-512, shipped defaults on both arms — `search_mode` is the only YAML line that differs.

**where the latency crossover sits: ~120–140K rows.** five scales, post-compact steady state:

| live rows | `flat_ivf` p50 / R@10 | `ivf` p50 / R@10 | faster |
| --- | --- | --- | --- |
| 106K | **1.63 ms** / 0.944 | 2.28 ms / 0.998 | flat |
| 141K | 2.41 ms / 0.937 | **1.39 ms** / 0.990 | ivf |
| 189K | 3.47 ms / 0.955 | **1.59 ms** / 0.995 | ivf |
| 336K | 6.90 ms / 0.946 | **2.51 ms** / 0.997 | ivf |
| 992K | 20.1 ms / 0.938 | **6.2 ms** / 0.988 | ivf |

the two curves have different shapes, which is the real finding: the scan is deterministic and linear at **~18–20 µs per 1000 rows**, while ivf's law-served p50 is essentially scale-flat in a 1.1–2.5 ms band up to ~350K (non-monotonic across corpus draws — law stamping noise) before rising to ~6 ms at 1M. the crossover is where the line pierces the band. one honest asterisk: this is the default-config comparison; at matched recall (~0.94), ivf at nprobe=8 serves 0.83 ms at 100K — warm ivf wins matched-recall latency everywhere. the flat case below the crossover is "fastest default AND smallest"; above it, it is memory and cold behavior only.

**what serving actually holds (RSS).** the flat figure is pinned and sufficient; the ivf figure is a small pin plus a warm working set that is effectively its hidden index:

| rows | `flat_ivf` pinned (plane + manifest) | `ivf` (pin + warm working set) |
| --- | --- | --- |
| 106K | **153 MiB** (77 + 76) | 86 MiB + ~337 MiB |
| 141K | ~180 MiB (103 + 76) | ~87 MiB + ~380 MiB |
| 336K | ~320 MiB (245 + 76) | ~90 MiB + ~437 MiB |
| 992K | **841 MiB** (726 + 115) | 109 MiB + ~3.05 GiB |

stating the multiple honestly, because it depends on the comparison: **2–4× floor-vs-floor** (against ivf's minimum warm working set — 2.2× at 100K, 3.6× at 1M), **4× per row** against any resident 2 B/dim plane and ~6–7× against the graph bundle, and ~9–10× against what the bench run actually held (flattering — that cache was not tuned down). the property that doesn't fit the table: squeeze memory and ivf's warm p50 slides toward its cold numbers; the scan's p50 does not move, because there is nothing to evict.

**cold economics** (316K point, representative): the plane's cold story is one striped blob fetch at open — cold-1st 756 ms / 4 GET, after which cold-steady equals warm: **6.6 ms / 0 GET, $1.59/1M queries**. ivf steady-cold is **724 ms / 165 GET per query, $117.81/1M** until its cache rebuilds. for scale-to-zero deployments, where cold is the steady state, this dwarfs the warm crossover.

**drain cadence.** the undrained-delta penalty is ~constant, not proportional — fixed per-superfile fanout over the commit wave, not delta-row scan cost (the rows themselves are ~0.2 ms at the scan's rate):

| rows | post-drain p50 | post-delta (+~6% undrained) | penalty |
| --- | --- | --- | --- |
| 106K | 1.56 ms | 3.21 ms | +1.65 ms |
| 141K | 2.37 ms | 4.03 ms | +1.66 ms |
| 189K | 3.21 ms | 5.21 ms | +2.00 ms |
| 336K | 6.16 ms | 8.03 ms | +1.87 ms |
| 992K | 19.1 ms | 22.7 ms | +3.58 ms |

so cadence should bound **undrained commits, not rows**: draining every ~4–8 commits keeps post-delta within ~1.5–2× the drained p50, and recall does not suffer in between (0.95 post-delta — the delta is served exactly). a drain on the flat arm costs ~32 s at 100K and ~200 s at 1M wall, of which the plane rebuild is roughly 0.2 s per 1000 rows (rebuild-always, one pass).

**operational takeaways.** `flat_max_docs` is already the switch — past it the mode serves ivf automatically. set it ~150K for always-fastest-default semantics; leave it at 1M when the deployment is memory-bound or scale-to-zero, where the scan above the crossover is slower but bounded (≤20 ms) and keeps the pinned-footprint and zero-GET properties. the 0.80 register floor did its job at every scale: the bare plane registered at shipped defaults with no overrides (gate recall 0.9045–0.9345 across the sweep), which the previous 0.98 default made impossible.

Issue #630, opened 2026-08-24
vector: build and persist the centroid-graph router at drain time


muralikpbhat:
Follow-up to #588 (centroid-graph ivf router). Makes the `centroid_graph` router **self-driving**: its HNSW-over-fine-centroids is kept correct across the table's whole life, built once at drain, persisted with the manifest, and loaded (mmap) by every query — never rebuilt on the hot path.

## The problem
#588's router builds its graph over the resident fp32 fine centroids **once per handle, at the first query**, in a build-once cache that is **never invalidated**. Two consequences:

1. **Correctness — the graph goes stale after a drain.** Its nodes map to `(superfile, flat cluster id)` valid *only* for the generation they were built against. A drain/compaction rewrites cells (split/merge/supersede) and renumbers fine clusters, so the cached graph then routes against superseded ids — wrong clusters, collapsed recall, or an out-of-range panic. Measured on a streaming recall-over-time bench that drains at every checkpoint: recall@10 = **0.774** at the first checkpoint (fresh graph), then **0.008**, **0.003** as later drains leave the graph stale.
2. **Latency — a per-handle first-query build.** That first query pays the build (the residual cold q0 gap in #588: LAION 100M q0 19.4 s vs `stamped`'s 15.4 s; steady cold already at parity ~5.4 s), and every node/handle/restart repeats it.

## The design
**Generation-stamped, persisted as a slow-state section, mmap-loaded — the query never builds (except legacy).**

- **Invalidation is a generation stamp, not a cache-clear.** Every table change bumps the hidden manifest generation (`manifest_id`). The graph is tagged with the `(generation, column)` it was built for. A query reads the generation off the manifest snapshot it *already holds* — a field access, no lookup, no lock, no poll — and uses the graph iff the tag matches. A mismatch means the table changed; the query loads/builds the graph for its own generation. No "clear on drain", no background poller, no per-query storage poll. (Stamp keyed by column too: a table with two vector columns must not share one graph.)

- **The graph lives with the manifest.** At drain, once the fine centroids settle at the final generation, the graph is built once and **published as a section in the slow-vector-state file, alongside the centroid section** — the same per-generation CAS-object model that already carries the centroids. It reuses the existing HNSW serialization (self-describing byte section) and the memory-mapped bundle path (zero-copy `Plane::Shared`). Only the graph topology + `node_map` are persisted; the `Fp32Scorer` is reconstructed from the resident centroid section on load, so the centroids aren't duplicated.

[...]

PR #673, opened 2026-08-28
vector: self-driving centroid-graph ivf router

muralikpbhat:
Implements #630. Turns #588's opt-in `centroid_graph` router into a **self-driving** router: its HNSW-over-fine-centroids is kept correct across the table's life, built once at drain, persisted with the manifest, mmap-loaded by every query (never rebuilt on the hot path), calibrated per-table by measured recall, metric- and codec-aware, and scale-gated.

> **Nothing changes by default.** `search_mode` defaults to `ivf` and `ivf_router` defaults to `stamped`; `centroid_graph` and the new `auto` are opt-in. This PR does not touch `search_mode` (the hnsw↔ivf axis) and does not flip any default — it makes the *ivf router* self-driving **when you opt into `auto`**. Defaulting `auto` is a tracked follow-up (#714).

## Two axes (this PR is only the second)

| Axis | Config | Default | This PR |
|---|---|---|---|
| index type | `search_mode` (`ivf` / `hnsw_ivf`) | `ivf` | **untouched** |
| ivf router | `ivf_router` (`stamped` / `centroid_graph` / `auto`) | `stamped` | adds calibration + `auto` |

- **`stamped`** (default) — grid routing + the manifest's stamped per-cell width. Unchanged.
- **`centroid_graph`** (opt-in) — score fine centroids via an HNSW over resident fp32 centroids, read the calibrated fanout of clusters. A cold-read / warm-compute win at scale.
- **`auto`** (opt-in) — pick `centroid_graph` per hidden-vector table, at query time, **only where it wins**: the calibrated fanout is concentrated (≪ cluster count) **and** the corpus is at/above the scale floor. Otherwise `stamped`.

**Not in scope:** the `search_mode` transitions (the `hnsw_ivf`→`ivf` fallback above `hnsw_max_docs` already exists, unchanged), and any `flat → hnsw_ivf → ivf` auto-ladder — a future `search_mode = auto` that composes on this resolver.

### 1. Correct invalidation — no more stale-graph collapse
#588 cached the graph build-once and never invalidated it, so after a drain/compaction renumbered the fine clusters the cached graph routed against superseded ids (measured collapse: recall@10 **0.774 → 0.008 → 0.003** across successive drains). The cache is now stamped `(manifest generation, column)`; a query reads the generation off the snapshot it already holds and uses the graph only on a match, else loads/rebuilds for its own generation. Single-flight build, no per-query poll.

### 2. Persisted graph in object storage, mmap-loaded — build once, query never builds
Built at the drain/compaction settle and published as a **per-generation section in the slow-vector-state file** (reusing the existing HNSW serialization + mmap path). A query mmap-loads it for its generation; the in-memory build is now a legacy fallback.

**Designed for B+ scale.** The section stores only topology + node-map (keyed by stable `superfile_id`) and reconstructs the scorer from the resident centroids — so its size tracks the *cluster count*, never the corpus. It stays **KB-scale at 100M** and remains small into the billions, so a query at any scale **mmap-loads a small object instead of rebuilding a graph over the corpus on the hot path**. Single-node, multi-node, and a restarted process are one identical, content-addressed path. Magic/version-guarded (corrupt/absent/drifted → in-memory build, never a panic); the section URI is GC-protected.

> *Next in the series:* the exact B+ numbers — section size and mmap-load time at a billion-plus rows — land in a follow-up.

### 3. Per-table fanout calibrated by MEASURED recall
Replaces the scale-blind global `global_fine_fanout` constant with a per-table stamp, calibrated the way the HNSW `ef` law is: at the post-commit settle, sample held-out queries, compute their exact top-k ground truth over the resident plane, then **sweep fanout and stamp the smallest that clears `target_recall`** (else a "loses" sentinel → `auto` keeps `stamped`). Two correctness properties that took real work:
- **Scale-safe.** The sweep re-runs the reader's selection at every fanout rung; done naively it read growing cluster sets from object storage per rung and never terminated at 100M. The rungs are **nested**, so it now reads the top-`max_fanout` clusters **once** per query and evaluates every fanout as a **prefix** of that single ranked read, ladder capped at a √N ceiling. Measured at 100M: the sweep is ~0.03 s; calibration is bounded by one ground-truth pass (~47 min).
- **Codec-correct ground truth.** The calibrator reconstructs stored rows to score recall; it was decoding every row on the fixed `[-1,1]` Sq16 grid — right for cosine, wrong for L2/NegDot (whose rows use the `Sq16Adaptive` per-cluster ruler) — distorting the ground truth and collapsing measured L2 recall to ~0.5. Rows are now decoded through their own codec's ruler (the served path's decode), so L2/NegDot calibrate correctly.

### 4. Metric- and codec-aware — NegDot and L2Sq, not just cosine
Scores centroids and reconstructs rows by the column's configured metric (Cosine / NegDot / L2Sq), removing #588's cosine-only guard. Cosine is provably unchanged.

### 5. Scale-gated `ivf_router = auto` — self-driving selection (opt-in)
Picks `centroid_graph` only where it wins — **concentration** (calibrated fanout ≪ cluster count) **and** **scale** (docs ≥ the floor), else `stamped`. Two config-tunable thresholds. **Default stays `stamped`**; explicit `stamped`/`centroid_graph` bypass the gate.

## Safety / backward-compat
- New manifest fields (graph ref, calibrated fanout) are additive `serde(default)` → old manifests decode cleanly.
- Every correctness path falls back to the in-memory build / the constant — nothing depends on the persisted section or a stamp existing.
- Reviewed across passes; every finding fixed (GC live-set, `superfile_id` keying, single-flight, streaming/OOM-safe ground truth, stable-id dedup, metric+codec-correct decode, sentinel/per-k handling).

## Tests
Generation/column rebuild across drain+compaction; persist round-trip per metric; reordered-superfile load; GC survival; metric-nearest-centroid selection; **codec-space decode (cosine + `Sq16Adaptive` L2)**; nested-prefix == per-fanout; fanout knee (cosine-vs-raw, completeness-gated overwrite, per-k zero → None, √N cap); and the `auto`-gate decision matrix.

## Validation status
- **Cosine, 10M (Cohere):** end-to-end 0.986; the calibrator stamps a real recall knee.
- **L2/NegDot:** unit-tested (the decode fix); the router itself is proven on L2 at 100M elsewhere (~0.976, #490).
- **Terminates at 100M:** calibration ~47 min (one ground-truth pass), down from a non-terminating sweep.
- **Pending before any default-flip:** an end-to-end 100M stamped-fanout number — blocked on the separate #554 codec fix (#713), else compaction saturates. The default stays `stamped`, so this PR ships the machinery without changing behavior.

## Follow-ups
- **#713** — the #554 codec union-grid fix (compaction quantizer saturation on L2/NegDot at scale). A dependency for a clean 100M run and for defaulting `auto` on non-cosine tables.
- **#714** — make `ivf_router = auto` safe to **default**: the 100M end-to-end validation, raising the 10M scale floor to the measured crossover, a cost-aware gate (today's gate guarantees recall via the sentinel but not that the graph is *faster*), and cache-fit confirmation. This PR intentionally keeps `auto` opt-in.

## Not in this PR (deliberate)
The default stays `stamped`; the `search_mode` axis (hnsw↔ivf) is untouched; the outer `search_mode = auto` mode ladder composes later (#630).


savannahar68 (approved):
overall lgtm
in case you are blocked for other PRs, then go ahead and merge it. 
If there are any minor comments i'll post it here and we can address it later

PR #675, opened 2026-08-28
size the build reserve by rows, not by the buffer they share

ashishmishra26:
## Why

`buffer_scalar_bytes` answers two questions, and the correct measure is opposite for each:

- **How much memory is held?** Capacity. A slice pins its parent's whole allocation, so that is genuinely resident, and the auto-flush threshold should see it.
- **How much scratch will the build need?** Not capacity. It scales with the rows a builder reads, so billing a shared allocation overstates it.

The build reserve was reading the held-memory figure.

Over an Arrow IPC stream that is catastrophic. The decode points every child array at one shared message buffer, and `get_array_memory_size` bills each child for all of it, so a nested schema with many children reports its message size multiplied by the number of children.

Measured on a 65-column document corpus with several nested levels (`list<struct<...>>`, `struct<struct<...>>`):

| | |
| --- | --- |
| batch on the wire | 23.7 MiB |
| `get_array_memory_size` after IPC decode | **3771.5 MiB** (~160x) |
| reserve asks (2.5x) | 9.21 GiB |
| connection budget | 5.40 GiB |

So a batch that fits comfortably is refused. Reported from the field as a 413 on `append` against a hosted table, with the figure not moving when the input shrank from a directory to a single file.

**Only remote tables were affected.** Decoding parquet locally gives each column buffers of its own, so capacity runs about 1.4x visible there. The reporter's loads against local object storage worked throughout, which is what pointed at the decode rather than the data.

## What changed

The visible figure was already being computed and discarded. `ingested_byte_legs` returns it for the priced legs, and the comment there already makes the argument for the FTS leg:

> The FTS leg moves with the priced legs deliberately: it weights the build-scratch reserve, and the builder's scratch scales with the text it will actually index — a slice's visible rows — not with the parent allocation the slice shares.

That argument covers the scalar leg identically. It is now tracked alongside the others and handed to the reserve. Held-memory accounting stays on capacity, where the same comment says it belongs.

No new measurement, no new helper, and the multipliers are untouched: the FTS leg has been visible-based with its own multiplier for a while, so there is in-tree precedent that the constants work against visible bytes.

## Test

Appends an IPC-decoded nested batch under a budget of exactly one capacity-measure, so 2.5x the visible size fits inside it and 2.5x the capacity cannot. The fixture asserts it actually exhibits the inflation before relying on it.

Reverting just the one changed argument makes it fail with the reported error:

```
OverBudget("append: during ingest, over connection memory budget:
requested 1235555 B with 0 B in use of a 434908 B limit")
```

`make check` clean. Full suite 2317 passed; the single failure is
`reader_synchronous_with_storage_upgrades_lazy_hidden_entry`, which passes in
isolation and is unrelated to this diff — it appears to be an existing
order-dependent flake and is worth a separate look.

## Left alone

`writer.rs` also measures capacity for the superfile fanout heuristic. Its comment says capacity is "fine for a fanout heuristic that is clamped to the pool anyway", which still reads as true, so it is unchanged.

The visible figure is measured on the caller's columns, without the engine-minted `_id`. That omits 16 bytes a row of `Decimal128` against a payload measured in kilobytes; measuring a second batch to recover it would cost more than it is worth.

### DIFF
--- a/src/supertable/writer.rs
+++ b/src/supertable/writer.rs
@@ -384,6 +384,19 @@ pub struct SupertableWriter {
     /// allocation. (`buffer_vector_bytes` is the exact f32 payload on
     /// either reading; slices share nothing wider than their rows.)
     buffer_fts_bytes: usize,
+    /// Visible (slice-aware) scalar bytes across `buffer`: what a build
+    /// will actually read, rather than the allocation the buffer holds.
+    ///
+    /// Separate from `buffer_scalar_bytes` because the two answer
+    /// different questions and the right measure is opposite for each.
+    /// Held memory is capacity: a slice pins its parent's whole buffer,
+    /// so that *is* the resident cost, and the auto-flush threshold
+    /// should see it. Build scratch scales with the rows the builder
+    /// touches, so it must not. Feeding capacity to the reserve made a
+    /// batch decoded from an Arrow IPC stream look ~160x its size,
+    /// because that decode points every child array at one shared
+    /// message buffer and capacity bills each child for all of it.
+    buffer_scalar_visible_bytes: usize,
     /// Ingested work for the batches sitting in `buffer`, computed at
     /// `append` time from the caller's own batch and held here until the
     /// commit that publishes them returns `Ok`. Counting at append time is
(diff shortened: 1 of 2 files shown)


ashishmishra26:
Parked as a draft rather than pursued now. The analysis, reproduction, and fix instructions are in #681; this branch carries a working change plus the regression test if someone picks it up.

Not ready to merge as-is: it touches the commit path, so the bench suite is required and has not been run.

PR #677, opened 2026-08-29
chore: add spans for optimizer and vector path

savannahar68:


src/supertable/writer.rs
```diff
@@ -3666,6 +3666,10 @@ async fn materialized_user_rows_for_drain(
 /// window where it is drained out of the user arm but not yet in the graph —
 /// invisible to both. Recall-quality overlays belong in B, where lagging only
 /// costs a temporarily wider serving law, never a missing row. Origin: OPANN #422.
+#[cfg_attr(
+    feature = "detailed-tracing",
+    tracing::instrument(name = "drain", skip_all)
```
Pranav2612000, line 3671: Is there a way to include the name of the table in the span?

Pranav2612000 (approved):

PR #680, opened 2026-08-30
fts: faster BM25 ranked top-k (threshold-first walk, f32 block-max, _id sidecar)

vinaykakade:
## Summary

Faster BM25 ranked top-k, focused on the two costs that dominate common-term
queries: **decoding blocks that can't matter**, and **resolving hit `_id`s**.
Four independent improvements, each versioned and back-compatible, each guarded
by correctness oracles:

1. **Threshold-first single-term ranked walk** (+ a coarse two-level block-max skip)
2. **Exact-`f32` block-max bound** (replaces quantized fixed-point)
3. **Tighter MaxScore leftmost-essential block skip** (per-block, not global, bound)
4. **Stable-id sidecar** for `_id` resolution (flat slice read, no id-column decode)

All ride a single new on-disk version (**`V5`**); `V1`–`V4` blobs are read
unchanged. No public API change.

---

### 1. Threshold-first single-term ranked walk + coarse block-max skip
`search_single_term_bmw` previously walked the term's posting blocks in doc-id
order and only started skipping once the k-heap filled — so for a very common
term it **decoded and scored a large fraction of blocks before the k-th
threshold had risen**, most of which could never enter the top-k.

`V5` appends a per-term **coarse block-max table** (one entry per span of
blocks). The walk now:
- **seeds the k-th threshold high** by scanning the coarse level for the best
  candidate spans *before* the main scan, so blocks below the seeded threshold
  are skipped from the start (seed depth scales with `k`); and
- uses the coarse level as the **outer of a two-level skip**, jumping a whole
  span in one compare when its coarse max can't beat the threshold.

**Effect:** far fewer decoded blocks for common-term ranked queries; largest at
small `k`. Only touches the single-term path — multi-term shapes are unchanged.

### 2. Exact-`f32` block-max bound
The per-block block-max BM25 was stored as `ceil(max × 1000)` in a `u32` and
recovered by dividing. That `ceil` rounds the bound **up** by up to ~0.002 —
which for a low-idf common term (BM25 ≈ 0.1–0.3) is a large *relative*
inflation, so the skip test `block_max ≤ threshold` fails on blocks a tight
bound would skip.

`V5` stores the block-max (and coarse-span max) as an **exact `f32`** — same 4
bytes, matching the precision of the reader's per-doc scoring. The bound is now
tight, so the skips above actually fire.

**Effect:** this was the single largest lever for very common single terms. On a
~5M-doc English-Wikipedia-scale corpus, the very common term `the` (infino
before → after, exact-`f32` vs the fixed-point bound, same build/box otherwise):

| ranked top-k | before (fixed-point) | after (exact `f32`) | speedup |
|---|--:|--:|--:|
| top-10  | ~195 µs  | ~38 µs  | ~5× |
| top-100 | ~1145 µs | ~154 µs | ~7× |
| top-1000| ~5225 µs | ~2175 µs| ~2.4× |

Other query shapes (multi-term union / intersection / phrase) are **neutral** —
the bound only affects how aggressively the single-term and MaxScore walks skip,
and a tighter *upper* bound can never change which docs are returned.

### 3. Tighter MaxScore leftmost-essential block skip
In the multi-term MaxScore walk, the block-skip test bounded the non-leftmost
terms by their **global term-max**. That's loose. It now bounds them by each
term's **per-block max over the leftmost term's current block range**
(`block_max_in_range`) — a strictly tighter yet still-sound upper bound, so the
skip fires on more blocks. Correctness is unchanged (any valid upper bound
preserves the exact top-k).

### 4. Stable-id sidecar for `_id` resolution
Resolving a hit's stable `_id` previously decoded the compressed (`ZSTD`) `_id`
column. `V5` writes a packed, fixed-width **sidecar region** so `_id` resolution
is a **flat slice read** at a computed offset — no page decompression. The
sidecar is written on **both** the build path and the merge/compaction path, is
**KV-gated** (readers without it fall back to the column), and adds ~80 MB per
5M docs.

**Effect:** large win for **cold** hit-resolution (nothing cached); warm-cache
resolution is neutral.

---

## Compatibility & format

- New on-disk version **`V5`**; `V1`–`V4` blobs read unchanged (the reader gates
  the coarse table, the `f32` bound, and the sidecar on the version / a KV flag).
- Block-max `f32` is the **same width** as the old fixed-point `u32` (**+0 bytes**);
  the coarse table adds **~0.15%**; the `_id` sidecar adds **~80 MB / 5M docs**.
- No public API change (`public-api.txt` unchanged).

---

## Why the performance holds

The wins are structural, not incidental:

- **Fewer decoded blocks.** For a common-term ranked query, the coarse block-max
  table lets the walk seed the k-th threshold high *before* the main scan and jump
  whole block-spans in one compare — blocks that can never enter the top-k are
  never decoded or scored. The exact-`f32` bound is what makes those skips
  actually fire: the old fixed-point `ceil` inflated the bound enough to block
  them for low-idf common terms. The two compound on the single-term path.
- **A tighter MaxScore skip.** Bounding the non-leftmost terms by their per-block
  max (not their global term-max) lets the multi-term walk skip more blocks.
- **A flat slice read for `_id`.** Cold hit-resolution reads a fixed-width sidecar
  at a computed offset instead of decompressing the id column.

Measured infino before → after (very common term `the`, ~5M-doc
English-Wikipedia-scale corpus, exact-`f32` vs the fixed-point bound, same build
otherwise):

| ranked top-k | before | after | speedup |
|---|--:|--:|--:|
| top-10  | ~195 µs  | ~38 µs   | ~5× |
| top-100 | ~1145 µs | ~154 µs  | ~7× |
| top-1000| ~5225 µs | ~2175 µs | ~2.4× |

Multi-term shapes (union / intersection / phrase) are **neutral by
construction**: the changes only tighten *upper* bounds and skip logic on the
single-term and MaxScore walks, add no work to the other kernels, and a tighter
upper bound can never change which docs are returned. The `_id` sidecar's win is
**cold-only** — a warm cache already avoids the id-column decode. These are
measured with the in-repo suite (`cargo bench --bench bench -- superfile fts`);
the single-term figures show the shape of the win.

## Why the results stay correct

The ranked order is preserved **by construction.** Every optimization here only
ever *tightens a sound upper bound* or *seeds the threshold*: a block or doc is
skipped solely when its bound proves it cannot beat the current k-th best, so the
exact top-k — and its order — is identical to the unoptimized walk. The `f32`
bound and the per-block MaxScore bound are strictly *tighter* than what they
replace, never looser, so they can only skip *more* of the provably-irrelevant
blocks, never drop a real result. The `_id` sidecar changes only how a stable id
is *fetched*, not which rows match.

That guarantee is enforced, not just argued:

- **Brute-force BM25 oracles** (`fts::brute_force_oracle`) compare top-k against
  the textbook BM25 formula on planted corpora, sharing no code with the
  optimized walks. This PR extends them with **oracles at
  block-max-optimization scale** — multi-block corpora large enough that the
  coarse skip, threshold seed, and `f32` bound actually engage — including
  **tie-free, strictly-ordered heads**, so a dropped *or reordered* result is
  caught, not merely a set-membership change.
- **Property-test fuzz oracle** (`fts::fuzz_oracle`) — random corpora and random
  clause/phrase queries diffed against brute-force (exact set + per-doc scores,
  and the top-k score multiset under truncation).
- **`_id` sidecar** — KV-gated, so a reader without it falls back to the id
  column; the merge/compaction path writes it and is asserted by invariant;
  crash-safety and CRC-corruption coverage is unchanged.
- **Versioning** — the `V1`–`V4` read paths are exercised alongside `V5`, so old
  files still read correctly.
- `make ci` is green (fmt, clippy `-D warnings`, the test lanes).


savannahar68 (approved):
ill have all the review comments today in sometime
feel free to pick those if any in separate PR

PR #684, opened 2026-08-31
fts: trim ranked-path decode/walk overhead + batch boolean-clause atom build

vinaykakade:
## Summary

Two behavior-preserving optimizations on the BM25 top-k read path, plus one
attempt that was measured, reverted, and kept in history for the record. No
format or public-API change; results are bit-identical (gated by the existing
oracles).

### 1. Batch the boolean-clause atom build
The must / should / negation path (and the count path) built its bare term
cursors **one term at a time** — each reopening the FST dictionary and issuing
its own serialized postings-range fetch. A k-term clause list therefore paid
**k dictionary opens and k serial fetch waves.**

A new arity-preserving core, `build_term_cursors_opt`, resolves every term
against **one FST open** and fetches all their posting ranges in a **single
parallel fan-out**, returning one slot per input term (`None` where a term is
absent, so callers still know which one missed). `build_term_cursors` keeps its
miss-dropping contract as a thin wrapper, so its many callers are unchanged.
The clause path now resolves all bare terms through one batched call — cutting
the planned dictionary ranges from one-per-term to one and overlapping the
fetches. Term order and absent-term (`None`) slots are preserved exactly.

### 2. Trim per-decode and per-walk overhead
Small hot-path cleanups on the top-k read path:
- **Single-term walk** (the most common query shape): keep its two per-query
  block decode buffers on the **stack** — they are `BLOCK_LEN`-sized and never
  cross an await, so the per-query heap alloc/free was needless.
- **Windowed MaxScore accumulate**: bind the four-wide SIMD window as one
  **subslice** per array, so the fixed-length lane reads drop their bounds
  checks (eight element checks become two range checks in the per-block inner
  loop).
- **Ranked-OR cursor**: reuse the block-bytes slice already held in the tf
  probe instead of recomputing the same subslice for the tf decode.
- Drop a **dead `pos >= block_n` guard** from `current_doc_id` (a live cursor
  always satisfies it; a debug assertion keeps the tripwire), so the hottest
  walk accessor does one comparison, not two.

### Reverted (kept in history): the cached unpacking dispatcher
An earlier commit built the `BitPacker4x` unpacking dispatcher once in a process
static. It was reverted: release builds pin `target-cpu`, so the feature-detect
folds to a compile-time constant and the per-call constructor already inlines
away to nothing — routing every decode through a `LazyLock` only **added** an
atomic-guarded init check to the hot path. Benchmarking the ranked path
confirmed it regressed, so it's out; the cleanups above stay. Kept as a commit
so the history shows the measurement.

## Why it's correct

Every change is **behavior-preserving** — same posting decode, same term
resolution, same clause semantics, same top-k. This is enforced, not just
argued:
- **Brute-force BM25 oracles** — top-k compared against the textbook BM25
  formula on planted (multi-block) corpora, sharing no code with the optimized
  walks.
- **Property-test fuzz oracle** — random corpora and random clause/phrase
  queries diffed against brute-force (exact set + per-doc scores).
- The batched atom build additionally preserves the **planned-range op-stats
  pins** (which assert the dictionary/postings fetch plan) — so the
  one-open/one-fan-out change is verified to keep the same term order and
  absent-term slots.
- `make ci` (fmt, clippy `-D warnings`, the test lanes).

## Performance

These remove work from the hottest read path rather than change any algorithm's
result: per-query heap allocations, per-block SIMD bounds checks, a redundant
subslice recompute, and a redundant walk-accessor comparison. The batched atom
build is the one structural win — a k-term boolean clause goes from **k FST
opens + k serial fetch waves to one open + one parallel fan-out.** The single
piece that benchmarked *negative* (the cached dispatcher) was reverted, so the
net is neutral-to-positive on the ranked path with the structural win on
multi-term clause building.

### DIFF
--- a/src/superfile/fts/posting.rs
+++ b/src/superfile/fts/posting.rs
@@ -54,21 +54,11 @@
 //!   `assert!` preconditions plus the caller's contract carry safety
 //!   in production.
 
-use std::sync::LazyLock;
-
 use bitpacking::{BitPacker, BitPacker4x};
 
 /// Number of `(doc_id, tf)` pairs per encoded block. Fixed at 128 to
 /// match `BitPacker4x::BLOCK_LEN`.
 pub const BLOCK_LEN: usize = BitPacker4x::BLOCK_LEN;
-
-/// Process-wide bit-unpacking dispatcher, built once. `BitPacker4x::new()`
-/// runs a runtime SIMD-feature probe and is not `#[inline]`, so calling it
-/// per block was an out-of-line call on every decode — twice per full
-/// scoring decode (doc ids + tfs), in the hottest read path. The selected
-/// instruction set is process-invariant, so resolve it a single time and
-/// reuse the (one-byte, `Copy`) dispatcher.
-static BITPACKER: LazyLock<BitPacker4x> = LazyLock::new(BitPacker4x::new);
 
 /// Header size in bytes (doc_count + delta_bits + tf_bits + encoding +
 /// base_doc_id).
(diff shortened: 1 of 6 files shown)


naman-modi (approved):
LGTM

PR #686, opened 2026-08-31
fts: fix capacity-overflow panic on unbounded-k ranked search

vinaykakade:
## Problem

The single-term ranked BM25 path panics with `capacity overflow` when called with a very large `k`. The threshold-first seed heap preallocated `BinaryHeap::with_capacity(k)` with no clamp; `k` is caller-controlled and may be `usize::MAX` (the unbounded "return everything" request a brute-force top-k comparison passes), so the allocation overflows and aborts the query.

```
thread 'main' panicked at library/alloc/src/raw_vec/mod.rs:28:5:
capacity overflow
```

This is a live crash on `main`: the superfile FTS bench runs a `BMW==brute-force` correctness check that issues an unbounded-`k` search, so the bench aborts before producing results.

## Fix

Clamp the seed heap's preallocation to the corpus size via the existing `top_k_initial_capacity` helper — the same clamp every sibling top-k heap already uses. The seed heap never holds more than the docs in scope, so this only bounds the up-front allocation; behavior is otherwise unchanged.

## Test

Adds a regression test that runs an unbounded-`k` search over a corpus large enough to engage the seed path (a term in every one of 35k docs, ~273 blocks, past the seed's `MIN_BLOCKS = 256` gate). It reproduces the exact `capacity overflow` panic before the fix and passes after.

### DIFF
--- a/src/superfile/fts/reader/search.rs
+++ b/src/superfile/fts/reader/search.rs
@@ -932,8 +932,13 @@ impl FtsReader {
             }
             let m = m_want.min(cand.len());
             cand.select_nth_unstable_by(m.saturating_sub(1), |a, b| b.0.total_cmp(&a.0));
-            // Decode those blocks, score, keep a k-sized seed heap.
-            let mut seed_heap: BinaryHeap<TopKEntry> = BinaryHeap::with_capacity(k);
+            // Decode those blocks, score, keep a k-sized seed heap. Clamp the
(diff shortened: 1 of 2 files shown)


vinaykakade:
Redundant: the same fix already landed on main via the #684 squash-merge (search.rs seed-heap clamp + the regression test are both on main). Closing.

PR #687, opened 2026-08-31
fts: faster ranked rare∧common AND via a rarest-driven bitset-probe walk

vinaykakade:
## Problem

A ranked multi-term AND that mixes one discriminating term with common ones
(`+the +book +of +life`, `+secretary +of +state`, `+lord +of +the +rings`) runs
the block flat-merge, which decodes each common term's 128-doc posting block just
to align it against the rare term — the dominant cost on these queries, since the
common terms' blocks are large and mostly irrelevant.

## Change

Route those conjunctions to a rarest-driven membership walk instead: drive the
rarest term, and for each of its docs test the common terms for presence with a
bitset bit-test (`contains`), reading a term's frequency only on a full match via
a popcount-rank (`tf_at_contained`) — **never decoding the common terms' doc-id
blocks.** The walk keeps the flat-merge's Block-Max-AND heap-bar skip (amortized
per window), so it doesn't forfeit top-k pruning.

Routing (`and_prefer_membership`) is two-tier on the rarest term's document
frequency: a very sparse rarest (< 1/64 of the corpus) always takes the walk; a
moderately-sparse rarest ([1/64, 1/16)) takes it only when the other terms are
collectively ≥ 8× denser, which is exactly when skipping their block decode beats
iterating the driver. Denser-rarest and all-common conjunctions stay on the
flat-merge, unchanged.

The same idea extends to the **2-term** case (`+the +X`): when the common term is
bitset-encoded (`is_bitset_dense`) and the rarer one is not, the walk bit-tests
the common term instead of the specialized two-pointer merge expanding its dense
blocks. A PFOR common term (e.g. `+american +funds`) keeps that two-pointer merge,
which the walk doesn't beat when the probe must still decode.

## Results

Measured on a 5M-document English-Wikipedia full-text index (single compacted
segment), ranked TOP-k, branch vs. `main`, same index:

- `+the +book +of +life`: TOP_10 **-39%**, TOP_100 **-39%**, TOP_1000 **-37%**
- 300 distinct intersection queries, TOP_1000: 3-term **-19%**, 4-term **-29%**,
  aggregate **-10%**
- 2-term `+the +X` (bitset common): `+the +book` TOP_1000 **-51%**,
  `+the +incredibles` **-27%**, `+the +who` TOP_1000 **-21%**.
- PFOR-common 2-term (`+american +funds`, `+body +painting`) and all-common
  conjunctions (`+to +be +or +not +to +be`): at parity — they stay on the
  flat-merge.

`COUNT` and every match set are byte-for-byte identical to `main`.

## Testing

Read-path only — no format change, and no new `unsafe`.

The walk is checked against a textbook BM25 brute-force scorer (independent of the
optimized walks):

- The existing membership-path oracles exercise the walk end to end — the
  presence probe and its reject/short-circuit path, the per-match term-frequency
  read on **both** block encodings (bitset popcount-rank and PFOR binary-search),
  multi-`other` scoring, and the Block-Max-AND heap-bar skip on a truncated top-k.
- New oracle `oracle_and_membership_middle_tier_multiblock_matches_brute_force`
  covers what those didn't: the **density-gated middle routing tier** (a
  moderately-sparse rarest term with much-denser companions — the shape this PR
  targets, which the pre-existing oracles never routed to) over a **200-doc
  intersection spanning ~32 posting blocks**, so the window-skip pruning has to
  cross blocks correctly. Both the full match set and a truncated top-k are
  compared to brute force.
- The property-fuzz BM25 oracle (uniform and skewed vocabularies) exercises the
  router branches across randomized corpora and queries.

The routing *decision* (walk vs. flat-merge) is a performance choice — both paths
return identical results — so it is guarded by the benchmarks above rather than
the correctness oracles.

## Note on code placement

The membership walk shares its module with the block flat-merge, whose hot
doc-walk is measurably sensitive to its own instruction placement. `#[cold]` on
the walk and keeping `tf_at_contained` past the hot cursor methods hold that
layout steady, so an all-common conjunction — which never routes to the walk —
stays at parity rather than drifting from the added code.

### DIFF
--- a/src/superfile/fts/reader/cursor.rs
+++ b/src/superfile/fts/reader/cursor.rs
@@ -728,54 +728,6 @@ impl TermCursor {
         Some(self.block_tfs[rank as usize])
     }
 
-    /// Tf for `doc` on a cursor a preceding [`Self::contains(doc)`] just confirmed
-    /// present. `contains` already advanced `current_block` to `doc`'s block (and,
-    /// on a PACKED block, decoded it), so this skips the block-advance and the
-    /// presence bit-test that [`Self::bitset_probe_tf`] repeats, doing only the tf
-    /// lookup: a popcount-rank into the tf array on a bitset block, or a binary
-    /// search over the decoded doc ids on a PACKED one. Only valid immediately
-    /// after `contains(doc)` returned `true` with no intervening advance.
-    pub(super) fn tf_at_contained(&mut self, doc: u32) -> u32 {
-        // Inline (df=1) cursor: single pre-decoded posting.
-        if self.bytes.is_empty() {
-            return self.block_tfs[0];
-        }
-        let block = self.blocks[self.current_block];
-        let raw = &self.bytes[block.block_byte_offset..block.block_byte_end];
-        if raw[posting::ENCODING_OFF] == posting::ENCODING_BITSET {
-            let base = read_u32_le(&raw[4..8]);
-            let bit = (doc - base) as usize;
-            let tf_bits = raw[2] as usize;
-            let tfs_size = BLOCK_LEN * tf_bits / 8;
-            let bitset_end = raw.len() - tfs_size;
-            let word_idx = bit / 64;
-            let word_at = posting::HEADER_SIZE + word_idx * 8;
-            let word = u64::from_le_bytes(raw[word_at..word_at + 8].try_into().expect("8 bytes"));
-            let presence = &raw[posting::HEADER_SIZE..bitset_end];
-            let mut rank: u32 = 0;
-            for w in presence[..word_idx * 8].chunks_exact(8) {
-                rank += u64::from_le_bytes(w.try_into().expect("8 bytes")).count_ones();
-            }
-            let below = if bit.is_multiple_of(64) {
-                0u64
-            } else {
-                (1u64 << (bit % 64)) - 1
-            };
-            rank += (word & below).count_ones();
-            if self.tf_decoded_block != self.current_block {
-                posting::decode_block_tfs(raw, &mut self.block_tfs);
-                self.tf_decoded_block = self.current_block;
-            }
-            self.block_tfs[rank as usize]
-        } else {
-            // PACKED: `contains` decoded this block's doc ids and tfs. Locate doc.
-            let pos = self.block_doc_ids[..self.block_n]
-                .binary_search(&doc)
-                .expect("contains(doc) confirmed presence");
-            self.block_tfs[pos]
-        }
-    }
-
     pub(super) fn is_exhausted(&self) -> bool {
         self.current_block >= self.blocks.len()
     }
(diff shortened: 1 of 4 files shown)


src/superfile/fts/reader/scorers.rs
```diff
@@ -582,22 +600,47 @@ impl FtsReader {
...
+    /// matching tf by popcount-rank, with no doc-id block decode** — short-
+    /// circuiting on the first miss. Only a doc present in *every* term is scored
+    /// (`Σ` per-term BM25, identical to the flat-merge) and emitted through the
+    /// generic sink. This skips the flat-merge's per-leader-doc `skip_to` that
+    /// fully decodes a common term's 128-doc block just to align it — the
+    /// profiled cost on an n≥3-term AND with a common term.
+    ///
+    /// Two properties keep it competitive across the sparsity range the gate
+    /// ([`and_prefer_membership`]) admits:
+    /// - **Presence is tested before any tf is read.** A selective AND misses on
```
naman-modi: nit: Its a bit difficult to understand this entire comment, can we make it precise, to the point, and maybe add an example to make it easier for the reader?

tests/superfile/fts/brute_force_oracle.rs
```diff
@@ -815,6 +815,98 @@ async fn oracle_and_membership_rejects_partial_matches() {
     }
 }
 
+#[tokio::test]
+async fn oracle_and_membership_middle_tier_multiblock_matches_brute_force() {
```
naman-modi, line 819: This test does aa good job at invoking the `skip` in case 2 when k=5 is < docs_matching=200, but all the scores are the same, so the actual motive of the test to "see if the right doc with a different score comes back" is not met.  So can we make it actually gate the skip:

- vary the length so scores differ
- run the exact set-check at a small k (say 20) over the 200-doc intersection so the heap fills and the skip fires
- assert the returned doc-id set, not just the score multiset

src/superfile/fts/reader/cursor.rs
```diff
@@ -973,6 +973,59 @@ impl TermCursor {
...
+    /// lookup: a popcount-rank into the tf array on a bitset block, or a binary
+    /// search over the decoded doc ids on a PACKED one. Only valid immediately
+    /// after `contains(doc)` returned `true` with no intervening advance.
+    ///
+    /// Kept at the end of the impl, past the doc-cursor hot methods
+    /// (`skip_to`, `next`, `decode_current_block`, `current_doc_id`), so adding
+    /// it doesn't shift their code offsets — those methods drive the flat-merge
+    /// AND path, which is measurably sensitive to its own instruction layout.
+    pub(super) fn tf_at_contained(&mut self, doc: u32) -> u32 {
+        // Inline (df=1) cursor: single pre-decoded posting.
```
naman-modi, line 1000: Can we add a test to test this method for tf>1? From my understanding all the tests are with tf=1. Pls ignore if its already tested in tests not in this PR.

src/superfile/fts/reader/cursor.rs
```diff
@@ -973,6 +973,59 @@ impl TermCursor {
...
+        if self.bytes.is_empty() {
+            return self.block_tfs[0];
+        }
+        let block = self.blocks[self.current_block];
+        let raw = &self.bytes[block.block_byte_offset..block.block_byte_end];
+        if raw[posting::ENCODING_OFF] == posting::ENCODING_BITSET {
+            let base = read_u32_le(&raw[4..8]);
```
naman-modi, line 1007: Just checking if this is similar to `bitset_probe_tf` [ref](https://github.com/infino-ai/infino/pull/687/changes#diff-e0c7a1989510ecfc186c6851e297d0b02ac40c4798164e012a96b48ad96a5ab0R679-R704) and can be extracted out?

naman-modi (approved):
Few comments, else LGTM Vinay.

src/superfile/fts/reader/scorers.rs

tests/superfile/fts/brute_force_oracle.rs
vinaykakade: Done

src/superfile/fts/reader/cursor.rs
vinaykakade: Done

PR #692, opened 2026-09-01
fix: retain block cache on reopen

savannahar68:
## What is the change about?

During vector query we download block cache instead of fetching the entire hidden table superfile. We explicitly clear this valid cache file during reopen of table because we don't have the index (roaring bit map) persisted with the blocks. 

### how does this affect the performance

Without this change, every open/reopen/restart results in 1st vector query to be slow even if the files are part of cache dir
After this change the queries after open should be warm instead of cold.

### DIFF
--- a/src/supertable/reader_cache/block_source.rs
+++ b/src/supertable/reader_cache/block_source.rs
@@ -175,14 +175,11 @@ impl BlockCachedSource {
         }
         self.state
             .get_or_init(|| {
-                // Publish a FRESH inode rather than truncating in place. Remove
-                // the old name first: a reader still holding the previous
-                // `.blocks` inode keeps its bytes (POSIX unlink-while-open),
-                // while this source gets its own inode to fill. `create_new`
-                // then refuses to reopen an existing inode, so a concurrent
-                // same-path source fails cleanly (degrading to a passthrough
-                // read) instead of truncating a live reader's blocks.
+                if let Some(adopted) = self.try_adopt(size) {
+                    return Some(adopted);
+                }
                 let _ = fs::remove_file(&self.path);
+                let _ = fs::remove_file(self.idx_path());
                 let file = fs::OpenOptions::new()
                     .read(true)
                     .write(true)
@@ -197,6 +194,58 @@ impl BlockCachedSource {
                 })
             })
             .as_ref()
+    }
+
+    fn idx_path(&self) -> PathBuf {
+        let mut p = self.path.clone().into_os_string();
+        p.push(".idx");
+        PathBuf::from(p)
+    }
+
+    /// Adopt a prior generation's sparse file when its persisted index proves
+    /// which ranges are valid (immutable superfile, so size-match is enough).
+    fn try_adopt(&self, size: u64) -> Option<BlockFile> {
+        let meta = fs::metadata(&self.path).ok()?;
+        if meta.len() != size {
+            return None;
+        }
+        let idx_bytes = fs::read(self.idx_path()).ok()?;
+        let bitmap = RoaringBitmap::deserialize_from(idx_bytes.as_slice()).ok()?;
+        let file = fs::OpenOptions::new()
+            .read(true)
+            .write(true)
+            .open(&self.path)
+            .ok()?;
+        let adopted_bytes: u64 = bitmap.iter().map(|b| Self::block_len(size, b)).sum();
+        if self.owns_accounting
+            && let Some(store) = self.store.upgrade()
+        {
+            store.account_adopted_bytes(adopted_bytes);
+        }
+        self.filled_bytes.store(adopted_bytes, Ordering::Release);
+        *self.filled.lock().expect("filled bitmap mutex poisoned") = bitmap;
+        Some(BlockFile {
+            file,
+            size,
+            mmap: OnceLock::new(),
+        })
+    }
+
+    /// Persist the filled-block bitmap, always after the blocks it describes so
+    /// the index never claims a block that was not written.
+    fn persist_idx(&self) {
+        let filled = self.filled.lock().expect("filled bitmap mutex poisoned");
+        let mut buf = Vec::with_capacity(filled.serialized_size());
+        if filled.serialize_into(&mut buf).is_err() {
+            return;
+        }
+        let idx = self.idx_path();
+        let mut tmp = idx.clone().into_os_string();
+        tmp.push(".tmp");
+        let tmp = PathBuf::from(tmp);
+        if fs::write(&tmp, &buf).is_ok() {
+            let _ = fs::rename(&tmp, &idx);
+        }
     }
 
     /// Byte length of block `b` (the trailing block may be partial).
@@ -302,6 +351,7 @@ impl BlockCachedSource {
         if !store.lazy_block_entry_is_current(&self.uri, &self.entry_token) {
             return Ok(false);
         }
+        let mut filled_any = false;
         for (rb0, rb1) in self.missing_runs(b0, b1) {
             let run_start = u64::from(rb0) * CACHE_BLOCK_BYTES;
             let run_end = (u64::from(rb1) + 1) * CACHE_BLOCK_BYTES;
@@ -327,11 +377,15 @@ impl BlockCachedSource {
             }
             let newly = self.mark_filled(bf.size, rb0, rb1);
             self.filled_bytes.fetch_add(newly, Ordering::AcqRel);
+            filled_any = true;
             if self.owns_accounting && newly < run_len {
                 // A concurrent filler beat us to some blocks; its accounting
                 // stands, ours is released.
                 store.release_block_bytes(run_len - newly);
             }
+        }
+        if filled_any {
+            self.persist_idx();
         }
         Ok(true)
     }
@@ -339,9 +393,7 @@ impl BlockCachedSource {
 
 impl Drop for BlockCachedSource {
     fn drop(&mut self) {
-        // Last reader over this source is gone (entry evicted/replaced and
-        // no in-flight queries): release the accounted bytes and remove the
-        // sparse file. The store may already be gone at process teardown.
+        // Release accounted bytes only; the file and index persist for adoption.
         if self.owns_accounting
             && let Some(store) = self.store.upgrade()
         {
@@ -349,9 +401,6 @@ impl Drop for BlockCachedSource {
             if filled > 0 {
                 store.release_block_bytes(filled);
             }
-        }
-        if self.state.get().is_some_and(|s| s.is_some()) {
-            let _ = fs::remove_file(&self.path);
         }
     }
 }
@@ -595,13 +644,121 @@ mod tests {
             "trailing partial block accounts its real length"
         );
 
-        // Drop the source: accounting released, blocks file unlinked.
+        // Drop the source: accounting released, but the sparse file and its
+        // index persist for a later generation to adopt.
         let path = dir.path().join("t.blocks");
         assert!(path.exists());
         store.remove_block_entry_for_test(&uri);
         drop(src);
         assert_eq!(store.stats().current_bytes, 0);
-        assert!(!path.exists());
+        assert!(path.exists(), "blocks file survives drop for adoption");
+        let mut idx = path.clone().into_os_string();
+        idx.push(".idx");
+        assert!(
+            std::path::PathBuf::from(idx).exists(),
+            "index sidecar persisted alongside the blocks file"
+        );
+    }
+
+    /// Reopen/reconnect warm survival: a fresh source adopts a prior
+    /// generation's persisted blocks and serves filled ranges with zero refetch.
+    #[tokio::test]
+    async fn adopt_reuses_blocks_across_source_generations() {
+        const OBJ: usize = 3 * CACHE_BLOCK_BYTES as usize + 1000;
+        let dir = tempdir().expect("tempdir");
+        let store = test_store(dir.path(), u64::MAX);
+        let uri = SuperfileUri::new_v4();
+        let path = dir.path().join("t.blocks");
+
+        let inner1 = Arc::new(CountingSource::new(OBJ));
+        let src1 = BlockCachedSource::new(
+            Arc::clone(&inner1) as Arc<dyn LazyByteSource>,
+            Arc::downgrade(&store),
+            uri,
+            path.clone(),
+        );
+        store.install_block_entry_for_test(uri, src1.filled_bytes_handle(), src1.entry_token());
+        let start = 100u64;
+        let len = 2 * CACHE_BLOCK_BYTES + 500;
+        let first = src1.range(start, len).await.expect("gen1 read");
+        assert_eq!(inner1.calls(), 1);
+        let filled = src1.filled_bytes_handle().load(Ordering::Acquire);
+
+        store.remove_block_entry_for_test(&uri);
+        drop(src1);
+        assert_eq!(store.stats().current_bytes, 0);
+        assert!(path.exists());
+
+        let inner2 = Arc::new(CountingSource::new(OBJ));
+        let src2 = BlockCachedSource::new(
+            Arc::clone(&inner2) as Arc<dyn LazyByteSource>,
+            Arc::downgrade(&store),
+            uri,
+            path.clone(),
+        );
+        store.install_block_entry_for_test(uri, src2.filled_bytes_handle(), src2.entry_token());
+
+        let again = src2.range(start, len).await.expect("gen2 read");
+        assert_eq!(again, first);
+        assert_eq!(inner2.calls(), 0, "adopted blocks serve without refetch");
+        assert_eq!(
+            src2.filled_bytes_handle().load(Ordering::Acquire),
+            filled,
+            "adopted footprint re-accounted"
+        );
+        assert_eq!(store.stats().current_bytes, filled);
+
+        let tail_start = 3 * CACHE_BLOCK_BYTES + 10;
+        let _ = src2.range(tail_start, 50).await.expect("gen2 tail");
+        assert_eq!(inner2.calls(), 1, "an unfilled range still fetches");
+    }
+
+    /// An absent index sidecar makes adoption fall back to a fresh file rather
+    /// than trust the sparse bytes blindly.
+    #[tokio::test]
+    async fn adopt_falls_back_to_fresh_when_index_is_missing() {
+        const OBJ: usize = 3 * CACHE_BLOCK_BYTES as usize + 1000;
+        let dir = tempdir().expect("tempdir");
+        let store = test_store(dir.path(), u64::MAX);
+        let uri = SuperfileUri::new_v4();
+        let path = dir.path().join("t.blocks");
+
+        let inner1 = Arc::new(CountingSource::new(OBJ));
+        let src1 = BlockCachedSource::new(
+            Arc::clone(&inner1) as Arc<dyn LazyByteSource>,
+            Arc::downgrade(&store),
+            uri,
+            path.clone(),
+        );
+        store.install_block_entry_for_test(uri, src1.filled_bytes_handle(), src1.entry_token());
+        let _ = src1
+            .range(100, 2 * CACHE_BLOCK_BYTES + 500)
+            .await
+            .expect("gen1");
+        store.remove_block_entry_for_test(&uri);
+        drop(src1);
+
+        let mut idx = path.clone().into_os_string();
+        idx.push(".idx");
+        std::fs::remove_file(PathBuf::from(idx)).expect("drop the index");
+
+        let inner2 = Arc::new(CountingSource::new(OBJ));
+        let src2 = BlockCachedSource::new(
+            Arc::clone(&inner2) as Arc<dyn LazyByteSource>,
+            Arc::downgrade(&store),
+            uri,
+            path.clone(),
+        );
+        store.install_block_entry_for_test(uri, src2.filled_bytes_handle(), src2.entry_token());
+        let _ = src2
+            .range(100, 2 * CACHE_BLOCK_BYTES + 500)
+            .await
+            .expect("gen2");
+        assert_eq!(
+            inner2.calls(),
+            1,
+            "no index means fresh fetch, not stale reuse"
+        );
     }
 
     /// Regression for the transient `Required field type_ is missing` decode
--- a/src/supertable/reader_cache/block_source.rs
+++ b/src/supertable/reader_cache/block_source.rs
@@ -26,13 +26,15 @@
 //! land — so eviction sees a lazy entry's true footprint. On budget
 //! exhaustion, or once this source's cache entry has been replaced (eviction
 //! / mmap promotion), reads degrade to plain uncached passthrough instead of
-//! failing. The source releases its accounted bytes and unlinks its file on
-//! `Drop` (i.e. when the last in-flight reader over it goes away).
+//! failing. The source releases its accounted bytes on `Drop` (i.e. when the
+//! last in-flight reader over it goes away); the sparse file and its persisted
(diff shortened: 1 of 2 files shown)


src/supertable/reader_cache/block_source.rs
```diff
@@ -199,6 +196,58 @@ impl BlockCachedSource {
...
 
+    fn idx_path(&self) -> PathBuf {
+        let mut p = self.path.clone().into_os_string();
+        p.push(".idx");
+        PathBuf::from(p)
+    }
+
+    /// Adopt a prior generation's sparse file when its persisted index proves
+    /// which ranges are valid (immutable superfile, so size-match is enough).
+    fn try_adopt(&self, size: u64) -> Option<BlockFile> {
```
muralikpbhat, line 217: **Correctness — crash-safety: adopted blocks are trusted with no fsync barrier and no content validation.**

`fill_missing()` writes block bytes via `write_all_at` (page cache, never fsync'd); `persist_idx()` writes the `.idx` via `fs::write`+`rename` — a *separate* file, also not fsync'd. On power-loss / kernel-panic the `.idx` rename can reach disk while the `.blocks` data pages don't, and the file reverts to its sparse (zero) holes. On restart `try_adopt()` sees a size-match + a deserializable index, marks those blocks filled, and serves zeros. The lazy reader opens with `verify_crc:false`, so this is never caught → the vector query silently returns wrong results.

This hazard is newly reachable: before this change the block cache was discarded on restart. In-process reopen and a clean `systemctl` restart stay safe (kernel page cache survives process exit); only an *unclean* shutdown bites — but that's real, and restart-warmth is the whole point. Note the whole-file cache has a parquet footer/CRC backstop that the sparse-block path lacks.

Suggest one of: `fsync` the `.blocks` file before renaming the `.idx` (cheap, once per fill batch — makes the index's claim true after a crash); a per-block checksum in the `.idx`; or adopt only when a clean-shutdown marker is present.

src/supertable/reader_cache/block_source.rs
```diff
@@ -199,6 +196,58 @@ impl BlockCachedSource {
...
+            return None;
+        }
+        let idx_bytes = fs::read(self.idx_path()).ok()?;
+        let bitmap = RoaringBitmap::deserialize_from(idx_bytes.as_slice()).ok()?;
+        let file = fs::OpenOptions::new()
+            .read(true)
+            .write(true)
+            .open(&self.path)
+            .ok()?;
+        let adopted_bytes: u64 = bitmap.iter().map(|b| Self::block_len(size, b)).sum();
```
muralikpbhat: **Correctness: unchecked persisted block indices → `block_len()` unsigned underflow.**

The `.idx` is deserialized into a `RoaringBitmap` whose block ids are never range-checked against `ceil(size / CACHE_BLOCK_BYTES)`. A torn-but-structurally-valid index (see the tmp-file race on L244) containing a block `b` with `b * 512KiB >= size` makes `block_len()` evaluate `size - start` on `u64` → panic in debug, or a wrapped enormous value in release that is summed into `adopted_bytes` and pushed into `current_bytes` via `account_adopted_bytes`, poisoning the store accounting and the filled bitmap. Range-check block ids against the file size on adopt before trusting them.

src/supertable/reader_cache/block_source.rs
```diff
@@ -199,6 +196,58 @@ impl BlockCachedSource {
...
+            file,
+            size,
+            mmap: OnceLock::new(),
+        })
+    }
+
+    /// Persist the filled-block bitmap, always after the blocks it describes so
+    /// the index never claims a block that was not written.
+    fn persist_idx(&self) {
+        let filled = self.filled.lock().expect("filled bitmap mutex poisoned");
```
muralikpbhat, line 252: **Warm-path contention: `persist_idx` holds the `filled` mutex across the blocking write + rename.**

The guard is only needed for `serialize_into`, but it stays live through `fs::write(&tmp, &buf)` and `fs::rename` (L246-247). Every concurrent `range()` locks `filled` (via `all_filled` / `missing_runs` / `mark_filled`), so other queries block on the mutex for the duration of the file I/O — a regression on the exact warm path this PR speeds up. Snapshot the bitmap under the lock, drop the guard, then write.

src/supertable/reader_cache/block_source.rs
```diff
@@ -199,6 +196,58 @@ impl BlockCachedSource {
...
+        let mut buf = Vec::with_capacity(filled.serialized_size());
+        if filled.serialize_into(&mut buf).is_err() {
+            return;
+        }
+        let idx = self.idx_path();
+        let mut tmp = idx.clone().into_os_string();
+        tmp.push(".tmp");
```
muralikpbhat: **Correctness: fixed per-URI tmp path races under concurrent persists.**

`tmp` derives solely from `self.path`, so it is identical across concurrent `fill_missing() -> persist_idx()` calls on one shared source. Two `fs::write` (truncate+write) interleave their bytes, then both `rename` — the surviving `.idx` can be a garbled bitmap. Best case it fails to deserialize next time (safe refetch); worst case it deserializes to a wrong bitmap and drives the L219 underflow / zero-serve. Use a unique tmp name (nonce / pid / thread-id).

src/supertable/reader_cache/disk.rs
```diff
@@ -1127,8 +1131,7 @@ impl DiskCacheStore {
             let Some(name) = path.file_name().and_then(|n| n.to_str()) else {
                 continue;
             };
-            if name.ends_with(BLOCKS_FILE_SUFFIX) {
-                let _ = fs::remove_file(&path);
+            if name.ends_with(BLOCKS_IDX_SUFFIX) || name.ends_with(BLOCKS_FILE_SUFFIX) {
```
muralikpbhat: **Cost/budget: restart preserves `.blocks`/`.idx` but never counts their bytes.**

`scan_cache_root` now `continue`s past them without adding their size to `current_bytes`, and `unindexed` / `evict_unindexed` only track whole-file copies. After a restart a heavily-warmed table's multi-GB block cache is invisible to the disk budget; superfiles never re-queried (and not yet GC'd) stay uncounted and unevictable, so across repeated restarts the on-disk cache can grow past the configured budget. Account the surviving block bytes at scan time and make them eviction candidates.

muralikpbhat (commented):
Reviewed the retain-block-cache-on-reopen change — the direction is right and it targets the reopen/restart cold-start correctly, with good adopt / fallback / restart tests. Seven inline findings:

- **#1-2 (block_source.rs) are the material correctness risks**: after an *unclean* restart the adopted sparse cache can serve zeroed/torn blocks as valid (no fsync barrier, no checksum, `verify_crc:false`), silently returning wrong query results — the exact restart path this PR makes reachable. In-process reopen and clean restart are safe.
- **#3-4 are concurrency defects on the warm path** (mutex held across disk I/O; racy fixed tmp path).
- **#5-7 are disk-budget / cleanup regressions** (uncounted retained blocks after restart; `.idx` cleanup gaps; unbudgeted adopt accounting).

Nothing here blocks the core idea. #1 — a durability barrier or a content check on the adopted sparse cache — is the one I'd want resolved before merge, since correctness outranks the warm-start win.

src/supertable/reader_cache/disk.rs
```diff
@@ -1254,6 +1258,7 @@ impl DiskCacheStore {
 
         let _ = fs::remove_file(self.cache_path(uri));
         let _ = fs::remove_file(self.blocks_path(uri));
+        let _ = fs::remove_file(self.blocks_idx_path(uri));
```
muralikpbhat, line 1334: **Consistency: `.idx` cleanup missed in two sibling paths.**

This path (and `erase_superfile_local_copy`, the cached-eviction path, and gc) removes `blocks_idx_path`, but `evict_unindexed` (~L1289) and the open-failure branch of `try_reuse_cached_file` (~L1243) still delete only `cache_path` + `blocks_path`, leaking orphan `.idx` sidecars for URIs never re-fetched. Harmless to correctness, but inconsistent with the rest of the PR — add the `.idx` unlink there too.

src/supertable/reader_cache/disk.rs
```diff
@@ -1304,6 +1309,20 @@ impl DiskCacheStore {
...
+    fn blocks_idx_path(&self, uri: &SuperfileUri) -> PathBuf {
+        self.config.cache_root.join(format!(
+            "{}{BLOCKS_FILE_SUFFIX}{BLOCKS_IDX_SUFFIX}",
+            uri.cache_filename()
+        ))
+    }
+
+    /// Account bytes for an adopted block cache whose data is already on disk; a
+    /// later reservation evicts if this pushes the store over budget.
+    pub(super) fn account_adopted_bytes(&self, bytes: u64) {
```
muralikpbhat: **Cost/budget: `account_adopted_bytes` adds the whole footprint with no budget check or eviction.**

On the `owns_accounting` vector path, the first query after reopen `fetch_add`s the entire adopted footprint. Re-opening a heavily-warmed large table (GBs of retained blocks) can push `current_bytes` far over `disk_budget_bytes` in one step; unlike `reserve_block_bytes` nothing evicts victims, so the store stays over budget until a later reservation happens to catch up — and a read-only table never fills, so it never does. Reserve/evict to budget on adopt, or cap the adopted bytes.

src/supertable/reader_cache/block_source.rs
savannahar68: thank fixed it

src/supertable/reader_cache/block_source.rs
savannahar68: fixed it in latest commit. Please check

src/supertable/reader_cache/block_source.rs
savannahar68: Done

src/supertable/reader_cache/block_source.rs
savannahar68: fair done

src/supertable/reader_cache/disk.rs
savannahar68: done added logic to consider blocks as part of storage budget bytes

src/supertable/reader_cache/disk.rs
savannahar68: thanks for this fixed now

src/supertable/reader_cache/disk.rs
savannahar68: fixed

savannahar68:
@muralikpbhat i've addressed all the review comments. 
please feel free to take a look

src/supertable/reader_cache/disk.rs
```diff
@@ -2724,6 +2824,10 @@ async fn lazy_background_fill(
                     fill_spawned: AtomicBool::new(true),
                     last_access_us: AtomicU64::new(store.now_us()),
                 });
+                if !block_source_retained {
```
muralikpbhat, line 2868: **New — budget leak on promotion.** This `!block_source_retained` cleanup deletes `.blocks`/`.idx` with raw `fs::remove_file` but never clears the scan-counted `block_files[uri]`. After a restart `scan_block_file` counted this file's bytes into `current_bytes` + `block_files`; promoting the whole file here leaves that entry behind, so `current_bytes` stays inflated by F bytes (the effective disk budget silently shrinks) until a later `evict_block_files` happens to reconcile a phantom entry against an already-deleted file. Call `store.drop_block_file(&uri)` instead of the two raw removes.

src/supertable/reader_cache/block_source.rs
```diff
@@ -199,6 +200,61 @@ impl BlockCachedSource {
...
+            return None;
+        }
+        let idx_bytes = fs::read(self.idx_path()).ok()?;
+        let (bitmap, adopted_bytes) = indexed_filled_bytes(size, &idx_bytes)?;
+        let file = fs::OpenOptions::new()
+            .read(true)
+            .write(true)
+            .open(&self.path)
+            .ok()?;
+        if self.owns_accounting
```
muralikpbhat: **New — double-count + can unlink a live promoted reader's file.** A promotion-path source has `owns_accounting=false`, so a successful `try_adopt` here skips `account_adopted_bytes` and never clears `block_files[uri]`, while the promoted entry also reserves the full object size → the block bytes are counted twice. Under budget pressure `evict_block_files` can then `remove_file` the `.blocks`/`.idx` out from under the live `HoleFallback` reader still serving them (reads survive via the open fd/mmap, but the warm cache is silently discarded and `persist_idx` recreates a stray `.idx` with no data file). The adopt path needs to reconcile `block_files[uri]` even when `owns_accounting` is false.

src/supertable/reader_cache/disk.rs
```diff
@@ -1127,8 +1133,15 @@ impl DiskCacheStore {
             let Some(name) = path.file_name().and_then(|n| n.to_str()) else {
                 continue;
             };
-            if name.ends_with(BLOCKS_FILE_SUFFIX) {
-                let _ = fs::remove_file(&path);
+            if name.ends_with(BLOCKS_IDX_SUFFIX) {
+                continue;
+            }
+            if let Some(body) = name.strip_suffix(BLOCKS_FILE_SUFFIX) {
```
muralikpbhat, line 1153: **New — disk-leak regression.** A `.blocks` with no valid `.idx` (crash between `set_len` and the first `persist_idx`) makes `scan_block_file` return `None`, so this scan neither counts nor deletes it — it's left in place. Pre-PR, scan removed *every* `.blocks` sidecar, so an un-adoptable sparse file for a superfile never queried again now leaks across every subsequent restart with no reclaim path. Delete `.blocks` (and any `.idx`) when the index is absent/unreadable.

src/supertable/reader_cache/block_source.rs
```diff
@@ -199,6 +200,61 @@ impl BlockCachedSource {
...
+            let mut buf = Vec::with_capacity(filled.serialized_size());
+            if filled.serialize_into(&mut buf).is_err() {
+                return;
+            }
+            buf
+        };
+        let idx = self.idx_path();
+        let seq = PERSIST_COUNTER.fetch_add(1, Ordering::Relaxed);
+        let mut tmp = idx.clone().into_os_string();
+        tmp.push(format!(".tmp.{}.{seq}", process::id()));
```
muralikpbhat, line 264: **New — orphan tmp accumulation.** These `<name>.blocks.idx.tmp.<pid>.<seq>` files match none of `scan_cache_root`'s reclaim rules (not `.idx`, not `.blocks`, and `from_cache_tmp_filename` strips only `.tmp`, not `.tmp.<pid>.<seq>`), so a crash/failed-rename after the write orphans them permanently — and the per-call unique `<seq>` means each crash leaves a *distinct* orphan (the old fixed-name tmp was simply overwritten next time). Either make the tmp name reclaimable by `scan_cache_root` or sweep `*.idx.tmp.*` on scan.

muralikpbhat (commented):
Re-review of `eaaf88c5`. **All 7 prior findings are genuinely resolved** — verified, not superficial:

| # | prior finding | fix |
|---|---|---|
| 1 | adopted cache: no fsync + no validation | `sync_data()` gates the idx persist; torn idx → deserialize fails → safe refetch |
| 2 | `block_len()` underflow | `indexed_filled_bytes` rejects a bitmap whose `max() >= n_blocks` |
| 3 | mutex across write+rename | serialize scoped; guard dropped before I/O |
| 4 | fixed tmp-path race | `.tmp.<pid>.<seq>` |
| 5 | uncounted restart blocks | `scan_block_file` counts them; `evict_at_least` evicts them |
| 6 | `.idx` cleanup gaps | unified `drop_block_file` wired into evict/reuse/gc |
| 7 | unbudgeted adopt | release-then-add is net-zero |

The crash-safety hole (the one that mattered) is closed. The fix did introduce **6 new issues**, flagged inline — **two are material and both live in the fix's own new accounting paths**:
- `disk.rs:2827` — promotion cleanup never clears `block_files[uri]` → permanent `current_bytes` inflation (budget shrinks).
- `block_source.rs:223` — `owns_accounting=false` adopt double-counts and lets `evict_block_files` unlink a `.blocks` a live promoted reader is still serving.

The other four are lower severity (an un-adoptable-`.blocks` disk-leak regression, orphan `.tmp.<pid>.<seq>` files, no sidecar format-version, and an O(n²) persist). I'd hold merge on the two accounting bugs.

src/supertable/reader_cache/block_source.rs
```diff
@@ -327,21 +384,36 @@ impl BlockCachedSource {
...
             }
         }
+        if filled_any && bf.file.sync_data().is_ok() {
+            self.persist_idx();
+        }
         Ok(true)
     }
 }
 
+pub(super) fn indexed_filled_bytes(size: u64, idx_bytes: &[u8]) -> Option<(RoaringBitmap, u64)> {
```
muralikpbhat, line 429: **New (defense-in-depth) — no format/version stamp on the sidecar.** `CACHE_BLOCK_BYTES` is a compile-time const with no marker in the `.idx`. If a later binary changes the block size and reuses a warm cache dir, `indexed_filled_bytes` reinterprets old block ids under the new granularity: an id still `< n_blocks` now maps to a hole in the sparse file, so `all_filled` passes and `read_local` returns zeroed bytes as cached data — a silent wrong-result. A 4-byte format version compared on adopt makes it fail-closed (refetch).

src/supertable/reader_cache/block_source.rs
```diff
@@ -327,21 +384,36 @@ impl BlockCachedSource {
...
             let newly = self.mark_filled(bf.size, rb0, rb1);
             self.filled_bytes.fetch_add(newly, Ordering::AcqRel);
+            filled_any = true;
             if self.owns_accounting && newly < run_len {
                 // A concurrent filler beat us to some blocks; its accounting
                 // stands, ours is released.
                 store.release_block_bytes(run_len - newly);
             }
         }
+        if filled_any && bf.file.sync_data().is_ok() {
```
muralikpbhat: **New (cost) — whole-index rewrite + fsync per fill.** `fill_missing` now re-serializes the entire (growing) roaring bitmap and `fs::write`+`rename`s it, plus `sync_data()`, after *every* fill that touches any block. For n progressive fills that's n full-index rewrites (~O(n²) bytes) and n fsyncs on the read path. The bitmap is small so absolute cost is modest, but per the cost-as-first-class rule, coalescing/throttling the persist (or persisting on drop) avoids the repeated whole-index rewrites.

savannahar68:
JFYI - i've raised gh issue for making some path of code where we are doing IO to use tokio blocking pool. Will pick that up separately

src/supertable/reader_cache/block_source.rs
savannahar68: done

src/supertable/reader_cache/block_source.rs
savannahar68: done

src/supertable/reader_cache/block_source.rs
savannahar68: donew

src/supertable/reader_cache/disk.rs
savannahar68: donew

src/supertable/reader_cache/block_source.rs
savannahar68: done

src/supertable/reader_cache/disk.rs
savannahar68: done

savannahar68:
@muralikpbhat addressed the next set of comments. Please check

muralikpbhat (changes requested):
Re-reviewed against `fix-block-cache`. Five of the seven are fully fixed, with tests — the block-id range check + `IDX_HEADER` stamp, the bitmap snapshot before IO in `persist_idx`, the unique `.tmp.{pid}.{seq}` names, scanning `.blocks`/`.idx` into `current_bytes` and making them evictable, and routing all deletes through `drop_block_file`. Thanks for those.

Two still need work before this is safe to merge.

**1. `account_adopted_bytes` adds to the budget without reserving or evicting.** At head the whole body is:
```rust
pub(super) fn account_adopted_bytes(&self, uri: &SuperfileUri, bytes: u64) {
    self.release_scanned_block_file(uri);
    self.current_bytes.fetch_add(bytes, Ordering::Release);
}
```
`release_scanned_block_file` removes the URI from the evictable `block_files` set, then the bytes are `fetch_add`ed with no `reserve` / `evict_at_least` / cap. The evictability added for *scanned* files doesn't cover *adopted* bytes. So reopening a heavily-warmed, read-only table adopts its entire cached footprint in one step, pushing `current_bytes` past `disk_budget_bytes`; and because a read-only table issues no later reservation, nothing ever reclaims it. The adopt path needs to reserve/evict to budget (or cap what it adopts).

**2. The fsync barrier isn't atomic against concurrent fills.** `sync_data()` now gates `persist_idx()` in both the fill and Drop paths — that closes the single-writer hazard, good. But `persist_idx` serializes the *shared* `filled` bitmap, and a second in-flight `range()` can `write_all_at` + mark a new block in the window between one caller's `sync_data()` returning and its bitmap snapshot inside `persist_idx`. The persisted `.idx` then names a block whose bytes were not in that fsync; after a crash it adopts as sparse zeros and is served with `verify_crc:false` — silent wrong data. Either persist only the blocks proven durable under the same critical section that established the fsync, or add the per-block/index checksum (the alternative from the original thread) so adoption fails closed.

The rest of the second-round items check out. Once these two are addressed I'll re-review.

savannahar68:
Both fixed and pushed in 62036e4c.

1. account_adopted_bytes and budget. You are right that a read-only reopen could stay over budget. One correction on where the bytes enter: at head account_adopted_bytes is net zero for a scanned file, because release_scanned_block_file subtracts the scanned bytes and then we add the same count back. The over-budget actually lands at scan, where we add every retained .blocks file into current_bytes with nothing reserved or evicted. So I added a synchronous trim_cold_to_budget that frees cold candidates (unindexed files first, then block files) back under budget, and call it both at the end of scan and on adopt. Only cold, unheld files are freed; live entries are left to the async reservation path, and the file being adopted was just claimed out of block_files so it cannot evict itself. New test: reopen_trims_retained_block_files_to_budget.

2. fsync barrier. Real race, thanks for catching it. persist_idx used to re-read the shared bitmap after sync_data, so a second fill could write and mark a new block in that gap and the index would name a block the fsync had not flushed. It now snapshots the bitmap before the fsync and persists that exact snapshot, in both the fill and drop paths. Every block in the snapshot was written before it was marked, so before the snapshot, so the fsync covers it; a fill that marks a block after the snapshot is simply absent, which is a safe undercount.

91 reader_cache tests pass, clippy and fmt clean. Ready for another look.

muralikpbhat (approved):
Re-reviewed at `62036e4c` — both remaining findings are fixed. Clearing the earlier change request.

- **#7 (adopted bytes over-budget):** `account_adopted_bytes` now calls `trim_cold_to_budget()`, which evicts the overage (unindexed caches first, then block files) whenever `current_bytes > disk_budget` — so a reopened warm read-only table can't sit permanently over budget. Preferring unindexed evictions before the just-adopted indexed blocks also keeps the retention this PR is for.
- **#1 (block-index fsync race):** `persist_idx` now takes a caller-captured `snapshot`, and both the `fill_missing` and `Drop` paths capture it via `snapshot_index()` **before** `sync_data()` and persist that exact snapshot. Since a block is in the snapshot only after its `write_all_at` (pre-snapshot), `sync_data` flushes every block the `.idx` names — a concurrent filler's later write can't sneak into this index. The lock is held only during serialize, not across the IO, which also resolves the earlier mutex-across-IO note.

Earlier findings (#2–#6) remain fixed. LGTM.

PR #694, opened 2026-09-01
chore: rotate storage creds in place

savannahar68:
add support to swap creds in place instead of building a new connection on refreshing creds

### DIFF
--- a/src/catalog/mod.rs
+++ b/src/catalog/mod.rs
@@ -60,6 +60,7 @@ use crate::{
     storage::{
         AzureStorageProvider, GcsStorageProvider, LocalFsStorageProvider, S3StorageProvider,
         StorageError, StorageProvider,
+        gcs::{GCS_BEARER_TOKEN_OPTION, SwappableGcpCredential},
     },
     superfile::{
         builder::FtsConfig,
@@ -115,11 +116,23 @@ pub fn connect_with(
         return connect_remote(backend, options);
     }
     let usage_meter = UsageMeter::new();
+    let gcs_credential = match &backend {
+        Backend::Gcs { .. } => options
+            .storage_options
+            .get(GCS_BEARER_TOKEN_OPTION)
+            .map(|bearer| SwappableGcpCredential::new(bearer.clone())),
+        _ => None,
+    };
     let store = match &backend {
         Backend::Memory => CatalogStore::Memory(Mutex::new(HashMap::new())),
         _ => {
-            let root = backend_to_provider(&backend, &options, Arc::clone(&usage_meter))?
-                .expect("non-memory backend yields a storage provider");
+            let root = backend_to_provider(
+                &backend,
+                &options,
+                Arc::clone(&usage_meter),
+                gcs_credential.as_ref(),
+            )?
+            .expect("non-memory backend yields a storage provider");
             // Opt-in probe: fail at connect on bad credentials, not first use.
             if options.validate {
                 bridge_sync_to_async(read_catalog(root.as_ref()))?;
@@ -149,6 +162,7 @@ pub fn connect_with(
             store,
             connection_memory_budget,
             usage_meter,
+            gcs_credential,
         }),
     })
 }
@@ -211,6 +225,8 @@ struct ConnectionInner {
     /// Sole object-store usage ledger for this connection (shared into every
     /// table provider). Benches and billing snapshot this meter.
     usage_meter: Arc<UsageMeter>,
+    /// Swappable GCS credential shared by every provider on this connection.
+    gcs_credential: Option<Arc<SwappableGcpCredential>>,
 }
 
 /// Where the `name → table` map lives. Durable backends persist it on the
@@ -394,6 +410,7 @@ impl Connection {
                     &self.inner.backend.join(&location),
                     &self.inner.options,
                     Arc::clone(&self.inner.usage_meter),
+                    self.inner.gcs_credential.as_ref(),
                 )
                 .map_err(|e| e.with_context("create_table", Some(name)))?
                 .expect("non-memory backend yields a storage provider");
@@ -536,6 +553,7 @@ impl Connection {
                     &self.inner.backend.join(&entry.location),
                     &self.inner.options,
                     Arc::clone(&self.inner.usage_meter),
+                    self.inner.gcs_credential.as_ref(),
                 )
                 .map_err(|e| e.with_context("open_table", Some(name)))?
                 .expect("non-memory backend yields a storage provider");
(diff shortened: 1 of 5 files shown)


muralikpbhat (commented):
Full review of the in-place cred-rotation change — **the core design is right**: `ArcSwap<GcpCredential>` implementing `CredentialProvider` means the GCS client reads the token per request (verified: no `TokenCache` wraps a custom provider, so `store()` is picked up by the next request while in-flight requests keep a consistent `load_full` snapshot), one shared credential rotates all of a connection's providers, and the trait default `Ok(false)` correctly makes S3/Azure/Local signal "reopen". Together with #692 (retain cache on reopen) and the platform wiring, this closes the credential-refresh churn.

5 findings inline — **#1 (Debug token leak) is the one to fix before merge**; #2 is a silent rustdoc mis-attribution; #3–5 are API-contract clarity (redundant trait path, partial-update-reported-as-success, undocumented shared-vs-opts precedence).

src/storage/gcs.rs
```diff
@@ -49,7 +50,34 @@ const GCS_REQUEST_TIMEOUT: Duration = Duration::from_secs(300);
 /// `object_store` config-key, so it is applied programmatically through the
 /// client's credential provider. When present it is consumed here rather than
 /// folded on as a `google_*` config key (which would be rejected as unknown).
-const GCS_BEARER_TOKEN_OPTION: &str = "google_bearer_token";
+pub(crate) const GCS_BEARER_TOKEN_OPTION: &str = "google_bearer_token";
+
+/// A GCS bearer credential the client reads per request, so a rotation swaps the token in place.
+#[derive(Debug)]
+pub struct SwappableGcpCredential {
```
muralikpbhat, line 56: **Security — plaintext bearer token leaks via derived `Debug`.** `object_store::gcp::GcpCredential` (0.13.2) is `#[derive(Debug)]` over `pub bearer: String` with **no redaction**; `arc_swap`'s `Debug` formats the loaded inner value; and both `SwappableGcpCredential` and `GcsStorageProvider` `#[derive(Debug)]` — while the `StorageProvider` trait *requires* `fmt::Debug`. So any `{:?}`/tracing of a provider (or an error transitively holding one) prints `GcpCredential { bearer: "<live token>" }` to logs. Pre-PR the token lived only inside the `object_store` HTTP client, never a Debug-derived field — this PR introduces the leak surface. Fix: hand-write `Debug` for `SwappableGcpCredential` that redacts `bearer`, and don't derive `Debug` on anything exposing it.

src/catalog/mod.rs
```diff
@@ -591,6 +610,23 @@ impl Connection {
     /// # let _ = posts;
     /// # Ok::<(), Box<dyn std::error::Error>>(())
     /// ```
+    /// Rotate the object-store credential in place; `false` if unsupported and the caller should reopen.
```
muralikpbhat: **Doc bug — `update_storage_credentials` was inserted between `open_table`'s doc block and its `fn`.** rustdoc now attaches `open_table`'s summary + ```` ```rust ```` example to `update_storage_credentials`, and this method's own one-line summary lands *after* the closing fence. `cargo doc` renders the credential method with "Open an existing table by name … [db.open_table(\"posts\") example] … Rotate the object-store credential in place" — wrong, and silent (the doctest still compiles). Move this method's `///` summary directly above its `fn`, and keep `open_table`'s doc+example attached to `open_table`.

src/storage/gcs.rs
```diff
@@ -426,6 +472,16 @@ impl StorageProvider for GcsStorageProvider {
     fn usage_meter(&self) -> Arc<UsageMeter> {
         Arc::clone(&self.meter)
     }
+
+    fn update_credentials(&self, opts: &StorageOptions) -> Result<bool, StorageError> {
```
muralikpbhat: **Redundant, non-composing rotation path.** `StorageProvider::update_credentials` (this override + the `Ok(false)` default) is never called in-tree — `Connection::update_storage_credentials` rotates by mutating the shared `gcs_credential` `Arc` directly and bypasses the trait method entirely. Two mechanisms for one job invite drift (a future edit to one won't affect the other). Either have `Connection::update_storage_credentials` fan out via `provider.update_credentials()`, or drop the trait method and keep only the shared-`Arc` path.

src/catalog/mod.rs
```diff
@@ -591,6 +610,23 @@ impl Connection {
...
+    pub fn update_storage_credentials(&self, storage_options: &[(String, String)]) -> bool {
```
muralikpbhat, line 627: **Misleading return contract.** This rotates only the `google_bearer_token` key and silently ignores every other key in `storage_options`, yet returns `true` ("rotated") — a caller passing a changed endpoint/service-account + bearer gets a silent partial update reported as full success. Conversely it returns `false` (documented as "unsupported, reopen") on a perfectly rotatable GCS connection whenever the caller just didn't include a bearer key, prompting an unnecessary reconnect. Either apply/validate all supplied keys, or document that only the bearer is honored and distinguish 'unsupported backend' from 'no bearer supplied'.

src/storage/gcs.rs
```diff
@@ -78,21 +107,31 @@ impl GcsStorageProvider {
...
-        // A bearer token has no object_store config-key, so pull it out and set
-        // it on the client's credential provider; the rest fold on as config keys.
         let mut config_opts = opts.clone();
-        if let Some(bearer) = config_opts.remove(GCS_BEARER_TOKEN_OPTION) {
-            builder =
-                builder.with_credentials(Arc::new(StaticCredentialProvider::new(GcpCredential {
-                    bearer,
-                })));
+        let bearer = config_opts.remove(GCS_BEARER_TOKEN_OPTION);
+        let credential = match shared {
```
muralikpbhat, line 151: **Undocumented precedence — an opts bearer is silently dropped when `shared` is `Some`.** `match shared { Some(shared) => Some(shared), None => bearer.map(...) }` ignores the (already-removed-from-config) `bearer` whenever a shared credential is supplied. Harmless inside `Connection` (shared derives from the same opts), but this is a `pub` constructor — an external caller passing both, expecting the explicit per-provider bearer to win, has it silently discarded. Document that `shared` wins, or error when both are supplied and differ.

savannahar68:
@muralikpbhat the Pr is open for review 
addressed all the comments. Mostly they were docs / styling

muralikpbhat (approved):
Re-reviewed after the updates — all five earlier findings are addressed:

- **Token-leak via `Debug` (the blocker): fixed.** Hand-written redacting `Debug` on both `SwappableGcpCredential` and `GcsStorageProvider`, and `debug_output_redacts_the_bearer_token` asserts the real token is absent from both renderings — not just that the placeholder is present.
- **Doc mis-attribution: fixed** — the summary now sits directly above `update_storage_credentials`.
- **Redundant trait path: fixed** — the unused `StorageProvider::update_credentials` override is gone; the shared-`Arc` path is the single mechanism.
- **Return contract & shared-vs-opts precedence: documented** — bearer-only + reopen-on-`false`, and `shared` is authoritative over an opts bearer.

Core design confirmed sound: `ArcSwap<GcpCredential>` read per request means an in-flight operation picks up a swap on its next request while keeping a consistent `load_full` snapshot, and one shared credential rotates every provider on the connection. LGTM.

PR #695, opened 2026-09-01
fts: faster multi-term intersection — dedup repeated terms + selective-first probe order

vinaykakade:
Two independent improvements to the multi-term full-text intersection path (ranked and count).

## 1. Deduplicate repeated query terms (fold query-term-frequency into idf)

The scored path built one term cursor per query *token*. When a query repeats a term — e.g. `+to +be +or +not +to +be`, or the disjunction `to be or not to be` — each occurrence became its own cursor that independently decoded the term's whole posting list and independently added its block-max to the conjunction's score bound. So a repeated term cost **double the decode** and **a looser AND bound** (the doubled block-max defeated block-max pruning, exactly on the most common terms). Repeated stopwords are the worst case.

**Fix:** collapse repeats to one cursor per distinct term, carrying a per-term query-term-frequency folded into the precomputed `idf_x_k1p1` weight. BM25 is linear in that factor, so one weighted cursor is numerically identical to N duplicate cursors — same scores, same docs — while decoding once and keeping a tight bound. Distinct-term queries hit the same code with all-ones weights (a no-op).

## 2. Probe the most-selective term first in the membership intersection walk

The rarest-driven membership walks (ranked and count) probe the non-driver terms with a short-circuiting all-present test, but left those terms in arbitrary query order. Ordering them **rarest-first (ascending doc frequency)** rejects a non-matching driver doc in the fewest probes and avoids consulting a very common term's large presence structure for docs a rarer companion has already excluded. Count and scores are unchanged (intersection and the score sum are order-independent).

## Correctness

- New brute-force oracle: `+common +common` returns the same docs as `+common` with scores ×2, checked as an exact doc-id set at small `k` (varied doc lengths so scores differ and the set-equality gates).
- All existing FTS brute-force oracles pass; full FTS integration layer green (110 tests).
- Full sweep over 962 representative queries × {COUNT, TOP_10/100/1000}: **0 result differences** vs `main` (identical counts and returned doc sets) for both changes.

## Performance

Measured on a single-superfile full-text index over a ~5M-doc English-Wikipedia corpus, per-query median latency, branch vs `main`, each binary run in isolation.

**Dedup** (the dominant win): overall 0.967 across the 962×4 sweep. Repeated-term queries — `+to +be +or +not +to +be` **TOP_100 −93%**, TOP_1000 −92%, TOP_10 −89%; `to be or not to be` TOP_1000 −80%, TOP_100 −73%. Broad ranked wins wherever a query repeats a term (intersection TOP_100 aggregate ≈ 0.90); distinct-term and phrase queries unchanged.

**Probe order:** makes intersection COUNT order-independent — a query whose terms happened to be written common-term-first drops ~25% to match the selective-first order — for a ~−6.6% aggregate on intersection COUNT, no result change. Per-query variation elsewhere is within the usual measurement noise.

No format change; no public API change.

### DIFF
--- a/src/superfile/fts/reader/search.rs
+++ b/src/superfile/fts/reader/search.rs
@@ -1162,9 +1162,10 @@ impl FtsReader {
     /// dictionary fetch and a single overlapped range wave, not one of each
     /// per term.
     /// `qtf`, when present, is a per-term query-term-frequency weight (parallel to
-    /// `terms`): each built cursor's `idf_x_k1p1` is multiplied by its weight, so a
-    /// deduplicated repeated term scores identically to the duplicates it replaced
-    /// (BM25 is linear in `idf_x_k1p1`). `None` leaves idf unweighted.
+    /// `terms`): each built cursor folds its weight into the effective idf, which
+    /// scales both the score and the BlockMaxWAND skip ceilings, so a deduplicated
+    /// repeated term ranks identically to the duplicates it replaced (BM25 is
+    /// linear in idf). `None` leaves idf unweighted.
     pub(super) async fn build_term_cursors_opt(
         &self,
         column_id: u32,
@@ -1253,11 +1254,14 @@ impl FtsReader {
                         false => tf,
                     };
                     let dl_norm_k1 = col_meta.dl_norm_k1.get(doc_id);
-                    let mut cursor =
-                        TermCursor::new_inline(doc_id, tf, self.n_docs as u64, dl_norm_k1, gidf);
-                    if weight > 1 {
-                        cursor.idf_x_k1p1 *= weight as f32;
-                    }
+                    let cursor = TermCursor::new_inline(
+                        doc_id,
+                        tf,
+                        self.n_docs as u64,
+                        dl_norm_k1,
+                        gidf,
+                        weight,
+                    );
                     cursors.push(Some(cursor));
                 }
                 Some(Resolved::Pfor {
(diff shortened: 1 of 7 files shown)


src/superfile/fts/reader/search.rs
```diff
@@ -1204,28 +1253,31 @@ impl FtsReader {
...
-                        doc_id,
-                        tf,
-                        self.n_docs as u64,
-                        dl_norm_k1,
-                        gidf,
-                    )));
+                    let mut cursor =
+                        TermCursor::new_inline(doc_id, tf, self.n_docs as u64, dl_norm_k1, gidf);
+                    if weight > 1 {
+                        cursor.idf_x_k1p1 *= weight as f32;
```
naman-modi: Hey Vinay, this scales the score by the repeat count but not `block_max_bm25` / `term_max_bm25`. Those are the pruning ceilings, and they're still 1x, so for a repeated term the "highest possible score in this block" is now half the real value. The skip then drops blocks, which it shouldn't, and gives wrong results.

A possible test : `+common +common +rare`, common in even docs (df 200, 2 blocks), tf=2 except tf=6 at doc 100 (block 1) and tf=11 at doc 300 (block 2). at k=1:
- doc 100 sets the bar at its real 2x score
- block 2's bound uses common's 1x block_max, comes out below that bar, so block 2 gets skipped
- doc 300 (the real top-1) is dropped → engine returns [100], brute force returns [300]

I think we need to scale block_max_bm25 (each block) and term_max_bm25 by weight too, and also test for this.

naman-modi (changes requested):

src/superfile/fts/reader/search.rs
vinaykakade: Good point, fixed and added a test

naman-modi (approved):
LGTM, vinay.

PR #698, opened 2026-09-02
superfile/vector: re-derive fine-cluster count when merging cells

muralikpbhat:
## Problem

Vector search over a large table can end up scanning most of the corpus on a cold query. The root cause is in hidden-index compaction: when per-cell fragments merge, the merged cell kept the *source* fine-cluster count instead of re-deriving it from the merged row count.

Fine clusters (the sub-IVF runs within a cell) are sized so each holds about one fine-run byte target of encoded rows; at drain time the count is `ceil(rows / rows_per_run)`. But merge had two escape hatches:

- same-shape fragments **byte-splice** cluster-*i* onto cluster-*i* and keep the source `n_cent`, so the merged clusters hold the *sum* of both sources' rows;
- the rebuild paths used `n_cent = max(sources)`.

Neither re-fit the target, so a cell's fine clusters fatten as it grows through drain generations. A fat cluster's summary centroid drifts to the blob's center of mass, far from its boundary rows, so cell routing misranks it — recall caps out and cold reads balloon.

## Fix

One shared `fine_run_target_n_cent(dim, codec, n_rows)` is now the single sizing authority for both drain and merge. The byte-splice fast path is gated: it runs only when the merged union still fits the source width (`fine_run_target_n_cent(merged_rows) <= source n_cent`) — two under-target clusters legitimately merge into one. When the union crosses the target, the merge falls through to the rebuild path and re-clusters to the re-derived width. Both rebuild branches (different-shape and tombstone) re-derive `n_cent` from the merged row count.

## Correctness / recall

On the streaming-ingest recall bench (200K docs, two drains with a synchronous optimize between them, synthetic 1024-d Sq16, fine-run target scaled down so a small corpus exercises the multi-cluster merge path):

| checkpoint | recall@10 | over-target fine clusters |
| --- | --- | --- |
| after 1 drain | 1.000 | 0 |
| after 2 drains + merge | 0.992 | 0 |

Before the fix, the same run left hundreds of merged cells byte-spliced to a single fat cluster. After, every over-target merge re-clusters to `ceil(rows / rows_per_run)`, no fine cluster exceeds the target (`over_cap = 0`), and recall@10 holds at 0.992 (≥ the 0.99 bar).

## Tests

- `multi_cell_merge_resplits_when_union_exceeds_fine_run_target` — two same-shape fragments whose union outgrows the fine-run target; byte-splices to `n_cent=1` on the old code, re-clusters to the target width on the fix.
- `fine_run_target_n_cent_tracks_row_count` — pins the sizing contract (under target → 1, crossing → `ceil`, monotonic).
- Updated `multi_cell_merge_rebuilds_cells_with_mismatched_fine_width` — a small merged cell now re-derives to a single fine run instead of inheriting the widest source count.

## Performance / cost

The gate adds one `ceil` per merged cell. Over-target same-shape cells now take the rebuild path (k-means re-cluster) instead of the byte-splice — extra work only for cells that were being mis-sized, and it is exactly what restores routing accuracy and cuts the cold read.

The change makes a cell's fine clusters *finer and closer to target* as it grows; finer sub-IVF runs sharpen cell routing rather than degrade it, so the direction of the recall effect is positive. Recall@10 holds at 0.992 in the streaming-ingest bench above, and CI's `supertable_vector` (1M) bench plus the recall test exercise the drain+merge path at scale.

### DIFF
--- a/src/superfile/builder.rs
+++ b/src/superfile/builder.rs
@@ -109,8 +109,9 @@ use crate::superfile::{
         cell_posting::{CellPostingBuilder, MaterializedIvfRow},
         distance::Metric,
         ivf_merge::{
-            MergedIvfSubsection, Sq8IvfMergeInput, merge_sq8_ivf_subsections,
-            merge_sq8_ivf_subsections_from_parsed, stable_ids_in_merged_local_order,
+            MergedIvfSubsection, Sq8IvfMergeInput, fine_run_target_n_cent,
+            merge_sq8_ivf_subsections, merge_sq8_ivf_subsections_from_parsed,
+            stable_ids_in_merged_local_order,
         },
         layout::VectorLayout,
         reader::{ColumnReader, VectorReader},
@@ -970,9 +971,10 @@ impl SuperfileBuilder {
 
         if any_tombstones {
             // Materialize → filter by file-local tombstone id → rebuild per cell.
-            // Also track the max fine-cluster count seen per cell so rebuilds
-            // keep the source IVF width (empty clusters stay empty).
-            let mut by_cell: HashMap<u32, (usize, Vec<MaterializedIvfRow>)> = HashMap::new();
+            // The fine-cluster count is re-derived from the surviving row count at
+            // rebuild time (see the build loop) so a merged cell is re-clustered
+            // to the fine-run byte target rather than inheriting a source width.
+            let mut by_cell: HashMap<u32, Vec<MaterializedIvfRow>> = HashMap::new();
             for (reader_idx, (reader, deleted)) in readers.iter().enumerate() {
                 let v = reader.vec().ok_or(BuildError::VectorReadError)?;
                 let superseded = superseded_per_reader.get(reader_idx);
@@ -995,20 +997,19 @@ impl SuperfileBuilder {
                     if rows.is_empty() {
                         continue;
                     }
-                    let entry = by_cell.entry(cell_id).or_insert_with(|| (0, Vec::new()));
-                    entry.0 = entry.0.max(col.n_cent as usize);
-                    entry.1.extend(rows);
+                    by_cell.entry(cell_id).or_default().extend(rows);
                 }
             }
 
             let mut cell_ids: Vec<u32> = by_cell.keys().copied().collect();
             cell_ids.sort_unstable();
             for cell_id in cell_ids {
-                let (n_cent, mut rows) = by_cell.remove(&cell_id).expect("cell present");
+                let mut rows = by_cell.remove(&cell_id).expect("cell present");
                 for (i, row) in rows.iter_mut().enumerate() {
                     row.local_doc_id = i as u32;
                 }
                 let stable_ids: Vec<i128> = rows.iter().map(|r| r.stable_id).collect();
+                let n_cent = fine_run_target_n_cent(vec_cfg.dim, vec_cfg.rerank_codec, rows.len());
                 let merged = build_merged_subsection_from_materialized(
                     vec_cfg.clone(),
                     n_cent.max(1),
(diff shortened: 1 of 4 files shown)


savannahar68 (commented):
Reviewed this and the fix direction is right. Couple of things i wanted to put here which i discussed with claude to understand which i think we will have to consider 

#### 1. The splice gate compares against a target that can never actually get stored, so some cells rebuild on every compaction forever

`fine_run_target_n_cent` in `src/superfile/vector/ivf_merge.rs` computes the fine cluster count purely from the byte target. But there is a separate, older cap in `materialized_centroids` (`src/superfile/vector/builder.rs:1242`) that clamps any cell with 80,000 rows or fewer down to at most 64 fine clusters, no matter what the byte target asks for.

The new gate at `src/superfile/builder.rs:1066` does not know about that cap. It compares the uncapped byte target against the cell's stored `n_cent`, which is already capped. So for a cell whose true byte target is above 64 but whose row count stays under 80,000, this happens every time it gets compacted:

1. The gate computes a target above 64, sees it does not fit the stored 64, and rejects the cheap splice.
2. It falls to the rebuild path, which recomputes the same above-64 target.
3. That target gets clamped back down to 64 by the cap in `materialized_centroids`.
4. The cell is stored with `n_cent = 64` again, identical to before the rebuild.

Next compaction repeats all four steps. The cell never reaches a state where the gate says yes, so it never gets to skip the expensive rebuild. Worked example at dim 1024 with the Sq8Residual codec: a cell with 70,000 rows computes a target of 74 fine clusters, gets clamped to 64 on write, and every later merge pays a full materialize and k-means rebuild only to land back on 64. The row band where this happens starts around 61,000 rows at dim 1024, and starts lower for wider vectors.

Suggested fix: have the gate compare against `min(byte target, small cell cap)`, the number the system will actually store, not the raw byte target formula.

---

#### 2. The eager copy done before the gate check is wasted on most cells now

`src/superfile/builder.rs:1040` parses and copies the full byte contents of every source fragment, plus decodes all stable ids, before the gate even runs. Given point 2 above, the gate rejects the splice for most cells now, so that copy gets thrown away and the rebuild path reads the same range from the reader again from scratch.

The gate only needs `n_cent`, row count, dimension, and codec, all of which are already available cheaply on `ColumnReader` without copying anything. Worth restructuring so the eligibility check runs first, and the full parse only happens for cells that actually end up splicing.

muralikpbhat:
Thanks, good catch on #1 — that perpetual-rebuild band is real. Fixed in `3742809c`.

**#1 (gate targets a count the build won't store):** I extracted the build's cap switch — the `consolidated`-vs-`n_cent_row_count_cap` logic that both `materialized_centroids` and the training-sample sizing duplicated — into a single `effective_cell_n_cent(requested, n_docs)`. The splice gate (and the rebuild derivation) now compare against `effective_fine_n_cent` = the byte target passed through that same cap, i.e. the count the build actually writes. So a sub-threshold cell whose byte target sits above the small-cell cap now stores the capped count and the gate accepts the splice, instead of rebuilding to the same 64 every compaction. Both build clamp sites route through the one helper now, so the gate can't drift from the build. Added `effective_fine_n_cent_applies_the_small_cell_cap` covering the 70K-row / dim-1024 case (target 74 → stored 64) and the uncapped consolidated case.

**#2 (eager parse/copy before the gate):** agreed. With #1 the gate now accepts the splice for those capped cells, so the wasted-copy case narrows to cells that genuinely re-split — but the structural point stands: the eligibility check only needs `n_cent`/rows/dim/codec, all cheap on the reader, so the full `sq8_ivf_merge_input` parse should run only for cells that actually splice. It's a self-contained restructure of the merge loop; I'll do it as a follow-up commit here rather than fold it into this fix.

savannahar68 (approved):

PR #704, opened 2026-09-03
fix: proper offloading of tokio workers

savannahar68:
* some async function uses std::fs instead of tokio::Fs or doesn't use tokio::spawn_blocking where it should use. This PR attempts to fix couple of those issue so that throughput will improve.

### DIFF
--- a/src/supertable/compaction/mod.rs
+++ b/src/supertable/compaction/mod.rs
@@ -27,13 +27,13 @@ use futures::{
 };
 use roaring::RoaringBitmap;
 use tempfile::NamedTempFile;
-use tokio::time;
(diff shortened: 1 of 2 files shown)


Pranav2612000 (approved):

savannahar68:
not merging this because after the bench runs there looks like regression

muralikpbhat:
ok, would be good to add numbers in the PR description, especially for performance improvements claims. Especially for vectors, 'bench supportable vector', recall-while-ingest and VectorDBBench are very reliable tools that I have been using and helps to clearly call out the wins.

savannahar68:
ill close this for now 
Will pick this up in future

PR #707, opened 2026-09-03
supertable/query: attach the row-filter predicate once; keep LIMIT out of filtered scans

### LINKED ISSUE #705: SQL scan row filter evaluates the predicate twice (predicate=p AND p)
When a `WHERE` predicate cannot be bounded by the index (scalar column, range, negation), `SupertableProvider::scan` enables Parquet row filters on the scan and also builds the predicate itself from the same filters (`row_group_predicate`) and sets it on the `ParquetSource`. DataFusion 54's physical `FilterPushdown` rule then offers the `FilterExec`'s predicate to the source; `ParquetSource::try_pushdown_filters` conjoins it with the predicate already present. `EXPLAIN` shows the result:

    DataSourceExec: ... predicate=category@2 = y AND category@2 = y

Parquet's `RowFilter` evaluates conjuncts in order, each over the rows the previous one kept, so the duplicate costs one more evaluation of the predicate over every surviving row. Negligible for a selective predicate; close to a full second filter pass for a dense one. The SQL bench cells the merge gate watches (`AVG(rating)` under a majority `category`, `SUM(rating)` under `bucket IN (all)`) are dense scans of exactly this shape.

Fix: enable row filters on the unbounded path and let DataFusion attach the predicate — it does so once and drops the `FilterExec`. On the bounded path nothing changes (row filters off; DataFusion still stores the predicate on the source for statistics pruning). Found while writing the EXPLAIN test for #701; independent of that branch.

### LINKED ISSUE #706: LIMIT on an index-unbounded predicate can return deleted rows
Reproduce on `main` with a superfile large enough for DataFusion to byte-range split it (over 10 MiB, the production shape; a smaller file gets a round-robin `RepartitionExec` that happens to hide it):

    -- delete the first row, then
    SELECT title FROM supertable WHERE _id > 0 LIMIT 3

returns the deleted row first.

Mechanism: the predicate is not index-bounded, so the provider enables Parquet row filters and DataFusion folds the `FilterExec` into the scan. With no `FilterExec` left to hold the fetch, `LimitPushdown` embeds `LIMIT 3` in the `DataSourceExec`. The Parquet opener then runs `prune_by_limit`: when statistics prove a row group fully matches the predicate (`_id > 0` does for every group), it rebuilds the access plan as `scan(row group)` for enough groups to cover the limit — discarding the `RowGroupAccess::Selection` the provider installed for that superfile's tombstones. Those rows are also exempt from the row filter ("fully matched" runs skip it), so the deleted row is decoded and returned.

The provider already withholds its own scan-level limit whenever filters are present (`effective_limit`), but that guard assumed a `FilterExec` would absorb any fetch DataFusion pushed. Fix: the meter the provider wraps around a filtered scan refuses limit pushdown (`supports_limit_pushdown` false, `with_fetch` None), so DataFusion keeps a limit node above the scan and the access plans stand. Filter-less scans are unaffected (no predicate, so no row group is ever "fully matched").

Found by the adversarial review of the row-filter change above; predates it.

ekechinwokah:
Fixes #705 , fixes #706.

**Commit 1 — attach the row-filter predicate once.** On the unbounded scan path the provider set its own copy of the predicate on the `ParquetSource` and enabled row filters; DataFusion's filter pushdown then conjoined the `FilterExec`'s identical predicate onto it, so the Parquet row filter evaluated `p AND p` — a second pass of `p` over every row the first kept. The provider now only enables row filters there and lets DataFusion supply the predicate once; the hand-built predicate and its helper are gone. Row filters stay off for filter-less scans (an empty filter list also lowers to `Unbounded`), so DataFusion's dynamic TopK/join/aggregate filters keep their pruning-only role there, as on `main`. Test: `query_sql_row_filter_carries_the_predicate_once` — EXPLAIN of `WHERE category = 'y'` shows one predicate with no self-conjunction and no `FilterExec`, the rows come back exactly, and the bounded control keeps its `FilterExec`.

**Commit 2 — keep a `LIMIT` out of a filtered scan's access plans.** Pre-existing: with the `FilterExec` folded into the scan, `LimitPushdown` embedded the fetch in the scan and the Parquet opener's limit pruning replaced the tombstone row selection with whole "fully matching" row groups, returning deleted rows. The meter around a filtered scan now refuses the fetch, so DataFusion keeps its limit node above the scan. Test: `sql_limit_under_a_row_filter_keeps_deleted_rows_out` — a superfile past the 10 MiB byte-range split threshold, first row deleted, `WHERE _id > 0 LIMIT 3` excludes it; the plan shape that makes the check load-bearing is asserted; verified to fail without the guard.

Bench: the CI Benchmark checks on this PR are the comparison against `main`. Commit 1 should move the dense scalar scans in `supertable_sql` (`AVG(rating) WHERE category=?`, `SUM(rating) bucket IN (all)`) down; commit 2 touches only `LIMIT` queries with a `WHERE`, where the limit node now sits above the scan instead of inside it. No format, manifest or public-API change.

### DIFF
--- a/src/supertable/query/exec/metered_exec.rs
+++ b/src/supertable/query/exec/metered_exec.rs
@@ -74,6 +74,11 @@ use crate::runtime_metrics::{
 pub(crate) struct MeteredExec {
     input: Arc<dyn ExecutionPlan>,
     op_stats: Option<Arc<OpStatsCollector>>,
+    /// Whether the planner may carry a `LIMIT` fetch past this node into
+    /// the child. The one place this is `false` is the meter the table
+    /// provider wraps around a scan that carries filters — see
+    /// [`Self::without_limit_pushdown`].
+    limit_pushdown: bool,
 }
 
 impl MeteredExec {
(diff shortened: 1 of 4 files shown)


naman-modi (approved):
LGTM, Ekechi.

PR #709, opened 2026-09-03
chore: add logs in compaction

savannahar68:
added some logs around compaction which will help in debugging

### DIFF
--- a/src/supertable/compaction/mod.rs
+++ b/src/supertable/compaction/mod.rs
@@ -28,7 +28,11 @@ use futures::{
 use roaring::RoaringBitmap;
 use tempfile::NamedTempFile;
 use tokio::time;
-use tracing::{info, warn};
+#[cfg(not(feature = "detailed-tracing"))]
+use tracing::Span;
+#[cfg(feature = "detailed-tracing")]
+use tracing::info_span;
+use tracing::{Instrument, info, warn};
 use uuid::Uuid;
 
 use crate::{
@@ -481,6 +485,10 @@ impl Supertable {
     }
 
     /// Merges the given superfiles into one
+    #[cfg_attr(
+        feature = "detailed-tracing",
+        tracing::instrument(name = "merge_superfiles", skip_all, fields(inputs = superfiles.len()))
+    )]
     pub(crate) async fn merge_superfiles(
         &self,
         superfiles: &[Arc<SuperfileEntry>],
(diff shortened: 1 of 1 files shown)


src/supertable/compaction/mod.rs
```diff
@@ -428,7 +428,20 @@ impl Supertable {
...
+                "compaction jobs planned"
+            );
+            for job in jobs {
+                info!(
+                    role = table.role().as_str(),
+                    inputs = job.inputs.len(),
+                    partition_key = ?job.partition_key,
+                    estimated_output_bytes = job.estimated_output_bytes,
+                    "running compaction job"
+                );
```
Pranav2612000: Is it better to add a 'span' on the "run_compaction_job" method instead? Will also include the time the fn took.

Pranav2612000: Also helps gate it behind the 'tracing' feature of the engine, so its easy to disable and does not affect everyone.

savannahar68: yes makes sense
changing it

src/supertable/compaction/mod.rs
```diff
@@ -788,6 +803,12 @@ impl Supertable {
                     return Ok(());
                 }
                 Err(CommitError::WriteContentionExhausted) if attempt + 1 < max_retries => {
+                    info!(
+                        superfile_id = %merged_superfile_id,
+                        attempt,
+                        max_retries,
+                        "compaction commit lost race, retrying"
+                    );
```
Pranav2612000: Is this going to add a lot of noise? Thoughts on using through 'spans' to gate it behind the 'tracing' feature of the crate?

Pranav2612000 (approved):

Pranav2612000:
This may add a lot of noise, so we may want to revert it once we're done with the debugging we need this for. Would also prefer to use spans instead of logs so that it does not affect users who do not need these logs.

src/supertable/compaction/mod.rs
savannahar68: i'll monitor it. In case it creates a lot of noice we can remove it. Primarily need for debugging

savannahar68:
> This may add a lot of noise, so we may want to revert it once we're done with the debugging we need this for. Would also prefer to use spans instead of logs so that it does not affect users who do not need these logs.

fair moved from logs to spans

PR #711, opened 2026-09-03
fts: per-column stored flag; analyzer folded onto FtsConfig

vinaykakade:
Two changes on one branch, both about per-column FTS configuration.

## 1. `stored(false)` — index-only text columns

An FTS column can now be declared index-only: its text is tokenized into
the FTS blob at ingest but never written to the Parquet body, so large
search-only text (logs, chat, scraped pages) skips the stored copy
entirely. The column stays part of the append/update contract and
remains fully searchable (ranked BM25, token match, phrase when
positions are on), but it cannot be selected via SQL, named in a search
projection, or used in a predicate — those resolve against the stored
schema and fail up front like any unknown column, with clean messages.

Public API: `IndexSpec::fts` now takes `impl Into<FtsField>`, a
per-column builder (`analyzer`, `stored`); `From<&str>` supplies the
defaults so `.fts("body")` callsites compile unchanged, and
`fts_with_analyzer` folds into the new door (`public-api.txt` updated —
this is the breaking edge, so the release is a 0.x minor). Node and
python bindings expose the option.

Persistence is additive and back-compatible: `inf.fts.columns` gains
`"stored":false` emitted only when false (all-stored files stay
byte-identical; readers default a missing key to true; format version
1.0.0 → 1.1.0), the catalog entry gains a serde-defaulted `fts_stored`
vec, and the options hash appends its tagged block only when a column
is index-only, so existing manifests keep verifying.

Merges no longer re-tokenize: every merge path (generic reader merge,
streaming FTS merge, sq8 byte-splice, multi-cell) carries each input's
prebuilt postings and stored doc-lengths across — an index-only column
has no Parquet text to re-tokenize, and carrying is cheaper and
byte-faithful for stored columns too. The multi-cell merge reorders
rows by stable id, so its carry remaps postings through the packed
order and runs the FTS accumulator in spill mode (whose finish sorts
triples by term/doc); doc-lengths scatter into output order. The
vector merge machinery (sq8 splice, multi-cell assembly, fallback
vector re-encode) is untouched.

## 2. Analyzer folded onto `FtsConfig`

The per-column analyzer now lives on `FtsConfig` as a name (the same
record `inf.fts.columns` persists), built via `FtsConfig::new` plus
chained setters. This deletes the `fts_tokenizers` vec that ran
parallel to `fts_columns` on `BuilderOptions` and `SupertableOptions`
— an alignment invariant every construction site had to maintain — and
the redundant default-tokenizer field seeded from the first column's
analyzer in three places. Names are validated once at options
construction (typed error naming the valid set) and resolved once at
builder construction. The options-hash analyzer block reads the names
off `fts_columns` — same bytes, so existing hashes are unchanged
(golden tests pin this).

Also fixes a latent per-field-analyzer bug: the filtered vector search
tokenized its predicate with the table-wide default tokenizer; it now
uses the filter column's own analyzer.

## Tests

- Round-trip: stored schema drops the column; search and phrase work;
  KV, catalog, and reopen persist the flag.
- Merge ≡ fresh-build equivalence (hits + scores) on the generic and
  streaming paths, with tombstones; sq8 and multi-cell carries pinned
  at the unit level, including the stable-id reorder with phrase
  positions and doc-lengths.
- Options-hash golden stability for all-default tables; legacy catalog
  entries decode; clean SQL/projection/predicate rejections with no
  engine internals in messages.
- Analyzer plumbing: standard-tokenizer oracle, mixed per-column
  analyzers across reopen, unknown-analyzer rejection at both layers.

## Bench comparison vs main

`cargo bench --bench bench -- superfile fts` (1M docs), main vs this
branch on the same host: every warm ranked / top-1000 / COUNT / phrase
cell within run-to-run noise (±5%, mixed sign); ingest equal (1-writer
9.85 s → 9.89 s, positions builds identical); index size identical at
884.72 MiB (the bench corpus keeps stored text, so this is the
no-regression check — index-only columns are opt-in). Correctness
self-checks (BMW ≡ brute-force) green on both sides. No vector-path
changes, so the recall gate is unaffected.

### DIFF
--- a/src/superfile/builder.rs
+++ b/src/superfile/builder.rs
@@ -73,14 +73,13 @@ use std::{
     collections::{BTreeSet, HashMap, HashSet},
     fmt,
     io::{BufReader, BufWriter, Cursor, Error, Seek, SeekFrom, Write},
-    mem,
     str::from_utf8,
     sync::Arc,
 };
 
 use arrow::compute::{concat_batches, take};
 use arrow_array::{Array, ArrayRef, Decimal128Array, LargeStringArray, RecordBatch, UInt32Array};
-use arrow_schema::{DataType, Schema};
+use arrow_schema::{DataType, Field, Schema};
 use parquet::basic::{Compression, ZstdLevel};
 use roaring::RoaringBitmap;
 use tempfile::{NamedTempFile, tempfile};
@@ -130,6 +129,19 @@ pub struct FtsConfig {
     /// opt-in. Columns without positions answer phrase queries with a
     /// typed error, never a silent bag-of-words fallback.
     pub positions: bool,
+    /// Keep the raw text in the Parquet body (the default). When
+    /// `false` the column is index-only: it is tokenized into the FTS
+    /// blob but never written to Parquet, so it cannot be read back —
+    /// not via SQL, not via a search projection, not in predicates.
+    /// The trade is file size: large text corpora that are only ever
+    /// searched skip the raw-text copy entirely.
+    ///
+    /// At ingest the column must still be present in the builder
+    /// schema (the text has to arrive to be indexed); the builder
+    /// drops it from the Parquet-bound rows. When rebuilding from an
+    /// existing superfile the column is legitimately absent from the
+    /// stored schema and its postings are carried across instead.
+    pub stored: bool,
 }
 
 // `VectorConfig` (the per-column vector config used by
@@ -359,6 +371,7 @@ impl BuilderOptions {
                             FtsConfig {
                                 column: c.name.clone(),
                                 positions: c.positions,
+                                stored: c.stored,
                             },
                             Arc::clone(&c.tokenizer),
                         )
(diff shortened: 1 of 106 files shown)


vinaykakade:
Rebased onto main (0.5.15 + the LIKE-on-FTS-columns pushdown) and bumped the crate to 0.6.0 — the removal of `fts_with_analyzer` is the breaking edge on the 0.x line. Rebase conflicts were test fixtures only: the LIKE-oracle fixtures that landed with the pushdown now select their analyzer by name through the new `FtsConfig` constructor. Interaction note: `LIKE` on an index-only (`stored(false)`) column still fails at plan time — the stored schema drops the column before the pushdown sees it — which the query-surface test pins; term-dictionary-backed `LIKE` for index-only columns is a possible follow-up since the dictionary exists for them. Full suites re-run green after the rebase (lib 2393 / superfile 186 / supertable 325, public-api snapshot clean).

src/supertable/query/vector.rs
```diff
@@ -4863,11 +4863,12 @@ impl SupertableReader {
...
-        // Tokenize the predicate once with the index tokenizer (the same
-        // tokenizer used at build time, so the terms match the postings AND
-        // the manifest term blooms). No tokens (empty / punctuation-only) ⇒
+        // Tokenize the predicate once with the FILTER COLUMN's analyzer
+        // (the same one used at build time, so the terms match the
+        // postings AND the manifest term blooms). A non-FTS filter column
+        // matches nothing; no tokens (empty / punctuation-only) ⇒
         // nothing matches.
-        let Some(tokenizer) = manifest.options.tokenizer.as_ref() else {
+        let Some(tokenizer) = manifest.options.try_fts_tokenizer_for(filter.column) else {
```
naman-modi, line 4871: This is the analyzer bug fix (filter predicate now tokenized with the filter column's own analyzer, not the table-wide default), but there's no test for it. Can we add a test around it (like 2 columns in a table with different analyzers), so we have a regression test.

src/superfile/builder.rs
naman-modi, line 495: Here should we also add a check for the "analyzer type" for the fts columns?

naman-modi (approved):
Lgtm, few minor comments.

src/supertable/query/vector.rs
vinaykakade: Thanks, done

src/superfile/builder.rs
vinaykakade: Done

PR #713, opened 2026-09-04
vector: cover every input in the merge quantizer grid (fixes #554)

### LINKED ISSUE #554: Compaction/drain merge sizes the destination Sq8 grid from a subset, not the union — components clamp (silent recall loss) on L2/NegDot at scale

muralikpbhat:
Fixes #554.

## Problem

The fitted-ruler rerank codecs — `Sq16Adaptive` (default for L2/NegDot since #563) and `Sq8Residual` — fit each cluster's quantizer grid to that input's own `[min, max]`. When a compaction merges cells whose contributions to the same cluster carry different grids, the merge sized the destination grid to the **first** contributing input and re-encoded the rest onto it, **clamping any out-of-range component**. The merge already flagged this as deferred in a comment ("*Sizing the destination grid to cover every input's range … is a separate future change*").

At scale it is not rare: a 100M compaction merges cells whose L2/NegDot magnitude ranges diverge, so **hundreds of millions of components saturate**. The drain logs `component(s) saturated their destination quantizer`, and affected rows' recall silently degrades (observed: LAION-100M/L2 recall cratered after a compaction clamped ~370M components).

## Fix

Size the destination grid to the per-component **union** of every contributing input's range, so no component clamps. Each input `i` covers `[offset_i, offset_i + scale_i·code_max]`, so per cluster component:

- `dst_offset = min_i offset_i`
- `dst_scale  = (max_i (offset_i + scale_i·code_max) − dst_offset) / code_max`

over the inputs that contribute to the cluster. Gated on the fitted-ruler codecs (`fits_per_cluster_ruler`); the fixed-quantizer path (one pinned grid shared by every input) keeps its copy-first behavior. `code_max` comes from the codec (`65535` for the u16 planes, `255` for the u8 residual family), so the union math is correct for both families. A zero-span union (every contributing input pinned the same constant offset) falls back to the smallest input scale, else `f32::EPSILON`.

## Tradeoff

The union grid is intentionally **coarser** (its step spans the combined range instead of each input's tighter span) but **never clamps**. Coarser-but-covering is the correct trade against catastrophic silent saturation: a small per-component 16-bit precision increase versus total loss of hundreds of millions of out-of-range components.

## Test

Adds `sq16_adaptive_merge_union_ruler_never_clamps`: merges two `Sq16Adaptive` L2Sq cells whose cluster rulers cover disjoint magnitude bands, asserts the pre-fix first-input grid *would* have clamped, then asserts the union merge produces **zero** saturated components (via the process-wide clamp tripwire) and round-trips the union extremes within the merged step.

## Related

- #563 added the `Sq16Adaptive` codec and made it the default for L2/NegDot.
- #568 tracks the saturation warning mislabeling this path as "Sq8" (not addressed here).

### DIFF
--- a/src/superfile/vector/ivf_merge.rs
+++ b/src/superfile/vector/ivf_merge.rs
@@ -195,22 +195,88 @@ pub(crate) fn merge_sq8_ivf_subsections_from_parsed(
         (vec![1.0f32; n_cent * dim], vec![0.0f32; n_cent * dim])
     };
     if has_cluster_quant {
-        // Copy the first contributing input's per-cluster ruler as the
-        // destination grid; per-row transcode below re-encodes the other inputs
-        // onto it (clamping any out-of-range component). This matches the
-        // existing merge behavior for every fitted/fixed cluster-quant codec.
-        // (Sizing the destination grid to cover every input's range so no
-        // component clamps is a separate future change, not bundled here.)
-        for c in 0..n_cent {
-            for inp in parsed {
-                let (_, count) = cluster_entry(&inp.sub, inp.cluster_idx_off, c);
-                if count == 0 {
+        if codec.fits_per_cluster_ruler() {
+            // Fitted/adaptive cluster-quant codecs (`Sq8Residual`,
+            // `Sq16Adaptive`) fit each cluster's ruler to that input's own
+            // `[min, max]`, so different inputs contributing to the same merged
+            // cluster carry DIFFERENT grids. Sizing the destination to the FIRST
+            // input's grid clamps every later input's out-of-range components —
+            // at 100M a compaction merges cells with divergent magnitude ranges,
+            // so hundreds of millions of components saturate and recall craters.
+            // Instead size the destination grid to the per-component UNION of
+            // every contributing input's range, so no component clamps.
+            //
+            // Input `i` covers `[offset_i, offset_i + scale_i * code_max]` for a
+            // component (its ruler was fit as `scale_i = (max - min) / code_max`,
+            // `offset_i = min`). The union is `[min offset_i, max hi_i]`, and the
+            // merged ruler is `offset = lo`, `scale = (hi - lo) / code_max`.
+            let code_max = codec.code_max();
+            let mut lo = vec![0f32; dim];
+            let mut hi = vec![0f32; dim];
+            let mut min_scale = vec![0f32; dim];
+            for c in 0..n_cent {
+                let base = c * dim;
+                let mut any = false;
+                for inp in parsed {
+                    let (_, count) = cluster_entry(&inp.sub, inp.cluster_idx_off, c);
+                    if count == 0 {
+                        continue;
+                    }
+                    let s = &inp.scale[base..base + dim];
+                    let o = &inp.offset[base..base + dim];
+                    if !any {
+                        for d in 0..dim {
+                            lo[d] = o[d];
+                            hi[d] = o[d] + s[d] * code_max;
+                            min_scale[d] = s[d];
+                        }
+                        any = true;
+                    } else {
+                        for d in 0..dim {
+                            lo[d] = lo[d].min(o[d]);
+                            hi[d] = hi[d].max(o[d] + s[d] * code_max);
+                            min_scale[d] = min_scale[d].min(s[d]);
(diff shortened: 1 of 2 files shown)


src/superfile/vector/ivf_merge.rs
```diff
@@ -195,22 +195,88 @@ pub(crate) fn merge_sq8_ivf_subsections_from_parsed(
...
+                    dst_offset[base + d] = lo[d];
+                    // Guard a zero-span union (every contributing input pinned
+                    // the same constant offset with `scale == 0`): fall back to
+                    // the smallest input scale when it is positive, else a tiny
+                    // epsilon, so decode never divides by zero / yields a
+                    // zero-width grid. A positive scale in any input forces
+                    // `hi > lo`, so this branch is defensive.
+                    dst_scale[base + d] = if hi[d] > lo[d] {
+                        (hi[d] - lo[d]) / code_max
+                    } else if min_scale[d] > 0.0 {
```
savannahar68: i think this is a dead code

savannahar68 (approved):
lgtm

src/superfile/vector/ivf_merge.rs
muralikpbhat: Good catch — you're right, it's unreachable. A zero-span union (hi == lo) forces every contributing input's scale to 0 at that dimension, so min_scale is 0 there too and the `min_scale > 0` branch never fires. Removed it (and the min_scale tracking); it falls straight to the epsilon guard now. Fixed in dcdd9d9.

PR #722, opened 2026-09-04
fts: carried-postings order; bench: tempdir leak; ci: tolerant A/B baseline

vinaykakade:
Two independent bugs, both surfaced by the 10M/1M supertable bench lanes; the second is the cause of the red `supertable_sql` / `supertable_vector` Benchmark runs on `main` since #711.

## Bug 1 — bench harness: cold-tier cache tempdirs leak until the disk fills

**Symptom:** 10M supertable FTS cold runs die with `StorageFull` panics at the disk-cache tempdir creation in `benches/utils/tiers.rs`, with several 20–50 GB `.tmp*` cache trees alive at once under the OS temp dir — plus orphaned bench RustFS daemons pinning them after each crash.

**Root cause:** the cold-tier guards (`SupertableColdGuard`, `SupertableVecColdGuard`, `SupertableSqlColdGuard`) and the metered cold-store open tuples declare the cache `TempDir` *before* the consumer. Fields drop in declaration order, so `remove_dir_all` runs while the consumer's disk cache is still flushing fetched ranges; the removal races new file creation, fails with "directory not empty", and `TempDir::Drop` swallows the error. Every fresh-cache cold iteration (23 queries × 5 iters × 2 phases) can leak a multi-GB tree.

**Fix:** reorder so the consumer drops first, and wrap the cache dir in `RetryingTempDir` whose `Drop` retries `remove_dir_all` until the tree is actually gone (bounded at 60 s, then warns) — absorbing writes still landing from background tasks that outlive the consumer handle. Harness-only; cleanup runs between iterations, outside every timed window.

## Bug 2 — engine: compaction's carried postings can come out unsorted (regression from #711)

**Symptom:** `main`'s Benchmark lanes for `supertable_sql` and `supertable_vector` panic during `optimize` at `src/superfile/fts/posting.rs:197` (bitset posting-block write indexes out of bounds). First failing run is #711's; runs before it are green.

**Root cause:** #711 made merges carry prebuilt postings instead of re-tokenizing (an unstored column has no Parquet text to re-tokenize). The multi-cell merge writes output rows in packed stable-id order, so its carry feeds each term's doc ids out of order and relies on the spilled accumulator's finish sorting triples by `(term, doc)`. That guarantee only held on the `n < 256` comparison-sort fallback: the radix path used by every real corpus is a stable counting sort **by term only**, preserving arrival order within a term. #711's merge-equality tests used corpora below the threshold, so they exercised the correct fallback and passed while the 1M lanes panicked. Downstream, `encode_block` sizes a bitset block from `doc_ids[count-1]` (assumed max), so an unsorted list under-allocates and the write walks out of bounds.

**Fix:** after the radix scatter, a linear pass walks each term run and sorts a run by doc id only if an inversion was observed — the always-ascending ingest shape pays a single O(n) scan and never triggers a sort, and the sorted-chunk contract `(lex_rank, doc_id)` now holds regardless of feed order (the external-merge heap key already used that pair across chunks).

## Validation

- New regression test `radix_sort_orders_docs_within_term_for_unsorted_feeds` drives the radix path (≥ 256 triples) with a descending per-term feed; fails before the fix, passes after.
- **1M-doc `supertable sql` bench (multi-cell vector index, `-C debug-assertions=on`): reproduced the panic pre-fix** — `posting list not sorted by doc_id` in `encode_and_emit_term`, backtrace through `build_from_multi_cell_sq8_ivf_readers_to` → `finish_to_spilled` → `merge_sorted_spill` — **and the identical invocation post-fix runs end-to-end clean** (optimize completes, exit 0, no panics).
- Lib/superfile/supertable test suites green locally; `make check` fmt (crate-granularity imports) fixed in the follow-up commit.

## Fix 3 — CI: the A/B bench lanes could not go green while main was broken

**Symptom:** the same-host A/B bench lanes fail in the "Same-host baseline (A/B main arm)" step whenever `main`'s bench crashes — which skips the PR arm entirely and reds the lane until main is fixed, even for the PR that fixes main (this one).

**Fix:** capture the baseline arm's exit status and downgrade its failure to a warning. The PR arm always runs; shapes without baseline metrics render `(new)`; staged JSONs from shapes that completed before the crash still diff; and the merge gate already passes explicitly ("no gate verdict produced") when baseline metrics are absent. A broken **PR** arm still fails its own step, so both-arms-broken stays red. Note: the bench lanes run on `pull_request_target`, which always takes the workflow files from the **base branch** — so this fix cannot apply to this PR's own lanes and they stay structurally red until it lands on `main` (every PR's baseline arm is broken main, and the paths filter deliberately re-benches workflow changes, so a CI-only PR is equally red). The three A/B lanes on this PR are red **only** in the main-arm baseline step; the PR arms and every other check are green.

### DIFF
--- a/benches/utils/tiers.rs
+++ b/benches/utils/tiers.rs
@@ -11,7 +11,10 @@
 //! `s3` reads `INFINO_REAL_S3_BUCKET`, `azure` reads
 //! `INFINO_REAL_AZURE_CONTAINER`, `gcs` reads `INFINO_REAL_GCS_BUCKET`.
 
+use std::fs;
 use std::sync::{Arc, OnceLock};
+use std::thread;
+use std::time::{Duration, Instant};
 
 use bytes::Bytes;
 use infino::{
@@ -54,6 +57,53 @@ const BENCH_COLD_FETCH_CHUNK_BYTES: u64 = 8 * MIB_BYTES;
 const MMAP_TIMER_DISABLED_SECS: u64 = 0;
 
 const SUPERFILE_BENCH_PREFIX: &str = "infino-superfile-bench";
+
+/// How long [`RetryingTempDir`] keeps retrying removal before giving up.
+const CACHE_DIR_REMOVE_MAX_WAIT: Duration = Duration::from_secs(60);
+/// Pause between [`RetryingTempDir`] removal attempts.
+const CACHE_DIR_REMOVE_RETRY_INTERVAL: Duration = Duration::from_millis(500);
+
+/// Cache tempdir whose removal survives in-flight background writes.
+///
+/// `TempDir`'s `Drop` swallows `remove_dir_all` errors. A cold-tier
+/// consumer's disk cache can still be flushing fetched ranges to disk when
+/// its guard drops, so a one-shot removal races the cache's file creation,
+/// fails with "directory not empty", and silently leaves a multi-gigabyte
+/// cache tree behind — repeated fresh-cache iterations then fill the disk
+/// until the run dies of `StorageFull`. Guards must declare this field
+/// *after* the consumer (fields drop in declaration order) so the cache is
+/// shut down first; the retry loop then absorbs any writes still landing
+/// from background tasks that outlive the consumer handle.
+pub struct RetryingTempDir(Option<TempDir>);
+
+impl From<TempDir> for RetryingTempDir {
+    fn from(dir: TempDir) -> Self {
+        Self(Some(dir))
+    }
+}
+
+impl Drop for RetryingTempDir {
+    fn drop(&mut self) {
+        let Some(dir) = self.0.take() else { return };
+        let path = dir.keep();
+        let deadline = Instant::now() + CACHE_DIR_REMOVE_MAX_WAIT;
+        loop {
+            let _ = fs::remove_dir_all(&path);
+            if !path.exists() {
+                return;
+            }
+            if Instant::now() >= deadline {
+                eprintln!(
+                    "[bench] WARNING: cache dir {} still present after {}s of removal retries",
+                    path.display(),
+                    CACHE_DIR_REMOVE_MAX_WAIT.as_secs()
+                );
+                return;
+            }
+            thread::sleep(CACHE_DIR_REMOVE_RETRY_INTERVAL);
+        }
+    }
+}
 
 /// Storage tier exercised by a search bench row.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
--- a/benches/utils/tiers.rs
+++ b/benches/utils/tiers.rs
@@ -90,13 +90,22 @@ impl Drop for RetryingTempDir {
         let path = dir.keep();
         let deadline = Instant::now() + CACHE_DIR_REMOVE_MAX_WAIT;
         loop {
-            let _ = fs::remove_dir_all(&path);
+            // Keep the last outcome for the give-up warning: a persistent
+            // error (e.g. permissions) reads very differently from an `Ok`
+            // whose tree reappeared under a still-running fill.
+            let last = fs::remove_dir_all(&path);
             if !path.exists() {
                 return;
             }
             if Instant::now() >= deadline {
+                let outcome = match &last {
+                    Ok(()) => "last removal returned Ok but the tree persisted \
+                               (still being recreated by a live fill?)"
+                        .to_string(),
+                    Err(e) => format!("last removal error: {e}"),
+                };
                 eprintln!(
-                    "[bench] WARNING: cache dir {} still present after {}s of removal retries",
+                    "[bench] WARNING: cache dir {} still present after {}s of removal retries; {outcome}",
                     path.display(),
                     CACHE_DIR_REMOVE_MAX_WAIT.as_secs()
                 );
--- a/benches/utils/tiers.rs
+++ b/benches/utils/tiers.rs
@@ -771,7 +771,7 @@ pub fn open_superfile_cold_reader(
     storage: Arc<dyn StorageProvider>,
     uri: &SuperfileUri,
     known_size: u64,
-) -> (TempDir, Arc<SuperfileReader>) {
+) -> (RetryingTempDir, Arc<SuperfileReader>) {
     let (cache_dir, cache) = fresh_superfile_cache(storage);
     let offsets = superfile_cold_size_hint(known_size);
     let reader = block_on(async move {
@@ -780,7 +780,9 @@ pub fn open_superfile_cold_reader(
             .await
             .expect("cold reader")
     });
-    (cache_dir, reader)
+    // The open above starts background section fills; hand callers a
+    // removal that survives writes landing after the reader drops.
+    (cache_dir.into(), reader)
 }
 
 fn fresh_disk_cache_with_mode(
--- a/benches/utils/tiers.rs
+++ b/benches/utils/tiers.rs
@@ -11,10 +11,12 @@
 //! `s3` reads `INFINO_REAL_S3_BUCKET`, `azure` reads
 //! `INFINO_REAL_AZURE_CONTAINER`, `gcs` reads `INFINO_REAL_GCS_BUCKET`.
 
-use std::fs;
-use std::sync::{Arc, OnceLock};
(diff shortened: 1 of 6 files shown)


vinaykakade:
Validation follow-up: the 1M-doc `supertable sql` bench (multi-cell vector index, `-C debug-assertions=on`) now completes end-to-end — the `optimize` pass that previously panicked (`posting list not sorted by doc_id` in `encode_and_emit_term`, via `build_from_multi_cell_sq8_ivf_readers_to`) runs clean, and the full run exits 0 with no panics. Same invocation, same corpus scale as the failing `main` Benchmark lane.

vinaykakade:
Note on the red bench lanes: the failing step in `bench (supertable_vector, 1M, l2sq)` is **"Same-host baseline (A/B main arm)"** — the lane benches this PR's arm and then current `main` on the same host, and it is the `main` arm that panics (`src/superfile/fts/posting.rs:197`, the exact bug this PR fixes; `main` doesn't have the fix). This PR's own arm completed cleanly on the same host. The `supertable_sql` and other `supertable_vector` lanes will hit the same baseline-arm failure — the A/B lanes cannot go green until the fix lands on `main`. `supertable_fts` and `superfile` lanes don't exercise the multi-cell compaction path and should pass both arms.

src/superfile/fts/builder.rs
```diff
@@ -1069,6 +1069,35 @@ fn radix_sort_records_by_lex_rank<const N: usize>(triples: &mut Vec<[u32; N]>, l
         }
     }
 
+    // Pass 4: repair within-rank doc order. The counting scatter above is
```
ekechinwokah, line 1080: correct and the right layer: superfile/builder.rs:974 already promised (term, doc) from the spilled finish and only the <256 fallback honored it. the empty case is excluded by the threshold branch, the boundary test short-circuits before out[n], inversions are compared within a rank only, and one record per (term, doc) (the claim map) makes the unstable per-run sort safe for 3- and 5-lane records since positions ride by blob offset.

two asks. (a) this is a third read of every triple on the finish hot loop and the header at 997-1000 still sells two reads and one write; post `superfile fts build` vs main, and if it's outside noise gate the pass on a per-column prebuilt-fed flag so tokenizing ingest pays nothing. (b) header point 2 (987-996) and the pass-3 comment (1056-1061) still say the within-rank order is free, and add_prebuilt_term_posting's doc (1598-1600) still says callers feed each term's docs ascending, which the multi-cell merge explicitly doesn't (superfile/builder.rs:1419-1425). update all three: ingest arrives ascending, carry paths may not, pass 4 restores it.

src/superfile/fts/builder.rs
```diff
@@ -3992,6 +4021,38 @@ mod tests {
...
 
+    /// The radix path (n >= `RADIX_SORT_MIN_TRIPLES`) must deliver
+    /// `(lex_rank, doc_id)` order even when a term's docs arrive out of
+    /// order — the compaction carry paths feed postings remapped through
+    /// a packed row order, unlike the always-ascending ingest path. A
+    /// term-only stable scatter once preserved the unsorted arrival
+    /// order, and `encode_block` then indexed a bitset block out of
+    /// bounds on the merged output.
+    #[test]
+    fn radix_sort_orders_docs_within_term_for_unsorted_feeds() {
```
ekechinwokah, line 4045: covers the sort, not the merge. multi_cell_merge_carries_fts_postings_through_reorder and the unstored variant (superfile/builder.rs ~2840) missed this because 10 rows never leave the comparison fallback. push one of them past RADIX_SORT_MIN_TRIPLES so the end-to-end contract is pinned where it broke, not only in the helper.

benches/utils/tiers.rs
```diff
@@ -55,6 +60,53 @@ const MMAP_TIMER_DISABLED_SECS: u64 = 0;
...
+/// `TempDir`'s `Drop` swallows `remove_dir_all` errors. A cold-tier
+/// consumer's disk cache can still be flushing fetched ranges to disk when
+/// its guard drops, so a one-shot removal races the cache's file creation,
+/// fails with "directory not empty", and silently leaves a multi-gigabyte
+/// cache tree behind — repeated fresh-cache iterations then fill the disk
+/// until the run dies of `StorageFull`. Guards must declare this field
+/// *after* the consumer (fields drop in declaration order) so the cache is
+/// shut down first; the retry loop then absorbs any writes still landing
+/// from background tasks that outlive the consumer handle.
+pub struct RetryingTempDir(Option<TempDir>);
```
ekechinwokah, line 79: the superfile tier still has the bug: SuperfileColdGuard (benches/utils/superfile.rs:807), SuperfileVecColdGuard (:1721), SqlColdGuard (:2432) and open_cold_consumer (:2384) all put the cache dir before the reader/table and open with the fts/sql fill flag on, so background fills race removal the same way. smaller caches, same leak shape. convert those too or say why that tier is exempt. supertable.rs:1739 (fts_quality) drops in the right order but with a one-shot TempDir; same wrapper for consistency.

benches/utils/tiers.rs
```diff
@@ -55,6 +60,53 @@ const MMAP_TIMER_DISABLED_SECS: u64 = 0;
...
+    }
+}
+
+impl Drop for RetryingTempDir {
+    fn drop(&mut self) {
+        let Some(dir) = self.0.take() else { return };
+        let path = dir.keep();
+        let deadline = Instant::now() + CACHE_DIR_REMOVE_MAX_WAIT;
+        loop {
+            let _ = fs::remove_dir_all(&path);
```
ekechinwokah: keep the last remove_dir_all error and print it in the give-up warning below; it's the only diagnostic when the 60s bound trips, and a permissions failure looks identical to a slow fill otherwise.

ekechinwokah (commented):
both fixes are correct and minimal; the fts one sits at the right layer (superfile/builder.rs:974 and :1419 already promise (term, doc) from the spilled finish, only the <256-triple fallback kept it) and the bench drop order is right at all six sites, cold_store holds the tuple whole so consumer goes first.

before merge:
- post `superfile fts build` (and supertable ingest) vs main in the body. pass 4 is a third read of every triple on the finish hot loop; the repo rule is a bench comparison in every pr body.
- the superfile-tier cold guards still have the cache dir declared first (inline comment on tiers.rs:79).
- the two doc contracts the patch now contradicts (inline on builder.rs:1072).

nice to have: a merge-level regression test that crosses RADIX_SORT_MIN_TRIPLES (inline on builder.rs:4032); keep the last remove error in the give-up warning (tiers.rs:93).

merges clean onto current main (2719f762); subjects follow the layer convention; no trailers.

src/superfile/fts/builder.rs
vinaykakade: (a) here are the build time changes in branch vs main - they are within noise. So, IMO, per-column prebuilt-fed flag isn't needed.

| Build (1M docs)      | main run 1 | branch run 1 | main run 2 | branch run 2 | mean main | mean branch | Δ mean |
|----------------------|-----------:|-------------:|-----------:|-------------:|----------:|------------:|-------:|
| 1 writer             | 9.33 s     | 9.42 s       | 9.68 s     | 9.35 s       | 9.51 s    | 9.39 s      | −1.3%  |
| 16 writers           | 1.44 s     | 1.35 s       | 1.56 s     | 1.41 s       | 1.50 s    | 1.38 s      | −8.0%  |

(b) thanks - updated the comments

benches/utils/tiers.rs
vinaykakade: Done - in commit https://github.com/infino-ai/infino/pull/722/changes/9485fbe7bcba05209c0fb7efac9482dd837d45b3

benches/utils/tiers.rs
vinaykakade: Good call — done in ee3292ca.

src/superfile/fts/builder.rs
vinaykakade: Done - added in https://github.com/infino-ai/infino/pull/722/changes/2551e52bdf423d446edfbf6170760acb11db1c42

ekechinwokah (approved):

PR #724, opened 2026-09-04
fts: global BM25 statistics by default

vinaykakade:
## Problem

BM25 idf needs each term's document frequency across the whole table. With per-superfile statistics — the current default — each superfile computes idf from its own df, so a fragmented table ranks with as many score scales as it has superfiles and merges them into one top-k. A document's rank then depends on which superfile its commit landed in. On the 10M-doc realistic-corpus quality bench (textbook-BM25 oracle, 18 query shapes × k ∈ {10, 100, 1000}), per-superfile stats collapse on any shape whose df varies across superfiles:

| shape (k=10) | recall, global stats | recall, per-superfile stats |
|---|---|---|
| stopword | 0.90 | 0.00 |
| common | 1.00 | 0.10 |
| three_stopword_or | 0.80 | 0.10 |
| mid | 1.00 | 0.50 |
| rare | 1.00 | 0.60 |

## Solution

`Bm25Stats` defaults to `Global`: every superfile scores with one corpus-wide idf, so scores are comparable by construction. Average document length stays per-superfile — it is baked into each superfile's norm tables, and it is not what breaks comparability; idf is.

Global idf is never stored. Each query computes it from the manifest snapshot it reads: N from the manifest, df per term through three tiers, cheapest first:

- **Idf cache.** The table handle caches idf per (column, term), keyed by manifest generation. A commit changes the generation, so stale entries simply stop matching. No invalidation logic.
- **Term-stats sidecar.** `optimize` writes one small content-addressed artifact per table (`term-stats/stats-<hash>.bin`, an FST of `(column, term) → df` summed over a recorded superfile set; ~4 MiB at 10M docs). A query reads corpus df in one lookup instead of asking every superfile. Appends keep it valid — new superfiles are a small tail the query tops up from their own dictionaries. Removing superfiles (compaction) drops the reference; the next `optimize` rebuilds it. Versioned, GC-integrated, and a failed load falls back to reading dictionaries.
- **Fused gather.** When df must come from superfile dictionaries (no sidecar yet, or the tail), the reads fold into the reads the query already does: each superfile kept for scoring looks up its terms and fetches their postings once, reports df, and reuses those same bytes for scoring. Pruned-but-present superfiles answer with a dictionary lookup only. A global-stats query issues the same reads as a per-superfile query.

Scoring keeps all of its pruning: stored block-max bounds bake in local idf, the score is linear in idf, so multiplying every bound by `global_idf / local_idf` keeps them exact — block skipping, phrase bounds, and the single-term fast path prune as tightly as before. Since scores are now on one scale, superfiles also share their running kth-best live during the search (published only from superfiles with no tombstoned rows, since kernel heaps precede tombstone filtering), so each walk skips work the others have ruled out.

Compatibility:
- Query-time change only; nothing on disk changes for existing tables. Old manifests read byte-identically; the sidecar appears after the next `optimize` (older readers ignore the field).
- Match sets are unchanged — only result *ordering* changes, and only on multi-superfile tables with df skew. `Bm25Stats::PerSuperfile` stays available and is bit-identical to before.
- Three surfaces that had `PerSuperfile` hardwired now follow the default: the SQL `bm25_search` table function, the BM25 leg of `hybrid_search`, and the internal `bm25_hits` kernel. Bindings treat an omitted `stats` as "engine default"; explicit values behave as before.
- Deleted docs keep counting toward df until compaction rewrites dictionaries — the same semantics as before, and as Lucene.

## Quality improvement

10M-doc realistic corpus, 18 shapes × 3 k, graded against the streaming textbook-BM25 oracle:

| | global (this PR's default) | per-superfile (old default) |
|---|---|---|
| mean recall @10 / @100 / @1000 | **0.90 / 0.96 / 0.97** | 0.62 / 0.76 / 0.82 |
| worst shape | **0.80** | 0.00 |
| cells won or tied (of 54) | **52** | — |

A single-superfile or fully compacted table produces identical results under both modes.

## Performance (no regression)

The authority is the CI A/B: both arms build on one VM and run the full battery at 1M docs on the realistic corpus. **Merge gate: PASS — 0 blocking, 0 informational regressions** at the gate's >50%-and->5ms bar. Warm p90, main vs this branch, representative rows:

| Query | main | branch | Δ |
|---|---:|---:|---:|
| single_rare | 310 µs | 195 µs | −37% |
| single_common | 173 µs | 193 µs | +12% |
| ten_term_or | 3.63 ms | 3.23 ms | −11% |
| forty_term_or | 12.30 ms | 12.08 ms | −2% |
| ten_term_and | 1.73 ms | 2.41 ms | +39% |
| must_common_should_common | 677 µs | 698 µs | +3% |
| phrase_two_common | 40.69 ms | 34.07 ms | −16% |
| phrase_two_mixed | 852 µs | 754 µs | −12% |
| phrase_three_common | 38.33 ms | 42.02 ms | +10% |
| phrase_plus_must_term | 43.23 ms | 43.86 ms | +1% |

Read those as a band, not as precision: the same shape swings by tens of percent between runs of identical code on this VM class, which is why the gate requires a regression to clear both 50% and 5 ms before it blocks, and why nothing here approaches it (`ten_term_and`, the widest mover, is a 0.7 ms absolute change). The steadier signal is the aggregate CPU counters, which average many shapes per row: OR search 2.58M → 2.53M cycles, AND search 6.49M → 6.28M, phrase search 227.0k → 227.8k.

Cold adds one fixed cost, and it is the honest price of corpus-wide statistics: a fresh cache fetches the term-statistics artifact once, which the CI cold leg measures as **+3.6 MiB and +1–2 GETs, near-constant across the battery** (for example `three_similar_or` 8.26 → 11.88 MiB, `phrase_three_common` 46.18 → 50.31 MiB). It is paid per cache, not per query, so it amortizes over every query that cache serves — and it replaces a document-frequency gather that would otherwise probe all 75 superfiles' dictionaries. Per-superfile statistics pay neither, because they never compute a corpus-wide figure at all; that is the trade this change makes. Cold latency is reported per shape but not gated, and at one fresh-cache iteration per shape it is a single sample: the CI cold leg is green with no blocking movers.

Warm is where a served workload lives, and there the artifact is already resident: the warm battery measured −3.6% at 10M docs on a maintained table, with the read plan for a covered query identical to the per-superfile plan. A table whose maintenance has not run yet pays a one-time dictionary gather per (term, generation) instead, which the idf cache then retires.

The maintenance itself is a stats-only pass inside `optimize` (dictionary walks and batched header probes — no posting bodies): ~2 minutes at 1M realistic docs on the bench VM, producing the ~4 MiB artifact.

## Testing

- **Oracles**: the zipfian brute-force test has two arms — per-shard oracles vs explicit `PerSuperfile` (unchanged semantics) and a whole-corpus oracle vs `Global` at a 90% recall floor. The 10M quality harness (18 shapes × 3 k) is within gates on both CI arms.
- **Sidecar lifecycle**: on a covered fragmented table, global OR and AND queries plan exactly the per-superfile read ranges (first query included); ranking equals a never-optimized control through publish → tail appends → republish → merging optimize; round-trip and malformed-artifact decode tests.
- **Plan accounting**: op-stats tests pin that the gather fans once per manifest generation and that global plans equal per-superfile plans on covered tables.
- **Live floor soundness**: a regression test deletes every high-scoring phrase doc in one superfile and asserts the survivors still fill the top-k — it fails if a tombstoned superfile is ever allowed to publish into the shared floor.
- **Maintenance cache hygiene**: an intermittent vector-lane failure traced (via per-request read traces on parallel CI dispatches) to the stats pass opening superfiles with background cache fills enabled — whole-table fills outlived `optimize` on real object storage and bled into the next measurement. Fixed (the pass opens readers without fills) and pinned by a regression test that fails if a stats-only optimize ever warms the disk cache like a query would.
- Full suites green (lib 2411, supertable 340+, superfile 190; clippy, fmt); CI: both fts legs, sql, superfile, and vector bench lanes green.

### DIFF
--- a/src/superfile/fts/reader/search.rs
+++ b/src/superfile/fts/reader/search.rs
@@ -7,8 +7,9 @@
 //! BlockMaxWAND fast path, and the atom (phrase-aware) scored walk.
 //! Its own `impl FtsReader` block, split from the reader `core`.
 
-use std::collections::BinaryHeap;
+use std::collections::{BinaryHeap, HashMap};
 
+use bytes::Bytes;
 use rustc_hash::FxHashMap;
 
 use super::{
@@ -38,6 +39,57 @@ use crate::{
         },
     },
 };
+
+/// One scored term's open-wave fetch result: the dictionary resolution
+/// plus, for a PFOR term, its full postings range and header df.
+/// Cloning is cheap — `Bytes` is a refcounted slice.
+#[derive(Clone)]
+pub(crate) enum FetchedTermSlot {
+    /// df=1 inline dictionary slot (no postings-region bytes exist).
+    Inline { doc_id: u32, tf: u32 },
+    /// PFOR term: the fetched metadata+skip+blocks range and its df.
+    Pfor {
+        bytes: Bytes,
+        /// The dictionary slot carried no length hint, so the fetch paid
+        /// a header probe before the body — two planned ranges, which
+        /// the cursor built from these bytes must keep reporting.
+        header_probed: bool,
+        df: u64,
+    },
+}
+
+/// One superfile's open-wave fetches for a global-stats query's scored
+/// terms, keyed by term. A key that is present with a `None` value is a
+/// *resolved miss* (the term is absent from this superfile) — the walk
+/// wave must not re-probe the dictionary for it. Built by
+/// [`FtsReader::fetch_scored_terms`]; consumed by the cursor builds via
+/// the `prefetched` parameter.
+pub(crate) struct FetchedTermMemo {
+    map: HashMap<Box<str>, Option<FetchedTermSlot>>,
+}
+
+impl FetchedTermMemo {
+    /// `Some(entry)` when the open wave resolved this term (entry `None`
+    /// = absent from this superfile); `None` when the term was not part
+    /// of the fetch (e.g. a negated term — unscored, never gathered).
+    pub(crate) fn lookup(&self, term: &str) -> Option<Option<FetchedTermSlot>> {
+        self.map.get(term).cloned()
+    }
+
+    fn contains(&self, term: &str) -> bool {
+        self.map.contains_key(term)
+    }
+
+    /// This superfile's df contribution per term (0 for a miss or an
+    /// un-fetched term, 1 for an inline slot).
+    pub(crate) fn df(&self, term: &str) -> u64 {
+        match self.map.get(term) {
+            Some(Some(FetchedTermSlot::Inline { .. })) => 1,
+            Some(Some(FetchedTermSlot::Pfor { df, .. })) => *df,
+            _ => 0,
+        }
+    }
+}
 
 /// Threshold-first seed tuning for the single-term ranked walk. The walk
 /// fills its top-k heap in doc-id order, so its skip threshold rises only
@@ -415,7 +467,13 @@ impl FtsReader {
         if lists.has_phrases() {
             // Phrase-bearing query: the heterogeneous atom walks.
             let (must_atoms, must_dict) = self
-                .build_atom_cursors(column_id, lists.musts, lists.must_phrases, lists.global_idf)
+                .build_atom_cursors(
+                    column_id,
+                    lists.musts,
+                    lists.must_phrases,
+                    lists.global_idf,
+                    lists.prefetched,
+                )
                 .await?;
             if must_atoms.iter().any(Option::is_none) {
                 // A must atom can never match in this superfile. The
@@ -435,13 +493,20 @@ impl FtsReader {
                     lists.shoulds,
                     lists.should_phrases,
                     lists.global_idf,
+                    lists.prefetched,
                 )
                 .await?;
             let should_atoms: Vec<AnyCursor> = should_built.into_iter().flatten().collect();
             // Negatives are a hard exclusion filter, not scored, so their
             // idf is irrelevant — always build them local.
             let (negative_built, negative_dict) = self
-                .build_atom_cursors(column_id, lists.negatives, lists.negative_phrases, None)
+                .build_atom_cursors(
+                    column_id,
+                    lists.negatives,
+                    lists.negative_phrases,
+                    None,
+                    None,
+                )
                 .await?;
             let negative_atoms: Vec<AnyCursor> = negative_built.into_iter().flatten().collect();
             let postings_bytes = atom_cursor_bytes(&must_atoms)
@@ -477,7 +542,7 @@ impl FtsReader {
             // Negatives are a hard exclusion filter, not scored, so their
             // idf is irrelevant — always build them with local stats.
             _ => Some(ExcludeFilter::new(
-                self.build_term_cursors(column_id, lists.negatives, None, false, None)
+                self.build_term_cursors(column_id, lists.negatives, None, false, None, None)
                     .await?,
             )),
         };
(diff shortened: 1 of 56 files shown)


src/supertable/manifest/term_stats.rs
```diff
@@ -0,0 +1,259 @@
...
+    superfile::{
+        SuperfileReader,
+        fts::dict::{DictBuilder, make_key},
+    },
+    supertable::manifest::{RoutingRef, part::ContentHash},
+};
+
+/// Object-store directory prefix for term-stats artifacts, sibling to
+/// the superfile data and slow-vector-state prefixes.
+pub(crate) const STORAGE_PREFIX: &str = "term-stats/";
```
naman-modi, line 46: Can we please update the [on-disk-layout](https://github.com/infino-ai/infino/blob/main/docs/architecture/on-disk-layout.md) to match this change?

src/supertable/query/fts.rs
```diff
@@ -697,77 +822,166 @@ impl SupertableReader {
...
+        // uncovered tail (recent commits) — and vanishes entirely on a
+        // table whose maintenance is current, restoring the single fully
+        // overlapped dispatch of the per-superfile plan. A load failure
+        // degrades to the full query-time wave.
+        let sidecar = self.term_stats_sidecar().await;
+        let covered: HashSet<Uuid> = sidecar
+            .as_ref()
+            .map(|s| s.covered().iter().copied().collect())
+            .unwrap_or_default();
+        debug_assert!(
```
naman-modi: This debug_assert will not work in the release build. Was this intentional? I believe this check should even run in the release builds, else the df would go wrong.

src/superfile/fts/reader/search.rs
```diff
@@ -906,8 +1068,19 @@ impl FtsReader {
...
-        let idf_t = bm25::idf(self.n_docs as u64, term_meta.df);
+        let local_idf = bm25::idf(self.n_docs as u64, term_meta.df);
+        let idf_t = global_idf.unwrap_or(local_idf);
         let idf_x_k1p1 = idf_t * (bm25::K1 + 1.0);
+        // Stored block-max and coarse entries bake in the LOCAL idf. Scores
+        // below use the (possibly overridden) effective idf, and the score
+        // is linear in idf, so rescaling every stored bound by the ratio
+        // keeps bounds exactly as tight as the per-superfile path — the
+        // same exact rescale `TermCursor::new` applies (see cursor.rs).
+        let bound_rescale = if local_idf > 0.0 && idf_t != local_idf {
```
naman-modi, line 1079: Can we add a test for single-term version: query "common", small k, fragmented table, assert top-k by value == single-superfile? To test this modified skip path.

tests/supertable/query/term_stats.rs
```diff
@@ -0,0 +1,397 @@
+// SPDX-License-Identifier: Apache-2.0
```
naman-modi, line 1: This majorly covers append + merging-optimize. What do you think about adding some tests around:

- delete on a covered table (delete some rows, global query == control that deleted the same rows) -> df stays gross until compaction
- update on a covered table (update some rows, == control) -> old row stays covered+gross, new row is tail, so it's counted in both. locks that's intended and matches the gather

naman-modi (approved):
LGTM, after some comments.

tests/supertable/query/term_stats.rs
vinaykakade: Good point, done

src/supertable/manifest/term_stats.rs
vinaykakade: Done

src/superfile/fts/reader/search.rs
vinaykakade: Done

src/supertable/query/fts.rs
vinaykakade: Done

PR #729, opened 2026-09-07
catalog: sql cap logical operator count and run SQL planning on infino's threads

naman-modi:
### What does this PR do?
A boolean predicate with thousands of `AND` / `OR` operators used to crash the process during planning. It is now refused with a query error past 1024, and planning runs on infino's own threads instead of the caller's.

DataFusion's planner recurses once per `AND` / `OR`, and a stack overflow aborts without unwinding, so one oversized query kills the embedder instead of returning an error. Infino cannot control how deep a query can go on the calling thredas stack.

### How do we do this change?
- Queries and `update` / `delete` predicates are refused past 1024 `AND` / `OR` operators. The check costs nothing on normal queries.
- Planning is spawned onto the query runtime's workers, now with 16 MiB stacks: 2 MiB dies near 1000 operators, 16 MiB near 8000, so the cap has ~8x headroom.

### Performance
- Warm SQL battery vs main at 1M docs: -3.1% total, every shape within noise; the fastest planning-bound queries (~125 us) pay ~1 us for the thread hop.
- All 43 benches queries pass the guard on the length compare alone; the 16 MiB stacks are virtual reservations.

### Notes
- A Rust embedder calling delete/update from inside its own tokio runtime still resolves on its own workers (an existing block_on preference); the shipped surfaces are unaffected.

### DIFF
--- a/src/catalog/mod.rs
+++ b/src/catalog/mod.rs
@@ -34,7 +34,19 @@ use std::{
 use arrow::record_batch::RecordBatch;
 use arrow_schema::SchemaRef;
 use dashmap::DashMap;
-use datafusion::{config::Dialect, error::DataFusionError, execution::context::SQLOptions};
+use datafusion::{
+    common::tree_node::{TreeNode, TreeNodeRecursion},
+    config::Dialect,
+    error::DataFusionError,
+    execution::context::SQLOptions,
+    logical_expr::{BinaryExpr, Operator},
+    prelude::Expr,
+    sql::sqlparser::{
+        dialect::GenericDialect,
+        keywords::Keyword,
+        tokenizer::{Token, Tokenizer as SqlTokenizer},
+    },
+};
 use futures::future::try_join_all;
 pub use index_spec::IndexSpec;
 use manifest::{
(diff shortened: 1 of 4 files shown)


src/catalog/mod.rs
```diff
@@ -967,16 +995,120 @@ fn read_only_sql_options() -> SQLOptions {
...
+    Ok(())
+}
+
+/// Exact `AND` / `OR` keyword count over the token stream; `None` when the text does not tokenize.
+fn connective_count(sql: &str) -> Option<usize> {
+    let tokens = SqlTokenizer::new(&GenericDialect {}, sql).tokenize().ok()?;
+    Some(
+        tokens
+            .iter()
+            .filter(|token| {
```
Pranav2612000, line 1059: Will this also match AND and OR used in other places in the query? ( E.g CASE statements )

src/catalog/mod.rs
```diff
@@ -967,16 +995,120 @@ fn read_only_sql_options() -> SQLOptions {
...
+/// Mutation-predicate side of [`ensure_sql_within_connective_cap`]: count `AND` / `OR` nodes in
+/// a built `Expr`. The stack-protected `TreeNode` walk is safe on the input it refuses; stops at
+/// the first node past the cap.
+pub(crate) fn ensure_expr_within_connective_cap(predicate: &Expr) -> Result<(), InfinoError> {
+    let mut connectives = 0usize;
+    predicate
+        .apply(|expr| {
+            if let Expr::BinaryExpr(BinaryExpr {
+                op: Operator::And | Operator::Or,
+                ..
```
Pranav2612000, line 1075: Does this include AND and OR used in other places in the SQL query?

Pranav2612000 (approved):

src/catalog/mod.rs
naman-modi: Yes.

src/catalog/mod.rs
naman-modi: Yes.

naman-modi:
SQL bench looks good, verified no regression as tested locally, so merging.

PR #735, opened 2026-09-08
fts: analyzer explicit everywhere, and default it to standard

vinaykakade:
## What

The default FTS analyzer moves from `ascii_lower` to `standard` (Unicode
UAX #29), and every place that *inferred* an analyzer is deleted first, so
no line of code has to remember which name used to be the default.

`ascii_lower` silently drops every non-ASCII token — `café`, `naïve`, all
CJK — so the old default was a recall hole rather than a preference. It
stays a fully supported analyzer; it just has to be asked for by name.

## Making the analyzer explicit

Each of these sites used to mean "the default", which is exactly what was
about to change under them:

| Site | Before | Now |
|---|---|---|
| Superfile footer `inf.fts.columns[].tokenizer` | `serde(default)` → `ascii_lower` | required — a missing key is a malformed file, not an old one |
| Table record's analyzer list | short list → `ascii_lower` per column | must name one analyzer per full-text column; `open_table` rejects and names the table |
| Options hash | analyzer block emitted only for a non-`ascii_lower` column | always emitted for a table with full-text columns |
| Remote transport | bare column name when `ascii_lower` + stored | always `{column, analyzer}` |
| `fts_tokenizer_for` | `unwrap_or(AsciiLowerTokenizer)` | deleted; callers use `try_fts_tokenizer_for` |

On the last one: a fallback analyzer could only ever prune with tokens the
index never wrote, so callers now treat a column with no full-text index as
unbounded for pruning, and `InvalidQuery` for a search. `bm25_hits_async`
and `exact_match` on the superfile reader fail where the reason is still
nameable rather than after a pass with some other column's analyzer.

## The options-hash bridge

Always emitting the analyzer block changes the digest of every existing
all-`ascii_lower` table. A hash stamped under the earlier rule is still
accepted for one release, and `build_list` re-stamps the current rule on the
next commit, so nothing has to be rewritten by hand. The acceptance is
labelled in-place with the condition for its removal. The drain
checkpoint's identity check goes through the same acceptance, so a drain
spanning an upgrade resumes instead of wedging on a re-encoded digest.

## Existing tables

Unaffected. They open, query, append and compact with the analyzer they
recorded — `BuilderOptions::new_from_reader` recovers it from the source
superfile, so compaction and `optimize` carry it forward whatever the
engine default is, and `check_fts_carry_compat` still refuses to merge two
analyzers. Nothing rewrites what a table recorded about its analyzer.

## Bindings

No signature change: Python's `analyzer=None` and Node's omitted `analyzer`
already meant "engine default", so both inherit the new one. Docs updated;
the fixtures that specifically exercise ASCII-only tokenization now name
`ascii_lower` explicitly instead of leaning on the default.

## Version

Minor bump to 0.7.0 across crate, Node and Python (`scripts/release_prep.py
--package all`), since a bare column declaration now tokenizes differently.

## Tests

- Footer without `tokenizer` is rejected, and the error names the field.
- `open_table` on a record whose analyzer list is short errors, naming the
  table and what is missing.
- The analyzer block is part of an all-`ascii_lower` table's identity; the
  earlier block-less digest still verifies, and is exact rather than a
  blanket pass (a table with a different analyzer still mismatches). The
  golden that pinned the pre-flag byte stream now pins it through the
  superseded encoding, which is what it was always describing.
- The transport names every column's analyzer, whichever it resolved to.
- `bm25_hits_async` / `exact_match` on a column with no full-text index
  error and name the column.
- End to end: a column declared without an analyzer keeps non-ASCII text,
  and diverges from an explicit `ascii_lower` on the same corpus.

The `remote`-feature lanes are the ones to watch here: the wire tests
compile only under `--features remote`, so `make test-remote` (not a plain
`cargo test`) is what grades the transport change. `make version-sync` and
`make public-api` green.

## Follow-up

The one-release hash acceptance comes out once no stored digest predates
the current rule.

### DIFF
--- a/src/catalog/index_spec.rs
+++ b/src/catalog/index_spec.rs
@@ -8,7 +8,7 @@
 
 use crate::superfile::{
     builder::FtsConfig,
-    fts::tokenize::ASCII_LOWER_TOKENIZER,
+    fts::tokenize::STANDARD_TOKENIZER,
     vector::{builder::VectorConfig, distance::Metric},
 };
 
@@ -46,22 +46,23 @@ pub struct FtsField {
 
 impl FtsField {
     /// Declare `column` as full-text indexed with the defaults: the
-    /// `ascii_lower` analyzer (ASCII split + lowercase, non-ASCII
-    /// dropped) and the raw text stored. The column must be a UTF-8
-    /// string column in the table schema.
+    /// `standard` analyzer (Unicode UAX #29 word segmentation, full
+    /// Unicode lowercasing) and the raw text stored. The column must be
+    /// a UTF-8 string column in the table schema.
     pub fn new(column: impl Into<String>) -> Self {
         Self {
             column: column.into(),
-            analyzer: ASCII_LOWER_TOKENIZER.to_string(),
+            analyzer: STANDARD_TOKENIZER.to_string(),
             stored: true,
         }
     }
 
-    /// Pick the column's analyzer by name (`"ascii_lower"` or
-    /// `"standard"` — the Unicode-aware UAX #29 tokenizer that keeps
-    /// non-ASCII text). The analyzer is per column: each FTS column is
-    /// tokenized with its own, so columns in one table may use
-    /// different analyzers.
+    /// Pick the column's analyzer by name — `"standard"` (the default)
+    /// or `"ascii_lower"`, which splits on ASCII alphanumerics and
+    /// drops every non-ASCII token. The analyzer is per column: each
+    /// FTS column is tokenized with its own, so columns in one table
+    /// may use different analyzers. It is recorded with the table and
+    /// cannot be changed afterwards.
     pub fn analyzer(mut self, name: impl Into<String>) -> Self {
         self.analyzer = name.into();
         self
--- a/src/catalog/index_spec.rs
+++ b/src/catalog/index_spec.rs
@@ -127,7 +127,7 @@ impl IndexSpec {
     /// use infino::{FtsField, IndexSpec};
     /// let spec = IndexSpec::new()
     ///     .fts("title")
-    ///     .fts(FtsField::new("body").analyzer("standard").stored(false));
+    ///     .fts(FtsField::new("body").analyzer("ascii_lower").stored(false));
     /// # let _ = spec;
     /// ```
     pub fn fts(mut self, field: impl Into<FtsField>) -> Self {
--- a/src/catalog/index_spec.rs
+++ b/src/catalog/index_spec.rs
@@ -27,14 +27,14 @@ struct VectorIndex {
 /// One full-text (BM25) indexed column, with its per-column options.
 ///
 /// Passed to [`IndexSpec::fts`]. A plain column name converts with all
-/// defaults (`ascii_lower` analyzer, stored text), so the common case
+/// defaults (`standard` analyzer, stored text), so the common case
 /// stays `.fts("body")`; build a `FtsField` to change an option:
 ///
 /// ```
 /// use infino::{FtsField, IndexSpec};
 /// let spec = IndexSpec::new()
 ///     .fts("title")
-///     .fts(FtsField::new("body").analyzer("standard").stored(false));
+///     .fts(FtsField::new("body").analyzer("ascii_lower").stored(false));
 /// # let _ = spec;
 /// ```
 #[derive(Debug, Clone)]
--- a/src/superfile/builder.rs
+++ b/src/superfile/builder.rs
@@ -63,8 +63,8 @@
 //! `register_column_with_tokenizer` sets a per-column analyzer — and
 //! dispatches per (column, doc) at `add_doc` time.
 //!
-//! Two tokenizers ship: `AsciiLowerTokenizer` (the default) and the
-//! Unicode-aware `StandardTokenizer`, selectable per column. The
+//! Two tokenizers ship: the Unicode-aware `StandardTokenizer` (the
+//! default) and `AsciiLowerTokenizer`, selectable per column. The
 //! `inf.fts.columns` JSON persists each column's tokenizer name, so a
 //! column is re-tokenized at rebuild / compaction with the analyzer it
 //! was indexed with. Further analyzers (language-specific stemmers, …)
@@ -130,7 +130,7 @@ use crate::superfile::{
 pub struct FtsConfig {
     pub column: String,
     /// Analyzer (tokenizer) name applied to this column —
-    /// `"ascii_lower"` (the default) or `"standard"`. Resolved to a
+    /// `"standard"` (the default) or `"ascii_lower"`. Resolved to a
     /// tokenizer instance once, at builder construction; an unknown
     /// name is a build error. Per column: each FTS column is tokenized
     /// with its own analyzer, so columns in one table may differ.
--- a/src/superfile/builder.rs
+++ b/src/superfile/builder.rs
@@ -98,7 +98,7 @@ use crate::superfile::{
     fts::{
         builder::FtsBuilder,
         reader::ColumnMeta,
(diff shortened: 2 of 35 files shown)


vinaykakade:
`Format, Build & Test` was red on the first push. Real failure, now fixed
in 336aab6c.

The three `index_spec_to_json` tests in `catalog/remote/wire.rs` still
expected `ascii_lower` in the emitted entries. They compile only under the
`remote` feature — `make ci`'s `test-remote` lane runs
`cargo test --features remote --lib catalog::remote::wire` — so a default
`cargo test` never built them, and the stale expectations survived the
default flip.

`index_spec_json_names_every_columns_analyzer` now names both analyzers
explicitly rather than leaning on the default for one of them, so it keeps
proving that each column carries its own analyzer instead of passing by
coincidence whenever the default happens to match. It also asserts
separately that a column declared without an analyzer crosses the wire
with the default *named*, not as a bare string the server would resolve
itself — which is the property this PR is actually about.

Correcting the description above: "Full `cargo test` (lib + integration)
green" was written from a run that did not include the `remote`-feature
lanes, so it overstated the evidence. The gate to trust is this job.

src/catalog/index_spec.rs
pratyushlokhande, line 38: nit: doctest needs to be updated here. these still mention `ascii_lower` as default

src/superfile/builder.rs
pratyushlokhande: nit: points to `AsciiLowerTokenizer` as default, need to be updated

pratyushlokhande (approved):
Few corrections on doctest and comments. LGTM.

vinaykakade:
## Ingest cost of the new default — measured, same box, same run

Baseline (main arm) vs this branch, `superfile fts` build cells, 1M docs
Zipfian / 200 tokens per doc / 10K vocab, Intel Xeon 6973P-C 4C/7T:

| Build cell | main | this branch | Δ time | throughput |
|---|---|---|---|---|
| 1 writer | 14.89 s | 19.48 s | **+30.8%** | 67.2 → 51.3 K/s |
| 7 writers | 3.74 s | 4.39 s | **+17.4%** | 267.1 → 227.9 K/s |
| 1 writer (positions) | 27.34 s | 31.67 s | **+15.8%** | 36.6 → 31.6 K/s |
| 7 writers (positions) | 5.48 s | 6.31 s | **+15.1%** | 182.3 → 158.5 K/s |

Stored size is **byte-identical** on both arms (618.43 MiB positionless,
884.72 MiB positional). This corpus is lowercase ASCII, so `standard`
segments it into exactly the same tokens as `ascii_lower` — the whole
delta is tokenizer CPU, not extra index data, and it is ingest-only:
no read cell regressed (`0 blocking` on the warm p90 gate, and the FTS
warm/cold supertable lanes both pass).

Cause is structural, not incidental. `AsciiLowerTokenizer` scans with
`simd_skip_non_token` / `simd_scan_token_run` (16 bytes per step, upper-case
and non-ASCII detection folded into the same pass). `StandardTokenizer`
goes through `text.unicode_words()` — per-character UAX #29 property
lookups — then re-walks each word with `word.chars().all(...)` to decide
whether it can borrow instead of case-folding. Making `standard` the
default therefore moves every ASCII-only corpus off the SIMD path onto
the general segmenter.

**Follow-up worth doing:** an ASCII fast path inside `StandardTokenizer`.
Within ASCII, the UAX #29 word-break rules collapse to a small fixed set
(`ALetter` `[A-Za-z]`, `Numeric` `[0-9]`, and the `MidLetter` /
`MidNumLet` / `MidNum` / `ExtendNumLet` punctuation classes), so an
all-ASCII chunk can be segmented with the same 16-byte SIMD scan and only
fall back to `unicode_words` when a non-ASCII byte appears. That is not a
one-liner and it must not be eyeballed: the boundary rules have to agree
with `unicode_words` exactly, or text gets indexed under one tokenization
and queried under another. The existing analyzer oracles plus a proptest
comparing the fast path against `unicode_words` over mixed ASCII input
are the way to gate it.

### Note on the red bench job

The `bench (superfile, …)` job failed on a transient object-store PUT that
the retry layer re-issued, not on a regression:

```
gate=warm p90, 0 blocking, 1 informational, 13 advisory regressions,
7 improvements, 1 failure line(s), baseline=yes
failure: … WARN infino::storage::retry: transient object-store error;
re-issuing attempt=0 … Error performing PUT https://…blob.core.windows.net/…
```

The gate counts a matching WARN line as a failure line. The single
informational mover is unrelated to FTS (`WHERE title = ?` warm p99,
7.97 → 17.67 ms).

src/catalog/index_spec.rs
vinaykakade: done

src/superfile/builder.rs
vinaykakade: done

vinaykakade:
## Ingest parity restored

The ingest cost I reported earlier is gone. Same box, same run, main arm
vs this branch, `superfile fts` build cells, 1M docs:

| Build cell | main (`ascii_lower`) | this branch | delta |
|---|---|---|---|
| 1 writer | 14.49 s / 69.0 K/s | **14.21 s / 70.4 K/s** | −1.9% |
| 7 writers | 3.73 s / 268.2 K/s | 3.69 s / 271.0 K/s | −1.1% |
| 1 writer (positions) | 26.23 s / 38.1 K/s | 26.73 s / 37.4 K/s | +1.9% |
| 7 writers (positions) | 5.46 s / 183.2 K/s | 5.37 s / 186.2 K/s | −1.6% |

All four cells inside ±2%, and stored size stays byte-identical
(618.43 MiB positionless, 884.72 MiB positional). The gate reports
`0 blocking, 0 informational, 1 advisory regression, 9 improvements,
0 failure lines` — pass. For contrast, the run before this fix was
`1 informational, 13 advisory regressions, 7 improvements`.

Progression on the single-writer cell: **+30.8%** (default flipped, no
fast path) → **+29.3%** (byte-at-a-time ASCII scan) → **−1.9%** (16-byte
vector lane).

### What actually cost the time

Not the general segmenter, which is what I first assumed. It was giving
up the 16-byte scan: the ASCII analyzer segments with vector compares,
and a byte-at-a-time replacement — even one driven by a class table
rather than Unicode property lookups — does not get close on x86. The
first fast path bought ~1.5 points, inside run-to-run noise, which is
why it needed replacing rather than tuning.

The vector lane rests on one observation: a 16-byte window holding no
joiner (`_ : , ; . '`) and no non-ASCII byte contains only letters,
digits and separators, and over just those classes the word-break rules
collapse to maximal `[A-Za-z0-9]` runs — exactly the ASCII analyzer's
rule. Those windows are driven from bitmasks; only the neighbourhood of
a joiner steps a byte at a time.

### Correctness

A wrong boundary here is silent recall loss — text indexed under one
tokenization and queried under another — so the lane is held to a
reference rather than eyeballed:

- The byte-at-a-time segmenter is kept as a test-only reference; the
  vector lane is compared against it exhaustively over every
  arrangement of the five bytes straddling the window boundary (three
  fillers, five lengths) and over a randomised sweep of every length
  across three strides, comparing the case-fold flag as well as the
  boundaries.
- The reference is held to the general segmenter exhaustively. That is
  complete rather than partial: the join rule reads only a four-byte
  window, so covering every four-to-six-byte class arrangement covers
  every rule combination reachable at any length.
- The property tests that compare both scans against the general
  segmenter now generate up to 52 bytes. At their previous 14 they
  never reached the vector lane at all.
- Byte-identical index output, and the bench's own oracle on the built
  artifact reports self-consistency plus BMW == brute-force.

### Two things I am *not* claiming

On an Apple M4 Max this branch measured 10.9% *faster* than the ASCII
analyzer, which suggested that analyzer's own scan had headroom (it
takes two vector passes per token, reloading the chunk each time). On
x86 it is a wash, so that reading does not hold up and there is no
follow-up to chase from it.

The earlier comment reporting a 15–31% ingest cost is superseded by the
table above.

PR #738, opened 2026-09-09
fix arrow memory: scratch memory requirements for arrow IPC data

### LINKED ISSUE #681: build reserve measures buffer capacity, so a nested batch over IPC reports ~160x its size
## Symptom

`append` fails with `OverBudget` for a batch that comfortably fits, when the table
has nested columns and the batch arrived as an Arrow IPC stream:

```
append: during ingest, over connection memory budget:
requested <N> B with 0 B in use of a <limit> B limit
```

Measured below on a public corpus: a batch whose wire size is **23.7 MiB** makes the
reserve ask for **9.21 GiB**. No per-connection budget one would sensibly set for a
batch that size will admit it.

Two details that were diagnostic:

- the requested figure did **not** change when the input shrank from a directory of 86 files to a single file
- loading the same data to local object storage worked throughout; only the remote path failed

## Root cause

`SupertableWriter::buffer_scalar_bytes` answers two questions, and the correct measure is opposite for each:

- **held memory** (drives the auto-flush threshold) — capacity is right, since a slice pins its parent's whole allocation and that is genuinely resident
- **build scratch** (drives `reserve_build_scratch`) — capacity is wrong, since scratch scales with the rows a builder reads, not with an allocation those rows share

`reserve_build_scratch` reads the held-memory figure.

Over an Arrow IPC stream this is catastrophic. The decode points every child array at one shared message buffer, and `get_array_memory_size` sums `Buffer::capacity()`, so **every child is billed for the whole message**. A nested schema has many children, so the batch reports its message size multiplied by roughly the number of child arrays.

## Reproduction

Public dataset: `wikimedia/structured-wikipedia`, file `enwiki/data/enwiki_namespace_0_0.parquet` (354 MB, 177,499 rows, 87 row groups, 65 columns, nested `list<struct<...>>` and `struct<struct<...>>`).

Decode one 2048-row batch, round-trip it through `StreamWriter`/`StreamReader`, and measure:

| | |
| --- | --- |
| decoded from parquet, capacity | 32.6 MiB |
| slice-aware measure | 23.7 MiB |
| **on the wire (IPC bytes)** | **23.7 MiB** |
| **after IPC decode, capacity** | **3771.5 MiB** |
| inflation | **~159x** |
| reserve asks (2.5x) | 9.21 GiB |

159 is essentially the number of nested child arrays in that schema. Per-column growth across the round trip is confined to nested columns; flat columns do not move:

```
growth MiB   factor  column                type
    1399.0    2829x  version               Struct(... editor: Struct(...))
     658.6     119x  references            List(Struct(...))
     308.1    1033x  event                 Struct(...)
     213.2     785x  additional_entities   List(Struct(aspects: List(Utf8) ...))
     189.7    2137x  image                 Struct(...)
```

Decoding parquet locally gives each column buffers of its own (capacity ≈ 1.4x visible), which is why only remote connections are affected.

## Fix

The visible figure is **already computed and discarded**. `ingested_byte_legs` returns it for the priced legs, and the comment there already argues this exact case for the FTS leg:

> The FTS leg moves with the priced legs deliberately: it weights the build-scratch reserve, and the builder's scratch scales with the text it will actually index — a slice's visible rows — not with the parent allocation the slice shares.

The same argument covers the scalar leg. Concretely, in `src/supertable/writer.rs`:

1. add a field alongside `buffer_scalar_bytes` for visible scalar bytes
2. accumulate `ingested_byte_legs`'s scalar leg into it at append
3. pass **that** to `reserve_build_scratch` instead of `buffer_scalar_bytes`
4. save/clear/restore it with the other buffer counters
5. leave `buffer_scalar_bytes` on capacity — held memory and the auto-flush threshold are correct as they are

No new measurement or helper. The multipliers should not need retuning: the FTS leg has been visible-based against its own multiplier already, so there is in-tree precedent that the constants work on visible bytes. Worth confirming rather than assuming.

Deliberately unchanged: the superfile fanout heuristic also measures capacity, and its comment says capacity is "fine for a fanout heuristic that is clamped to the pool anyway", which still reads as true.

## Test

Append an IPC-decoded nested batch under a budget of exactly one capacity-measure, so `2.5x` the visible size fits and `2.5x` the capacity cannot. Have the fixture assert it actually exhibits the inflation before relying on it.

With only the one argument reverted, the test fails with the reported error shape:

```
OverBudget("append: during ingest, over connection memory budget:
requested 1235555 B with 0 B in use of a 434908 B limit")
```

## Before merging

This touches the commit path, so the bench suite is required and the comparison against `main` belongs in the PR body. That has not been run.

## Existing work

A working change with that test is on branch `build-scratch-visible` (`make check` clean, full suite green apart from an unrelated order-dependent flake in `reader_synchronous_with_storage_upgrades_lazy_hidden_entry`, which passes in isolation and is worth its own issue).

## Impact

Any table with nested columns fed from an IPC stream is effectively unloadable at a
sensible batch size: the reserve scales with the inflated figure rather than with the
data. Working around it on the caller's side means shrinking appends by roughly the
inflation factor, so a corpus that should load in tens of commits takes thousands —
leaving enough superfiles to slow queries afterwards.

Pranav2612000:
- At the moment, the calculations to find out the scratch memory needed rely on the `buffer_scalar_bytes` field. This field is calculated using arrowrs's `get_array_memory_size` fn https://github.com/apache/arrow-rs/blob/b06b9c12c17dca4f575c9c3e5a091fdb9412bab7/arrow-array/src/record_batch.rs#L804-L816
- Now, an Arrow array doesn't own bytes directly. It holds a Buffer, which is three things:
`Buffer = (pointer into a memory block, length, a handle to the whole block)`. It's possible ( as in the case with IPC ) that the same buffer stores columns and their child arrays ( with the Arrow array for each having the appropriate pointers ).
- In such scenarios, the fn overestimates and returns higher values. The fn also states this. It documents that it "might overstimate, since multiple columns can share the same buffers or slices thereof, the memory used by the shared buffers might be counted multiple times"
- This is what happens with arrow IPC, as it is zero copy. Here's an example representation
```
one 29,120-byte allocation (the IPC message)
[###############################################################]
 ^f0.offsets ^f0.values ^f1.offsets ^f1.values ...  32 windows total

Each of those 32 windows, asked .capacity(), answers 29,120 — the whole block. Sum them:

32 windows × 29,120 B = 931,840 B
```
- so `buffer_scalar_bytes` points to 931,840 B whereas the actual data is 29,120B. Scratch memory needed is also 29,120B + some overhead.

### What does the PR do?
- The PR fixes the above behaviour to ensure that ArrowIPC byte consumption calculations are done correctly. We're now able to append nested data.
- Fixes https://github.com/infino-ai/infino/issues/681

### How do we do this change?
- We introduce a new variable `self.buffer_visible_scalar_bytes` which is closer to the actual data being ingested. This improves the scratch memory calculation.
- Unit tests added to verify the changes.

### Questions
Q. Is the `buffer_scalar_bytes` calculation incorrect right now? What is its impact?
A. Yes it is. For a 1000 slices of a 100MB parent, the `buffer_scalar_bytes` is 100 * 1000 = 100GB, but the actual size is 100MB. At the moment, it'll have the impact of many small superfiles being generated ( as we use this value to determine number of superfiles to split the data into )

### Next steps ( not in this PR )

This PR unblocks ingestion, but more cleanup is required for the best behaviour.
- Use Arrow's MemoryPool/claim ( behind a feature flag ) to calculate and decide the held memory once
- Deprecate ( and remove ) the `buffer_scalar_bytes` field

### DIFF
--- a/src/runtime_metrics/ingest.rs
+++ b/src/runtime_metrics/ingest.rs
@@ -38,9 +38,17 @@ pub fn classify_batch_bytes(batch: &RecordBatch) -> (u64, u64) {
 /// zero-copy window into a larger buffer is billed for the rows it
 /// actually carries rather than for its parent's whole allocation.
 ///
-/// This is the right measure wherever bytes are *priced*. It is the wrong
-/// one for held-memory accounting (an auto-flush threshold, a scratch
-/// reserve), where the parent allocation really is what is resident.
+/// This is the right measure wherever bytes are *priced*, and wherever a
+/// figure scales with the rows something will read — the build-scratch
+/// reserve included, since the blobs and serialized file a build allocates
+/// are sized by the rows it decodes, not by the parent allocation those
+/// rows are a window into.
+///
+/// It is the wrong measure for held memory (the auto-flush threshold),
+/// where a slice really does pin its parent's whole allocation and cannot
+/// free it. Note that capacity is not right there either: summed across
+/// arrays that share one allocation it counts the same bytes once per
+/// array, so neither figure is the resident total.
 ///
 /// Two families need handling of their own, because Arrow's slice-aware
 /// sizing is not uniformly slice-aware:
--- a/src/supertable/writer.rs
+++ b/src/supertable/writer.rs
@@ -376,6 +376,12 @@ pub struct SupertableWriter {
     buffer_scalar_bytes: usize,
     /// Held f32 vector payload bytes across `buffer`.
     buffer_vector_bytes: usize,
+    /// scratch memory increases with the rows read, not with the parent
+    /// allocation. this field tracks bytes a builder will actually read.
+    /// Its possible for multiple columns, and their nested arrays, to share the allocation, causing
+    /// the get_array_memory_size() to report a much larger size than actually
+    /// needed.
+    buffer_visible_scalar_bytes: usize,
     /// Byte size of the FTS-indexed text columns within `buffer`. A
     /// subset of `buffer_scalar_bytes`, not extra held memory; tracked
     /// only to weight the build-scratch reserve, since the FTS index
@@ -1000,6 +1006,7 @@ impl Supertable {
                 inner: Arc::clone(self.inner()),
                 buffer: Vec::new(),
                 buffer_scalar_bytes: 0,
+                buffer_visible_scalar_bytes: 0,
                 buffer_vector_bytes: 0,
                 buffer_fts_bytes: 0,
                 pending_ingest: IngestTally::default(),
@@ -1148,6 +1155,7 @@ impl SupertableWriter {
 
         self.buffer.push(BufferedBatch { scalar, vectors });
         self.buffer_scalar_bytes += scalar_bytes;
+        self.buffer_visible_scalar_bytes += scalar_bytes_u64 as usize;
         self.buffer_vector_bytes += vector_bytes;
         self.buffer_fts_bytes += fts_bytes;
 
@@ -1794,7 +1802,7 @@ impl SupertableWriter {
         // Held until this function returns, i.e. past `publish_superfiles` below.
         let _build_guard = reserve_build_scratch(
             &self.inner.options.connection_memory_budget,
-            self.buffer_scalar_bytes,
+            self.buffer_visible_scalar_bytes,
             self.buffer_vector_bytes,
             self.buffer_fts_bytes,
         )?;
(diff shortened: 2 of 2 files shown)


naman-modi (approved):
LGTM, thanks for fixing.

PR #739, opened 2026-09-09
chore: reduce gcs calls for doc counts

savannahar68:
before this PR in order to fetch the number of docs at a cell level, we used to fetch superfiles from object store, and then count, leading to network calls. We already have this information in manifest. This PR uses this to reduce get calls

### DIFF
--- a/src/supertable/writer.rs
+++ b/src/supertable/writer.rs
@@ -5015,52 +5015,71 @@ async fn group_multicell_rows(
     Ok(out)
 }
 
-/// Per-cell doc counts from a packed (or legacy) entry. Legacy returns one
-/// `(partition_hint_or_0, n_docs)` pair.
-async fn cell_doc_counts_for_entry(
+/// Per-cell physical doc counts held by `entry`, as `(cell, n_docs)` pairs,
+/// with superseded cells excluded. Read straight off the manifest's
+/// per-entry [`VectorSummary`] — no superfile open. Every entry this engine
+/// writes carries one (built from the just-written bytes at commit time), so
+/// a missing or empty summary is treated as a genuine error rather than a
+/// silent degrade.
+fn cell_doc_counts_for_entry(
     inner: &SupertableInner,
     entry: &Arc<SuperfileEntry>,
     superseded: Option<&BTreeSet<u32>>,
 ) -> Result<Vec<(u32, u32)>, BuildError> {
-    let storage = inner
-        .options
-        .storage
-        .as_ref()
-        .ok_or_else(|| BuildError::Store("cell maintenance requires storage".into()))?;
-    let reader = open_reader(
-        &inner.options.store,
-        inner.options.disk_cache.as_ref(),
-        Some(storage),
-        entry,
-        true,
-    )
-    .await
-    .map_err(|e| BuildError::Store(e.to_string()))?;
-    let v = reader
-        .vec()
-        .ok_or_else(|| BuildError::Store("IVF entry missing vector index".into()))?;
-    // A superseded cell's on-disk blocks are dead (replaced by a split's
-    // children elsewhere), so it contributes no live docs — excluding it here
-    // keeps split-selection and parent-discovery from re-counting the same
-    // rows that already live in the child cells.
-    let is_superseded = |cell: u32| superseded.is_some_and(|s| s.contains(&cell));
-    if v.is_multi_cell() {
-        Ok(v.packed_cell_ids()
-            .iter()
-            .filter(|&&cell| !is_superseded(cell))
-            .filter_map(|&cell| {
-                let n = v.packed_cell_n_docs(cell)?;
-                Some((cell, n))
-            })
-            .collect())
-    } else {
-        let cell = entry.partition_hint.unwrap_or(0);
-        if is_superseded(cell) {
-            Ok(vec![])
-        } else {
-            Ok(vec![(cell, entry.n_docs as u32)])
-        }
+    let column = vector_index_column(inner)
+        .ok_or_else(|| BuildError::Store("no vector index column configured".into()))?;
+    let summary = entry.vector_summary.get(&column).ok_or_else(|| {
+        BuildError::Store(format!(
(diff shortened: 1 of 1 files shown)


muralikpbhat (commented):
Reviewed — the mechanism is right and the win is real: per-cell counts now come off the manifest's per-entry `VectorSummary` (summed `clusters.counts`, superseded cells excluded with the same semantics as before), which deletes a superfile open per entry per maintenance pass at both call sites (`scan_cell_parents`, `recalibrate_probe_laws`). The function going `async → sync` because the awaits fell out is the honest signature of removing I/O. Tests cover the drained two-cell shape, superseded exclusion, both error paths, the legacy single-cell shape, and the column fallback; the count sum is overflow-checked. CI green.

**Holding approval on one thing — the missing-summary hard-error can wedge maintenance on existing tables.**

The new code errors when an entry has no `VectorSummary`, and the doc-comment justifies it with "every entry this engine writes carries one." Two problems with that claim:

1. **The write side doesn't guarantee it.** Both staging sites insert conditionally — `if let Some(summary) = build_column_vector_summary(...)` at writer.rs:2871 and the WAL update pipeline (pipeline.rs:689), whose own doc says a `None` from the reader skips the column. So an entry without a summary is a shape the engine itself can produce.
2. **Entries written before #422 (which introduced `vector_summary`) have none at all.** A hidden manifest carrying any such entry now hard-fails `optimize()` — split selection *and* law recalibration — where the old code worked by opening the superfile. And the recovery path is compaction, which itself calls `scan_cell_parents`, so an affected table has no way out. With long-lived customer tables in the wild, this is worth being certain about.

Ask, either of:
- keep the reader-open path as a **warn + fallback** for a missing-summary entry (the fast path stays the default; the fallback logs loudly so it's visible), or
- verify no live hidden manifest predates #422 or carries a summary-less entry, **and** make the write-side invariant true by construction (insert unconditionally or error at write time) so the read-side error claim actually holds going forward.

One more thing to confirm (a sentence in reply is fine): the byte-splice merge path rebuilds each output entry's summary from the spliced result rather than carrying a source entry's forward — the metadata-only restage at writer.rs:2948 cloning `old.vector_summary` is correct *there*, but a spliced entry whose counts came from one source would drift the counts this PR now serves.

savannahar68:
> Reviewed — the mechanism is right and the win is real: per-cell counts now come off the manifest's per-entry `VectorSummary` (summed `clusters.counts`, superseded cells excluded with the same semantics as before), which deletes a superfile open per entry per maintenance pass at both call sites (`scan_cell_parents`, `recalibrate_probe_laws`). The function going `async → sync` because the awaits fell out is the honest signature of removing I/O. Tests cover the drained two-cell shape, superseded exclusion, both error paths, the legacy single-cell shape, and the column fallback; the count sum is overflow-checked. CI green.
> 
> **Holding approval on one thing — the missing-summary hard-error can wedge maintenance on existing tables.**
> 
> The new code errors when an entry has no `VectorSummary`, and the doc-comment justifies it with "every entry this engine writes carries one." Two problems with that claim:
> 
> 1. **The write side doesn't guarantee it.** Both staging sites insert conditionally — `if let Some(summary) = build_column_vector_summary(...)` at writer.rs:2871 and the WAL update pipeline (pipeline.rs:689), whose own doc says a `None` from the reader skips the column. So an entry without a summary is a shape the engine itself can produce.
> 2. **Entries written before #422 (which introduced `vector_summary`) have none at all.** A hidden manifest carrying any such entry now hard-fails `optimize()` — split selection _and_ law recalibration — where the old code worked by opening the superfile. And the recovery path is compaction, which itself calls `scan_cell_parents`, so an affected table has no way out. With long-lived customer tables in the wild, this is worth being certain about.
> 
> Ask, either of:
> 
> * keep the reader-open path as a **warn + fallback** for a missing-summary entry (the fast path stays the default; the fallback logs loudly so it's visible), or
> * verify no live hidden manifest predates #422 or carries a summary-less entry, **and** make the write-side invariant true by construction (insert unconditionally or error at write time) so the read-side error claim actually holds going forward.
> 
> One more thing to confirm (a sentence in reply is fine): the byte-splice merge path rebuilds each output entry's summary from the spliced result rather than carrying a source entry's forward — the metadata-only restage at writer.rs:2948 cloning `old.vector_summary` is correct _there_, but a spliced entry whose counts came from one source would drift the counts this PR now serves.

I actually had fallback earlier, removed it because i thought it won't be needed, still think its not needed. But added it back with a warning message, we will have an alert on this and remove the fallback path in case warns doesn't show up.

I don't feel remaning points are issue 

On your merge question: merged files rebuild their own summary from the merged data, so the counts stay correct. Nothing gets copied over from the old files.

muralikpbhat (approved):
Approving. The round matches the ask precisely, and I verified it in code:

- **Warn + fallback landed in the right shape**: `cell_doc_counts_from_summary` is now a pure `Option` fast path, and *every* shape that previously hard-errored — no summary for the column, empty cells, even a u32 overflow (`.ok()?`) — declines to a loud `warn!` + `cell_doc_counts_via_reader`, which is the original open-the-superfile logic restored byte-for-byte. Maintenance can no longer wedge on a summary-less entry, and the warn gives you the alert signal to later retire the fallback with evidence.
- **The fallback test is the strong version**: it asserts the summary path declines on both broken shapes *and* that the fallback result equals the reader path's counts — equivalence, not just no-error.
- Splice answer accepted: merged files rebuilding their own summaries from the merged data is exactly what keeps these counts trustworthy.

One nit, fix at your leisure (or in a follow-up): the function's doc-comment now carries both generations — the old paragraph still says a missing/empty summary "is treated as a genuine error rather than a silent degrade," directly above the new paragraph describing the fallback. The stale half should go; as written the comment contradicts the code it documents.

On the alert plan: agreed, and worth pairing the warn with a date — if it never fires across a few weeks of sweeps on the long-lived tables, that's the evidence to make the write-side insert unconditional and delete the fallback; if it does fire, we learned which manifests still carry pre-summary entries.

PR #740, opened 2026-09-09
fts: say global where the BM25 statistics default is documented

vinaykakade:
Making `Bm25Stats::Global` the default flipped the behavior but left six places describing the old default. Two of them are the bindings' **published API surface**, and both shipped that way in 0.7.0.

### The user-visible half

- `infino-node/infino/index.ts` — the `Bm25Stats` type comment and `Bm25SearchOptions.stats` both named per-superfile as the default.
- `infino-python/src/lib.rs` — the `bm25_search` docstring named `"per_superfile"` as the default and called it "fastest".

A caller who omits `stats` was told they get segment-local idf while the engine scores with corpus-wide idf — a silent ranking difference, documented backwards, in a release that is already on npm and PyPI.

### The rest of the trail

- `ClauseLists::global_idf` field doc (`src/superfile/fts/reader/core.rs`)
- the doc comment on `bm25_global_stats_scores_across_superfiles` (`src/supertable/query/vector.rs`)
- the Python smoke test's comment
- the FTS quality bench's column legend — the bench's own code was already corrected to emit `recall (per-superfile stats)`, so the module doc and the checked-in table headers were the only stale copies. The numbers under that column are unchanged; only the label was wrong.

### One of them was not just a comment

`bm25_global_stats_keeps_the_default_modes_membership` took its per-superfile arm from `bm25_hits`, which scores with `Bm25Stats::default()` — now `Global`. **The test was comparing global statistics against themselves and asserting nothing.** It has been passing vacuously since the default moved.

It now requests `PerSuperfile` explicitly via `bm25_hits_stats`, so it exercises the cross-mode membership invariant it was written to guard, and is renamed for what it actually compares. Verified failing-to-meaningful: the test passes with a real two-mode comparison.

`infino-python/python/infino/_infino.pyi` is deliberately untouched — it declares the `Literal` but makes no default claim.

### Verified

`cargo fmt --check`, `cargo clippy --all-targets` clean, and `cargo test --test supertable bm25_global_stats` passes.

### DIFF
--- a/benches/README.md
+++ b/benches/README.md
@@ -229,11 +229,11 @@ JSON metrics land in `target/infino-bench/<bench>.json` (local only).
 
 _Host: unknown CPU · 16C/16T · macos/aarch64_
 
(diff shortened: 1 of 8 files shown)


naman-modi (approved):
LGTM

PR #742, opened 2026-09-09
fts: per-column BM25 parameters, declared and overridable per query

vinaykakade:
Adds the `k1` / `b` BM25 similarity knobs: declared per FTS column at `create_table`, and overridable per search. Stopword and stemming options are a separate change.

## Why both declaration sites

Scoring parameters are naturally a query-time concern — retuning relevance should not require rebuilding an index. What makes that awkward here is the bound representation: the build writes the exact `f32` BM25 score per block, so a stored bound belongs to the parameters that produced it, and a query scoring with a different pair would prune against bounds that no longer hold.

Rather than change that representation, this keeps it and corrects the bound. Bound tightness is the measured single-term lever on this engine — the exact-`f32` bound is what won it — and a representation that defers the bound computation to query time was built and measured here earlier: a dead heat overall, with the union top-10 case regressing. So a column declares its pair and the build bakes bounds at it, and a query may score with a different pair with the bounds corrected for the difference.

## The correction factor

`k1` / `b` do not enter the score linearly, so there is no exact ratio — but there is a computable supremum. `idf` cancels, and the per-doc ratio between two parameter sets is monotone in `tf` tending to 1, so its supremum over `tf ≥ 1` is `max(1, (1+A)/(1+B))` per length bucket, where `A` and `B` are the two norm decode tables' entries. The max over the buckets documents actually occupy gives a scalar `R ≥ 1` that keeps every stored bound an upper bound under the new pair — loosening, never under-bounding, and exactly `1.0` when the pairs agree.

`R` composes into a multiply that was already there. The reader already rescales stored bounds for a query-time parameter the build cannot know — global-idf statistics, and the repeated-term `weight` — in `TermCursor::new`'s `idf_rescale` and `search_single_term_bmw`'s `bound_rescale`. Both factors are linear multipliers on a stored bound, and both are `1.0` on the default path, so the default path performs the same multiply it already performed, with one more factor folded into it at cursor-build time.

## Scope of the change

The hot kernel is already parameterized: `score_with_dl_norm_k1` is `idf_x_k1p1 · tf / (tf + dl_norm_k1)`, so the scoring loops — scalar, SIMD, union, AND, phrase — are untouched and only its two precomputed inputs move. `Bm25Params` owns that arithmetic; `NormTable` derives its 256-entry decode table from it.

**One latent bug fixed along the way**: `idf_x_k1p1` was built from the `bm25::K1` constant in four places, so a column with a declared pair would have multiplied by `1.2 + 1` while dividing by a norm table built at its own `k1`. Cursors now build it from the effective pair, and `TermCursor` keeps the bare `idf` so the phrase cursor composes member idfs directly instead of recovering them by dividing by a `(k1 + 1)` it would have to assume.

## Format and compatibility

No layout change, no new region, no section-version step. The pair is written into `inf.fts.columns` **unconditionally, defaults included** — unlike `positions` and `stored`, whose absence has exactly one possible meaning. A scoring parameter is the provenance of the stored bounds, and a reader that infers it will infer wrong the day the recommended default moves; the `rerank_codec` note in `supertable::manifest::options_hash` records that lesson from a real mismatch. A missing pair therefore means the standard values and nothing else, frozen as a format constant, so **every file written before this reads unchanged** and every derived quantity reduces to what it was.

The pair rides `new_from_reader` alongside the analyzer, positions and stored flag. Dropping it there would rebake a merged file's bounds at the standard values while the source superfiles kept their own, so one column would score two ways depending on which superfile a document landed in — and a compaction, not any user action, would be what changed the ranking. Same failure shape as the `rerank_codec` omission that path already had a regression test for.

## Signature changes

None of these appear in the published surface: `SupertableReader` is declared inside `test_visible!`, which is `pub` only under the `test-helpers` feature and `pub(crate)` otherwise, so a default-feature build — which is how `public-api.txt` is generated — has zero entries for it. `Supertable::bm25_search` *is* published, but its signature does not move here; it has taken `Bm25SearchOptions` since #465. In-repo call sites moved with them.

**That is not the same as "no caller breaks", and one did.** A crate that enables `test-helpers` gets the wider surface, and a change to any `test_visible!` item is invisible to both `cargo-public-api` and `cargo-semver-checks` — the latter is run with `default-features` precisely so it does not grade test-helpers internals. The bench engine wrapper enables that feature and calls `reader.bm25_search(...)`; it failed to compile against this branch until its call site was updated. No automation covers those consumers, so they have to be swept by hand.

- `SupertableReader::bm25_hits(column, query, k, mode)` → `(column, query, k, opts)`, and **`bm25_hits_stats` is removed**. That method existed only to pass one extra knob past `mode`; the statistics scope is now a field of `Bm25SearchOptions`, so keeping both would have been a parallel method for no remaining reason — and adding the parameter override on top would have needed a third variant.
- `SupertableReader::bm25_search(column, query, k, mode, stats, projection)` → `(column, query, k, opts, projection)`, and `Supertable::bm25_search` likewise. This is what lets a third knob arrive without a fourth positional.
- `SuperfileReader::prepare_clauses`, `run_prepared` and `bm25_search_or_range_prebuilt` each take `bm25: Option<Bm25Params>`, and a new private `fts_scored` resolves it. `prepare_clauses` and `run_prepared` must be given the *same* override: the cursors carry `idf · (k1 + 1)` from the pair they were built with, and scoring divides by a norm table derived from that pair.
- `FtsBuilder::register_column_with_tokenizer` takes the column's pair; `bm25::score` takes it too.

`bm25_hits` gets more verbose at simple call sites — `bm25_hits("t", "q", K, BoolMode::Or)` becomes a four-line struct construction. If that is the wrong trade, `impl Into<Bm25SearchOptions>` with a `From<BoolMode>` would keep the concise form compiling; say so and I will switch it.

## Surface additions

- `FtsField::bm25(k1, b)` and `FtsConfig::bm25`, validated at `SupertableOptions::new` with a typed error naming the column; persisted in the catalog's `TableEntry` and restored on reopen under the same frozen fallback.
- `Bm25SearchOptions::with_bm25(k1, b)` for the per-search override, validated before the fan-out starts.
- `Bm25Params` re-exported at the crate root, because it is named by the public `Bm25SearchOptions::bm25` field and its only other path (`superfile::fts::bm25`) is `pub` solely under the `test-helpers` feature — the type would otherwise appear on the contract while being unnameable in a default build. Its scorer internals (`dl_norm_k1`, `idf_x_k1p1`) stay `pub(crate)`.
- Both bindings: python `IndexSpec.fts(k1=, b=)` and `bm25_search(k1=, b=)`, with the hand-maintained `.pyi` updated and the typing sample exercising them; node `FtsOptions.k1/b` and `bm25Search`. Each requires the pair together or not at all — the two interact through the length norm, so a half-override scores with a combination the caller never chose.

## Versions: nothing further to bump

**Superfile format: no version change, and none is needed.** `src/superfile/format/` is untouched, `FORMAT_VERSION` is unchanged, and no FTS section version moves — every file this writes is still V5. Nothing in the reader derives behaviour from a version distinction here, because the pair is read from the column entry. Older files read unchanged: a missing pair can only mean the standard values, and every derived quantity then reduces to what it was.

The one on-disk difference is additive and forward-only: a newly written column entry carries `,"k1":1.2,"b":0.75` (18 bytes per FTS column) even at the defaults, and a *declared* column's stored bounds are that column's true maxes under its own pair. Files already on disk are never rewritten.

**Crate version: `0.7.0`, already stamped by #735 — this needs no further bump.** The published surface does move, though, so it is worth being precise rather than claiming otherwise: `public-api.txt` gains `Bm25Params`, `Bm25SearchOptions` gains `#[non_exhaustive]`, the `bm25` field and `with_bm25`, `FtsField::bm25` is added, and `Bm25SearchOptions` **loses its `Eq` derive**. The last two of those are breaking on a `0.x` line, which is exactly what the already-stamped minor covers. Practically nothing breaks — every construction site in this repo and in the only Rust consumer uses the builders, and nothing depended on `Eq` — but `cargo-semver-checks` grades the surface rather than the blast radius, and it is satisfied by the version #735 already set.

## Tests

**The correction is exact where it has to be.** An override scores identically to a file *baked* at that pair — same doc order, same scores — across four parameter pairs, four query shapes and three values of `k`. This is the test that catches a bound left too small, since a bound that under-states a score makes the block-max skip drop a qualifying document.

**Graded against the reference implementation at the pair.** `BruteForceBm25::with_params` makes the brute-force oracle score with the same pair, so a comparison means something rather than grading one similarity function against another. Two gates on it: a column declaring 1.6 / 0.4 ranks exactly as the reference does at 1.6 / 0.4 across single-rare, single-common and two-term-union shapes; and a query-time override on a *standard* table matches the reference at the override's pair for three pairs including `b = 0` — the pruning-correction path, where every stored bound is inflated by the factor.

**Every kernel that reads a stored bound.** Phrase and must-clause queries route through code the single-term and union tests never reach — the phrase cursor composes its members' idfs and its own term-level bound, and a `+must` clause runs the ranked-AND membership walk. A table baked at a pair is compared against a default table overridden to it, document-for-document and score-for-score, over a phrase, a two-must query and a phrase-plus-term mix.

**The composed factor.** What ships is `(idf / local_idf) · R`, and global statistics are the default, so the composed form is the common path rather than an edge case. An override is checked against a baked table under both `Global` and `PerSuperfile`.

**Pruning happens, and drops nothing.** A small `k` fills the top-k heap and engages the block-max skips where a `k` covering the whole match set does not, so the small-`k` head must equal the head of the unpruned run. Four parameter pairs, two small `k` values.

**The derivation itself.** `R` is never below any rescored score across 7 parameter pairs × 50 lengths × 6 term frequencies — the supremum property the whole correction rests on, pinned directly rather than through its uses. Plus `R == 1.0` for an unchanged pair and `>= 1.0` otherwise.

**Validation.** Seven out-of-range declared pairs — including `NaN` and infinity — are refused with the column named and both bounds in the message, and the four boundary values that must be accepted (`b = 0`, `b = 1`, tiny and large `k1`) are. The same bounds are enforced on the query-time override, checked before the fan-out starts.

**Persistence.** The pair round-trips through the column entry; a missing pair reads as the standard values; the decode table matches the closed form against the quantized length the scorer actually sees; and the pair survives a rebuild, so a compaction cannot silently rescore a column.

**Bindings, functionally and not just by type.** `mypy --strict` proves the signature exists; three python tests prove it works — a declared column and an overridden default table agree on titles and scores, passing one parameter without the other is refused on both the search and the create surfaces, and out-of-range pairs are refused.

## CI

`make check` (fmt, clippy `-D warnings` including the `remote` feature, bench-gate, doc-check), `make doctest`, `make test-remote`, `make public-api`, `scripts/check_api_parity.py` (20 covered), `make python-typecheck`; lib suite 2473 pass, supertable integration 337 pass.

`make coverage` did not finish locally — I terminated the run — so I am not quoting a number for it.

## Measurement

A same-box A/B of this branch against `main`, 962 queries across five command shapes, with the branch re-measured in a later slot to separate a real delta from measurement-position bias. Two results are solid because they do not depend on timing:

- **Identical results.** Zero count mismatches against `main` across all 962 queries × 5 command shapes. The default path returns what it returned before.
- **Index size +127 bytes** on 4.13 GB (0.000003%) — the always-written pair in `inf.fts.columns`, and independent confirmation that the two arms really were different binaries.

**On speed I am claiming nothing.** Every command except one flips sign between the two slots (TOP_10 `+4.5%` then `-2.8%`; TOP_100 `-16.3%` then `+2.9%`; COUNT `-4.9%` then `+0.9%`), which is the signature of position bias rather than a code effect. The run's noise floor is wider than any difference it shows, so it establishes neither a regression nor a win. An earlier run had reported a TOP_10 regression consistently in both slots; that did not reproduce here.

### DIFF
--- a/src/superfile/fts/reader/metadata.rs
+++ b/src/superfile/fts/reader/metadata.rs
@@ -26,36 +26,105 @@ use crate::superfile::fts::{bm25, tokenize::Tokenizer};
 #[derive(Debug, Clone)]
 pub struct NormTable {
     /// Per-doc quantized length bucket. Empty for a column with no docs.
-    bytes: Vec<u8>,
-    /// Bucket → `K1·(1 - B + B·dequantize_len(bucket)/avgdl)`. A fixed
+    /// Parameter-free — the buckets are lengths, not norms — and shared
+    /// rather than copied so [`NormTable::rescored`] costs one `Arc`
+    /// bump plus a 1 KiB table instead of a pass over every doc.
+    bytes: Arc<[u8]>,
+    /// Bucket → `k1·(1 - b + b·dequantize_len(bucket)/avgdl)`. A fixed
     /// 256-entry table, boxed so `ColumnMeta` stays pointer-sized (it is
     /// scanned by non-scoring paths — column lookup, listing) while the
     /// `u8` bucket index into a fixed-length array lets the compiler drop
-    /// the bounds check in `get`.
-    lut: Box<[f32; 256]>,
+    /// the bounds check in `get`. This is the only parameter-dependent
+    /// part of the table.
+    lut: Arc<[f32; 256]>,
+    /// Lowest and highest bucket any doc in this column actually
+    /// occupies, tracked at build so [`NormTable::bound_scale`] takes
+    /// its supremum over lengths that occur rather than over all 256
+    /// representable ones. `(0, 0)` for an empty column.
+    occupied: (u8, u8),
 }
 
 impl NormTable {
     /// Build from a column's per-doc lengths and average length. An
     /// `avgdl` of `0.0` (empty column) yields an empty table; it is
     /// never indexed because `search` short-circuits on empty columns.
-    pub(super) fn new(doc_lengths: impl Iterator<Item = u32>, n_docs: usize, avgdl: f32) -> Self {
+    pub(super) fn new(
+        doc_lengths: impl Iterator<Item = u32>,
+        n_docs: usize,
+        avgdl: f32,
+        params: bm25::Bm25Params,
+    ) -> Self {
         if avgdl <= 0.0 {
             return Self::empty();
         }
-        let inv_avgdl = 1.0_f32 / avgdl;
-        // Fill the boxed table in place so the 256 f32s land on the heap
-        // directly rather than being built on the stack and moved.
-        let mut lut = Box::new([0.0_f32; 256]);
-        for (b, slot) in lut.iter_mut().enumerate() {
-            let dl = bm25::dequantize_len(b as u8) as f32;
-            *slot = bm25::K1 * (1.0 - bm25::B + bm25::B * dl * inv_avgdl);
-        }
         let mut bytes = Vec::with_capacity(n_docs);
+        let mut lo = u8::MAX;
+        let mut hi = u8::MIN;
         for dl in doc_lengths {
-            bytes.push(bm25::quantize_len(dl));
+            let bucket = bm25::quantize_len(dl);
+            lo = lo.min(bucket);
+            hi = hi.max(bucket);
+            bytes.push(bucket);
         }
-        Self { bytes, lut }
+        let occupied = if bytes.is_empty() { (0, 0) } else { (lo, hi) };
+        Self {
+            bytes: Arc::from(bytes),
+            lut: build_lut(avgdl, params),
+            occupied,
+        }
+    }
+
+    /// The same per-doc buckets decoded at different parameters — for a
+    /// query that overrides what its column declared. Shares `bytes`,
+    /// so the cost is one 256-entry table.
+    pub(super) fn rescored(&self, avgdl: f32, params: bm25::Bm25Params) -> Self {
+        if self.bytes.is_empty() {
+            return Self::empty();
+        }
+        Self {
+            bytes: Arc::clone(&self.bytes),
+            lut: build_lut(avgdl, params),
+            occupied: self.occupied,
+        }
+    }
+
+    /// The factor `R >= 1` by which every bound built at `baked` must be
+    /// inflated to stay an upper bound under `query`'s parameters, where
+    /// `self` is the norm table at `baked` and `other` the one at
+    /// `query`.
+    ///
+    /// The stored per-block bound is the block's true max of
+    /// `idf·tf·(k1+1) / (tf + dl_norm_k1)`. Between two parameter sets
+    /// the per-doc ratio is
+    ///
+    /// ```text
+    ///   (k1'+1)/(k1+1) · (tf + A) / (tf + B),
+    ///       A = lut_baked[bucket],  B = lut_query[bucket]
+    /// ```
+    ///
+    /// with `idf` cancelling — it carries no parameters. For a fixed
+    /// bucket that is monotone in `tf` and tends to 1, so its supremum
+    /// over `tf >= 1` is `max(1, (1+A)/(1+B))`; taking the max over the
+    /// buckets docs actually occupy gives the supremum over the column.
+    /// Loosening, never under-bounding, and exactly `1.0` when the two
+    /// parameter sets agree.
+    pub(super) fn bound_scale(
+        &self,
+        other: &NormTable,
+        baked: bm25::Bm25Params,
+        query: bm25::Bm25Params,
+    ) -> f32 {
+        if baked == query || self.bytes.is_empty() {
+            return 1.0;
+        }
+        let (lo, hi) = self.occupied;
+        let mut worst = 1.0_f32;
+        for bucket in lo..=hi {
+            let a = self.lut[bucket as usize];
+            let b = other.lut[bucket as usize];
+            worst = worst.max((1.0 + a) / (1.0 + b));
+        }
+        worst * (query.k1 + 1.0) / (baked.k1 + 1.0)
     }
 
     /// `dl_norm_k1` for a doc (length quantized): one per-doc byte load
@@ -79,10 +148,23 @@ impl NormTable {
     /// read.
     pub(super) fn empty() -> Self {
         Self {
-            bytes: Vec::new(),
-            lut: Box::new([0.0; 256]),
+            bytes: Arc::from(Vec::new()),
+            lut: Arc::new([0.0; 256]),
+            occupied: (0, 0),
         }
     }
+}
+
+/// Decode table for one parameter set: bucket → `dl_norm_k1`. Filled in
+/// place on the heap so the 256 `f32`s are not built on the stack and
+/// moved.
+fn build_lut(avgdl: f32, params: bm25::Bm25Params) -> Arc<[f32; 256]> {
+    let mut lut = Box::new([0.0_f32; 256]);
+    for (bucket, slot) in lut.iter_mut().enumerate() {
+        let dl = bm25::dequantize_len(bucket as u8);
+        *slot = params.dl_norm_k1(dl, avgdl);
+    }
+    Arc::from(lut)
 }
 
 /// Per-column metadata, indexed by column_id (declaration order).
@@ -101,6 +183,19 @@ pub struct ColumnMeta {
     /// `dl_norm_k1.get(d)` and multiplies-out to `idf · tf · (K1+1) /
     /// (tf + dl_norm_k1.get(d))`.
     pub dl_norm_k1: NormTable,
+    /// The parameters this column is being *scored* with. Equal to the
+    /// pair recorded in the KV entry unless the query overrode it, in
+    /// which case `dl_norm_k1` has been re-decoded to match and
+    /// `bound_scale` carries the correction for the stored bounds.
+    pub params: bm25::Bm25Params,
+    /// Factor to apply to every bound read out of the skip table or the
+    /// coarse table before comparing it against a score. `1.0` unless
+    /// the query overrode this column's parameters, in which case it is
+    /// [`NormTable::bound_scale`] between the baked pair and the
+    /// query's — the stored bounds belong to the baked pair, and
+    /// inflating them by this keeps them upper bounds under the pair
+    /// actually being scored.
+    pub bound_scale: f32,
     /// Whether this column's index carries token positions (from
     /// `inf.fts.columns`); phrase queries require it.
     pub positions: bool,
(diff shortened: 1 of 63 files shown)


tests/supertable/query/brute_force_oracle.rs
```diff
@@ -339,6 +382,90 @@ static STANDARD_FIXTURE: LazyLock<StandardFixture> = LazyLock::new(|| {
...
+fn oracle_query_time_override_matches_the_reference_at_that_pair() {
+    let f = &*STANDARD_FIXTURE;
+    let corp = corpus_with_prefix_terms();
+    for params in [
+        OracleBm25Params { k1: 1.6, b: 0.4 },
+        OracleBm25Params { k1: 0.6, b: 0.9 },
+        OracleBm25Params { k1: 1.2, b: 0.0 },
+    ] {
+        let oracles = build_oracles_with_params(&corp, SUPERFILES, params);
+        for (query, k, head) in [
```
naman-modi: Can we add a few more query types like:
- phrase eg: "new york"
- AND eg: +rust +async 

These seem important to be tested with non-default and compared with brute force.

src/supertable/query/fts.rs
```diff
@@ -3312,6 +3502,196 @@ mod tests {
...
+    }
+
+    /// A small `k` fills the top-k heap and engages the block-max skips;
+    /// a `k` covering the whole match set does not. Under an override
+    /// every stored bound is inflated by the correction factor, so a
+    /// factor that came out too small would skip a block holding a
+    /// qualifying document — visible only as the small-`k` result
+    /// disagreeing with the head of the unpruned one.
+    #[test]
+    fn an_override_prunes_without_dropping_hits() {
```
naman-modi, line 3772: Can we have atleast a few 100 docs, so the block skipping path (taking into the new change of multiplier by R) is actually enforced? Right now, I dont think that is tested.

src/superfile/fts/reader/metadata.rs
```diff
@@ -138,12 +233,41 @@ pub struct FtsColumnConfig {
...
+    pub fn params(&self) -> bm25::Bm25Params {
+        bm25::Bm25Params::new(self.k1, self.b)
+    }
 }
 
 pub(super) fn default_stored() -> bool {
     true
 }
 
+pub(super) fn default_k1() -> f32 {
```
naman-modi, line 263: The old superfiles (written before this PR) don't have k1/b in inf.fts.columns, so on read they fall back to default_k1() = bm25::K1. that's fine today since it's 1.2, as ealrier.

But `bm25::K1` is also the recommended default, which can be changed anytime. So if someone later sets K1 = 1.4, default_k1() moves to 1.4 too, and those old fieldless files (still written at 1.2) get read as 1.4 -> norm table at 1.4 vs bounds baked at 1.2 -> wrong scores/pruning, silently.

So, can we split it: a separate FROZEN_STANDARD_K1 = 1.2 (and _B) that default_k1/default_b return, frozen forever, and keep bm25::K1 as the recommendation that's free to move?

naman-modi (approved):
Minor comments, else looks good to me.

vinaykakade:
Added both shapes to the parameter tests in `2c77c9a2` — and to *both* arms, so the declared pair and the query-time override now grade identical ground. Good call: those two read a stored bound the covered shapes never touch (the must-clause runs the ranked-AND membership walk; the phrase composes its members' idfs against a term-level bound), which is exactly where a mis-scaled bound would surface first.

Two things worth flagging from doing it.

**The obvious version of the phrase case had no teeth.** I first added it with the same head of 2 the neighbouring cases use, then mutated the oracle to score the phrase as an intersection to check it would fail — and it passed. `"rust async"` and `+rust +async` agree on the top two; the document that separates them, `rust trait dyn impl async` (both terms, not adjacent), ranks third and was never compared. Both new cases now compare the whole result instead of a prefix. Re-mutated to confirm they fail:

- oracle scores the phrase as an intersection → `declared_phrase: supertable=[0, 20] oracle=[0, 20, 22]`
- engine query drops the `+` sigils → `declared_two_term_and: supertable=[0, 20, 22, 52, 58, ...] oracle=[0, 20, 22]`

**The clause structure is stated explicitly rather than parsed.** The phrase suite feeds the oracle from the engine's own parser (`tok.parse(query).into_clauses(mode)`), which is convenient but means a parsing change moves both sides together and cannot fail. Since these cases grade *scoring* at a non-standard pair, I wrote the decomposition out so a parser that stopped honouring a sigil fails here instead of agreeing with itself.

Two incidental notes: the fixture column now indexes positions, because a phrase query is otherwise rejected outright — that changes what the column can be asked, not how a term scores, and every pre-existing case still passes. And the merge shared by the oracle wrappers is now one helper; the sort/tie-break/truncate was copied three times and these shapes would have made it five.

337 supertable tests pass, fmt and clippy clean.

tests/supertable/query/brute_force_oracle.rs
vinaykakade: Done

vinaykakade:
You're right, and it was worse than under-tested — the assertion was unreachable. `BLOCK_LEN` is 128, so three documents are a single partial block and the block-max skip never runs. That test passed on *any* correction factor.

Fixed in `5c57dead` with a corpus spanning five whole blocks (`5 * BLOCK_LEN`): `quick` is in every document so its posting list splits into whole blocks with stored per-block maxima, term frequency varies (`i % 3` repetitions) and document length sweeps 1..24 padding tokens, so per-block maxima differ and both halves of the score move when the pair changes. Small `k` (1, 2, 10, 50) against a match set of 640 fills the heap early, so the kth-best score climbs above whole blocks' maxima and the skip actually runs. It now also asserts scores match, not just ids.

Rather than assume the path is engaged, I measured it — forcing the correction factor to `1.0` in `NormTable::bound_scale`, an under-bound that should drop qualifying documents:

```
an_override_prunes_without_dropping_hits_across_blocks ... FAILED
  assertion failed: k=2 at k1=2 b=1: pruning changed the top-2 head
an_override_prunes_without_dropping_hits (3 docs) ....... ok
```

The new case catches the broken bound; the old one does not notice. That is the gap you spotted, quantified.

I kept the three-document case — it covers the partial-block tail, which has its own bound handling — and renamed the new one to say which regime it covers.

Worth noting the idf-rescale path already had exactly this shape of coverage in `global_stats_small_k_pruning_matches_single_superfile`, which sizes its corpus to 160 past a block for the same reason. The parameter correction is the direct analogue and should have matched it from the start.

2481 lib tests pass, fmt and clippy clean.

src/supertable/query/fts.rs
vinaykakade: Done

src/superfile/fts/reader/metadata.rs
vinaykakade: thanks, when we change the k1/b defaults, we can increase the superfile version and handle this there. (trying to avoid "default for a default" right now).

PR #753, opened 2026-09-10
fts: per-column stopwords and stemming on the analysis chain

vinaykakade:
## What

An FTS column's **analysis** grows from just a tokenizer into a tokenizer plus an optional
stopword set and an optional stemmer, applied in that order. Each is its own option,
declared through the same `FtsField` door as the column's other options, and all three
apply to both sides of a search — so a term is queried in the form it was indexed under.

> **A note on names.** *Tokenizer* here means the thing that splits text (`standard`,
> `ascii_lower`); *analyzer* means the whole chain — that tokenizer plus its filters.
> Before filters existed the two were interchangeable in this engine, and four surfaces
> still say "analyzer" for a value that is a bare tokenizer name: the public
> `.analyzer(...)` option, `FtsConfig`'s field, the wire key, and the catalog record. The
> persisted superfile key says `"tokenizer"` and is the one that has it right. This PR
> **does not rename anything** — that would break three ecosystems plus the hosted spec
> for a cosmetic win — but it does make every doc-comment say tokenizer where the value is
> a tokenizer, and states the vocabulary outright in `analysis.rs`.

**Both filters are opt-in and the default is untouched.** That is deliberate: the default
is the parity surface, and it already matches Lucene's default. `IndexWriterConfig`
defaults to `StandardAnalyzer`, whose no-arg constructor is documented as building "an
analyzer with no stop words" and which never stemmed — so tokenize-and-lowercase, which
is what `standard` does here, is the right default and this PR does not move it. A column
declaring no filter is not wrapped at all: it keeps its plain analyzer name and takes the
same monomorphized ingest scan, so it is byte-identical *and* code-path-identical to one
declared before chains existed. Closing any remaining gap in `standard` itself belongs to
the tokenizer and the scorer, not here.

This PR is scoped to stopwords and stemming only.

```rust
IndexSpec::new().fts(
    FtsField::new("body")
        .stopwords(Stopwords::English)
        .stemmer(Stemmer::English)
        .positions(true),
);
```

Both bindings take the filters by name, matching how they already take `analyzer` and
`metric`:

```python
IndexSpec().fts("body", stopwords="english", stemmer="english", positions=True)
```

## Persistence: two additive fields, and a derived identity

A column persists its base tokenizer name plus, only when set, a `stopwords` and a
`stemmer` field:

```json
{"name":"body","tokenizer":"standard","k1":1.2,"b":0.75,
 "stopwords":"english","stemmer":"english","positions":true}
```

Both are ordinary additive fields. An old file has neither, serde defaults them off, and
that is correct because such a file really was built unfiltered — so the direction this
format promises (**new reader, old file**) needs no special handling. The catalog record
and the create-table request carry them the same way. A column with no filter is
byte-identical everywhere to one declared before filters existed, and there is **no
format-version change**.

The rule at the boundary: an unknown *field* is ignorable, so an engine predating a filter
reads the column as unfiltered and that column's results degrade until it rolls forward.
An unknown *value* of a field the reader does know — `"stopwords":"german"` on an engine
shipping no German list — fails the open, because analyzing without a set the postings
were built with is a different index, not a degraded one.

That split is deliberate and follows a precedent already set here: this codebase declined
to mark these files with a new FTS section version specifically because a lost-recall
regression is recoverable by rolling forward while an unopenable table is an outage.
Encoding the filters into the analyzer name would have made a rollback exactly that
outage, since an older engine cannot parse the name at all.

The chain still *has* an identity string — `"standard+stop=english+stem=english"` from
`chain_name` — but it is derived in memory and never written. It exists because three
places have to treat a column's analysis as one value, and reading three fields at each
would be three chances to forget one:

- the **`LIKE` prune lowering**, which recognizes analyzers by `Tokenizer::name` (see
  below);
- the **merge carry check**, which compares analyzer names to refuse merging
  differently-analyzed superfiles;
- the **options-hash**, which needs analysis in table identity — a one-line change that
  leaves existing hashes byte-identical, since a filterless column derives to its plain
  base name.

Nothing parses a name back, so a chain is only ever built from its components.

## Position holes, on both sides

Removing a token leaves a **hole**: the ordinal it would have occupied is skipped, so
phrase matching still knows the words it matched were not adjacent in the text.
`"new york"` does not match `new the york`.

Holes travel two new channels:

- **Index** — `Tokenizer::tokenize_each_positioned` yields `(token, position)` with
  gap-inclusive ordinals. The trait's default numbers emitted tokens consecutively, which
  is correct for a tokenizer that drops nothing; one that drops tokens overrides it.
- **Query** — a parsed query's phrases carry per-term offsets (`Phrase<T>`), so
  `"end of the world"` asks for its two surviving terms *three positions apart* rather
  than adjacent. `PhraseCursor` verifies `start + offsets[j]` instead of `start + j`.
  Offsets normalize to the first surviving term, which is both what makes them a pure
  relative constraint and what keeps the verifier's `checked_sub` from underflowing near
  the start of a document.

Without the query side the feature is silently wrong in exactly the direction the
fail-loud design exists to prevent: requiring adjacency matches nothing, and requiring
mere co-presence matches `end world`.

Doc length counts the tokens the chain **emits** — what Lucene's norms count
(`IndexingChain.PerField.invert` bumps `invertState.length` inside the `incrementToken`
loop, so a token `StopFilter` dropped never reaches it) and what a merge's carried
postings preserve.

A phrase reduced to one surviving term degrades to a plain term clause; one reduced to
zero contributes nothing. Both match Lucene.

## `LIKE` pruning is dropped for a chained column, not adapted

The `LIKE` lowering bounds a pattern fragment by the terms it tokenizes to, including a
lex-range walk for a prefix token. That is sound only while an indexed term is a
substring-preserving image of the text, and a stemmer breaks it: `LIKE '%runni%'` matches
the text `running`, whose indexed term is `run`, so a prefix walk for `runni` finds
nothing and **drops a superfile that genuinely matches**. Not a pruning inefficiency — a
wrong answer, and the same lowering feeds the vector kNN pre-filter, where an
over-restrictive candidate set costs recall.

The fix was already latent: `Analyzer::of` dispatches on `Tokenizer::name`, a chain's
derived name matches no arm, and `None` there already means `Unbounded`. So the correct
behaviour arrives by construction. It is now documented on `Analyzer::of` as a
*requirement* and pinned by a test that asserts the property **directly** — that a chain's
name is recognized by no arm — rather than by a `LIKE` case that would also pass if the
resolver simply returned no tokenizer at all. That distinction matters: the weaker version
of this test passed for the wrong reason during the refactor.

Equality and value-set pruning **stay**, sound for the opposite reason: the literal runs
through the same chain the postings did, so the terms it yields are terms the index really
holds and requiring them can only widen the candidate set. A literal the chain reduces to
nothing already falls to `Unbounded`.

## Two things found by reading the code this would touch

**`positions` had no public surface at all.** `FtsConfig::positions` existed, but
`IndexSpec::to_configs` never set it and `FtsConfig` is not re-exported — so every table
created through the public API had `positions = false` and answered **every** phrase query
with an error. None of the hole behaviour above was reachable, let alone testable, without
exposing it. Now on `FtsField` and recorded in the catalog record — required, not
optional, because the flag joins the options-hash, so losing it on reopen would fail the
table's own hash check rather than merely forget how to answer a phrase.

**A pre-existing bug in `ascii_lower`, independent of this feature.** Its zero-copy
`tokenize_each_query` override numbered emitted tokens consecutively, but the tokenizer
drops non-ASCII runs and the index side has always left a hole for one. So a phrase query
carrying a non-ASCII word asked for its surviving words closer together than they were
indexed: **`"new café york"` could never match the text it was copied from.** Fixed, and
both query entry points now share one scan so they cannot disagree about which tokens a
query has.

## New dependency

`rust-stemmers` 1.2 — the de-facto Rust Snowball port, dual MIT / BSD-3-Clause, and its
only transitive dependencies are `serde` + `serde_derive`, already direct dependencies
here, so the graph grows by one node rather than a subtree. Vendoring the stemmer tables
instead would put us on the hook for correctness fixes to algorithms we do not own. It is
also what the reference implementations we compare against use, so a stemming
disagreement is a genuine behaviour difference rather than a difference in Snowball ports.

`Stemmer::English` is named `english`, not `porter`, in both the API and the persisted
name, because the two are different algorithms: Snowball's `english` *is* Porter2, whereas
Lucene's `PorterStemFilter` is the original Porter.

## What the built-in filter names resolve to is frozen

A column persists the *name* `english`, not the words or the rules — so what those names
resolve to is on-disk state, and nothing was guarding it. Two silent-corruption paths were
open and are now pinned by golden tests:

- Editing `ENGLISH_STOPWORDS` would re-analyze every column already built under
  `stopwords: "english"` — add a word and queries for it start missing documents that
  contain it. The prior test only checked sortedness, dedup and a length bound, so a word
  could be added or removed and pass CI.
- Worse for the stemmer, whose rules live in `rust-stemmers` under a caret range admitting
  any `1.x`: a routine dependabot bump changing one Porter2 rule would change how an
  existing column *would* tokenize while its postings kept the old stems.

Both failure messages say what the failure means and what to do — never re-record the
expectation; reject the change, or ship a new name (`english2`) with its own frozen list.
The stemmer range stays a caret on purpose: what needs guarding is the crate's output, not
its version number.

## Shared test harness, and one rename for safety

Three test modules had grown identical copies of the closures that own a parsed query's
clauses and hand them to the oracle — twenty-odd lines each, with a fourth copy due with
the next oracle. They now share `oracle_top_k_atoms`, which also puts the "parse with the
same tokenizer the reader uses" invariant in one place instead of three. The chain
oracle's superfile fixture was a copy of the shared builder too; that builder now takes an
`FtsConfig` and the old entry point delegates.

`PhraseCursor::offsets` is renamed `position_offsets`, because it sat two fields from
`run_offsets` and `cached_run_offset` — which are **byte** offsets into the positions
blob. Two meanings of "offset" in one struct family, where confusing them is a correctness
bug rather than a type error.

## Correctness

New oracle (`tests/superfile/fts/analysis_chain.rs`) grades **four chains × fifteen
queries × both default operators** against the textbook BM25 scorer, on **scores as well
as match sets**. Scores are the load-bearing half: removing a stopword shortens the
document and so moves every score in the column through the length normalizer, so an
engine and oracle that counted length differently would keep identical match sets and rank
plausibly and wrongly.

Two more tests: phrase holes asserted against **corpus truth** rather than the oracle, so a
hole bug affecting both implementations alike still fails; and stemming's symmetry across
inflections plus its silence on irregular forms (`ran` stays its own term).

Then the boundaries the field-based persistence introduces, each of which fails silently if
broken:

- absent filter fields mean off, and an unrecognized *value* is refused — asserted on the
  deserializer, since neither case is reachable through our own writer;
- a catalog record naming an unresolvable filter refuses the table;
- the wire emits the filter keys only when set;
- the filters join the options-hash, while a filterless table's hash stays byte-identical
  to what it was before filters existed.

And the paths whose failure would be silent regardless of representation: a rebuild carries
the whole analysis (compaction would otherwise re-tokenize unfiltered), the `positions`
flag round-trips the catalog (it joins the options-hash, so losing it makes a positional
table fail its own hash check), a positionless column rejects a phrase naming the column,
and the three `FtsField` analysis setters commute.

`BruteForceBm25` had to change to grade any of this — it stored a dense `tokens:
Vec<String>` and scanned it for *contiguous* phrase occurrences, so a hole the engine left
would not exist for it. Now `Vec<(String, u64)>` walked by position, which reduces to the
old contiguity check when both sides are dense.

## Verification

Green: `make check` (fmt, both clippy lanes, api-parity, version-sync, doc-check),
`make doctest`, `make test-remote` (17), `cargo test --lib` (2498),
`--test superfile` (193, both fuzz oracles), `--test supertable` (337),
`supertable_concurrent_processes` (5), and the remaining standalone targets.

`public-api.txt` regenerated; the whole diff is additive — two `#[non_exhaustive]` enums
and three `FtsField` setters. The attribute goes on now because a second language is the
obvious next addition and adding an enum variant is otherwise a breaking change; callers
only ever construct these, never match on them.

Not run here: `make coverage`, `make miri` / `make asan`.

**`supertable_commit_crash_localfs` could not be graded on this machine.** The test leaks
its aborting children, so `cargo test` never returns even after assertions pass. Orphan
batches of those same children were already present from 1d21h and 22h before this branch
existed, alongside 21 orphan `rustfs` daemons from unrelated runs, so nothing here can be
attributed to this diff either way. Worth a clean run in CI.

## Performance

Same-host A/B from this PR's benchmark lanes: the PR ref and base `main` are built and run
side by side on one VM, so the comparison is against a `main` arm measured on the same
hardware in the same run. Verdict line:

```
gate=warm p90, 0 blocking, 0 informational, 11 advisory regressions,
8 improvements, 0 failure line(s), baseline=yes
```

### Ingest — flat on default columns

| build (superfile FTS, 1M docs) | main | PR | ratio |
| --- | --- | --- | --- |
| 1 writer | 15.28 s | 15.37 s | 1.006 |
| 7 writers | 3.85 s | 3.84 s | 0.997 |
| 1 writer (positions) | 27.45 s | 28.82 s | **1.050** |
| 7 writers (positions) | 5.69 s | 5.64 s | 0.991 |

The default (unfiltered) path is flat — +0.6% and −0.3% — which is the number that mattered
most here. FTS tokenizer ingest cost on this codebase is architecture-sensitive: a change
has previously measured ~5% on ARM and ~29% on an x86 CI host, so a local reading proves
nothing and this lane is the one that counts.

That is also what the diff predicts. The analysis chain adds one
`else if let Some(chain)` arm *after* the two existing downcast arms in each `add_doc`
path, and it is `None` for any column that declares no filter; the downcast itself is once
per doc per column, not per token.

**The +5.0% row is not claimed as clean.** Single-writer positional ingest sits exactly at
the 5% advisory threshold, and positions is the path this diff touches most. Against it
being systematic: the 7-writer positional build moved the *other* way (−0.9%) on the same
code and host, and each cell is one sample per arm. Worth a second reading rather than an
explanation.

### Warm query p90 — 15 of 23 cells faster, geomean 0.983

Nothing is near the gate. Slowest: `must_rare_should_common` 1.110 (283 → 314 µs),
`three_similar_or` 1.089, `two_term_and` 1.087, `single_df1` 1.080 (a 1 µs cell),
`two_term_or` 1.077 — all sub-millisecond or ~6 ms, inside the ±5–15% within-arm spread
these shapes show on repeated runs of identical code.

**All four phrase cells are at or below parity** — `phrase_two_mixed` 0.874,
`phrase_two_common` 0.935, `phrase_three_common` 0.959, `phrase_plus_must_term` 1.002 —
which is the specific risk in this diff: `PhraseCursor::verify_at_aligned` now reads an
indexed offset where it cast a loop index, once per member per aligned doc rather than per
position. No measurable cost.

### Two honest limits on the above

**The gate is weak, so its green is not the evidence.** It blocks only on warm p90 and only
past **>50% AND >5 ms**; build time is classified advisory and never blocks. A 30%
regression would pass. The numbers above are read from the two arms directly, not inferred
from the verdict. The **11 advisory regressions** are counted but not listed in the job log,
and the summary artifact was not retrievable (empty check-run output, no PR comment posted),
so I could not reconcile my extraction against the tool's own classification.

**An earlier comment on this PR is superseded.** It reported a reproducible 43% improvement
on `two_term_and_small` that I could not explain and explicitly declined to claim. On this
lane that cell is **1.057** — slightly slower. So it was a code-layout artifact of the local
build, not a real effect, and nothing in this PR should be credited with it.

### Cross-engine

A separate same-box cross-engine run put the branch's index within **16 bytes of main's on
4.13 GB**, confirming the byte-identity claim for a column with no filter. Its latency
buckets were mostly unmeasurable — the same-code control drifted 24–44% on COUNT — with one
bucket (TOP_100, +7–12%) clearing its control uniformly across every query shape, phrases
least affected. That shape is inconsistent with the only query-path change here, so it is
being re-run rather than attributed.

### DIFF
--- a/infino-node/src/lib.rs
+++ b/infino-node/src/lib.rs
@@ -48,7 +48,8 @@ use datafusion::execution::context::SessionContext;
 use datafusion::logical_expr::Expr;
 use infino::{
     Bm25SearchOptions, Bm25Stats, BoolMode, ColdFetchMode, CompactionSettings, GcError,
-    InfinoError, Metric, OptimizeError, OptimizeOptions as InfinoOptimizeOptions,
+    InfinoError, Metric, OptimizeError, OptimizeOptions as InfinoOptimizeOptions, Stemmer,
+    Stopwords,
 };
 
 // ---------------------------------------------------------------------------
@@ -129,6 +130,29 @@ fn metric_from_str(s: &str) -> Result<Metric> {
         other => Err(Error::new(
             Status::InvalidArg,
             format!("unknown metric {other:?}; use 'cosine', 'l2sq', or 'negdot'"),
+        )),
+    }
+}
+
+/// Parse a stopword-set name. Named built-in sets only; the error
+/// names the valid set rather than leaving the caller to guess.
+fn stopwords_from_name(s: &str) -> Result<Stopwords> {
+    match s.to_ascii_lowercase().as_str() {
+        "english" => Ok(Stopwords::English),
+        other => Err(Error::new(
+            Status::InvalidArg,
+            format!("unknown stopwords {other:?}; use 'english'"),
+        )),
+    }
+}
+
+/// Parse a stemmer name. Same shape as [`stopwords_from_name`].
+fn stemmer_from_name(s: &str) -> Result<Stemmer> {
+    match s.to_ascii_lowercase().as_str() {
+        "english" => Ok(Stemmer::English),
+        other => Err(Error::new(
+            Status::InvalidArg,
+            format!("unknown stemmer {other:?}; use 'english'"),
         )),
     }
 }
@@ -519,10 +543,19 @@ pub struct VectorFilter {
 #[napi]
 #[derive(Clone, Default)]
 pub struct IndexSpec {
-    /// `(column, analyzer, stored)`; `analyzer` `None` means the default.
-    fts: Vec<(String, Option<String>, bool, Option<f64>, Option<f64>)>,
+    /// One declared FTS column, as its options arrived.
+    fts: Vec<FtsDecl>,
     /// `(column, dim, metric)`.
     vectors: Vec<(String, u32, String)>,
+}
+
+/// One declared FTS column: the column name plus the options given for
+/// it, with the `Option`s still meaning "not given" so each falls back
+/// to the engine's own default rather than one restated here.
+#[derive(Clone)]
+struct FtsDecl {
+    column: String,
+    options: FtsOptions,
 }
 
 /// Per-column FTS options for `IndexSpec.fts`.
@@ -534,6 +567,33 @@ pub struct FtsOptions {
     /// split + lowercase, non-ASCII dropped). It is recorded with the
     /// table and cannot be changed afterwards.
     pub analyzer: Option<String>,
+    /// Remove this column's stopwords — the very common words whose
+    /// presence says almost nothing about what a document is about.
+    /// `"english"` is the only set; omit for none (the default).
+    ///
+    /// Applies to both sides: the words leave the index and they leave
+    /// a query, and each one removed leaves a hole in the token
+    /// positions, so an exact phrase still knows the words it matched
+    /// were not adjacent in the text. The trade is that once a word is
+    /// not indexed, no query can find it — declare it on prose, not on
+    /// short identifiers. Recorded with the table and unchangeable
+    /// afterwards, since it decides what is in the index.
+    pub stopwords: Option<String>,
+    /// Reduce this column's words to their stems, so a search for one
+    /// inflection finds the others (`running`, `runs`, `run`).
+    /// `"english"` is the only stemmer; omit for none (the default).
+    ///
+    /// Applies to both sides, like `stopwords`. The trade is precision:
+    /// stemming conflates words a reader would not, and there is no way
+    /// to ask for an unstemmed form on a stemmed column. Recorded with
+    /// the table and unchangeable afterwards.
+    pub stemmer: Option<String>,
+    /// Record token positions, which is what exact phrase queries
+    /// (`'"climate policy"'`) need. Default false: positions roughly
+    /// double the column's index footprint, so they are a per-column
+    /// opt-in. A column without them answers a phrase query with an
+    /// error naming the column, never a silent bag-of-words fallback.
+    pub positions: Option<bool>,
     /// Keep the raw text in the table (default true). `false` makes the
     /// column index-only: searchable, but the text is never stored, so
     /// it cannot be selected, projected, or filtered on (append/update
@@ -556,19 +616,16 @@ impl IndexSpec {
         Self::default()
     }
 
-    /// Mark `column` (a UTF-8 string column) as full-text indexed,
-    /// with optional per-column `options` (analyzer, stored, k1/b).
+    /// Mark `column` (a UTF-8 string column) as full-text indexed, with
+    /// optional per-column `options` (analyzer, stopwords, stemmer,
+    /// positions, stored, k1/b).
     #[napi]
     pub fn fts(&self, column: String, options: Option<FtsOptions>) -> Self {
         let mut next = self.clone();
-        let opts = options.unwrap_or_default();
-        next.fts.push((
+        next.fts.push(FtsDecl {
             column,
-            opts.analyzer,
-            opts.stored.unwrap_or(true),
-            opts.k1,
-            opts.b,
-        ));
+            options: options.unwrap_or_default(),
+        });
         next
     }
 
@@ -587,10 +644,27 @@ impl IndexSpec {
     /// Lower to the core `IndexSpec` builder.
     fn to_rust(&self) -> Result<infino::IndexSpec> {
         let mut spec = infino::IndexSpec::new();
-        for (column, analyzer, stored, k1, b) in &self.fts {
-            let mut field = infino::FtsField::new(column.clone()).stored(*stored);
+        for FtsDecl { column, options } in &self.fts {
+            let FtsOptions {
+                analyzer,
+                stopwords,
+                stemmer,
+                positions,
+                stored,
+                k1,
+                b,
+            } = options;
+            let mut field = infino::FtsField::new(column.clone())
+                .positions(positions.unwrap_or(false))
+                .stored(stored.unwrap_or(true));
             if let Some(a) = analyzer {
                 field = field.analyzer(a.clone());
+            }
+            if let Some(name) = stopwords {
+                field = field.stopwords(stopwords_from_name(name)?);
+            }
+            if let Some(name) = stemmer {
+                field = field.stemmer(stemmer_from_name(name)?);
             }
             // Both or neither: the two parameters interact through the
             // length norm, so half-overriding is a footgun.
--- a/infino-node/src/lib.rs
+++ b/infino-node/src/lib.rs
@@ -577,7 +577,10 @@ pub struct FtsOptions {
     /// were not adjacent in the text. The trade is that once a word is
     /// not indexed, no query can find it — declare it on prose, not on
     /// short identifiers. Recorded with the table and unchangeable
-    /// afterwards, since it decides what is in the index.
+    /// afterwards, since it decides what is in the index: the words it
+    /// removed were never written, so changing it means re-ingesting
+    /// from the source text. With `stored: false` that text is never
+    /// kept, so the combination is permanent.
     pub stopwords: Option<String>,
     /// Reduce this column's words to their stems, so a search for one
     /// inflection finds the others (`running`, `runs`, `run`).
@@ -586,7 +589,11 @@ pub struct FtsOptions {
     /// Applies to both sides, like `stopwords`. The trade is precision:
     /// stemming conflates words a reader would not, and there is no way
     /// to ask for an unstemmed form on a stemmed column. Recorded with
-    /// the table and unchangeable afterwards.
+    /// the table and unchangeable afterwards: a stem is not invertible
+    /// — the index holds `run`, never the `running` it came from — so
+    /// changing it means re-ingesting from the source text, and with
+    /// `stored: false` that text is never kept, making the combination
+    /// permanent.
     pub stemmer: Option<String>,
     /// Record token positions, which is what exact phrase queries
     /// (`'"climate policy"'`) need. Default false: positions roughly
--- a/infino-node/src/lib.rs
+++ b/infino-node/src/lib.rs
@@ -134,27 +134,41 @@ fn metric_from_str(s: &str) -> Result<Metric> {
     }
 }
 
-/// Parse a stopword-set name. Named built-in sets only; the error
-/// names the valid set rather than leaving the caller to guess.
+/// Parse a stopword-set name. Named built-in sets only; the error names
+/// the valid set rather than leaving the caller to guess.
+///
+/// **Exact match, deliberately not case-folded.** These names are format
+/// vocabulary: the engine's own resolver is exact, and a column persists
+/// the name it was given. Accepting `"English"` here would mean this
+/// binding has a wider vocabulary than the format and must remember to
+/// normalize before persisting — a coupling that is one forgotten call
+/// away from writing a file the reader refuses. It would also diverge
+/// from the python binding, so the same call would work in one and throw
+/// in the other.
+///
+/// The asymmetry decides it: accepting more spellings later is additive,
+/// while tightening later breaks every caller who relied on the loose
+/// one. (The older `metric` argument *is* case-folded, which is an
+/// inconsistency in the other direction — worth reconciling, but not by
+/// widening a new surface to match an old one.)
 fn stopwords_from_name(s: &str) -> Result<Stopwords> {
-    match s.to_ascii_lowercase().as_str() {
-        "english" => Ok(Stopwords::English),
-        other => Err(Error::new(
+    Stopwords::from_name(s).ok_or_else(|| {
+        Error::new(
             Status::InvalidArg,
-            format!("unknown stopwords {other:?}; use 'english'"),
-        )),
-    }
+            format!("unknown stopwords {s:?} (valid: \"english\")"),
+        )
+    })
 }
 
-/// Parse a stemmer name. Same shape as [`stopwords_from_name`].
+/// Parse a stemmer name. Same exact-match rule as
+/// [`stopwords_from_name`], and for the same reasons.
 fn stemmer_from_name(s: &str) -> Result<Stemmer> {
-    match s.to_ascii_lowercase().as_str() {
-        "english" => Ok(Stemmer::English),
-        other => Err(Error::new(
+    Stemmer::from_name(s).ok_or_else(|| {
+        Error::new(
             Status::InvalidArg,
-            format!("unknown stemmer {other:?}; use 'english'"),
-        )),
-    }
+            format!("unknown stemmer {s:?} (valid: \"english\")"),
+        )
+    })
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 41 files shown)


vinaykakade:
## Bench: `superfile fts`, 1M docs, base ↔ branch

Method: alternating base/branch, 3 reps each, so machine drift hits both arms alike rather
than only whichever ran second. Both binaries verified as the commits they claim (the
branch arm links `rust-stemmers`; the base has zero such symbols). Host: aarch64, 16C.
Ratio is median-of-3 branch ÷ median-of-3 base; lower is better.

**Noise band** is the max/min spread across the three reps of the *same* binary — the
measurement floor. A ratio inside it says nothing.

### Warm query p50 — 22 of 23 cells inside their own noise band

| Cell | base p50 | ratio | noise band | verdict |
| --- | --- | --- | --- | --- |
| single_rare | 7 µs | 1.011 | ±3% | within |
| single_df1 | 1 µs | 0.855 | ±36% | within |
| single_common | 27 µs | 1.394 | ±56% | within |
| two_term_or | 625 µs | 0.975 | ±14% | within |
| three_wide_or | 1.72 ms | 1.012 | ±9% | within |
| three_similar_or | 7.26 ms | 1.000 | ±6% | within |
| five_term_or | 11.15 ms | 1.001 | ±4% | within |
| ten_term_or | 20.33 ms | 1.000 | ±7% | within |
| twenty_term_or | 51.92 ms | 1.015 | ±3% | within |
| forty_term_or | 109.29 ms | 1.030 | ±7% | within |
| two_term_and | 1.40 ms | 0.971 | ±15% | within |
| **two_term_and_small** | **297 µs** | **0.570** | **±14%** | **outside — see below** |
| three_wide_and | 4.26 ms | 0.977 | ±10% | within |
| three_similar_and | 7.72 ms | 0.978 | ±11% | within |
| five_term_and | 9.59 ms | 0.983 | ±8% | within |
| ten_term_and | 13.03 ms | 0.964 | ±7% | within |
| must_common_should_common | 1.15 ms | 1.052 | ±13% | within |
| must_rare_should_common | 370 µs | 1.061 | ±18% | within |
| must_two_should_two | 10.97 ms | 0.975 | ±7% | within |
| phrase_two_common | 192.20 ms | 0.987 | ±7% | within |
| phrase_two_mixed | 11.33 ms | 0.910 | ±9% | within |
| phrase_three_common | 221.40 ms | 0.937 | ±8% | within |
| phrase_plus_must_term | 31.37 ms | 0.979 | ±6% | within |

Geomean 0.975 · min 0.570 · max 1.394. The two extreme ratios (`single_common` 1.394,
`single_df1` 0.855) are 27 µs and 1 µs cells whose own within-arm spread is ±56% and ±36%
— nothing is measurable there at this rep count.

**The phrase cells are the ones this change was most likely to cost**, since
`verify_at_aligned` now reads an indexed offset where it used to cast a loop index. All
four are ≤ 1.0 (0.910–0.987). No measurable cost.

### Ingest — build wall time

| Cell | base | branch | ratio |
| --- | --- | --- | --- |
| 1 writer | 15.98 s | 15.95 s | 0.998 |
| 16 writers | 2.40 s | 2.45 s | 1.021 |
| 1 writer (positions) | 21.14 s | 21.59 s | 1.021 |
| 16 writers (positions) | 4.04 s | 4.11 s | 1.017 |

Geomean 1.014.

**This is the number I would not trust from this host, and it should be read off CI
instead.** FTS tokenizer ingest cost on this codebase is strongly architecture-dependent —
a change measured at ~5% locally on aarch64 has previously come in at ~29% on an x86 CI
host, so a small local delta is not evidence of a small delta on x86. Three of four cells
sit at +1.7–2.1%, consistently positive rather than straddling zero, which is the shape
that warrants the CI reading rather than dismissal. The `bench (superfile, 1000000, sf)`
lane on this PR grades it on the CI host; I'd treat that as the gate, not this table.

Mechanically the ingest path should be unchanged for any column that declares no filter:
the chain adds one `else if let Some(chain)` arm *after* the two existing downcast arms,
and it is `None` for such a column; the downcast itself is once per doc per column, not
per token.

### The one outlier, unexplained

`two_term_and_small` is a reproducible **43% improvement**, not noise — base 298/295/297 µs
against branch 169/171/151 µs, tight within each arm across three reps.

I could not explain it, and I am not claiming it. My hypothesis was that `PhraseCursor`
gaining a `Vec<u32>` grew the `AnyCursor` enum that term cursors also live in, changing
layout for the atom walks. **That is refuted by measurement**: `TermCursor` is 168 B and
`PhraseCursor` 120 B, so `AnyCursor` is 168 B either way and the new field did not grow it.
`TermCursor` is untouched by this diff.

The residual hypothesis is a code-layout/inlining shift from the new module and the new
crate edge — this scoring path has shown alignment sensitivity of this magnitude before,
which is why PGO keeps coming up for it. Unconfirmed. Since it is an improvement on one
sub-millisecond cell it does not block, but it should not be counted as a win either, and
if a reviewer wants it chased down I am happy to do that rather than leave it as folklore.

### Not run locally

`make coverage`, `make miri`, `make asan`. And `supertable_commit_crash_localfs` could not
be graded on this machine — see the PR description.

infino-node/src/lib.rs
```diff
@@ -133,6 +134,29 @@ fn metric_from_str(s: &str) -> Result<Metric> {
     }
 }
 
+/// Parse a stopword-set name. Named built-in sets only; the error
+/// names the valid set rather than leaving the caller to guess.
+fn stopwords_from_name(s: &str) -> Result<Stopwords> {
+    match s.to_ascii_lowercase().as_str() {
```
pratyushlokhande: Node and Python disagree on the filter names. node package lowercases the argument; python doesn't. Built both from this branch and called them:

stopwords="English"   node: ACCEPTED    python: ValueError: unknown stopwords "English"
stopwords="ENGLISH"   node: ACCEPTED    python: ValueError

Node normalizes before persisting (fts_stopwords: ['english']), so nothing corrupt lands on disk — but it's the same call working in one binding and throwing in the other. Python is also inconsistent with itself: metric="Cosine" is accepted two functions away. Suggest exact-match in both, matching core Stopwords::from_name, which your own test pins as rejecting "English". Note the resolvers fire in to_rust() at create_table, not at fts() — that's where a test goes. Neither binding has one.

infino-python/src/lib.rs
```diff
@@ -254,17 +312,39 @@ impl IndexSpec {
...
+        analyzer = None,
+        stopwords = None,
+        stemmer = None,
+        positions = false,
+        stored = true,
+        k1 = None,
+        b = None,
+    ))]
+    #[allow(clippy::too_many_arguments)]
     fn fts(
```
pratyushlokhande, line 333: Python fts() positional signature changed. stopwords, stemmer, positions are inserted before stored, k1, b. Built both modules and compared:

main sig: (column, analyzer=None, stored=True, k1=None, b=None)
PR   sig: (column, analyzer=None, stopwords=None, stemmer=None, positions=False, stored=True, k1=None, b=None)

fts("body", "standard", False)             main: OK   PR: TypeError: argument 'stopwords': 'bool' object is not an instance of 'str'
fts("body", "standard", True, 1.2, 0.75)   main: OK   PR: TypeError

Fails loudly, never silently — the new slot is Option<String>, so a bool/float can't slide in. Every caller in this repo passes keywords past column, so CI won't catch it; only downstream positional callers break. Either append the three after b, or note it as a breaking change in the release notes.

pratyushlokhande (approved):
Couple of comments around drift b/w node and python package. Everything else looks good. Ok to merge post those are fixed.

infino-python/src/lib.rs
vinaykakade: Good point - done. And the new args are now keyword-only.

infino-node/src/lib.rs
vinaykakade: Good point, done

Issue #737, opened 2026-09-09
reader cache: simplify the cold-fetch path and the cache entry


naman-modi:
The reader cache works but has gotten hard to read.

Two problems:
- There are ~8 functions that all do the same fetch sequence (get bytes, write to disk, rename, open, insert) with minor differences. The crash-safety steps are copy-pasted across all of them.
- A cached entry has 5 optional fields (mmap, block source, token, fill flag, accounting) that you decode by hand everywhere to figure out which of 2 states it's in.

Proposal (behaviour-preserving, no functional change):
- Collapse the fetch functions into one `finalize` that owns the write -> rename -> mmap -> open -> insert tail. The rest just says how it gets the bytes.
- Replace the optional fields with one enum: an entry is either `Resident` (whole file mmapped) or `Lazy` (block-backed). Then `reader()` reads top to bottom as check-memory, check-disk, else-fetch.

Existing `disk_cache` tests are the contract and should pass unchanged. Bench flat.

PR #755, opened 2026-09-10
refactor disk reader_cache: merge cache entry states into a Residency enum

naman-modi:
### What does this PR do?

This PR replaces a cache entry's five loosely-coupled state fields with one `Residency` enum, without changing behavior. An entry is now explicitly one of `Buffered`, `Mapped`, or `Paged`, instead of a combination of optional fields. Part of #737.

The entry's real state, whether its bytes are in a heap buffer, lazily block-backed, or fully mmapped, was never written down. It lived as an implicit combination of `mmap`, `block_source`, `block_token`, and `fill_spawned`. Nothing stopped an illegal combination like a mmapped entry that also claimed to be source-owned, so the set of valid states existed only in people's heads.

### How do we do this change?

- `Residency` names the three states: `Buffered` (whole file in a heap buffer), `Mapped` (whole file mmapped, optionally keeping the large vector blob on the block cache), and `Paged` (byte ranges fetched on demand from object storage). The mmap, block source, and fill latch each move inside the variant they belong to, so an illegal combination no longer type-checks.

### Notes

- Behavior-preserving: the existing `disk_cache` suite is the contract and CI stays green with no assertion edits.

### DIFF
--- a/src/supertable/reader_cache/block_source.rs
+++ b/src/supertable/reader_cache/block_source.rs
@@ -164,9 +164,10 @@ impl BlockCachedSource {
         }
     }
 
-    /// Identity token installed on the cache entry that owns this source.
-    pub(crate) fn entry_token(&self) -> Arc<()> {
-        Arc::clone(&self.entry_token)
+    /// Whether `token` is this source's own identity token, compared by pointer. Lets the owning
+    /// cache entry check "is this source still current" without cloning the `Arc`.
+    pub(crate) fn owns_token(&self, token: &Arc<()>) -> bool {
+        Arc::ptr_eq(&self.entry_token, token)
     }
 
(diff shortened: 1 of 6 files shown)


Pranav2612000 (approved):

Pranav2612000:
@savannahar68 correct me if I'm wrong, but will vector queries be served by the "Paged" mode ( because we don't mmap the vector blocks )? And some of these fetches will be through the disk ( BlockFile ), not the object store necessarily? If yes, these definitions may need to be modified

> Residency names the three states: Buffered (whole file in a heap buffer), Mapped (whole file mmapped, optionally keeping the large vector blob on the block cache), and Paged (byte ranges fetched on demand from object storage). The mmap, block source, and fill latch each move inside the variant they belong to, so an illegal combination no longer type-checks.

PR #756, opened 2026-09-10
fts: correct BM25 corpus statistics, score scale, and tokenization

vinaykakade:
A review of the BM25 scorer against the model it is supposed to implement turned up four scoring divergences and two tokenization gaps. Three of the four are the same defect wearing different hats — corpus statistics computed over **rows** where BM25 is defined over **documents that have the field** — so the change is organized by mechanism rather than by symptom.

## What was wrong

**Statistics counted rows, not documents.** A row whose text cell is null is indexed as the empty string so the doc-lengths array stays aligned with Parquet, and both the average document length and the collection size counted those empty slots. Dividing the token total by the row count deflates the average by the column's fill rate, and the average sits under a saturating ratio, so a sparse column compresses term-frequency differences and over-rewards short documents. Using the row count as the collection size is worse: the document frequency it is compared against is *not* rescaled with it, so the column weights its common terms too heavily against its rare ones.

**The table-wide average was never table-wide.** Global statistics globalized the term weight but left the length normalizer per-superfile, so the same document scored differently depending on which commit it landed in — the one thing that mode exists to rule out. The quality bench has been paying for this with a `residual_tolerance` allowance since it was written.

**Scores carried a `(k1 + 1)` factor.** Constant across a query, so ranking-neutral, but every published score was a fixed 2.2x (at the default `k1`) of the conventional form of BM25 for the same `k1` and `b` — wrong for a score threshold, a weighted fusion against vector distances, or any comparison of absolute scores.

**The standard analyzer dropped emoji**, and **token length was unbounded.** The second is a scoring bug, not just a resource one: length is a token count, so an unbroken 100 KB run made a document of length 1 — the shortest document possible, taking the largest length boost the formula can award, while being the longest document in the corpus.

## How

**Statistics over documents.** The reader reads the exact per-doc length array in full to quantize it, so summing the lengths and counting the non-empty ones costs no extra read; the builder divides by the same count and bakes idf with it. A file written before this gets corrected statistics the next time it is opened, with no reindex.

**The average is baked at write time, table-wide.** Each superfile's `(token total, documents with tokens)` rides `FtsSummaryAgg` on the manifest as a tail append compatible in both directions. At commit the writer folds those over the current manifest (`ManifestSnapshot::fts_corpus_stats`) and hands them to the builder (`BuilderOptions::fts_corpus_stats` → `FtsBuilder::set_corpus_length_stats`), so every new superfile — buffered append, WAL update, packed/hidden-cell build, and compaction merge (over everything the merge does not replace, plus itself) — declares the running table-wide average as of its commit, rounded once through `bm25::stored_avgdl`, the same helper the reader uses. The reader scores a current-version file at the average it declares; its stored bounds are exact `f32` maxima at that average and need no correction. The table-wide collection size for idf is the same fold, at query time.

**Why write time and not query time.** A stored bound is a score frozen at the average the build divided by, and a common term's block maxima span ~2%, so scoring at any other average needs an inflation that admits most blocks (measured: 1% → 7x the blocks; the query-time version of this change doubled `single_common` latency in four CI runs). Recomputing the bound from stored `(tf, length)` points does not fit either: on a bench-shaped corpus every 128-doc block of a common term has ≥ 7 competitive points scoring within a fraction of a percent of each other, so two points in the existing 4-byte slot were 1.3–4% loose even at baked statistics and the 8-byte four-point side table this PR briefly carried (+2.2% index) covered about half the blocks. Baking the running table-wide average costs nothing at query time and nothing on disk. What it gives up: a table's first superfiles declare an average over less data, converging as the table grows and re-baked at the then-current table-wide value by every compaction that touches them, so a fragmented table matches a single superfile exactly for the last commit's documents and converges for the earlier ones — which is what the oracle test now asserts.

**One decode site.** `reader/bounds.rs` (`StoredBound`, `BoundDecoder`) interprets a blob's block-max slot by version for both the single-term walk and the multi-term cursors: a current file's exact `f32`; a V5 file's `f32` carrying `(k1 + 1)`, at its row-count average, corrected on open by the supremum factor and `1/(k1 + 1)`; a pre-coarse file's fixed point likewise. The term cursor takes its `ColumnMeta` instead of eleven arguments. The query-time average override is gone with the mechanism that needed it; `k1`/`b` overrides remain an opt-in that pays a supremum factor, as before, memoized per pair on the superfile reader.

**The score scale is version-stamped.** Dropping `(k1 + 1)` changes what a stored bound means, so the blob version goes to **V6** — the V5 layout, differing in the bound's scale and in what the declared average is. Older blobs keep their pruning power rather than merely staying correct. A test-only `BlobEra` lets the builder write faithful V5 and pre-coarse files, so the compatibility tests exercise the legacy read paths end to end (sparse V5 top-k parity with a fresh file, override composition on an older file) instead of stamping a version onto current bytes.

## Coverage

Every corpus the oracles and fuzz tests build was dense, so this whole class of defect was untestable; **all 118 FTS integration tests passed unchanged after the scoring fix**. Added: sparse columns at the superfile and table level (null rows change neither idf nor the average), the running average across commits (append and update paths), the manifest fold and its unknown-contributor rule, the fixed-point rounding shared by builder and reader (including an average the fixed point cannot represent exactly), bound exactness at declared statistics on a realistic block distribution, the decoder table, faithful legacy files, the memo, emoji as a phrase separator at the index level, the token cap's chop semantics and its effect on document length, and the manifest tail in both directions of skew.

## Compatibility and behaviour changes

- Existing superfiles are readable unchanged and need no reindex.
- **BM25 scores and rankings shift** on sparse columns and on fragmented tables. That is the correction; it belongs in the release note.
- **Absolute scores drop by a factor of `k1 + 1`** (2.2 at the default). Ranking-neutral.
- The analyzer changes affect what a **re-ingest** produces, not existing files.
- Public API surface unchanged (`make public-api` clean); the `superfile` module is public only under `test-helpers`. Nothing in the published surface breaks.
- **Release line: 0.9.0, coordinated across crate, Node and Python** (`release_prep.py --package all --version 0.9.0` after merge). Not for an API break — for the behaviour a release consumer receives: scores in a new scale, rankings shifted on sparse and fragmented tables, a new FTS blob version older readers cannot open, and analyzers that tokenize a re-ingest differently. The version bump is the release commit's, not this PR's.
- A V5 blob written by the released 0.8.1 builder is checked in as a fixture (`tests/fixtures/`); the current reader must rank it exactly like a fresh build, and the test-only V5 writer must reproduce it byte for byte.

## Measured

CI two-arm bench on `07981b51` ([run 34704616419](https://github.com/infino-ai/infino/actions/runs/34704616419)): main and PR built and run on one VM, 1M docs, realistic corpus, 16 commits, 75 superfiles. **Gate (warm p90): passed — 0 blocking, 4 advisory regressions, 9 improvements.** A local M4 run gave the same quality tables row for row.

**Index size 5.38 GiB on both arms.** Ingest 92.4 → 98.3 s (10.8 → 10.2 K docs/s); peak RSS 1.48 → 1.52 GiB.

**Quality**, `recall vs engine BM25` 1.0000 on every shape in both arms; worst `max score Δ` against the corpus-wide oracle, main → PR:

| k = 10 | k = 100 | k = 1000 |
|---|---|---|
| 5.00% → **1.06%** | 5.31% → **1.06%** | 5.31% → **1.06%** |

`recall vs BM25` at k = 1000: ten_common_or 0.876 → 0.908, three_stopword_or 0.873 → 0.898, phrase_stopwords 0.960 → 0.976, stopword 0.938 → 0.949, must_two_should_two 0.950 → 0.961; no shape down. At k = 10 three shapes move by one hit (two up, one down).

**Warm latency**, pre-compact, p50 / p90, main → PR:

| shape | main | PR |
|---|---|---|
| `single_common` | 210 / 232 µs | 211 / 237 µs |
| `single_rare` | 207 / 270 µs | 206 / 228 µs |
| `two_term_or` | 361 / 399 µs | 376 / 419 µs |
| `three_wide_or` | 521 / 539 µs | 483 / 516 µs |
| `ten_term_or` | 2.23 / 2.49 ms | 2.25 / 2.50 ms |
| `forty_term_or` | 9.57 / 10.00 ms | 9.24 / 9.89 ms |
| `two_term_and` | 335 / 368 µs | 332 / 385 µs |
| `three_wide_and` | 796 / 1010 µs | 891 / 924 µs |
| `must_common_should_common` | 348 / 371 µs | 340 / 376 µs |
| `must_two_should_two` | 1.17 / 1.52 ms | 1.21 / 1.48 ms |
| `phrase_two_common` | 20.96 / 22.71 ms | 20.38 / 22.85 ms |
| `phrase_two_mixed` | 756 / 815 µs | 682 / 785 µs |

`single_common` — the shape the query-time version of this change doubled — is flat. Count and top-1000 within ±3% on every shape. Post-compact tables and the cold lane (0 blocking; 37 advisory regressions, 36 improvements) scatter several-fold in both directions in both arms, with GET counts and bytes unchanged on most shapes.

**SBG same-box branch-vs-main** ([run 34703590051](https://github.com/infino-ai/search-benchmark-game/actions/runs/34703590051), branch at `45494c1a`, 962 queries, one c7i.2xlarge, page metric = mean of per-query min of 10; ms):

| metric | main | branch | Δ | lucene 10.5 |
|---|---|---|---|---|
| TOP_10 | 1.101 | 1.039 | −5.6% | 0.891 |
| TOP_100 | 1.345 | 1.502 | **+11.7%** | 1.237 |
| TOP_1000 | 2.016 | 2.021 | +0.3% | 1.797 |
| TOP_100_COUNT | 2.440 | 2.328 | −4.6% | 3.948 |
| COUNT | 0.919 | 0.918 | −0.1% | 1.181 |

Index 3.848 GiB both arms. The TOP_100 delta held in both of the branch's positions (first +11.7%, last +9.6%), concentrated in two-token intersections and unions, and absent at TOP_10 and TOP_1000. It **did not reproduce on the controlled A/B box** (c7i.2xlarge, AVX2-pinned, main `be681cad` and branch `07981b51` built with identical flags, both 5M-doc indexes rebuilt from the corpus, the 962-query set at NUM_ITER 10 run in both engine orders):

| metric | main → branch (order 1) | main → branch (order 2) | order-averaged ratio |
|---|---|---|---|
| TOP_10 | 1.239 → 1.225 ms | 1.250 → 1.232 ms | 0.987 |
| TOP_100 | 1.561 → 1.540 ms | 1.602 → 1.608 ms | 0.995 |
| TOP_1000 | 2.272 → 2.170 ms | 2.198 → 2.277 ms | 0.995 |

Per category, every (metric, category) pair is within 0.976–1.005 except the single-term query (n = 1, +4%); result counts match on all 2 886 (metric, query) pairs. The kernels in this PR differ from `main` by renames only, so the SBG run's TOP_100 gap is a measurement artifact of that run, not the change.


src/superfile/fts/tokenize.rs
```diff
@@ -60,9 +61,119 @@ const NON_ASCII_BYTE_MIN: u8 = 0x80;
...
+/// count, so an unbroken 100 KB run would otherwise be a document of
+/// length 1 — the shortest document possible, taking the largest length
+/// boost BM25 can award, while being the longest document in the
+/// corpus.
+///
+/// Counted in characters rather than bytes so the limit does not shift
+/// with the script: 255 bytes is 255 Latin characters but around 85 CJK
+/// ones, which would tokenize the same sentence differently depending
+/// on the language it is written in.
+pub const MAX_TOKEN_CHARS: usize = 255;
```
naman-modi, line 85: This 255-char cap is applied when we index (emit_capped) but I am not sure if it is applied on the query path. So if they query is more than 255 chars, it will not be tokenized in a similar way, and we would get 0 hits; would be a regression. Can we also add a test around this?

src/test_helpers/brute_force_bm25.rs
```diff
@@ -153,7 +153,13 @@ impl BruteForceBm25 {
...
         }
 
-        let n = docs.len() as u32;
+        // Corpus statistics are defined over the documents that carry
+        // tokens, not over every row handed in. A document whose text
+        // analyzes to nothing — trivially, one that is all stopwords
+        // once the chain has run — cannot match any term, so counting
+        // it would deflate the average length and inflate the
+        // collection size for every term in the column.
+        let n = docs.iter().filter(|d| d.dl > 0).count() as u32;
```
naman-modi, line 162: Mostly all the tests are very "dense". This PR is about sparse, so can we add a direct test: one corpus with every 3rd doc empty (""), fed to BOTH build_infino_superfile and BruteForceBm25::index, assert the scores match by value? Will help strengthen our test suite.

src/superfile/fts/reader/bounds.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// [`Self::Legacy`]. These blobs also predate the coarse table.
+    LegacyFixedPoint,
+}
+
+impl StoredBound {
+    pub(super) fn for_version(version: u32) -> Self {
+        match version {
+            format::fts::VERSION_V6 => Self::Exact,
+            format::fts::VERSION_V5 => Self::Legacy,
+            _ => Self::LegacyFixedPoint,
```
naman-modi: For my understanding, we will keep on updating these versions to match these enum. Like for a new VERSION_V7 - it would resolve to Exact, and then the VERSION_V6 would resolve to Legacy? I am think if this will be hard to understand rather than directly using versions itself?

One more point here; do we have enough tests that if someone later adds VERSION_V7 and does not update this section, it does not automatically resolve to the LegacyFixedPoint arm?

naman-modi (approved):
Looks good to me, post these comments.

src/superfile/fts/tokenize.rs
vinaykakade: You were right, and it was worse than one path: the cap was applied on the index side only, and all three query fast paths (both tokenizers' `tokenize_each_query`, and `ascii_lower`'s positioned one) emitted a run whole. A query token over 255 characters was looked up as a term the index never wrote — zero hits.

Fixed in f4582375 by making it structurally impossible to diverge again: one emitter (`emit_ascii_segment`) does the fold and the chop for every ASCII fast path, index and query alike; the two sides only differ in a small sink (index hands pieces to the builder without allocating, query borrows them from the query string). Pieces take consecutive positions on the query side as they do in the index.

Tests: `query_paths_chop_like_the_index_paths` (unit; both analyzers, both entry points, positions) and `tests/superfile/fts/token_cap.rs` (end to end through the superfile: a 300-char token in the document is found by the same token, its upper-case form, and each of its pieces in the query, for `ascii_lower` and `standard`).

vinaykakade: Done

src/superfile/fts/reader/bounds.rs
vinaykakade: Both points taken, in 2c6a8000.

Variants are now named by version — `StoredBound::{V6, V5, V1ToV4}` — so the arms read the same numbers the header carries, and a new version gets its own arm rather than shifting the meaning of an existing one. The builder's test-only era follows suit (`BlobEra::{V6, V5, V2ToV4}`).

On the fall-through: `for_version` is now exhaustive over the accepted versions and returns `None` for anything else, and the open path uses it as its accept list. So a `VERSION_V7` added to the format without an arm here does not decode under V1–V4 rules — it fails to open with an unsupported-version error, which the builder's own round-trip tests would hit immediately. `every_accepted_version_has_an_arm_and_the_next_one_is_refused` pins each of the six accepted versions to its arm and asserts `VERSION_V6 + 1` is refused.

vinaykakade: Done

Also, over the next week, will be migrate everyone to latest superfile version and deprecate old versions. We may need to do this exercise periodically (and deprecate older versions) to keep complexity bounded.

src/test_helpers/brute_force_bm25.rs
vinaykakade: Added in 03adf4a9: `oracle_sparse_column_scores_match_by_value` in `tests/superfile/fts/brute_force_oracle.rs`. Eight documents with a null row after every second one (plus two trailing nulls), fed to both `build_infino_superfile` and `BruteForceBm25::index`; four queries under OR and AND, match sets equal and every score within `BM25_SCORE_ABS_TOLERANCE` of the oracle's. Lengths stay below the quantizer's exact range so nothing else can hide a discrepancy — dividing by rows (18) instead of documents (8) moves every idf by more than the tolerance and fails this test.

vinaykakade: Done - oracle_sparse_column_scores_match_by_value in tests/superfile/fts/brute_force_oracle.rs

PR #760, opened 2026-09-11
vector: make the resident hnsw graph metric-aware (serve l2sq/negdot)

### LINKED ISSUE #757: Vector: hnsw_ivf serves the graph only for cosine (l2sq/negdot fall back to IVF)
## Summary

Under `search_mode = hnsw_ivf`, the resident HNSW graph is built and served only for **cosine** columns. For `l2sq` and `negdot` metrics the graph calibration declines at drain and every query silently falls back to the IVF scan path — so those metrics get IVF latency/throughput, never the graph's.

## Root cause

The HNSW graph ships a single production `NodeScorer` — `Sq16Scorer` — which decodes on the fixed `[-1, 1]` cosine grid and scores `−dot` (see `src/superfile/vector/hnsw.rs`, the module doc and `Sq16Scorer`; the alternative `Fp32Scorer` is unit-test only). The graph build path (`collect_hnsw_codes` → `Sq16Scorer::from_codes` in `src/supertable/query/vector.rs`) uses this scorer regardless of the column's metric.

For non-cosine columns the resident rerank codec is `Sq16Adaptive` (a per-cluster fitted ruler), which the IVF rerank path uses correctly. But the graph build reinterprets those adaptive codes on the fixed cosine grid and scores them as `−dot` — the wrong ruler *and* the wrong distance. The graph ends up built on mis-decoded vectors, calibration measures low recall, and the drain declines the graph (`hnsw probe: best recall below floor — graph-hostile, serving ivf`), so `hnsw_search` returns `None` and the query serves IVF.

## Observed

- An `l2sq` (euclidean) table under `hnsw_ivf` shows **flat recall and flat RPS across every `ef`** (128 → 2048) — the signature of no graph being walked; latency is the IVF scan's, not the graph's.
- Cosine tables (including 1536-dim) engage the graph normally (recall rises with `ef`), so this is metric-specific, not dimension-hardness.
- Diagnostic: `INFINO_LOG_NODES=1` prints the `hnsw_search` fallback reason and per-query layer-0 node visits.

## Fix

Add a metric-aware graph `NodeScorer`: an `Sq16Adaptive` scorer that carries the per-cluster scale/offset and scores the correct distance (`l2sq`, and `negdot`), so `Hnsw::build` / `Hnsw::search` grade against the geometry the column was created with. The graph is already generic over `NodeScorer`, so this is an additive scorer plus wiring it into the build/calibrate path for non-cosine columns — no change to the cosine path.

## Acceptance

- `l2sq` and `negdot` tables build and serve the resident graph under `hnsw_ivf`: recall rises with `ef`, and served latency is the graph's rather than the IVF scan's, meeting the recall bar.
- Existing cosine behavior is byte-for-byte unchanged.

muralikpbhat:
## Summary

Under `search_mode = hnsw_ivf` the resident HNSW graph was built and served only for **cosine** columns. `l2sq` and `negdot` columns were declined up front — the fitted-ruler `Sq16Adaptive` codec they store fails the graph's `is_sq16` gate, and both the build and serve paths carried a cosine-only metric gate — so every query on those metrics silently fell back to the IVF scan and got its latency/throughput, never the graph's.

This makes the graph metric-aware end to end. It closes #757.

## What changes

The single resident scorer now carries the column `Metric` and, for non-cosine, a fitted GLOBAL `(scale, offset)` ruler plus per-node `‖x̂‖²`, scored through the existing adaptive `Sq16Kernel`. The fixed `[-1, 1]` cosine grid stays the default (no ruler, scored `−dot`) and is byte-for-byte unchanged.

- **Build.** A non-cosine column decodes its per-cluster `Sq16Adaptive` rows to fp32 at drain (the same transient transcode the IVF merge already performs — no fp32 in the served path), fits ONE global ruler over the whole graph, re-encodes to a single-`u16` plane, and computes the norm table. The codec gate is relaxed to the fitted-ruler single-`u16` codec for non-cosine; the cosine-only `metric_supported` gate stays in place for the flat-IVF arm and the centroid router (out of scope here).
- **Calibration.** Calibration queries are synthesized in the plane's own geometry — decoded through the ruler, normalized only for cosine, with the off-node jitter scaled to the vector's magnitude for non-cosine so recall is not falsely optimistic.
- **Walk / refine.** The graph keeps its SQ8-walk / Sq16-refine shape for cosine. A fitted-ruler graph walks the metric-aware Sq16 plane directly: the SQ8 / 4-bit walk planes are code-space `−dot` proxies that assume the uniform fixed grid (one scale/offset for every dim), so on a per-dim ruler they mis-navigate — decode derives none and the serve path skips them. (A ruler-aware coarse walk is a follow-up.)
- **Serve.** The serve path rejects a graph built for a different metric than the column now declares, and publishes the metric's own distance (cosine `1 − dot`, `l2sq` squared distance, `negdot` `−dot`) so a cross-generation merge compares like with like.
- **Persistence.** A new bundle version stamps the metric, the ruler, and (l2sq) the per-node norm table after the existing sections. Cosine keeps writing the current version, so existing cosine bundles and `public-api.txt` are untouched.
- **Extend.** Incremental extend of a fitted-ruler graph is not yet wired, so a non-cosine drain rebuilds in full (correct, just not incremental); a ruler-inheriting extend is a follow-up.

## Validation (gist-960-euclidean, l2sq, 1M × 960)

Before, an `l2sq` table showed the signature of no graph being walked — **flat recall and flat rps across every `ef`**. After, recall rises monotonically with `ef` and the served latency is the graph's:

| ef | recall@10 | rps | p95 (ms) |
|----:|----:|----:|----:|
| 128 | 0.8319 | 3752 | 2.44 |
| 256 | 0.9183 | 2487 | 4.12 |
| 512 | 0.9645 | 1519 | 7.19 |
| 1024 | 0.9860 | 944 | 12.4 |
| 1536 | 0.9920 | 726 | 16.0 |
| 2048 | 0.9944 | 431 | 19.7 |

The broken baseline served the IVF scan: recall flat at 0.9872 across every `ef`, ~410 rps. The graph now clears the recall@10 ≥ 0.99 bar and is walked (thousands of rps at low `ef`). Cosine columns (including 1536-dim) are unchanged.

## Testing

- New regression tests: an `l2sq` table whose `hnsw_ivf` recall rises with `ef` (and which carries no coarse walk plane, so it walks the metric-aware Sq16 plane), and a metric-aware bundle round-trip covering `l2sq` (norm table) and `negdot`. Both fail before and pass after.
- The obsolete "hnsw declines non-cosine" assertion was updated: the hnsw arm now declines only the fixed-grid codec under a non-cosine metric (`CodecUnsupported`), while the flat arm keeps its cosine-only decline.
- `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, and the full library test suite pass locally; cosine behavior is byte-for-byte unchanged.

## CI note (second commit)

A second, self-contained commit adds `.config/nextest.toml` with `[profile.default] retries = 2`. It is unrelated to the vector change and can be split out, but it rides here because it unblocks this PR's own coverage job: the `Tests & coverage` gate recently moved to `cargo llvm-cov nextest` (process-per-test parallelism) with no nextest config, so transient cross-process contention on shared FS/storage resources — a different test each run, all in disk-cache / concurrent-writer code untouched here — was failing the whole coverage run non-deterministically. The retries report such a transient as flaky rather than failing the run; a test that fails every attempt still fails. No `test-groups` serialization for now.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,13 @@
+# nextest configuration.
+#
+# The coverage gate (`make coverage-ci`) runs the suite through
+# `cargo llvm-cov nextest`, which executes each test in its OWN process and many
+# in parallel — a different isolation model than `cargo test`'s single-process
+# threaded run. That parallelism can surface transient cross-process contention
+# on shared external resources (temp dirs, storage paths) that the threaded run
+# never exposed. Retry a failed test a couple of times so such a transient is
+# reported as FLAKY (visible, still investigable) rather than failing the whole
+# coverage run on one non-deterministic hiccup. A test that fails every attempt
+# still fails the run, so real regressions are unaffected.
+[profile.default]
+retries = 2
--- a/src/superfile/vector/hnsw.rs
+++ b/src/superfile/vector/hnsw.rs
@@ -49,9 +49,11 @@ use rayon::prelude::*;
 use crate::superfile::vector::{
     distance::{
         Metric, SQ4_CODE_MAX, SQ4_LOADING_SIGMAS, SQ4_RESIDUAL_CENTER, SQ4_RESIDUAL_DIVISOR,
-        SQ4_ROW_BLOCK, Sq4Kernel, Sq16Kernel, dequantize_sq16_into, distance, encode_sq16_row,
-        quantize_query_i8, sq8_walk_dot,
+        SQ4_ROW_BLOCK, Sq4Kernel, Sq16Kernel, dequantize_sq16_adaptive_into, dequantize_sq16_into,
+        distance, encode_sq16_adaptive_row, encode_sq16_row, quantize_query_i8, sq8_walk_dot,
+        sq16_adaptive_norm_sq,
     },
+    rerank_codec::SQ16_CODE_MAX,
     rotation::RandomRotation,
 };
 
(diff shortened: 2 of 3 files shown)


savannahar68 (approved):
lgtm

PR #762, opened 2026-09-11
chore: add split check logic in optimizer

savannahar68:
The split eligiblity right now is very simple and based on number of docs > 500k or modality set in config > 0 (min doc set to 128) this leads to a lot of superfiles being eligible for split check during optimizer. Which leads to optimizer making a lot of get_range calls on object store in every sweep even if there is no change in cells.

This Pr attempts to optimize the flow and introduce fingerprint which is basically hash of all the superfiles related to a cell, which we persist and check if there is any change during every sweep. We proceed with split check (ashman d calculation) only if hash has changed since last run.

### DIFF
--- a/src/supertable/gc/mod.rs
+++ b/src/supertable/gc/mod.rs
@@ -430,6 +430,7 @@ mod tests {
             Some(Manifest {
                 tombstone_seqs: Default::default(),
                 superseded_cells: Default::default(),
+                split_checks: Default::default(),
                 format_version: FORMAT_VERSION.into(),
                 manifest_id: TEST_MANIFEST_ID,
                 options_hash: ContentHash::of(b"options"),
@@ -503,6 +504,7 @@ mod tests {
             Some(Manifest {
                 tombstone_seqs: Default::default(),
                 superseded_cells: Default::default(),
+                split_checks: Default::default(),
                 format_version: FORMAT_VERSION.into(),
                 manifest_id: TEST_MANIFEST_ID,
                 options_hash: ContentHash::of(b"options"),
@@ -571,6 +573,7 @@ mod tests {
             Some(Manifest {
                 tombstone_seqs: Default::default(),
                 superseded_cells: Default::default(),
+                split_checks: Default::default(),
                 format_version: FORMAT_VERSION.into(),
                 manifest_id: TEST_MANIFEST_ID,
                 options_hash: ContentHash::of(b"options"),
--- a/src/supertable/handle.rs
+++ b/src/supertable/handle.rs
@@ -6124,6 +6124,7 @@ mod tests {
             drained_ranges: None,
             global_vector_index: None,
             superseded_cells_additions: None,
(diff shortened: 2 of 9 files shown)


muralikpbhat (approved):
Approving. Verified end to end, including against the reviewed design's three commitments:

- **Digest, not list**: blake3 over the sorted *live* (supersede-filtered) superfile ids, 12 bytes per cell — with order-independence, change-sensitivity, and supersede-exclusion each unit-tested.
- **Derived, not maintained**: recomputed from the manifest each pass, so no write path carries an obligation — and the mid-pass race fails safe: a verdict recorded against scan-time state that a concurrent drain then changed simply won't match the next derivation, costing one spurious re-check and never a wrongly-skipped one.
- **Version and parameters in the key**: `CELL_SPLIT_CHECK_VERSION` is folded into the hash *and* stored as a compared field, and `modality_d` is hashed too — a threshold tune or logic change auto-expires every verdict at once (tested). The deletes-by-construction rationale is written at the fingerprint with the version-bump obligation stated, which is exactly where it belongs.
- **Compat and atomicity done right**: `#[serde(default)]` with an explicit legacy-manifest decode test; verdicts land through the existing commit-metadata overlay (last-writer-wins per cell); no commit when nothing new; only whole-verdicts stored. I also chased the one semantic risk — a defensive no-op being recorded as a checked verdict — and it defuses: the `None` arm covers only deliberate leave-whole outcomes (thin cells, out-of-grid ids, both outcome-correct), while real errors propagate and abort the pass before the recording commit runs. A transient failure cannot mint a false verdict.

Three notes, none blocking:

1. **The headline claim deserves an I/O assertion.** The e2e test asserts the second optimize leaves the verdict unchanged — which would also pass if the cell were re-downloaded and re-checked to the same answer. A counting-storage assertion that the second optimize issues **zero reads** against the cell's superfiles pins the feature (skip the download) rather than the bookkeeping (keep the record).
2. **Split children are born unverdicted.** The split path already knows its children are unimodal leaves (it marks them unsplittable for the pass), but records no verdict for them — so the sweep after a bulk-load split pass re-downloads every fresh child once to conclude what the splitter just proved. Recording the child verdicts in the same commit as the split is the free win; fine as a follow-up.
3. **The memo never shrinks**: entries for cells that no longer exist carry forward indefinitely. Correct for safety (self-invalidating), but a retain-against-the-live-grid at rebuild would mirror how `superseded_cells` already cleans itself.

On the design question of whether an existing marker could have replaced the hash: no single one is a correct per-cell change detector — `manifest_id` is correct but global (any commit anywhere re-checks everything, killing the touched-5-of-755 case), a per-cell max `birth_version` watermark is the near-miss but holds still when a split supersedes a parent's blocks without adding a superfile to the cell, and tombstones are deliberately not part of the key. The derived hash over (live ids, threshold, version) is the minimal correct key, and it introduces no maintained state.

PR #772, opened 2026-09-14
feat: merge entry states in disk cache

naman-modi:
### What does this PR do?

This PR makes the disk reader cache read the way we describe it: check memory, then local disk, then an already open lazy reader, then the object store, and stop at the first hit. One read intent per call replaces the bool flags and the three separate reader functions. Along the way it fixes a case where a query kept fetching from the object store even though the whole file was already on local disk, and two accounting bugs where a lost race could leak budget or keep an evicted entry alive. Part of #737.

Three reader functions each walked memory, disk and source in their own order, with a bool deciding whether to warm the file, so the same lookup was written three ways and none of them read like the mental model. A lazy handle in memory could hide a full copy on disk, so FTS and vector reads kept issuing range GETs for a file sitting on NVMe. Putting an entry into the cache happened in four places and only one of them released the entry it replaced; cleaning up the fetch coordinator happened in eleven places and never on eviction. Each of those was a small hole that a race could fall through.

### How do we do this change?

- One `reader_tiered` walks memory, disk, open lazy reader, then source, and stops at the first hit. `ReadIntent` (`Stream`, `Warm`, `Load`) is the only policy; the bool flags and the three separate reader functions are gone.
- Disk is checked before a lazy entry is allowed to serve, so a lazy handle can no longer hide a finished file.
- `admit_entry` is the one place an entry enters the cache: it releases what it replaces, and a lazy entry never replaces a whole-file one. The fetch coordinator is dropped when its fetch completes and on eviction.

### Notes

- Behavior is otherwise preserved; the two accounting fixes and the shadow fix are the only intended changes.
- Next slice: consolidate the three cold-fetch shapes in `fetch.rs`.

### DIFF
--- a/src/supertable/reader_cache/disk/fetch.rs
+++ b/src/supertable/reader_cache/disk/fetch.rs
@@ -67,30 +67,27 @@ impl DiskCacheStore {
     ) -> Result<Arc<CachedEntry>, DiskCacheError> {
         let (mmap, bytes) = mmap_readonly_with_handle(path).map_err(DiskCacheError::Io)?;
         let reader = SuperfileReader::open_with(bytes, OpenOptions { verify_crc })?;
-        Ok(self.build_mmap_entry(Arc::new(reader), mmap, size, None, None, false))
+        Ok(self.build_mmap_entry(Arc::new(reader), mmap, size, None))
     }
 
-    /// Build a promoted, mmap-backed `CachedEntry`. The single place the whole-file entry shape is
-    /// written: `Eager` accounting (the whole superfile size is store-reserved) and a live mmap.
-    /// `block_token`/`block_source` stay `Some` only when a vector-hole fallback keeps part of the
-    /// object on the block cache after promote.
+    /// Build a promoted [`Residency::Mapped`] entry: the single place the mmap shape is written,
+    /// always `Eager` (the whole superfile size is store-reserved). `vector_source` is `Some` only
+    /// when the vector blob was left out of the mmap and still comes from the block cache.
     pub(crate) fn build_mmap_entry(
         &self,
         reader: Arc<SuperfileReader>,
         mmap: Arc<Mmap>,
         size: u64,
-        block_token: Option<Arc<()>>,
-        block_source: Option<Arc<BlockCachedSource>>,
-        fill_spawned: bool,
+        vector_source: Option<Arc<BlockCachedSource>>,
     ) -> Arc<CachedEntry> {
         Arc::new(CachedEntry {
             reader,
-            mmap: Some(mmap),
+            residency: Residency::Mapped {
+                mmap,
+                vector_source,
+            },
             size_bytes: Arc::new(AtomicU64::new(size)),
             accounting: EntryAccounting::Eager,
-            block_token,
-            block_source,
-            fill_spawned: AtomicBool::new(fill_spawned),
             last_access_us: AtomicU64::new(self.now_us()),
         })
     }
@@ -261,12 +258,10 @@ impl DiskCacheStore {
         //    cache hits serve the mmap reader instead.
         let entry = Arc::new(CachedEntry {
             reader: Arc::clone(&foreground_reader),
-            mmap: None, // hybrid foreground entry is in-memory; finalizer mmaps later
+            // Hybrid foreground: the whole file is in a heap buffer; the finalizer mmaps it later.
+            residency: Residency::Buffered,
             size_bytes: Arc::new(AtomicU64::new(size)),
             accounting: EntryAccounting::Eager,
-            block_token: None,
-            block_source: None,
-            fill_spawned: AtomicBool::new(false),
             last_access_us: AtomicU64::new(self.now_us()),
         });
         self.n_cold_fetches.fetch_add(1, Ordering::AcqRel);
@@ -388,11 +383,14 @@ impl DiskCacheStore {
         entry: &CachedEntry,
         storage: Option<&Arc<dyn StorageProvider>>,
     ) {
-        if skip_background_fill() || entry.mmap.is_some() {
+        if skip_background_fill() || entry.is_mapped() {
             return;
         }
-        if entry
-            .fill_spawned
+        // Only a Paged entry can start a fill, and its latch makes that happen at most once.
+        let Some(fill_spawned) = entry.fill_spawned() else {
+            return;
+        };
+        if fill_spawned
             .compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire)
             .is_err()
         {
@@ -403,8 +401,7 @@ impl DiskCacheStore {
         // block source, and reserve it here — the vector open never did, so the
         // promotion's mmap needs budget reserved before it downloads.
         let size = entry
-            .block_source
-            .as_ref()
+            .block_source()
             .map(|bs| bs.size())
             .filter(|&s| s > 0)
             .unwrap_or_else(|| entry.size_bytes.load(Ordering::Acquire));
@@ -638,7 +635,6 @@ impl DiskCacheStore {
         }
 
         let lazy_reader = Arc::new(lazy_reader);
-        let block_token = block_source_arc.entry_token();
         let (size_bytes, accounting) = if allow_background_fill {
             (Arc::new(AtomicU64::new(size)), EntryAccounting::Eager)
         } else {
@@ -649,14 +645,14 @@ impl DiskCacheStore {
         };
         let entry = Arc::new(CachedEntry {
             reader: Arc::clone(&lazy_reader),
-            mmap: None,
+            // Fill is modality-gated via [`Self::maybe_spawn_background_fill`] after the open
+            // returns, so it starts false here and vector opens never flip it.
+            residency: Residency::Paged {
+                block_source: block_source_arc,
+                fill_spawned: AtomicBool::new(false),
+            },
             size_bytes,
             accounting,
-            block_token: Some(block_token),
-            block_source: Some(block_source_arc),
-            // Fill is modality-gated via [`Self::maybe_spawn_background_fill`]
-            // after the open returns — vector never starts it.
-            fill_spawned: AtomicBool::new(false),
             last_access_us: AtomicU64::new(self.now_us()),
         });
         self.n_cold_fetches.fetch_add(1, Ordering::AcqRel);
@@ -700,7 +696,7 @@ impl DiskCacheStore {
                 verify_crc: self.config.verify_crc_on_open,
             },
         )?;
-        let entry = self.build_mmap_entry(Arc::new(reader), mmap, size, None, None, false);
+        let entry = self.build_mmap_entry(Arc::new(reader), mmap, size, None);
         self.install_promoted_entry(*uri, Arc::clone(&entry), &final_path, InstallMode::Fresh);
         self.n_cold_fetches.fetch_add(1, Ordering::AcqRel);
         reservation.commit();
@@ -833,7 +829,7 @@ async fn finalize_to_mmap(
         // Replace the in-memory-backed entry with the mmap-backed one. `ReplaceIfPresent` drops
         // the file instead of reinstating when a racing reservation evicted the slot mid-finalize
         // (reinstating an evicted, already-released entry would leak its reservation).
-        let entry = store.build_mmap_entry(Arc::new(reader), mmap, size, None, None, false);
+        let entry = store.build_mmap_entry(Arc::new(reader), mmap, size, None);
         store.install_promoted_entry(uri, entry, &final_path, InstallMode::ReplaceIfPresent);
         store.coordinators.remove(&uri);
         Ok::<(), DiskCacheError>(())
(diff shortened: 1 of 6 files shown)


src/supertable/reader_cache/disk/read.rs
```diff
@@ -149,135 +105,242 @@ impl DiskCacheStore {
...
+
+        // A Load rewrites the whole file, so a lazy handle is no use to it: drop it here so it can
+        // neither shadow the disk check below nor be served as the tier-3 hit. (majorly for compaction)
+        if intent == ReadIntent::Load {
+            self.drop_lazy_entry(uri);
+        }
+
+        // Tier 2, disk cache: the whole file is already on local disk (a prior run, or a lazy reader
+        // that has since been fully filled). Mmap it, zero object-store GETs.
+        if let Some(entry) = self
```
muralikpbhat, line 196: **Two structural consequences of the disk probe moving here (outside the single-flight), worth deciding deliberately:**

1. **Tier 2 lost its single-flight.** N concurrent tier-1 misses on one URI each stat + open + mmap the same cache file (+ a full-file CRC scan each when `verify_crc_on_open` is set — a 64-way fan-out re-opening a 2 GiB cell does 64 scans), and transiently charge N× the size before `admit_entry` releases the losers — enough to shove a tight budget into range-only degrade for unrelated opens. The old walk did the disk reuse exactly once, inside the coordinator. Moving the probe back inside the coordinator (or adding a per-URI once-gate on this tier) keeps the shadow fix and restores the single-flight.
2. **Every open of a live lazy entry now pays a blocking `fs::metadata` (usually ENOENT) before tier 3 serves** — per superfile, per fan-out wave, on the async runtime thread; the old path served the lazy hit from the map alone. And once a Warm/Load reader lands a whole file on disk, subsequent **Stream** opens promote it to a full Eager mmap — post-compaction, vector superfiles flip from SourceOwned block accounting to full-file Eager charges, which under a tight budget evicts other entries. The waiting `st-vec` bench lanes are exactly where both would show — this shouldn't merge before they report.

src/supertable/reader_cache/disk/read.rs
```diff
@@ -149,135 +105,242 @@ impl DiskCacheStore {
...
+            .map(|e| Arc::clone(&e))
+    }
+
+    /// Remove a cached lazy entry and give its budget back. Used when [`ReadIntent::Load`] finds a
+    /// lazy entry it cannot use and must re-fetch a full copy.
+    fn drop_lazy_entry(&self, uri: &SuperfileUri) {
+        if let Some((_, removed)) = self.cached.remove(uri) {
+            self.release_entry_accounting(&removed);
+        }
+        self.coordinators.remove(uri);
```
muralikpbhat, line 246: Error-arm herd: the cell is permanently initialized with `Err`, and each of N waiters does `coordinators.remove(uri)` — remove-by-key, which can strip a **newer** in-flight cell someone else just created — then retries `fetch_from_source` uncoalesced. One transient GET failure on a hot cold-miss ⇒ N independent HEAD+full-fetch retries, plus an N+1th caller starting yet another duplicate download after a waiter stripped the fresh coordinator. Compare-and-remove (remove only if the cell is the one this waiter joined) plus letting one waiter re-establish the coordinator would keep the retry coalesced.

src/supertable/reader_cache/disk/read.rs
```diff
@@ -20,124 +21,79 @@ use crate::{
...
+    /// must not come through here: a compaction read can never accept that degrade, so it uses
+    /// [`Self::reader_synchronous_with_storage`], which fails loudly instead.
+    pub async fn open_for_query(
         self: &Arc<Self>,
         uri: &SuperfileUri,
         storage_key: &str,
         offsets: Option<&SubsectionOffsets>,
         storage: Option<&Arc<dyn StorageProvider>>,
-        allow_background_fill: bool,
+        intent: ReadIntent,
```
muralikpbhat, line 55: The Load-must-not-degrade invariant is only a `debug_assert_ne` — in release, a future `Load` caller hitting `BudgetExceeded` silently falls into `open_range_only` and hands a rewrite path a partial-range streaming reader. Since `Load` means "compaction input," a hard error is the right release-mode behavior. Related hardening while here: a `Load` that coalesces onto an in-flight lazy fetch can be handed a Paged entry (the cell isn't intent-keyed; `drop_lazy_entry` runs before the cell exists under a race) — `intent == Load && !entry.has_whole_file()` after tier 4 → drop and re-fetch eagerly makes the whole-file contract enforced, not assumed.

muralikpbhat (commented):
Reviewed with three deep passes: per-call-site tier-walk equivalence, the concurrency/accounting claims, and the semantics this refactor must not break. The verdict up front: **the mapping is exact, the three intended fixes are real, and the preserved-semantics checks all pass — holding on one converged accounting-hole family plus the waiting bench lanes.**

**Verified sound:**
- **Call-site equivalence**: all 20+ movers map exactly (`true→Warm`, `false→Stream`, sync→`Load`); no caller's warming behavior or fallback changed beyond the intended shadow fix, which works and is genuinely tested (`tiers_never_refetch_local_data` asserts lazy-entry + complete-disk-file → zero cold fetches + promotion).
- **The eviction exemption for resident indexes is byte-identical** — pinned-URI machinery and `select_for_eviction` untouched, protected residency/accounting classes not folded, and `admit_entry`'s lazy-never-replaces-whole-file rule *adds* protection. The large-scale warm-serving behavior that depends on it is safe.
- **The claimed races are fixed where claimed**: replace-release is atomic under one shard lock (exactly-once for the loser), no resurrect-after-evict (Vacant-gated re-insert; eviction now drops the coordinator — a real fix, pinned by the new test), no waiter hangs, eviction-vs-serve is refcount-correct. op_stats scopes untouched; per-tier counters survive; the zero-length-file guards survive; bench/writer/wal/handle changes are mechanical; hygiene clean.

**Holding on: the `ReplaceIfPresent` family — all three review passes converged on it independently.**

1. **`install_promoted_entry`'s `ReplaceIfPresent` arm (fetch.rs:114) replaces the occupant without `release_entry_accounting`** — a second door that contradicts `admit_entry`'s own doc claim of being the one admission point. Pre-PR this was benign (the slot could only hold a SourceOwned entry; release is a no-op). **This PR's shadow fix makes it live**: tier 2 can now admit a fully-charged Eager whole-file entry while a background fill is still latched — narrow window (fill's rename → a concurrent tier-2 stat → install clobbers) and wide window (whole file already on disk from a prior run → tier 2 admits Eager → the latched fill re-downloads and installs over it). Every occurrence leaves the replaced entry's bytes phantom-charged until restart. Evict-then-readmit reaches the same arm. Fix is one unconditionally-safe line: release the replaced entry, same as `admit_entry` rule 2.
2. **Evict-during-fill leaks the fill's reservation** (fetch.rs:1226): when the install finds the slot Vacant, the entry is dropped and `let _ = reserved_bytes` never rolls back the charge reserved at fetch.rs:399 — and every `rollback_lazy_background_fill` path releases a SourceOwned entry (no-op) instead of the fill's reservation. Each occurrence permanently shrinks the usable budget toward BudgetExceeded/range-only.
3. **The fill rollback removes whatever entry is *current* for the URI, not the one the abandoned fill owns** (fetch.rs:1061) — under an evict-then-readmit race it destroys a valid, fully-downloaded entry another admission just installed and strands that entry's reservation, compounding (2). Scope the rollback to the fill's own entry (pointer identity or a generation tag).

These three are the same corridor the PR set out to fix — the accounting claims are right for the paths the tests cover, and one interleaving beyond them. None are covered by a test, and **there's no `current_bytes == Σ live-entry charges` assertion anywhere**, so all three would regress silently even after fixing; a debug-mode invariant check (asserted in the existing sequential tests) is the cheap durable guard.

**Also before merge: the six bench lanes are still waiting, and they're not a formality here** — the tier-2 restructuring is a real behavioral shift (details inline at read.rs:196): lost single-flight on the disk tier, a per-open stat on the hottest vector path, and Stream opens now promoting post-compaction vector superfiles from SourceOwned block accounting to full Eager charges. `st-vec` warm/cold is exactly where a delta would show.

**Non-blocking (inline + here):** the error-arm thundering herd with remove-by-key stripping newer coordinators; the `Load` contract as `debug_assert`-only plus the un-keyed coalesce race; the two half-true intent doc-comments; the local-put lost-race arm unlinking the **winner's** finished cache file (store_files.rs:118 — perf only: the next process re-downloads a file that was fully local); `open_lazy_reader`'s guard comment saying "explicitly Paged" while the code filters on `block_source().is_some()` (match the variant explicitly); and a "funciton" typo in 187403e4's subject for the eventual squash.

With the three-line family fixed, the invariant assertion added, and green benches, this is a clear approve — the one-walk/one-intent shape is the right destination, and the exemption/metering preservation work was done carefully.

src/supertable/reader_cache/disk/mod.rs
```diff
@@ -171,8 +171,23 @@ pub(crate) struct CachedEntry {
...
+/// entry serves `Stream`/`Warm` but not `Load`) and which fetch shape a real miss uses; nothing
+/// else branches on it.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ReadIntent {
+    /// Serve byte ranges on demand, stay remote-backed, never promote. Vector search.
+    Stream,
+    /// Serve now, and download the rest in the background toward a local mmap. FTS and SQL queries.
+    Warm,
+    /// Download the whole file and mmap it before serving. Compaction, which rewrites its input.
+    Load,
```
muralikpbhat, line 185: Both intent doc-comments are now half-true: `Stream` "never promotes" — except tier 2 promotes an on-disk file to a full mmap for Stream reads (the intended shadow fix, and the test asserts it); `Load` = "compaction" — except user-table compaction opens inputs with `Warm` (dispatch.rs:208); only the hidden-index path uses Load. A maintainer gating a future change on either sentence breaks the vector budget model or compaction. Tighten both.

src/supertable/reader_cache/disk/fetch.rs
Pranav2612000, line 118: Due to some of other changes, does this now require some "accounting" related changes when we replace an entry?

naman-modi: Yes, same thing Murali flagged. Since tier 2 can now admit a whole file from disk while a fill is still running, blindly replacing the slot would leave the old entry's bytes charged forever.

fixed it by making the fill only touch the entry it started from (pointer check on the reader):
- slot still has its own lazy entry: swap in the mmap, it inherits that entry's charge
- someone else's whole file is there: keep theirs, drop our download, release only what the fill reserved for itself
- slot empty (evicted mid-fill): drop the file, release our own reservation

src/supertable/reader_cache/disk/read.rs
naman-modi: Agreed, moved tiers 2 to 4 inside the single flight, so n concurrent misses do one stat/mmap/download. the per-open stat and stream-promoting-to-eager are left for the st-vec lanes to decide.

src/supertable/reader_cache/disk/read.rs
naman-modi: Fxed, waiters remove the cell only if it's still the one they joined (remove_if + ptr_eq), and the retry goes back through the single flight once before a solo attempt.

src/supertable/reader_cache/disk/mod.rs
naman-modi: My bad, updated them.

src/supertable/reader_cache/disk/read.rs
naman-modi: `open_for_query` now returns a hard error for Load, and a Load that joins a lazy walk walks again instead of serving the partial file.

naman-modi:
On your comment @muralikpbhat:

1. "replace if present": fixed, the fill now only swaps out the entry it started from (pointer check), otherwise keeps what's there and releases only its own reservation.
2. "evict-during-fill" leak: fixed, an empty slot on finish releases the fill's own reservation and drops the file.
3. rollback removing the current entry: fixed, rollback is remove_if on the fill's own entry, so someone else's whole file stays.
4. open_lazy_reader guard: now matches Residency::Paged explicitly.

muralikpbhat (approved):
Approving. Verified this round in code and ran the disk-cache suite locally (111 passing, including your four new race tests). The fill's owner-scoped swap is right — and keeping a stranger's whole file rather than deleting the shared path avoids a bug my original ask would not have caught. The RAII reservation with explicit commit closes the whole leak family, not just the cases I flagged. The Load re-walk and ptr-scoped waiter removal both check out.

One non-blocking note: no read-path numbers on the record. The single-flight adds coordination to the miss path (hits look untouched) — a quick before/after of the st-vec cold lane would close it, or a stated reason it can't matter: tier 4 is network-bound so the added locking is noise there, but a tier-2 disk hit (open + mmap of a file already cached) is fast enough that serializing concurrent readers of one hot file through the flight could show up under concurrency. Either the number or that reasoning on the record is fine.

PR #773, opened 2026-09-15
fts: blob V7 — a third smaller index (short-form terms, term-block dictionary, compact blocks, grouped positions, packed ids)

vinaykakade:
On a Zipfian text corpus the FTS index spent more than half of its postings region on the ~97% of terms whose whole posting list fits one block: a 92-byte fixed cost per term (header, skip entry, position sub-index row, coarse slot, block header) plus the block codec's padding of a partial block to 128 lanes, so two far-apart docs cost hundreds of bytes. The table layer then stored the dictionary a second time in the manifest's open blob and a third time in the term-stats sidecar. This PR is a size change; every query result is unchanged.

## What changes

**Blob version 7** (readers accept V1–V7; V1–V6 stay read-only legacy, tests write faithful older eras; `format::fts::BlobLayout::for_version` is the one table of what each version lays out):

- **Short form for single-block terms.** A term with `df ≤ 128` is a varint df, a tf==1 bitmap, group-varint doc deltas, the remaining tfs and, on a positional column, its position group inline — no header, skip table, sub-index, coarse table or block. The reader decodes a short body once into the pre-decoded single-block cursor the df=1 inline form already used, so ranked, count, phrase and compaction paths see the cursor they always did.
- **Term dictionary as front-coded term blocks** in place of the FST: sorted terms in blocks of 32, each stored as the suffix it does not share with the previous one plus a form byte, an offset delta and a length, behind a fixed-width first-key table searched in place (open reads the footer only). `TermDict` hands every caller decoded entries for either layout.
- **4-byte block header.** One word holds count, both widths, the encoding and a patched block's exception counts; a packed or patched block's base doc id is the previous block's last doc id from the skip table (a bitset block keeps its stored origin). Byte 3's low two bits are the encoding in both layouts, so the count kernels' one-byte probe is unchanged. `BlockHeader` parses either layout once per block.
- **Length-coded skip entries.** A `u16` block length replaces the absolute offset and a positionless column drops the positions field (14 bytes positional, 10 otherwise, from 16); each coarse slot gains its span's start offset, so sequential walks accumulate one length per block and a random block is one slot read plus at most 31 lengths.
- **Patched block encoding.** A block may pack its doc-id deltas and tfs each at the width most lanes fit, with the outlier lanes' high bits as exceptions — smaller than plain packing on the partial and sparse blocks of mid-frequency terms. The writer grants it per term, to terms with at most 16k postings: a patched block decodes about 15 ns slower than a plain one, and a common term's blocks are what every intersection and union expands in bulk, while a term under the bar is walked in full only when it is itself a query term. Never taken for a block the presence bitset claims. The decode unpacks the low bits, ORs the exceptions in and prefix-sums with an eight-lane SIMD log-step (`wide::u32x8`), measured fastest of the alternatives on the reference machine.
- **Positions are per-block groups**: each document's first position and each gap packed as two streams with exceptions. A phrase member locates a group once per block and reads one pair's run on its own, so a candidate costs its own tf lanes; the run-offset sub-index is gone.
- **Doc lengths are `u16`**, saturating at 65,535; the writer's bound and the reader's bucket both derive from the saturated value.

**Table layer / superfile:**

- The manifest inlines a superfile's open ranges only while they total ≤ 1 MiB (the fallback fetches them in one parallel wave).
- No term-stats sidecar for a one-superfile table (its own dictionary is the global statistics).
- **Frame-of-reference `_id` sidecar** (`inf.ids.layout = packed`): blocks of 1024 docs, each half of the id bit-packed from the block minimum. O(1) resolve preserved; the per-query resolve opens the sidecar by its header (the blocks are validated at superfile open).

**Shared helpers** consolidated along the way: one LEB128 varint module, format-level little-endian word readers, one exception-width planner in the bit-packing module, and the group index as the only position-group decoder.

**Tooling:** `SuperfileReader::size_breakdown` / `FtsReader::size_breakdown` and the `index_size_report` example attribute every byte to a region, a df band and a layout part, with histograms of patched blocks by exception count and bytes saved (`test-helpers`).

## Measured

1M Wikipedia docs, standard analyzer, positions on, text unstored, one compacted superfile:

| | before | after |
|---|---|---|
| FTS postings region | 532 MiB | 229 MiB |
| df ≤ 128 terms' share of it (their positions now inline) | 294.5 MiB | 39.9 MiB |
| position sub-index | 29.0 MiB | 0 |
| positions region | 387.3 MiB | 338.0 MiB |
| term dictionary | 23.0 MiB | 15.8 MiB |
| doc lengths | 3.8 MiB | 1.9 MiB |
| `_id` sidecar | 15.3 MiB | 1.25 MiB |
| manifest part + term-stats | 39.5 MiB | 0.07 MiB |
| **table directory total** | **1,001 MiB** | **587.4 MiB** |

5.3M docs (the full benchmark corpus, same build path): the directory goes from 4,131,354,106 to 2,777,706,541 bytes (−33%).

Term count, postings count and every query result are unchanged (the corpus indexes to the same 1,642,896 terms / 126,703,012 postings; the same-box A/B below reports zero count mismatches over the full query set).

## Benchmark comparison against `main`

Same box, both indexes rebuilt from the 5.3M corpus, the full 962-query set, `client.py` run in both engine orders and the orders averaged (mean of per-query minima, branch / main):

| Lane | ratio | union | intersection | phrase | negated |
|---|---|---|---|---|---|
| TOP_10 | 1.05 | 1.04 | 1.08 | 1.04 | 1.02 |
| TOP_100 | 1.02 | 1.04 | 1.06 | 0.99 | 0.99 |
| TOP_1000 | 1.03 | 1.04 | 1.06 | 1.00 | 1.00 |
| TOP_100_COUNT | 1.04 | | | | |
| COUNT | 1.09 | | | | |

The A/B's run-to-run floor is about ±2% per lane and ±2–4% per tag (the previous round on the same box read union 1.00–1.01, intersection 1.06–1.08, phrase 1.03–1.04). Ranked union queries are within that floor, and a second machine's 1M run of the same commit (same harness, main vs branch, both orders) has every ranked lane within 1–2% of main: TOP_10 1.01, TOP_100 1.02, TOP_1000 1.01.

What remains at 5.3M is a 5–8% residual on the ranked intersection tag and 7–8% on the count lanes. It is not the format: a source-line profile of both engines shows the difference spread over lines that are the same code in both — the flat-merge loop, bounds checks, the cursor advance — rather than on any branch-specific line, and the new dictionary lookup measures about 1.5 µs per query against the FST's 0.8. The remaining cause is code generation and cache layout inside the inlined query kernels, which the 1M working set does not expose; the first thing to try if it shows on the public page is moving the cursor's block-crossing decode out of the merge loop's inlined body.

Two more levers were tried and reverted on this evidence: reading one tf lane per bitset probe instead of unpacking the block's 128 tfs cost the union tag 7% and intersection 5%, because the union and intersection kernels probe the same dense block many times; a merge walk in phrase verification read as a 4% loss. Both boxes' scripts are kept so the next attempt starts from the same instrument. Eight rounds of this comparison drove the reader changes above (the first round was 1.4–2.5× slower before the per-run position reads, the header-only sidecar open and the fixed-width dictionary index).

## CI benchmark lanes

The PR's benchmark workflow compares the branch against a `main` arm built in the same run. Three lanes moved; each was traced to its cause and either fixed or shown to be a measurement shape rather than reader cost.

**Warm FTS lane (`phrase_two_common` +130%, `phrase_three_common`, `phrase_plus_must_term`) and SQL lane (`token_match (all rows)` +68%) — compaction shape, not the reader.** The warm rows are measured after `optimize()`. Both lanes compact with the default 1 GiB byte target, and this branch stores the realistic corpus at 1.60 GiB against main's 5.38 GiB, so the same target packs 3.4× more docs per output: the planner filled one job with 46 of the 75 superfiles (613k docs) and left 29 unmerged under the fill floor, where main got five 14-input jobs of 187k docs. Running the real planner on both size profiles reproduces the observed superfile counts (FTS 75 → 12 vs 75 → 26, SQL 112 → 3 vs 112 → 27). A query runs one thread per superfile, so post-compaction latency follows the largest superfile's doc count, 613k vs 187k. At equal shape — the same lanes' rows before compaction, 75 (FTS) and 112 (SQL) superfiles — the same box measures:

| Query (1M realistic, warm p50) | main | branch |
|---|---|---|
| `phrase_two_common` | 21.7 ms | 25.0 ms (+15%) |
| `phrase_three_common` | 28.3 ms | 31.0 ms (+9.5%) |
| `phrase_plus_must_term` | 31.5 ms | 35.3 ms (+12%) |
| `token_match (all rows)` (CI, Azure) | 56.5 ms | 59.6 ms (+5.5%) |

The bench now derives its compaction target from the table's measured bytes per doc, so every arm compacts to the same ~200k docs per superfile (what main's default produced on this corpus) and the warm rows compare readers. With it the branch compacts 75 → 5 superfiles of 200k docs and its post-compaction rows on the same box are `phrase_two_common` 44.4 ms, `phrase_three_common` 46.3 ms, `phrase_plus_must_term` 52.6 ms, against main's 35.2 / 41.0 / 45.3 ms on its 12-superfile shape whose largest holds 187k docs — +13–26%, of which ~7% is the larger superfile. Until this lands the `main` arm still compacts with its own bench code, so the two arms' shapes converge only after merge. With the pinned shape the PR's benchmark workflow is green on every lane (warm FTS `phrase_two_common` 40.4 ms vs main's 33.9 ms on its 12-superfile shape, SQL `token_match` inside the gate, cold lane informational only). The planner's own behaviour under a denser index — one full job plus stranded leftovers, and a byte target that ignores per-superfile query parallelism — is pre-existing and left for a separate change.

**Cold FTS lane (informational movers of 2–10× on single rows) — the open blob, fixed; the residual movers are storage noise.** The first version capped the open blob at 1 MiB *in total*, which dropped it entirely for any real superfile and sent every first cold query to fetch each open range itself. The cap is now 4 MiB *per range*: parquet tail, doc lengths and small dictionaries stay inline, only a large superfile's dictionary is left out, and the open wave fetches exactly the ranges the blob lacks. On the CI cold lane (Azure, 75 superfiles), per-query median of branch/main: **cold open 0.74×, cold first query 0.94×**, with GET bytes equal or lower on every query (`single_common` 5.06 → 4.85 MiB, `phrase_two_common` 35.1 → 29.7 MiB). The rows the lane still flags are single-sample Azure variance; main's own arm has the same query at 4.57 s before compaction and 0.28 s after.

## Compatibility and behaviour

- Existing superfiles read unchanged; compaction rewrites them to V7.
- Ranked hits resolve their stable ids in doc order through the packed sidecar (one block parse per touched block), scattered back to score order; results are identical.
- A V7 blob cannot be opened by older readers — a minor-version release when this lands, like the last blob bump.
- Scores and rankings are unchanged. A document past 65,535 tokens has its stored length saturated; its bound and bucket are computed from the same value on both sides.
- Public API surface unchanged; the size report is `test-helpers`-only.

## Verification

`cargo test --lib` (2,620), `--test superfile` (197, includes the fuzz oracle and CRC-corruption tests, with new ones for the short body and the compact header), `--test supertable` (342), `clippy --all-targets -D warnings`, `fmt` with the CI import configuration — green on the head commit.


src/superfile/bits.rs
```diff
@@ -0,0 +1,221 @@
...
+    per_exception: impl Fn(u32, u32) -> usize,
+) -> ExceptionPlan {
+    let plain = width_of(lanes.iter().copied().max().unwrap_or(0).into());
+    let mut best = ExceptionPlan {
+        width: plain,
+        exceptions: Vec::new(),
+        bytes: base(plain),
+    };
+    for width in 0..plain {
+        let mut exceptions = Vec::new();
```
pratyushlokhande: here plan_exceptions allocates fresh Vec per candidate width. Up to 33 allocs/stream; 2 streams per position group per block, 2 more per patched-block candidate, ~1M blocks. can we pass reusable &mut Vec scratch?

src/supertable/writer.rs
pratyushlokhande: This returns off..postings_offset — the whole term dictionary — plus the whole doc-lengths region. The measured table puts those at 15.8 + 1.9 MiB for one 1M-doc superfile, 17.7× OPEN_BLOB_INLINE_MAX_BYTES, so the blob is empty and every cold open pays the extra round trip. Narrow this to the footer + first-key table the V7 open actually reads, or report the cold-open number so the trade is on the record.

pratyushlokhande (approved):
Few minor comments, rest looks good. ok to merge.

src/superfile/bits.rs
vinaykakade: Done

src/supertable/writer.rs
vinaykakade: Done

vinaykakade: Numbers from the CI cold lane on `9c28e675` (Azure, 75 superfiles), per-query median of branch/main across the 23 queries: cold open 0.74×, cold first query 0.94×. GET bytes are equal or lower on every query (`single_common` 5.06 → 4.85 MiB, `phrase_two_common` 35.1 → 29.7 MiB). With the cap per range, the parquet tail, doc lengths and small dictionaries stay inline and only a large superfile’s dictionary is left out, fetched by the open wave as an uncovered range. The rows the lane still flags (+300–600% on single queries) are single-sample Azure variance: main’s own arm has `three_similar_and` first query at 4.57 s before compaction and 0.28 s after.

PR #776, opened 2026-09-16
tests: corpus generators pinned to old engine releases

pratyushlokhande:
## What

Adds five small generator crates, each pinned to one published `infino`
release, that write an identical corpus of 12,000 documents. Each release's
builder produces a different FTS superfile shape, so the same documents give
us one table per shape to test against.

| Generator | Shape it writes |
|---|---|
| `v0_1_5` | positionless, 48-byte header, no positions region |
| `v0_5_4` | 56-byte header, empty positions region |
| `v0_5_12` | bitset blocks, pre-coarse |
| `v0_8_0` | exact-f32 block-max + coarse table, positionless |
| `v0_8_2` | the same, with a positional column |

The generated tables and the assertions over them land in a follow-up PR;
this one is just the tooling.

## Why

Read-side compatibility code for superseded FTS shapes can only be removed
once there is a way to prove the current engine still reads files the old
writers actually produced. Reconstructing those bytes from the format spec
tests the spec, not what shipped — so the corpus is written by the real
releases.

## What the reviewer should look at

- **The shared corpus** (`shared/corpus.rs`) is included with `include!`
  rather than shared as a crate: each generator resolves its own dependency
  graph, and one crate would have to satisfy all of them at once.
- **`burst` term shaping.** A uniformly dense term does *not* produce bitset
  blocks — its deltas are all 1 and pack to a single bit. The run-and-gap
  pattern makes one block's max delta wide enough that the presence bitset
  encodes smaller than the packed deltas, which is what the `v0_5_12` shape
  needs. `N_DOCS` is sized so the corpus-wide term clears
  `BLOCK_LEN * COARSE_BLOCK_MAX_SPAN` postings and the coarse table holds
  more than one entry.
- **Each crate carries its own `[workspace]` table** so its pinned,
  year-old dependency graph never takes part in resolving the engine's own.
  `cargo metadata` at the root is unaffected. `target/` dirs are gitignored;
  lockfiles are checked in so the corpus is reproducible.

Notably, no dependency pinning was needed — every one of these releases
still builds clean on the current toolchain.

## Test

`cargo build --release` in each of the five crates, and each was run to
produce a table whose FTS blob version byte was checked against the shape it
is supposed to write. All five matched.


pratyushlokhande:
Closing — opened early. Reopening once the full migration work is in place.

PR #777, opened 2026-09-16
vector: widen the filtered undrained-tail probe by predicate selectivity

### LINKED ISSUE #774: vector: filter_query pushdown returns only a fraction of matching rows
`vector_search(..., filter_column=, filter_query=)` returns far fewer rows than
match the predicate when the matching rows are sparse and scattered. Raising `k`
does not recover them.

20,000 rows, `dim=384`, cosine, 50 superfiles. Exactly 10 rows carry the token
`XR4200`. Both FTS paths find all 10; the pushdown finds 0–3, per query:

```
token_match('body', 'XR4200')      -> 10 rows
bm25_search('body', 'XR4200', 20)  -> 10 rows
vector_search(k=10 , filter_query='XR4200') -> [0, 1, 3, 0, 1, 3, 1, 0]
vector_search(k=50 , filter_query='XR4200') -> [0, 1, 3, 0, 1, 3, 1, 0]
vector_search(k=200, filter_query='XR4200') -> [0, 1, 3, 0, 1, 3, 1, 1]
```

The counts vary by query vector and are flat in `k`, which points at the
predicate being applied within the probed IVF cells rather than to the candidate
set before probing — a matching row in an unprobed cell is unreachable at any
`k`.

Expected: with 10 matching rows and `k >= 10`, all 10 come back ranked by
distance. That is also what the pushdown is documented to do — the Python
docstring says the kNN "ranks only among the matching rows rather than
post-filtering the global top-k", which reads as a guarantee over all matching
rows.

Impact: a pushdown on a selective predicate — a document id, an error code, a
tenant tag — silently returns a fraction of the matches with no error and no
signal to the caller. Callers that need every match have to fall back to
`bm25_search`.

Environment: `infino` 0.7.1 (Python), macOS arm64, local path connection.

Repro (self-contained, ~25s):

```python
import tempfile, random
import numpy as np, pyarrow as pa, infino

N, DIM, B, RARE = 20_000, 384, 2_000, "XR4200"
VOCAB = ("latency throughput parquet embedding shard tenant invoice refund cluster vector "
         "keyword recall precision fusion rerank ingest compaction manifest snapshot "
         "predicate pushdown quantization centroid index segment").split()
rng = np.random.default_rng(42)

conn = infino.connect(tempfile.mkdtemp())
schema = pa.schema([pa.field("id", pa.large_utf8(), nullable=False),
                    pa.field("body", pa.large_utf8(), nullable=False),
                    pa.field("vec", pa.list_(pa.float32(), DIM), nullable=False)])
tbl = conn.create_table("t", schema, infino.IndexSpec().fts("body").vector("vec", DIM, "cosine"))

for start in range(0, N, B):
    v = rng.normal(size=(B, DIM)).astype(np.float32); v /= np.linalg.norm(v, axis=1, keepdims=True)
    tbl.append(pa.table({
        "id": pa.array([f"n{start+j}" for j in range(B)], pa.large_utf8()),
        "body": pa.array([" ".join(random.Random(start+j).sample(VOCAB, 8))
                          + (f" {RARE}" if (start+j) % 2000 == 0 else "")
                          for j in range(B)], pa.large_utf8()),
        "vec": pa.FixedSizeListArray.from_arrays(pa.array(v.reshape(-1)), DIM),
    }))

print("token_match:", tbl.token_match("body", RARE, projection=["id"]).num_rows)
print("bm25_search:", tbl.bm25_search("body", RARE, 20, projection=["id"]).num_rows)
qs = rng.normal(size=(8, DIM)).astype(np.float32); qs /= np.linalg.norm(qs, axis=1, keepdims=True)
for k in (10, 50, 200):
    print(k, [tbl.vector_search("vec", qs[i].tolist(), k, filter_column="body",
                                filter_query=RARE, projection=["id"]).num_rows for i in range(8)])
```

muralikpbhat:
Fixes https://github.com/infino-ai/infino/issues/774.

## Problem

Filtered vector search (`VectorFilter` + kNN) on a table that has not been optimized returns far fewer than `k` hits when the filter predicate is selective and its matching rows are scattered — even though enough matches exist. The same query after `optimize()` returns all of them. With the issue's repro (10 tagged rows among 5,000), the unoptimized table returned 0–3 of the 10 tagged rows.

## Root cause

The undrained filtered routing arm probes a fixed 4 cells (`FILTERED_USER_CELL_NPROBE`), chosen by query-vector proximity alone. The filter's allow bitmap plays no part in routing, so when the allowed rows are sparse and spread across cells, most of them sit in cells the probe never visits. Drained cells do not have this problem: their serving laws already carry a per-superfile inverse-selectivity widening (`filter_selectivity_mult`), so a selective filter widens the probe there.

## Fix

Bring the same selectivity math to the undrained arm, at the routing altitude where the decision is made:

- Export the count-based core of the superfile tier's multiplier as `selectivity_mult_from_counts(allowed, population)` (pure refactor of `filter_selectivity_mult`; drained behavior unchanged).
- In `vector_fanout_over_superfiles`, the filtered-undrained arm computes allowed-row and population counts from the allow bitmaps and superfile manifests, and scales the 4-cell base width by the inverse selectivity, capped by the existing `MAX_FILTER_SELECTIVITY_MULT` (64).

A broad filter (high selectivity) keeps today's narrow probe and cost; a needle-in-haystack filter widens toward scanning every cell, which is the only correct answer when the survivors are scattered. Unfiltered routing and drained routing are untouched — this is one `else if` arm plus a `pub(crate)` export, ~30 lines.

Not done here: probe-cell selection is still by query distance, so a dense filter with scattered matches stays approximate pre-drain (the review's second finding). That needs allow-aware cell selection with a per-cell membership summary, applies to both tiers, and the window is transient (drain converts the tail to calibrated cells) — tracked as https://github.com/infino-ai/infino/issues/780.

Review round: the tail's selectivity is now counted over the tail's own superfiles only (the allow map also carries drained superfiles' matches, which overstated selectivity and collapsed the widening back to the fixed floor on partially-drained tables — reviewer catch), and an empty tail allow-set now skips the fan instead of probing. New test covers the partially-drained shape.

## Verification

- New regression test `filtered_knn_finds_sparse_matches_outside_the_probed_cells`: 10 tagged one-hot rows scattered across a 256-row table must all be found at k=10 with ascending scores. Fails on the parent commit (8/10 found), passes with the fix.
- Issue repro end-to-end via a locally built wheel: Pratyush's exact script now returns 10/10 tagged rows on the unoptimized table at k=10, 50, and 200 (was 0–3); the optimized-table and common-token results are unchanged.
- Full local gates: clippy clean, lib tests 2582 passed / 0 failed, supertable integration 343 passed / 0 failed.
- Latency (same wheel, unoptimized table, k=10): the selective-filter query goes 0.24 ms → 0.57 ms (it now probes wide enough to be correct), the broad-filter query 1.08 ms → 1.48 ms; measured against a locally built baseline wheel, so treat deltas as directional.

Note: main's lib-test target currently doesn't compile (a test fixture missing the newly added `stem` field); #778 fixes that, and this PR's test-suite legs need it merged to go green.


src/supertable/query/vector.rs
```diff
@@ -5249,13 +5249,21 @@ impl SupertableReader {
...
+                // Filtered UNDRAINED-tail fan: cell routing follows the QUERY,
+                // but a sparse predicate's matches sit wherever they sit — so
+                // widen the [`FILTERED_USER_CELL_NPROBE`] floor by the filter's
+                // inverse selectivity (same capped math as the superfile tier's
+                // `selectivity_mult_from_counts`). Dense filters keep the floor.
+                let allowed: u64 = allow
+                    .as_ref()
+                    .map(|m| m.values().map(|bm| bm.len()).sum())
+                    .unwrap_or(0);
+                let population: u64 = superfiles.iter().map(|e| e.n_docs).sum();
```
pratyushlokhande, line 5273: `allowed` and `population` are counted over two different superfile sets.

`route_filtered_vector_hits_async` (`vector.rs:6545`) passes `undrained_user` — the tail subset — as `superfiles`, but hands down the **full** `user_allow` map, which still carries entries for the drained superfiles. So `allowed` sums the whole table's matches while `population` sums only the tail's rows.

Drained matches inflate `allowed`, selectivity is overstated, and `mult` shrinks. Once drained matches >= tail rows the ratio exceeds 1, `ceil(1/sel)` is 1, and this arm quietly degrades to the old fixed floor.

```rust
let allowed: u64 = allow
    .as_ref()
    .map(|m| {
        superfiles
            .iter()
            .filter_map(|e| m.get(&e.uri))
            .map(|bm| bm.len())
            .sum()
    })
    .unwrap_or(0);
```

Related, in the description: "a partially-drained table serves its undrained tail through the hidden-index arm (unchanged `FILTERED_HIDDEN_CELL_NPROBE` floor)". That call is on the user-table reader, so `hidden_vector_index` is false and the tail lands in *this* arm, not the hidden one — which is also why the miscount above is reachable rather than theoretical.

Separately on `population`: `e.n_docs` counts tombstoned rows, while the allow bitmaps have them subtracted (`subtract_tombstones`). A delete-heavy table therefore understates selectivity and over-probes.

src/supertable/query/vector.rs
```diff
@@ -5249,13 +5249,21 @@ impl SupertableReader {
...
+                let mult = selectivity_mult_from_counts(allowed, population).max(1);
+                let width = FILTERED_USER_CELL_NPROBE.saturating_mul(mult);
```
pratyushlokhande: This sizes the probe by the predicate's **global density**, but whether a match comes back depends on whether it lands in a probed cell — on how the survivors are **scattered**, not how many there are. A dense predicate whose query-relevant matches are few and spread out gets `mult` ~ 1 and still loses hits, so "Dense filters keep the floor" is the failure case rather than the safe one.

Repro against this branch, no local edits: 768 rows, 522 tagged, only 10 carrying dim-0 mass so the exact top-10 *is* those 10 rows. At `k = 10`:

```
allowed=522 population=768 mult=2 width=8
found 8/10   <-- fails
```

Forcing `width = 4096` in this same arm returns 10/10, so probe width is provably the binding constraint and this formula computes the wrong one. The new test passes because its predicate is sparse (10/256); raise the density and the original failure returns unchanged.

That makes the allow-bitmap-aware cell selection listed as a follow-up look like the actual fix rather than a cost optimization — a global-ratio multiplier cannot bound this error, because the quantity it measures is not the one that decides recall.

Nit on the same line, no behaviour change: `.max(1)` is dead code. `selectivity_mult_from_counts` returns 0 only for `allowed == 0`, and that cannot reach here — the allow-map builder drops empty bitmaps after tombstone subtraction (`.filter(|(_, bm)| !bm.is_empty())`), and this arm already sits behind `any_tagged`. Swapping the clamp for `assert_ne!(raw_mult, 0, ..)` and running `--lib` (2585 passed) plus `--test supertable` (343 passed) trips it zero times. For any `allowed > 0` the helper already returns >= 1, so the clamp can never change a value. Worth dropping; if the `EMPTY_FILTER_SELECTIVITY_MULT` sentinel is worth guarding at all, an assert states the invariant where a silent clamp hides it.

Can hand over the repro test if that is useful.

pratyushlokhande (commented):
Two correctness findings on the new routing arm, both validated against this branch. Confirmed first that the PR's own claims hold: the new test fails on the parent (8/10) and passes here, clippy is clean, and the supertable suite is 343/343.

src/supertable/query/vector.rs
muralikpbhat: Confirmed — fixed in 360666b7: both counts now come from this fan's superfiles only, and a zero-match tail skips the fan (no more `.max(1)` swallowing the empty sentinel). New test `filtered_knn_sizes_the_tail_probe_from_tail_counts_only` fails on the old counting, passes now. PR description corrected on the hidden-arm claim. Agreed on `n_docs`/tombstones — over-probe only, directional.

src/supertable/query/vector.rs
muralikpbhat: Agreed — the multiplier measures how many rows match, not where they sit, so it can't bound this. It fixes the fewer-than-k failure and restores parity with pre-drain unfiltered search; exactness needs filter-aware cell selection, on the drained path too. Filed https://github.com/infino-ai/infino/issues/780 with your repro as the acceptance test (low priority: the gap closes at the next drain). `.max(1)` fixed in 360666b7.

pratyushlokhande (approved):
LGTM

PR #784, opened 2026-09-17
fts: lazy bitset probes, anchor-driven two-term unions, cheaper top-k admission

vinaykakade:
## What changes

Ranked top-k over a single superfile spends its time in a few kernels; profiling the query shapes that trail on a 1M-doc corpus (a rare term unioned with a stopword, conjunctions of stopwords, two-token phrases) found four costs that were about how the work was done rather than how much of it there was:

- **Lazy bitset probes.** A skip into a bitset-encoded block expanded all 128 doc ids to serve one probe. The cursor now publishes the first set bit at or after the target as a one-doc block (masked `trailing_zeros` over the block's words) and reads its tf as one lane out of the packed tf array. A caller that keeps stepping or skipping inside the same block gets the full expansion on the second step; two expanded blocks in a row switch the cursor to eager expansion, and it retries a lazy probe every eighth block so a caller that turns sparse can leave that mode. This was tuned over five measured variants: a rare anchor's WAND touches each stopword block exactly twice, stopword conjunctions walk blocks by skips, and both had to come out ahead.
- **Two-term WAND tail.** Once the top-k threshold exceeds the weaker term's maximum score, every remaining candidate is a doc of the other term. The kernel now walks that term, checks the weak term's block bound and probes its tf, instead of re-sorting the pair, re-deriving the pivot and stepping the weak cursor past every scored doc.
- **Top-k admission in place.** Every kernel replaced the evicted heap entry with `pop()` then `push()` (two sifts); `peek_mut` does one. The two flat-merge loops and the conjunction's inner scan read subslices of the decoded prefix so the per-step bounds check folds into the loop bound; the block decode is no longer inlined into every per-doc step.
- **Short position runs.** Runs of up to four positions are read straight out of the two packed streams, exceptions applied per lane, bypassing the generic lane reader and its scratch.

Results and rankings are unchanged: the fuzz oracle passes, and every benchmark run compared the branch's output against main's with zero mismatches. Nothing filters or reweights any term.

## Measured

1M docs, single superfile, 962 queries, two reps with engine order reversed, per-query medians, this branch against `main` in the same run:

| lane | main | branch | ratio |
|---|---:|---:|---:|
| TOP_10 | 368 µs | 353 µs | 0.96 |
| TOP_100 | 477 µs | 455 µs | 0.95 |
| TOP_1000 | 735 µs | 706 µs | 0.96 |

Measured on 046bf695; the head adds only tests and helper extraction. Every query tag is at or under 1.0: unions 0.94–0.96 (two-token unions 0.84–0.94 in the previous round), phrases 0.95–0.96, intersections 0.98. Two rounds on consecutive heads agree within 1% on every lane. The shapes the changes target: `the movement` TOP_10 1275 → 304 µs, `the progressive` 1068 → 237, `the incredibles` 686 → 428, `"secretary of state"` 2308 → 1914; stopword conjunctions and multi-stopword unions within noise of main.

Two variants were tried and reverted on measurement: reading one tf lane on every probe (dense callers lost 5–7%), and a bisecting in-block skip (the windowed union lost 5%).

## Verification

`cargo test --lib` (2,634), `--test superfile` (197, includes the fuzz oracle), `--test supertable` (342), `clippy --all-targets -D warnings`, `fmt` with the CI import configuration — green on the head commit.

New tests, one per path the change adds:

- `lazy_bitset_skips_match_the_decoded_walk` — skips that land in a new block, inside the same block, and far ahead; then a 300-doc walk after a skip (crossing the expansion of the lazy block and into the next blocks); count-only and scored cursors.
- `dense_and_sparse_skip_patterns_match_the_decoded_walk` — one sequence of skips at stride 1, then one per block, then stride 2, then jumps past whole blocks: exercises the switch to eager expansion, the streak reset, and the periodic lazy retry.
- `lazy_block_expands_for_materialize_contains_and_walk` — a lazily entered block is expanded on demand by `materialize_at` (whole block, positioned on the doc, followed by a 200-doc walk), and by `contains` / `tf_at_contained`.
- `bitset_probe_helpers_agree_with_the_full_expansion` — `bitset_next_doc` and `bitset_tf_at` against the expanded arrays for every target from before the block's base to past its last doc, every tf width the packer produces, both header layouts.
- `wand_bmw_stopword_anchor_tail_agrees_with_bmm` — a term in every doc unioned with a rare term of varying tf, both term orders, k = 1 to 400 (including a k that never fills the heap, so the tail never triggers), against MaxScore+BMM.
- `short_runs_read_directly_agree_with_the_lane_reader` — position runs of one to four values with exceptions in the first stream, the gap stream, and both (including the block's first pair), against the generic lane reader.
- `decoders_refuse_a_header_width_past_32` — the width check moved from every header parse to the decoders; a compact header patched to width 33 still panics at decode.
- `and_heap_push_keeps_the_smaller_doc_on_a_tied_score` and `replace_worst_is_pop_then_push` — admission tie-break, and the in-place replacement producing the same heap as pop-then-push.
- `same_block_tf_probes_leave_the_lazy_doc_intact` — on a corpus whose tf varies within a block, a `tf_at_contained` / `bitset_probe_tf` probe into the block a cursor holds lazily must leave `current_doc_id()`/`current_tf()` intact and the walk on from there exact (review finding; the probe now reads its lane alone while the block is held lazily, and `current_tf()` asserts the buffers were not replaced).
- `wand_two_term_tail_fills_a_partial_heap` — the tail entered with an empty heap and a threshold seeded at the weak term's max score, k = 1, 5, 20: it must fill the heap before evicting and end with MaxScore+BMM's top-k (review finding; the tail now takes `k`).
- `decode_current_positions` asserts on entry that the member's block is fully decoded, the precondition the phrase path relies on `materialize_at` for (review finding).
- `bulk_resolve_matches_single_lookups_in_every_order_and_density` extended with a score-ordered hit list that touches every block out of order.


pratyushlokhande (commented):
Four notes on invariants this PR widens, two nits. None is a live bug — each is held off by a guard or convention elsewhere, none by an assertion.

src/superfile/fts/reader/cursor.rs
```diff
@@ -1167,6 +1215,51 @@ impl TermCursor {
...
+        let raw = &self.bytes[range];
+        self.block_doc_ids[0] = doc;
+        if !self.count_only {
+            self.block_tfs[0] = bitset_tf_at(raw, &hdr, rank);
+        }
+        self.block_n = 1;
+        self.pos = 0;
+        self.lazy_bit = doc - hdr.base;
+        self.decoded_block = usize::MAX;
+        self.tf_decoded_block = usize::MAX;
```
pratyushlokhande, line 1297: `publish_lazy` writes the lazy doc's tf into `block_tfs[0]` but leaves `tf_decoded_block = usize::MAX`. A same-block tf probe (`tf_at_contained`, `bitset_probe_tf`) then sees the mismatch, calls `decode_block_tfs`, and overwrites `block_tfs[0]` with the block's **rank-0** tf — while `block_doc_ids[0]` and `pos` still name the lazy doc. `current_tf()` then reports a different doc than `current_doc_id()`.

Reproduced on a corpus where tf varies *within* a block (`reps = 1 + doc_id % 5`):

```
skip_to(4_100)            -> block_n == 1, current_doc_id() == 4100, current_tf() == 1
contains(4_105); tf_at_contained(4_105)   // current_block unchanged
                          -> current_doc_id() == 4100, current_tf() == 2
```

2 is doc 4096's tf (rank 0). Same result via `bitset_probe_tf`.

Latent: every production probe site keeps the probed cursor disjoint from the stepped one. But that is held by the "must not also be iterated" convention in two doc comments and asserted nowhere — and before this PR `skip_to` always fully decoded, so `block_tfs` was always the whole block in doc order and this could not arise.

`lazy_block_expands_for_materialize_contains_and_walk` walks this exact sequence and cannot catch it: `two_column_reader`'s `common` tf is `1 + 2*((doc_id/128)%3 == 0)`, constant within a block, so rank-0's tf always equals every other doc's — and the test never re-reads `current_tf()` after the probe.

Fix: record the lazy rank and leave `tf_decoded_block` alone, or add `debug_assert!(self.lazy_bit == u32::MAX)` to `current_tf()` — the latter would have caught this.

src/superfile/fts/reader/cursor.rs
```diff
@@ -1167,6 +1215,51 @@ impl TermCursor {
...
+        let (hdr, range) = self.current_block_bytes();
+        let raw = &self.bytes[range];
+        self.block_doc_ids[0] = doc;
+        if !self.count_only {
+            self.block_tfs[0] = bitset_tf_at(raw, &hdr, rank);
+        }
+        self.block_n = 1;
+        self.pos = 0;
+        self.lazy_bit = doc - hdr.base;
+        self.decoded_block = usize::MAX;
```
pratyushlokhande, line 1296: This line is the only thing keeping the phrase path correct, and nothing says so.

`phrase.rs::decode_current_positions` reads `cursor.pos` as the in-block pair index and `cursor.block_tfs[..cursor.block_n]`, with no decoded-block check. With `block_n == 1, pos == 0` it reads the wrong position run and poisons the cached `group_block`. It survives only because `verify_at_aligned` calls `materialize_at` first, which re-expands *because* `decoded_block` is invalidated here.

Mutation test — removing both `materialize_at` calls: 4/8 lib and 10/15 integration phrase tests fail. Re-running with the lazy publication also short-circuited: 3 and 7 fail. So four tests break **only** when the lazy path is live:

- `position_groups_verify_phrases_in_long_and_short_terms`
- `must_phrase_skips_cooccurring_nonadjacent`
- `phrase_block_crossing_rejection`
- `phrase_oracle_agreement_multi_block`

Pre-PR a plain `skip_to` always left the block decoded, so the precondition was free. Now it depends on a caller two files away.

Fix: `debug_assert_eq!(self.cursor.decoded_block, self.cursor.current_block)` at the top of `decode_current_positions`.

Related: the comment at `phrase.rs:531` is now false — it says `materialize_at` is a no-op on the ranked path because `skip_to` already decoded the block. It doesn't any more, and that comment invites deleting the call.

src/superfile/fts/reader/scorers.rs
```diff
@@ -309,6 +310,36 @@ fn and_prefer_membership(has_bitset_blocks: bool, cursors: &[TermCursor]) -> boo
...
+        let norm = dl_norm_k1.get(doc);
+        let own = bm25::score_with_dl_norm_k1(strong.idf_weight, strong.current_tf(), norm);
+        weak.shallow_advance_block_to(doc);
+        if own + weak.inspect_block_max_bm25() > *threshold {
+            let mut score = own;
+            if let Some(tf) = weak.bitset_probe_tf(doc) {
+                score += bm25::score_with_dl_norm_k1(weak.idf_weight, tf, norm);
+            }
+            if score > *threshold {
+                replace_worst(heap, TopKEntry(score, doc));
```
pratyushlokhande: The tail only ever evicts, never `push`es, so it silently drops qualifying docs if entered with `heap.len() < k`.

Safe today: `threshold` is written in exactly three places in `run_wand_bmw` — the `0.0` init, and two sites both under `heap.len() == k` — and `replace_worst` does not change length, so `threshold > 0.0` implies `heap.len() == k`.

But the fn takes `heap` and `threshold` and not `k`, so nothing local says that. `run_wand_bmw`'s own doc comment records that seeding the pivot threshold from a floor was tried and reverted — that is exactly the change that arms this. Seeding it (`threshold = cursors.iter().map(|c| c.term_max_bm25).fold(f32::INFINITY, f32::min)`) gives:

```
panicked at scorers.rs:336: non-empty              # empty heap
tail kept [(111, 2.23)] but 5 docs qualify (k = 5) # partial heap
```

Fix: pass `k` and branch `if heap.len() < k { heap.push(..) } else { replace_worst(..) }`, or `debug_assert!(heap.len() == k)` at the top of the fn.

src/superfile/fts/reader/scorers.rs
```diff
@@ -309,6 +310,36 @@ fn and_prefer_membership(has_bitset_blocks: bool, cursors: &[TermCursor]) -> boo
...
+/// when the block bound allows and `weak` holds the doc. `weak` is only
+/// probed, never stepped, so a stopword's bitset blocks are never expanded.
+fn wand_two_term_tail(
+    weak: &mut TermCursor,
+    strong: &mut TermCursor,
+    heap: &mut BinaryHeap<TopKEntry>,
+    threshold: &mut f32,
+    dl_norm_k1: &NormTable,
+) {
+    while !strong.is_exhausted() {
```
pratyushlokhande, line 331: nit: no block skip on `strong`, and `own` forces a decode before the bound gate — 19,995 docs scored regardless of selectivity, vs 640 for BMM. Only safe because `two_term_has_rare_anchor` (df ratio 16) guarantees `strong` is the rare side; worth a line saying so.

src/superfile/fts/posting.rs
```diff
@@ -804,6 +804,77 @@ pub fn decode_block_tfs(bytes: &[u8], hdr: &BlockHeader, dest_tfs: &mut [u32]) {
...
+/// The first set bit of a bitset block at or after `from`, as `(doc, rank)`
+/// where `rank` is the number of set bits before it — the doc's index in
+/// the block's tf array. `None` when no doc of the block is `>= from`.
+///
+/// A probe that skips into a bitset block needs one doc, not the 128 the
+/// full expansion writes; this scans the block's words from the target's
+/// word with a masked `trailing_zeros`, at most a handful of u64s.
+pub fn bitset_next_doc(bytes: &[u8], hdr: &BlockHeader, from: u32) -> Option<(u32, usize)> {
+    debug_assert_eq!(hdr.encoding, ENCODING_BITSET);
+    let words = &bytes[hdr.payload()..bytes.len() - hdr.tfs_size()];
```
pratyushlokhande: nit: missing `hdr.check_widths()` and the length assert its siblings have — `tf_bits = 33` underflows here and panics `attempt to subtract with overflow` instead of `block header: bit width > 32`.

src/superfile/fts/reader/sink.rs
```diff
@@ -301,18 +301,59 @@ pub(super) fn and_heap_push(
...
+        *worst = TopKEntry(score, doc_id);
+    }
+}
+
+/// Replace the heap's worst entry in place — one sift, where a `pop`
+/// followed by a `push` costs a full sift-down and two sift-ups. The
+/// caller has already decided `entry` belongs in the top-k.
+#[inline]
+pub(super) fn replace_worst(heap: &mut BinaryHeap<TopKEntry>, entry: TopKEntry) {
+    if let Some(mut worst) = heap.peek_mut() {
```
pratyushlokhande, line 319: nit: no-op on an empty heap where the `pop()` + `push()` it replaces inserted, and every call site follows with `.expect("non-empty")`. Unreachable today (all 8 sites guarded); a `debug_assert!(!heap.is_empty())` would make the precondition explicit.

pratyushlokhande (approved):
Added few comments. ok to merge post the above comments are resolved.

src/superfile/fts/reader/cursor.rs
vinaykakade: Right — the tf array was being unpacked over the lazily published doc, and the convention that probed cursors are never stepped was the only thing hiding it. Fixed in 63057c90 the first way you suggest, plus the assertion: a tf probe (`tf_at_contained`, `bitset_probe_tf`) into a block that is held lazily reads its lane alone and leaves the buffers as they are (`bitset_tf_for_probe`), `current_tf()` debug-asserts that a lazy block’s tf buffer was not replaced by a whole-block unpack, and `same_block_tf_probes_leave_the_lazy_doc_intact` reproduces your sequence on a corpus with tf = 1 + doc % 5 (both probe paths, then a 300-doc walk from the lazy doc). No kernel takes this path today, so no cost change.

src/superfile/fts/reader/cursor.rs
vinaykakade: Agreed on both counts. In the same push: `decode_current_positions` now states the precondition in its doc comment and `debug_assert_eq!`s `decoded_block == current_block` on entry (after the inline-position early return); `publish_lazy` documents why it invalidates `decoded_block` and points at that assertion; and the comment at the `materialize_at` calls now says the opposite of what it said — on the ranked path a skip into a bitset block holds one doc, so those calls are not optional. Your mutation (dropping the two `materialize_at` calls) now fails the assertion in debug builds instead of reading the wrong run.

src/superfile/fts/reader/scorers.rs
vinaykakade: Yes — the tail assumed a full heap and nothing local said so. Fixed in the latest push: `wand_two_term_tail` takes `k`, pushes while the heap has room, replaces the worst entry once it holds `k`, and follows the heap with the threshold only from that point. `wand_two_term_tail_fills_a_partial_heap` runs your mutation directly — empty heap, threshold seeded at the weak term’s max score — for k = 1, 5, 20 and checks the result fills the heap and equals MaxScore+BMM.

src/superfile/fts/reader/scorers.rs

src/superfile/fts/posting.rs
vinaykakade: Done in the latest push: `bitset_next_doc` now calls `hdr.check_widths()` and asserts the header+tfs length before slicing, the same as `bitset_tf_at`; `bitset_probe_refuses_a_header_width_past_32` patches a bitset header to tf width 33 and expects the width panic rather than the subtraction overflow.

src/superfile/fts/reader/sink.rs
vinaykakade: Added in the latest push: `debug_assert!(!heap.is_empty())` at the top of `replace_worst`, with the doc comment now stating that, unlike pop-then-push, it does not insert into an empty heap.

PR #785, opened 2026-09-17
fix: write block idx file without any buffer

savannahar68:


savannahar68:
@naman-modi jfyi

Pranav2612000 (approved):
Now that we persist everytime, what are some things we expect to get worse?

Good to merge after adding PR description

PR #786, opened 2026-09-17
chore: add logs wherever we evict cached files

savannahar68:
* add logs wherever we try to evict files form cache
* this is to debug tierd cache logic


Pranav2612000 (approved):
Please add PR description before merge.

PR #787, opened 2026-09-18
fts: block-level bounds in the union window and a block-at-a-time phrase walk

vinaykakade:
## Summary

Three ranked kernels use per-block bounds where they used term-wide ones or none, and the phrase walk aligns its members a block at a time.

- **Windowed union: per-window block-level partition.** A term whose block maxima over the window cannot lift the weaker terms past the threshold is treated as non-essential for that window and only probed for the other terms' candidates; a window whose whole bound is under the threshold is skipped outright. Two guards, both from measurement: the partition is gated on the weakest essential term alone (the suffix sums only grow towards the stronger terms), and a term is demoted only when its estimated docs in the window are at least eight times the remaining essentials' — demoting a mid-frequency term trades SIMD accumulation of its docs for a probe per candidate, which only pays for the stopword shape. A stopword next to a rare term at large k was being scored doc by doc over every block; that query is 3.3× faster.
- **Phrase: block-level bound, once per block.** The rarest member's block is bounded by the members' block maxima scaled to the phrase idf and skipped whole when under the bar. Bites at small k, neutral above.
- **Phrase: block-at-a-time alignment.** The rarest member's current block is the candidate list; every other member filters the whole list in one pass over its own blocks (a bit test per doc on a dense block, one decode and a merge on a packed one, via a second cursor per member), and only the survivors are positioned — by a bare block seek — for their tf and positions. Leftovers serve the next call. Both the ranked and the unranked (count) walks use it.

Two alternatives were built, measured, and reverted inside this branch, and the commit history keeps their reasons: a bisecting/galloping packed-block tf probe (the completion pass probes a few docs apart, where a predictable scan beats mispredicted halvings; a quarter slower on a long query) and a block-at-a-time membership conjunction (its per-candidate cost is already one bit test; a fifth slower). The membership probe likewise stays on a bisection, measured against a scan from the last position.

## Measurements

5.3M-doc corpus, 962 queries, 10 iterations, two reps with reversed engine order, per-query medians, branch vs `main` at `437cb2e9` on the same index; zero result mismatches.

| lane | mean, branch / main |
|---|---:|
| TOP_10 | 0.91 |
| TOP_100 | 0.93 |
| TOP_1000 | 0.93 |
| TOP_100_COUNT | 0.97 |
| COUNT | 1.01 |

Per shape vs main: phrases 0.80 / 0.84 / 0.85 at TOP_10 / TOP_100 / TOP_1000 (two-token 0.78 at TOP_10); two-token unions 0.88 at TOP_1000; intersections, other unions and counts within ±2%. Geometric mean of per-query ratios 0.981. The stopword-plus-rare TOP_1000 shape goes from 27 ms to 6 ms.

## Verification

- New tests: the windowed union against per-candidate MaxScore on a corpus that demotes the stopword per window and skips windows whole once the rare term exhausts; packed-block probes against the walk across blocks; ranked phrase pruning against the unpruned walk; the batched ranked phrase walk against the unranked doc-at-a-time walk (consecutive, random skips, under a bar) on a corpus whose rarest member's blocks span dozens of the others' blocks.
- Every commit: FTS unit tests (378), the superfile fuzz oracle (197), `make check`.
- Reader-only change; no format change.


pratyushlokhande (approved):
LGTM

PR #788, opened 2026-09-18
bench: exact BM25 quality oracle, superfile quality phase, realistic text by default

vinaykakade:
## What

Makes the FTS quality bench's engine-model reference exact, adds the same quality phase to the superfile FTS cell, and switches the two FTS cells to the realistic text corpus by default. Bench and test code only; the two engine additions are reachable only under `test-helpers`.

### Exact engine-model reference

The quality phase graded the engine against a reference that normalized every document with the corpus-wide average length. The engine scores each superfile at the table-wide average as of that file's commit (all files committed before it, plus its own), so the gated columns carried a 5% tolerance. That tolerance also let a wrong document within 5% of the k-th score pass.

The reference now reproduces the engine's inputs and verifies them against the table:

- **Row-to-superfile mapping.** Manifest entries ordered by `id_min`, each assigned its `n_docs` rows contiguously. This is verified, not assumed: every file's stored document lengths (`FtsReader::column_doc_lengths`, new, test-visible) must equal the oracle's lengths for those rows, element for element. A mismatch is a hard error naming the file and row.
- **Declared average.** Read from each file's reader (`SupertableReader::open_superfile`, new, test-visible: the same tiered open the query fan-out uses) and checked against the writer's rule. Files of one commit share their "committed before" totals and the manifest does not record commit boundaries, so the grouping is derived: a file whose declared average matches the rule with every earlier file counted as committed starts a commit; one that matches only with the current commit's earlier files excluded continues it. Neither matching is a hard error.
- **Gates.** Recall vs engine BM25 must be 1.0 at the tie tolerance (1e-4 relative); max score delta ceiling is 1e-3. The slack, its scale rule and constants are removed. The module doc and report note are rewritten; they previously still described per-superfile normalization.

### Superfile quality phase

`cargo bench -- superfile fts quality` grades the cell's exact search artifact through `bm25_hits_async`, after verifying the file's lengths and declared average. With one file the reference is exact by construction, so this is where a kernel change gets graded first. The grading is shared between the tiers (`grade_with` over a hit source); the per-superfile-statistics column is omitted on the single-file tier. `search` now selects warm, cold and quality on both argument grammars.

### Realistic corpus by default for the FTS cells

`corpus::default_text_flavor_for_fts` sets a process default that the two FTS cells (and the FTS ingest-shape child) call before generating anything. An explicit `corpus=` still wins; `corpus=synthetic` now pins the uniform flavour explicitly. SQL and combined shapes keep the uniform default, so their history is unchanged. The FTS speed tables' history resets; their titles now name the corpus flavour.

Also fixed: the isolated ingest child never received the parent's `corpus=` tokens, so with `corpus=realistic` it ingested the uniform corpus under the realistic label. And the bench-utils crate's unit tests did not compile on `main` (a test helper predated the `Phrase` type from the analysis-chain change).

### CI

The PR bench's `superfile` leg runs every phase, so with the new default the single-file BM25 oracle gates every bench run alongside the supertable one in `supertable_fts_warm`. The FTS legs' explicit `corpus: realistic` is kept (now redundant) with an updated comment.

## Verification

- `cargo test -p infino-bench-utils --lib`: 99 passed. New test ingests the 600-doc realistic corpus in three commits with a two-thread writer and the byte split target disabled (two shards per commit), shows the corpus-wide average is measurably inexact there, recovers three commits and six files, and reaches recall 1.0 with deltas inside f32 noise on every shape.
- End to end, `supertable fts quality`: 100K docs (16 superfiles, one per commit) and 1M docs (75 superfiles from 16 commits, exercising the same-commit branch). Every cell recall 1.0, max delta 0.0000%.
- End to end, `superfile fts quality`: 100K and 1M docs. Every cell recall 1.0, max delta 0.0000%. The 1M run wrote the new README section.
- `cargo clippy --all-targets --features test-helpers -- -D warnings` clean; `cargo fmt` check clean; plain `cargo check --lib` warning-free (the new imports are cfg-gated).

## Performance

No query-path change; nothing to compare. The engine additions are test-visible accessors. The superfile FTS cell's fixture builder now takes the process flavour, which is the same bytes as before under `corpus=synthetic`.

## Follow-up

The README's supertable quality table still shows the 10M numbers recorded before the statistics correction (with the old 5% note). Regenerating it needs a 10M run; the superfile table in this PR is at its 1M reference scale.


pratyushlokhande (approved):

vinaykakade:
The bench differences in superfile vs main are due to changing the corpus to "realistic" - force merging even if there is superfile bench failure, as the corpus being measured is different.

Issue #783, opened 2026-09-17
vector: optimize() cost grows ~O(N) per call (quadratic over incremental ingest); drain scratch scales with the ingest delta


muralikpbhat:
## Summary

Billion-scale incremental ingest (append batches, then `optimize()` every K rows) surfaces two distinct scaling limitations in `optimize()` / compaction. Both may be partly by design, but there look to be real improvement opportunities.

1. **TIME:** each `optimize()` is ~O(N) in *total* rows (measured through 450 M), so the aggregate over an incremental build is ~**quadratic** (consistent with the measured cumulative).
2. **DISK:** the drain phase materializes the *entire* ingest delta to local scratch before uploading, so peak scratch is ∝ delta → disk pressure / failure at large deltas.

## Data — TIME (96-d cosine, ~billion-scale, `optimize()` every 50M rows, 16 vCPU / 64 GB, object store + disk cache)

| rows | optimize() | rows | optimize() |
|---|---|---|---|
| 50M | 1450 s | 300M | 9423 s |
| 100M | 2234 s | 350M | 11778 s |
| 150M | 3621 s | 400M | 14472 s |
| 200M | 5479 s | 450M | **16836 s (4.7 h)** |
| 250M | 7451 s | | |

- **Measured (50 M→450 M):** per-`optimize()` grows ~linearly in N — ~**+2400 s per 50 M step** (cost rose 11.6× while N rose 9×).
- **Aggregate is ~quadratic, and the measured cumulative bears this out:** total optimize time is 3684 s at 100 M vs **55908 s at 400 M** — a **15.2× rise for a 4× rise in N** (a clean quadratic would be 16×). By 450 M, ~**90 % of wall-clock is optimize** (ingest ~2.3 h vs cumulative optimize ~20 h).
- **Projected (beyond 450 M, not yet reached):** extrapolating the trend → ~5 h/optimize at 500 M, ~**12 h at 1 B**, full 1 B incremental build ~5–6 days.

## Phase attribution (per-phase timers + code reading; billion-scale breakdown to follow)

Instrumented `optimize()` with per-phase timers (drain / split / merge / recalibrate / router-cache). From small-scale runs and the code:

- **drain**: ~O(delta), roughly constant per step — not the growth.
- **`recalibrate_probe_laws`**: one full sweep scoring a **256-query sample** (`WIDTH_LAW_QUERY_SAMPLE`) over all cells — O(N) *read* but light compute (~minutes even at 250 M). Not the dominant cost.
[...]

PR #790, opened 2026-09-18
vector: graph-route drain cell assignment + drain phase instrumentation

muralikpbhat:
Part of infino-ai/infino#783.

## What

Three related changes to the hidden vector index's `optimize()`/drain:

**1. Graph-routed drain cell assignment.** Assigning each drained row to its nearest cell currently scores the row's 1-bit code against *every* cell centroid (a Hamming shortlist over the whole grid) then exact-rescores a 20%-of-grid window — O(n_cells) per row. This routes each row through an HNSW built once per drain over the cell centroids, visiting ~O(log n_cells) candidates and exact-scoring only those. The router HNSW is built serially (deterministic insert order) so placement is reproducible — see the Placement-determinism note under Correctness. Placement is unchanged: candidates are re-scored with the identical exact distance and fed to the same nearest-two + Voronoi replica closure, so the primary + replica set matches the current path.

**2. Drain sub-phase instrumentation.** Adds debug-gated (`diagnostics.drain_build_timers`) timers splitting the drain into materialize / assign / assemble / freeze / upload, so each phase is independently measurable.

**3. Deterministic width-law calibration sampling.** The drain's width/probe-law calibration draws its stand-in queries by reservoir-sampling rows in the order they are offered during the drain. That offer order is not stable run to run (materialization/enumeration order varies), so the sampled query set — and therefore the stamped width law that sets the cold-read fanout — drifted between otherwise-identical builds. This selects the calibration queries by a content min-hash (keep the `k` smallest-hash rows) so the query set is a function of the rows' contents alone, independent of offer order. The stamped law is now reproducible.

## Why

Per-`optimize()` cost grows with total rows (infino-ai/infino#783). Per-row cell assignment scales with the cell grid; routing it through a graph makes it ~O(log n_cells) while keeping placement exact-scored. (3) is a correctness prerequisite the graph placement surfaced — see Correctness.

## Correctness

- **Placement parity** with the exact path: graph candidates are re-scored with the identical exact distance, so the value fed to the replica closure is bit-identical; only which candidates are surfaced differs, and `ef` is sized to keep that recall high.
- **Empty cells** (count-0) are excluded from the graph via a node→cell map, exactly as the shortlist path excludes them — the graph can never route into an empty cell after a split creates them.
- **Cosine-gated.** Graph-assign runs only for cosine; L2Sq/NegDot take the exact shortlist path. On those metrics the coarse centroid HNSW is hub-dominated for the varying-norm geometry and **collapses the grid at assign time** — most cells starve and one becomes a 100k+ catch-all. recall@10 tolerates the catch-all, but *filtered* recall craters (1.00 → 0.68 at 1M, measured A/B on the real corpus), while the exact path stays balanced (filtered 1.00). It is a build-level property, not a beam-width one: it persists even at exhaustive ef. Non-cosine graph-assign returns in a follow-up once the coarse-router build is made metric-robust.
- **`ef` (cosine):** `ef(n) = max(16, round(√n/4))` → 16/16/16/32 at n_cent 256/1K/4K/16K. The floor of 16 (vs a bare √n/4 of 4–8 at small grids) holds primary-placement parity high enough that the post-drain cold-read block layout matches the exact path's.
- **Placement determinism.** The coarse-router HNSW is built with a serial (deterministic) insert order rather than the parallel build. A parallel build's concurrent-insert reordering shifts which candidates the graph surfaces for a thin tail of boundary rows, moving their placement between otherwise-identical drains — that instability collapsed *filtered* recall on the exact-parity metric. Serial insertion over these few thousand centroids is negligible and makes placement bit-reproducible. This is the placement half of the determinism; the calibration half is next.
- **Calibration determinism.** The graph's tighter placement nudges a thin tail of boundary rows to different cells than the exact path, which was enough to make the *pre-existing* per-run drift in the stamped width law cross the bench's cold-GET ceiling / filtered-recall floor — surfacing as a run-to-run flake rather than a fixed regression. Change (3) removes the drift at its source (order-independent calibration sampling); the stamped law, and thus the cold-read fanout, is now reproducible across identical builds.
- Behind `vector.drain_graph_assign` (default on) with automatic fallback to the exact path for very small grids (n_cent < 64) and for all non-cosine metrics.

## Measurement

**Isolated assign-kernel A/B** (768-d, synthetic grids; speedup + primary parity, which transfer across hardware):

| grid (cells) | assign speedup | primary parity |
|---|--:|--:|
| 256 | 1.44× | 0.994 |
| 1,024 | 2.77× | 0.997 |
| 4,096 | 8.9× | 0.997 |
| 16,384 | 17.3× | 0.990 |

**End-to-end 1M-row 768-d drain, paired (knob off vs on):** coarse cell-assign **3846 → 1619 ms (2.38×)**, recall@10 **0.9962 → 0.9986** (parity), grid 256.

## Scope (honest)

This addresses the **coarse cell-assignment slice** of the drain. At real scale that slice is ~10% of the drain and the cell grid grows slowly (256 default), so the assign win widens only gradually with scale. Instrumentation of the full drain (this PR) shows the drain is dominated by **assemble** (superfile encode/splice/write), with a further ∝N cost in the settle step still being characterized — those are larger, separate levers, tracked under #783. This PR does **not** touch the within-cell k-means build, the probe-law recalibration, or split — each is its own follow-up.

## Testing

`cargo fmt`, `cargo clippy` clean; unit tests for per-metric placement parity vs an exact reference, the empty-cell (node→cell) regression, the `ef` knee formula, the config toggle, and calibration-query determinism. The `supertable_vector` benches — which previously flaked run-to-run on the cold-GET ceiling and filtered-recall floor — now pass across three consecutive independent runs with change (3), on both the cosine (`st-vec`, where the flake surfaced) and L2 (`st-vec-l2`) shapes. The full billion-scale drain breakdown, including the larger assemble and settle levers this PR does not touch, is tracked in #783.


muralikpbhat:
Please hold review — the `supertable_vector` bench gate is flaking run-to-run (filtered recall@10 and a steady cold-GET ceiling), not a fixed regression: it traces to a pre-existing non-determinism in the drain's fine-clustering/calibration (the stamped width/fanout law varies per run) that this PR's graph-routed placement pushes over the gate threshold. Fixing that determinism first; will mark ready once it's green.

muralikpbhat:
Update: root-caused and fixed. The flake had two determinism roots in the drain, both fixed in this PR:

- **Placement.** The coarse-router HNSW was built in parallel; the concurrent-insert reordering moved a thin tail of boundary rows to different cells between otherwise-identical drains — enough to swing the filtered-recall floor. It is now built serially (deterministic insert order), so placement is bit-reproducible.
- **Calibration.** The width/probe-law calibration sampled its stand-in queries by reservoir-sampling rows in the order they are offered during the drain, and that offer order is not stable run to run — so the stamped width law (which sets the cold-read fanout) drifted between identical builds, swinging the cold-GET ceiling. The calibration queries are now chosen by a content min-hash (the `k` smallest-hash rows), so the query set — and the stamped law — is a deterministic function of the corpus, independent of offer order.

Both are pre-existing non-determinism this PR's tighter placement pushed across the gate; neither is a fixed regression. The `supertable_vector` benches now pass across three consecutive independent runs (the cosine `st-vec` shape, where the flake surfaced, and the L2 `st-vec-l2` shape). Ready for review.

savannahar68 (approved):
looks good
any idea why it didn't work with non cosine based

muralikpbhat:
> looks good any idea why it didn't work with non cosine based

to be honest, couldn't root cause that well and felt I could come back to it later given other priorities and hence implemented the gate for now.

PR #791, opened 2026-09-18
bench, tests: grade every posting form and routing branch, and the boundaries between them

vinaykakade:
## What

Expands BM25 accuracy coverage in three disjoint directions, on top of #788 (the exact quality oracle and the superfile quality phase). **Stacked on #788 — merge that first**; until it lands, this PR's diff includes its two commits.

Test and bench code only. The one engine addition is an accessor reachable only under `test-helpers`.

### 1. Quality battery: every posting form, every routing branch

The battery goes from 18 shapes to 58, chosen by document-frequency tier and clause shape so that at both reference scales it reaches every posting form and every routing branch, rather than sampling three tiers as before:

- **Terms** at ranks 1, 2, 30, 31, 100, 200, 300, 2000-2002, 5000, 20000, 40000 and 200000+, plus the per-doc token and a digit token. These land on bitset, packed, patched, short-form and inline lists; on lists above and below the seed floor; on lists with and without a coarse table; and, at rank 40000, on a term that a single 1M-doc file stores long form while a fragmented table stores it short form in every file, so the short-form boundary is graded from both sides.
- **Unions** keyed to the two-term anchor ratio (so one shape routes to WAND below the k cutoff and MaxScore above it), the union-count dominance threshold, and the full-bitset density gate; plus a union of ten short-form lists and a five-tier union.
- **Intersections** keyed to the membership walk's three sparsity tiers, including the middle tier both with a dense companion (routes to the walk) and without (stays on the flat merge); plus dense-dense, short-form-straddles-a-fan-out, inline-with-bitset, and a disjoint pair for the empty-result path.
- **Negation**, previously one shape, now seven: dense survivor set, tiny survivor set, negated short-form list, negated phrase, two negatives under intersection, and must/should/negative together.
- **Phrases**, previously two, now eight: a short-form member's inline positions group beside a bitset member, three tokens, a repeated term with overlapping starts, an empty-result pair, a phrase intersected with a term, and two phrases in one union.
- **Analyzer**: two shapes whose query text is capitalized or punctuated, which must grade identically to the bare term.
- **Parameters**: one shape at a query-time `(k1, b)` pair, scored by the oracle at the same pair. This is the path that rescales the stored per-block bounds; a rescale that came out too small prunes a qualifying document, which only an oracle at that pair can see.

`k` becomes {1, 10, 100, 128, 1000}. `k = 1` is the sharpest ranking check (the single best document, with no tie slack to hide in); `k = 128` is the largest `k` at which a two-term union routes to WAND, so it and 1000 bracket that dispatch.

### 2. Two more gated columns, and a tripwire on the textbook one

- **`count`** compares the unranked count kernels against the oracle's match count wherever the tier's count surface expresses the shape. The anchored, bitset and phrase count kernels were previously graded only by unit and small-oracle tests, never at bench scale against a whole-corpus reference.
- **`order`** requires hits to come back in descending engine score and adjacent pairs to agree with the reference's order beyond the tie tolerance. The score-delta gate bounds each score; this bounds the sequence a caller actually receives.
- The **textbook column** gains a floor of 0.80 from `k = 100`. It measures a design cost, not a bug, so the floor is a regression tripwire set below the recorded values with margin; at smaller `k` one boundary tie moves recall by a tenth, so those rows report only.

### 3. A coverage table, so the battery's claim is verified not assumed

A new section per tier lists the posting form every battery term took in every graded file: files holding it, its summed document frequency, long-form blocks by encoding, and the files where it took the short or inline form or carries a coarse table. It reads through a new test-visible `FtsReader::term_layout`.

This is what keeps the battery honest. The tier table above is derived from Zipf ranks and format thresholds; a corpus change, a format version or a threshold move could silently take a term off its intended form, and the recall numbers would still look fine. Now that shows up as a row in the table.

The superfile tier has no public entry taking a query-time parameter override, so that one shape is skipped there and named on stderr: 57 shapes graded on the superfile tier, 58 on the supertable tier.

### 4. Integration tests at the boundaries (`tests/`)

A new `tests/superfile/fts/boundaries.rs` plants one list on each encoding or kernel edge and grades every shape by match set and by score: a term at the short-form cap and one past it; patched blocks with a delta exception, a tf exception, and 31 versus 32 outlier lanes (the exception cap and the fallback past it); lists at and past the builder's patching cap; a bitset block whose word-aligned origin sits below its first document; lists whose density changes mid-way in both directions; ranked anchors in the last partial block and in the first block of the second coarse span; a phrase member positioned many blocks past its driver; and negation probes landing on a block's last document and past the list's end.

The fuzz oracle gains a tiered vocabulary spanning every encoding at once, pins `k` to 1, 128 and 129 alongside the uniform draw, and an opt-in large tier that reaches past the single-term seed floor.

The supertable oracle gains a per-file-average arm: each file's rows are recovered and verified against the corpus, its declared average checked against the commit rule (and against the value that counting an earlier same-commit flush would wrongly have given), then negation, must/should, phrase, union and intersection shapes graded against a table-wide-idf reference re-pointed at that file's average. Two arms, one file per commit and two files per commit.

### 5. Unit tests where there were none (`src/superfile/fts/`)

The cursor, filter and phrase modules had no unit tests at all. They now have property and planted coverage: seeks landing on the first document at every block edge of every encoding (planted and random, including the short-form and inline cursors), the eager-to-lazy probe retry, negation probes at block edges and past the list, a phrase member positioned by block seek, overlapping starts of a repeated word, and a three-member phrase rejecting a missing member.

The scorers' Block-Max-AND and two-term WAND bounds are property-tested as never below a real score, over random lengths and parameter pairs. A bound below a real score is silent recall loss, the one failure mode the end-to-end oracles can only approach indirectly.

The posting codec gains delta and tf exceptions coexisting in one block, the 31-versus-32 exception limit, and bitset probes on 64-bit word edges; `stored_avgdl` gains a rounding-boundary round trip.

No property failed against the engine. Three first-run failures were fixture premises (a planted list whose first lane became an unplanned exception, and a block-alignment assumption), since corrected.

## Verification

- Quality phases, both tiers at 100K docs and the superfile tier at 1M: every cell inside every gate, including the two new ones.
- `cargo test -p infino-bench-utils --lib`: 100 passed.
- `cargo test --lib superfile::fts::`: 388 passed.
- `cargo test --features test-helpers --test superfile fts::`: 135 passed, 1 opt-in ignored.
- `cargo test --features test-helpers --test supertable query::brute_force_oracle::`: 19 passed.
- `cargo clippy --all-targets --features test-helpers -- -D warnings`, the fmt check, and `make public-api`: all clean.

## Performance

No query-path change; nothing to compare. The battery is larger, so a quality phase costs more oracle atoms and more engine queries, which is bench time rather than product latency.

## Follow-up

The README's supertable quality table is still the one recorded before the statistics correction. Regenerating it needs a 10M run; the superfile tables here are at their 1M reference scale.

## Duplication pass

The three strands above were written in separate worktrees, so a review pass over the combined result found four copies of things that already existed once. All are consolidated here:

- **A private pair of token and scored-document totals** in the supertable oracle was the engine's `ColumnLengthStats` rewritten, and it had already drifted: it averaged in `f32` where the writer averages in `f64`, and mapped an empty column to a length of one where the engine reports zero. Nothing failed at the fixture's size, but a test that verifies the writer's rule has to do the rule's arithmetic the writer's way.
- **The clause-split-and-own block** that `BruteForceBm25::top_k_query` was added to replace survived in the bench's oracle test, and the boundary tests hand-rolled the phrase suite's `search_hits`. Both call the shared helper now.
- **The manifest's two ways of exposing its entries** (the flat view, or its parts when lazy) were walked in two places; one collector knows that now.
- **Both quality tiers verified a file the same way** — read its stored lengths, check them against the oracle, read the average it declares. That is one helper, which also removes the only identical block between the two tier entry points.
- **Four suites carried the same whitespace-split match-set helper**, three of them predating this work. These are the independent side of every assertion in those suites, so copies drifting apart would let neighbouring suites grade against different notions of a token. One generic copy now.

Two candidates were judged and deliberately left: the oracle restating the BM25 formula rather than sharing code with the kernels it grades (that is the point of an oracle, and it is documented), and the boundary tests restating the encoder's exception and patching caps as literals rather than importing them (a fixture pinned to the number fails when the number moves, which is what it is for).

One structural item is deferred: the commit rule is now implemented twice, once in the bench to *derive* commit boundaries from declared averages and once in the tests to *assert* them against a known fixture. The two cannot share code across the crate boundary today; hoisting the shared half into `test_helpers` is worth doing when one of them next changes.


tests/superfile/fts/fuzz_oracle.rs
```diff
@@ -139,6 +176,53 @@ fn skewed_corpus_strategy() -> impl Strategy<Value = Vec<Vec<usize>>> {
     prop::collection::vec(doc, 1..=max_docs())
 }
 
+/// A corpus whose vocabulary spans every posting encoding at once:
+/// per token draw the first four vocabulary terms are weighted by
+/// [`TIER_WEIGHT_DENSE`] down to [`TIER_WEIGHT_SHORT`], and the fifth
+/// is planted once, in the first document, after generation. The rest
+/// of the vocabulary never appears, so a query naming it hits the
+/// absent-term paths. `n_docs` sets the scale: at the CI default the
+/// tiers land on bitset, packed multi-block, single long-form block,
```
pratyushlokhande: At 400 docs, any term common enough to fill a block is also dense enough to get stored as a bitset — I measured zero packed and zero patched blocks across three seeds, so this lane has the same coverage as the two that already existed. Either reword this to what it actually builds, or plant one tier on a stride so the sparser formats show up.

tests/superfile/fts/fuzz_oracle.rs
```diff
@@ -72,6 +73,42 @@ const VOCAB: &[&str] = &[
...
+const TIER_MULTI_BLOCK: usize = 1;
+const TIER_ONE_BLOCK: usize = 2;
+const TIER_SHORT: usize = 3;
+const TIER_SINGLETON: usize = 4;
+
+/// Smallest corpus of the large tiered tier: past the single-term
+/// walk's seed floor of 256 blocks (32 768 postings), so the dense
+/// tier's list is seeded from its coarse table and skipped span by
+/// span, which no default-size case reaches.
+const LARGE_TIER_MIN_DOCS: usize = 33_000;
```
pratyushlokhande: Even at the top of the range (33,256 docs) the dense term only reaches 251 blocks, so it never crosses the 256 the seed path needs — this test can't hit the thing it was written for. Wants ~34,000, and planting the dense term in every doc instead of by weighted draw so the block count is predictable rather than sitting just under the line.

pratyushlokhande (commented):
Two notes on the fuzz oracle — the boundary tests look solid.

vinaykakade:
You're right, and it was worse than the doc being wrong. I measured the lane with the layout accessor this PR adds: across three seeds, every long-form tier is bitset, zero packed and zero patched. `gamma` isn't the "single long-form block" the doc claims either, it's two bitset blocks.

The cause is structural rather than a bad weight, which is why rewording alone would have been the wrong fix. A drawn tier dense enough to fill a block is dense enough for the bitset to claim it, and lowering a weight until its deltas pack narrower drops the term below the 128-posting cap on the way. Random placement is the reason: the widest gap in a block sets the delta width for all 128 lanes, and at these densities that gap is wide.

So I took your second option. The delta-coded formats now come from planted strides, whose gaps are uniform by construction: one tier at stride 3, the narrowest stride whose deltas beat a presence bitset (at stride 2 the 128 postings span 254 documents, which is four 64-bit words, exactly what two-bit deltas cost, and the encoder keeps the bitset on a tie). Every seed now gives a packed block deterministically.

Patched I could not reach at this corpus size, so the note now says so rather than claiming it. A block uneven enough to be worth patching spans enough documents that its presence bits are cheaper still; escaping that needs both a wide span and 128 postings to fill it, and a few hundred documents cannot supply both at once. The jumped tier reaches patched in the large tier, which I verified, and the planted corpora in `boundaries.rs` cover it exactly either way.

Two bugs fell out of writing this, both yours to thank for:

- The lane planted tokens by evicting a drawn one when a document was at the length cap, so a second plant into the same document evicted the first. The singleton was missing entirely from two of three seeds, and the strided tier's document frequency drifted below its stride. Drawn lengths now leave room for the planted tokens.
- The default lane drew corpora from one document upward, where the strided tier cannot reach the short-form cap at all. It now draws from the range where the forms it grades are actually present.

Both claims are asserted against the built index now, at each size, so the table cannot drift from what the generator does again. That assertion is the part I should have written in the first place.

tests/superfile/fts/fuzz_oracle.rs
vinaykakade: You're right, and it was worse than the doc being wrong. I measured the lane with the layout accessor this PR adds: across three seeds, every long-form tier is bitset, zero packed and zero patched. `gamma` isn't the "single long-form block" the doc claims either, it's two bitset blocks.

The cause is structural rather than a bad weight, which is why rewording alone would have been the wrong fix. A drawn tier dense enough to fill a block is dense enough for the bitset to claim it, and lowering a weight until its deltas pack narrower drops the term below the 128-posting cap on the way. Random placement is the reason: the widest gap in a block sets the delta width for all 128 lanes, and at these densities that gap is wide.

So I took your second option. The delta-coded formats now come from planted strides, whose gaps are uniform by construction: one tier at stride 3, the narrowest stride whose deltas beat a presence bitset (at stride 2 the 128 postings span 254 documents, which is four 64-bit words, exactly what two-bit deltas cost, and the encoder keeps the bitset on a tie). Every seed now gives a packed block deterministically.

Patched I could not reach at this corpus size, so the note now says so rather than claiming it. A block uneven enough to be worth patching spans enough documents that its presence bits are cheaper still; escaping that needs both a wide span and 128 postings to fill it, and a few hundred documents cannot supply both at once. The jumped tier reaches patched in the large tier, which I verified, and the planted corpora in `boundaries.rs` cover it exactly either way.

Two bugs fell out of writing this, both yours to thank for:

- The lane planted tokens by evicting a drawn one when a document was at the length cap, so a second plant into the same document evicted the first. The singleton was missing entirely from two of three seeds, and the strided tier's document frequency drifted below its stride. Drawn lengths now leave room for the planted tokens.
- The default lane drew corpora from one document upward, where the strided tier cannot reach the short-form cap at all. It now draws from the range where the forms it grades are actually present.

Both claims are asserted against the built index now, at each size, so the table cannot drift from what the generator does again. That assertion is the part I should have written in the first place.

tests/superfile/fts/fuzz_oracle.rs
vinaykakade: Confirmed, and widening the range alone would not have fixed it. The dense term was drawn at weight 40 of 57 over drawn lengths of 1 to 9, so it lands in about 95% of documents: 248 blocks at the top of the current range, and still only 253 at the 34,000 you suggested. It never crosses 256 at any corpus size the lane would plausibly use, so the seeded walk has never run in this lane.

So I took the second half of your suggestion, which is the part that actually fixes it. The dense tier is planted in every document now rather than drawn, making its posting count exactly the corpus size and its block count arithmetic instead of a draw that has to clear a line. The corpus floor is then derived from the walk's floor plus eight blocks of headroom rather than being a round number chosen near it, and a test asserts the crossing against the built index.

I restated the walk's 256-block floor as a constant here rather than importing it, on the same reasoning the boundary tests use for the encoder's caps: if the floor moves, this lane should fail and be re-sized deliberately rather than following it silently to wherever it went. That is exactly the failure mode you just caught, one level up.

Both of your comments were the same shape of bug, and it is worth naming: the lane asserted the engine's *answers* thoroughly and never asserted the *shape of the corpus* it was grading them on, so two claims in the doc drifted from what the generator built and nothing failed. There are now assertions on both, at both sizes.

pratyushlokhande (approved):

PR #792, opened 2026-09-19
vector: 1-bit-gated recalibration sweep (3.1x faster, law-parity)

muralikpbhat:
## What

Speeds up the compaction recalibration sweep — the pass that refits the vector probe-serving laws (width / fine / rerank) after a compaction — by replacing its exhaustive per-row fp32 scoring with a two-pass 1-bit gate.

## Why

Recalibration scores a frozen 256-query sample against every live row to measure how deep each query has to search to hit its true top-k, then stamps the laws the query path serves against. The scoring core was an exhaustive fp32 sweep: every row, every query, full-precision distance. That is O(N) fp32 work per optimize and dominates the sweep's wall time — on a single 16 vCPU / 64 GB box over a 5.3M-row 768-d corpus it is ~120s of CPU, and it grows linearly with the corpus.

## How

Two passes, gated by the cheap 1-bit RaBitQ estimate:

- Pass 1 (`shortlist_rows`) runs the 1-bit estimate over every row of every cell and keeps, per query, the top-CAP candidates by estimate. CAP is fixed well above the deepest law k, so the survivor set is guaranteed to cover every k the laws measure. Where the target supports it, the estimate uses the FastScan-LUT kernel (each query folded into nibble tables once at freeze, a cell's codes transposed once, 64 rows scored per SIMD permute); otherwise it falls back to the scalar per-row estimate. Shortlists are held in bounded min-heaps (top-CAP by estimate), so selection is O(rows·log CAP) with CAP-bounded memory rather than a full sort-and-truncate.
- Pass 2 (`score_survivors`) reloads only the cells that hold a survivor and exact-rescores just those rows, then observes fine ranks. A cell with no survivor contributes to no query's top-k and is skipped entirely.

The laws still come from the same finish reduction, and the rerank histogram bins the scalar estimate on both the FastScan and scalar paths, so the survivor rank the laws are fit against is byte-identical to the exhaustive scorer's. This equivalence is the correctness contract, asserted by a parity test across all three metrics (cosine / L2 / negdot) that keeps the exhaustive score as its reference oracle. A separate focused test exercises the bounded-heap eviction the parity test can't reach (its cap exceeds the row count).

## Validation

On the single 16 vCPU / 64 GB box, 768-d 5.3M-row corpus, both metrics:

- Laws identical to the exhaustive path (width `[1,1,1,1]` on both; rerank within a point), post-drain recall 0.988–0.990, filtered recall 0.993–0.999.
- Sweep wall 119.4s → 38.4s, a 3.1x reduction. The saving is the fp32 scoring the estimate culls.

## Known follow-up

The win here is CPU: the gate removes fp32 distance work, not row I/O. Pass 1 still materializes every cell to read its codes, and pass 2 reloads the survivor cells, so on an I/O-bound (cold object-store) recalibration the row-materialization cost is paid roughly twice. Loading only the codes in pass 1 (and caching survivor rows for pass 2) would close that gap; it is a natural follow-up and does not change the laws.

Part of #783.


src/supertable/opann.rs
```diff
@@ -1945,6 +2282,59 @@ fn truncate_ascending(cand: &mut Vec<(f32, u32, i128, f32)>, cap: usize) {
     }
 }
 
+/// Keep the `cap` largest-estimate `(estimate, cell, stable id)` candidates —
```
savannahar68: the comment doesn't look right
can you rewrite this ?

savannahar68 (approved):
lgtm

src/supertable/opann.rs
muralikpbhat: Rewritten — the stray duplicate doc block (it described the keep-cap operation, which `push_bounded` already documents) is dropped; `EstCand` now carries a single accurate comment. Pushed in 2e5217de.

PR #793, opened 2026-09-19
vector: optional recalibration policy on optimize() (Auto/Force/Skip)

muralikpbhat:
## What

Adds an optional recalibration policy to `optimize()` so a caller can defer the O(N) probe-law recalibration when it isn't needed yet, without changing any existing behavior.

## Why

`optimize()` re-fits the vector probe-serving laws (width / fine / rerank) whenever the hidden index membership changed or the rerank law lags its pool. That recalibration is a full-corpus sweep — it scores a frozen query sample against every live row — and it is the right thing to do before serving queries. But a workload that ingests in checkpoints and optimizes each one pays that sweep on every checkpoint even though no query is served until the very end. At large scale, repeated per-checkpoint recalibration dominates the ingest.

The storage side of `optimize()` (drain, split, merge) is necessary on every checkpoint and must always run. Only the query-serving recalibration is deferrable — so make it a policy the caller chooses.

## How

`OptimizeOptions` gains a `recalibrate` field:

- `Auto` (default) — the existing condition: recalibrate when the live superfile set changed since the pre-pass snapshot, or the rerank law lags its pool. Every existing caller keeps this behavior unchanged.
- `Force` — always recalibrate, regardless of the Auto condition (a final optimize before serving, when the laws must reflect the full corpus).
- `Skip` — never recalibrate this optimize; the storage-necessary drain/split/merge still run.

The policy threads from `optimize()` through compaction to the single `recalibrate_probe_laws` gate. `compact()` / `compact_async()` keep their `Auto` behavior for their existing (test) callers; a `compact_with()` / `compact_async_with()` pair carries the explicit policy, and the production `optimize()` path uses it. Because the field defaults to `Auto`, no existing caller changes.

The Node binding exposes it as `optimize({ recalibrate: "auto" | "force" | "skip" })`, omitted ⇒ `"auto"`; an unrecognized value is rejected rather than silently ignored.

## Correctness

The gate wraps exactly the `recalibrate_probe_laws` call and nothing else — all storage work runs before it unconditionally, so `Skip` cannot skip storage-necessary work. A test plants a lagging law (the same state the existing repair test proves `Auto` fixes), then asserts `Skip` leaves it lagging and a following `Force` repairs it — so the gate defers on demand and can be forced back on. Existing recalibration tests continue to pass under the default `Auto`.

Part of #783.


savannahar68 (approved):
looks good

PR #794, opened 2026-09-19
Use configured disk-backed temp directory

### LINKED ISSUE #539: Use a specified/configured temporary directory, rather than /tmp
For context, see PR comment here: https://github.com/infino-ai/infino/pull/531#discussion_r3720870475

adityasd314:
### What

Configure a dedicated scratch directory for temporary files.

### Why

Keep large temporary files on disk instead of relying on the system temporary directory.

### How

Use a configurable `scratch_root`, defaulting to `/opt/private/tmp`.

Closes #539


adityasd314:
@savannahar68 Can we assume that the `/opt/private/tmp` will always exist, that way we won't need to call `ensure_scratch_root` everytime we are doing `NamedTempFile::new_in`?

adityasd314:
Should we be having `mkdir -p -m1777 /opt/private/tmp` in the ci-runner so that the parent dir is created ?

savannahar68:
engine vs enterprise becomes tricky here. Let me think about this.
lets connect on Monday to close this. I'm thinking more of a default path for scratch dir, and having an option to pass so that it covers both. Become without proper permissions we might run into permission issues.

adityasd314:
okay, to avoid getting into permission issues some pre-determined directory must already be existing, Also I will remove the code which tries to create the parent directory if it does not exist, as anyways it will fail due to permission issue on `/opt` if it tries creating one there

src/config/config.yaml
```diff
@@ -72,6 +72,7 @@ supertable:
 storage:
   backend: none
   local_root:
+  scratch_root: /opt/private/tmp
```
savannahar68: Defaulting scratch_root to /opt/private/tmp breaks as /opt isn't writable without root, so create_dir_all fails right away. And it's a specific machine's path we shouldn't be using as the default.

src/config/mod.rs
```diff
@@ -188,6 +188,25 @@ pub fn global() -> &'static Config {
...
+/// Resolve the root used for temporary scratch space.
+///
+/// This function only resolves the path;
+/// callers that need to create files should use [`ensure_scratch_root`].
+pub(crate) fn scratch_root() -> PathBuf {
+    global()
+        .storage
+        .scratch_root
+        .clone()
+        .unwrap_or_else(env::temp_dir)
```
savannahar68, line 197: this is right approach, but it will never run

savannahar68 (changes requested):
Thanks for taking an attempt at fixing this.

Some changes which are required

Right now scratch_root defaults to Some("/opt/private/tmp") and ensure_scratch_root runs create_dir_all on it. That fails on a plain cargo run/cargo test as /opt isn't writable without root, so a normal dev, CI, a container, or macOS all hit EACCES (or a read-only rootfs) before anything useful happens. So the default path effectively makes the engine require root + a pre-made /opt/private/tmp, which most people running infino won't have.

The fallback you added (unwrap_or_else(env::temp_dir)) is the right behavior, but it never runs as the default config pre-fills scratch_root with Some(...), so unwrap_or_else always sees a value and the system-temp path is dead code.

What I'd do:

- Default scratch_root to None, and let scratch_root() resolve to env::temp_dir() when unset:
pub(crate) fn scratch_root() -> PathBuf {
    config().storage.scratch_root.clone().unwrap_or_else(env::temp_dir)
}
  env::temp_dir() already respects TMPDIR, always exists, and needs no permissions, so the default run is unchanged from today's behavior and gives ability to configure the path
- Delete DEFAULT_SCRATCH_DIR_ROOT, the scratch_root: /opt/private/tmp line in the embedded default config, and related two tests
- Drop the create_dir_all on the root. Keep NamedTempFile::new_in / tempdir_in as-is

src/config/config.yaml
adityasd314: understood, Should we print a warning somewhere in the code if the `scratch_root` defaults to `/tmp` in the code ?

src/config/mod.rs
adityasd314: understood, Either config.yml or `DEFAULT_SCRATCH_DIR_ROOT` will populate this both to `/opt/private/temp`

adityasd314:
okay, making the required changes !

adityasd314 (commented):
Made the necessary changes !

src/supertable/reader_cache/config.rs
```diff
@@ -146,7 +145,7 @@ const DEFAULT_MMAP_SWEEP_INTERVAL_SECS: u64 = 75;
 impl Default for DiskCacheConfig {
     fn default() -> Self {
         Self {
-            cache_root: temp_dir().join("infino-disk-cache"),
+            cache_root: crate::config::get_scratch_root().join("infino-disk-cache"),
```
savannahar68: disk cache should stay durable. if someone points scratch_root at a folder that gets auto-cleaned, this deletes the live cache. can we leave cache_root on temp_dir as it is

src/config/mod.rs
```diff
@@ -188,6 +188,18 @@ pub fn global() -> &'static Config {
     })
 }
 
+/// Resolve the root used for temporary scratch space.
+///
+/// This function only resolves the path;
+/// callers that need to create files should use [`ensure_scratch_root`].
```
savannahar68: i think ensure_scratch_root doesn't exist. So we can remove it from comments

src/config/mod.rs
```diff
@@ -188,6 +188,18 @@ pub fn global() -> &'static Config {
...
+pub(crate) fn get_scratch_root() -> PathBuf {
```
savannahar68: lets rename this as scratch_root. In rust get_ prefix is not preferred
https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

savannahar68 (commented):
looks good
some minor comments

src/config/mod.rs
adityasd314: oh yes, my bad, i forgot to cleanup!

src/config/mod.rs
adityasd314: understood !

src/supertable/reader_cache/config.rs
adityasd314: understood, will do that !

adityasd314 (commented):
minor changes !

src/superfile/builder.rs
```diff
@@ -2322,12 +2322,16 @@ fn stream_index_blobs_to_scratch(
     cell_posting_builder: Option<CellPostingBuilder>,
     prebuilt_multi_cell: Option<Vec<(u32, MergedIvfSubsection)>>,
 ) -> Result<(NamedTempFile, NamedTempFile), BuildError> {
-    let fts_file = NamedTempFile::new().map_err(BuildError::Io)?;
-    let vec_file = NamedTempFile::new().map_err(BuildError::Io)?;
+    let scratch_root = crate::config::scratch_root();
```
savannahar68: lets avoid fully qualified path and import at top

src/supertable/slow_vector_state.rs
```diff
@@ -87,6 +87,8 @@ pub(crate) enum SlowVectorStateError {
     HashMismatch,
     #[error("state parse: {0}")]
     Parse(String),
+    #[error("IO error")]
```
savannahar68: lets pas the actual details to this error 
#[error("io error: {0}")]

Check BuildError::IO

src/supertable/slow_vector_state.rs
```diff
@@ -434,7 +436,9 @@ pub(crate) async fn fetch_centroid_section(
             meta.size
         )));
     }
-    let spill = NamedTempFile::new().map_err(io_err)?;
+
+    let scratch_root = crate::config::scratch_root();
+    let spill = NamedTempFile::new_in(&scratch_root)?;
```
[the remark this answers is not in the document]
adityasd314: understood, on it!

src/supertable/slow_vector_state.rs
adityasd314: okay

src/superfile/builder.rs
adityasd314: understood

PR #797, opened 2026-09-19
fts: score-first ranked conjunctions and unions

vinaykakade:
Ranked conjunctions and ranked unions were the two weakest ranked shapes in our measurements, while
unranked `COUNT` over the same queries was already fast. A gap that appears only when a top-k bar
exists, and shrinks as the bar loosens, is a pruning gap rather than a per-candidate cost gap.

Every kernel here was doc-first: establish which docs match, then score them. The scores were the
last thing computed, so they could not be used to avoid work. This PR makes all of them score-first —
compute what is already known about a doc's score early, and use it to stop paying for the rest.

## Conjunctions

Pruning came only from the block-max bound, which sees the driver's whole 128-doc block. For a rare,
high-idf driver that is loose by nearly a full idf, because one block spans the driver's entire
term-frequency and length-norm range. Instrumented counts on a three-term query at k=10 showed the
walk visiting about three quarters of the driver's postings and fully scoring a third of them.

- **Screen the driver doc on its own score before probing.** Its term frequency is already decoded
  and the length norm is a byte load plus an L1 lookup, so the screen costs a fraction of the
  membership probes it replaces. Survivors carry both values into the full score.
- **The same screen in the general merge.** There the walk over the other terms' postings is
  irreducible, but the per-doc-per-term bookkeeping around it — a slice rebuild and a position
  store-back — is not, and the profile put it at about a sixth of that kernel.
- **Fold each term's score in as it is probed**, re-testing against the bound on the terms not yet
  seen, so a doc already too far behind is abandoned mid-way.
- **Widen the routing gate.** Which conjunctions take the walk instead of the merge was calibrated
  against a walk that probed every driver doc. Halving that cost moves the break-even. Two-term
  conjunctions could not reach the sparsity tiers at all, so a very sparse rarer term is now driven
  even against a packed common term.
- **Skip the screen where it cannot reject**, i.e. when the bar sits below what the other terms could
  contribute alone.
- **Let a sink whose bar is discounted decline the screen outright.** The must+should walk reports a
  bar that is the k-th best *minus* every optional clause's ceiling, so that a doc carrying all the
  required clauses and no optional one stays reachable. A required-only leader score tested against
  that bar is short by exactly the discount, which is the room the screen needed to reject anything,
  so no threshold tightening fixes it. The sink knows whether its bar is the real one; it now says so.

## Unions

The profile put these queries on the single-dominant-clause path, which has no windowing, so every
cost is per candidate. Two things dominated: rebuilding the non-essential bound on every candidate,
and probing every companion clause for any candidate that cleared a very loose test.

- **Measure the non-essential bound per block, not per candidate.** That sum changes only when a
  candidate crosses one of those clauses' block boundaries, and when the essential clause is the
  rare one — the shape this path exists for — almost every rebuild reproduced the previous answer.
- **Stop probing a candidate once it cannot reach the bar.** Fold each clause's contribution in as
  it is probed and re-test against the suffix bound on the clauses still unprobed, strongest first
  so the largest uncertainty resolves first. Probing was 15–24% of these queries; scoring was ~6%.

Reader-only. No format change, no public API change, no new dependency.

## Benchmarks

A 5 M-document Wikipedia corpus and a 962-query workload, five commands, two repetitions with the arm
order reversed on the second and pooled. Measured on this head; baseline is current `main`; both arms
read the same index (2.59 GiB either way).

Lane means, `main` → branch:

| lane | main µs | branch µs | ratio |
|---|---:|---:|---:|
| TOP_10 | 1133 | 996 | **0.88** |
| TOP_100 | 1524 | 1413 | **0.93** |
| TOP_1000 | 2108 | 2025 | **0.96** |
| TOP_100_COUNT | 2721 | 2596 | **0.95** |
| COUNT | 1224 | 1220 | 1.00 |

Geometric mean of per-query ratios over all five commands: **0.956**.

By query shape, branch against `main`, mean / geometric mean of per-query ratios:

| shape | n | TOP_10 | TOP_100 | TOP_1000 |
|---|---:|---:|---:|---:|
| intersection | 300 | **0.63 / 0.71** | **0.79 / 0.91** | 0.93 / 0.99 |
| intersection, 2 tokens | 198 | **0.67 / 0.70** | 0.84 / 0.90 | 0.94 / 0.96 |
| intersection, 3 tokens | 83 | **0.62 / 0.73** | 0.82 / 0.94 | 0.96 / 1.04 |
| intersection, >3 tokens | 19 | **0.60 / 0.79** | 0.65 / 0.96 | 0.86 / 1.06 |
| union | 301 | **0.83 / 0.90** | **0.89 / 0.95** | 0.95 / 0.97 |
| union, 2 tokens | 198 | 0.95 / 0.97 | 0.98 / 1.00 | 0.98 / 0.99 |
| union, 3 tokens | 83 | **0.75 / 0.79** | **0.82 / 0.87** | 0.93 / 0.95 |
| union, >3 tokens | 20 | **0.80 / 0.77** | 0.86 / 0.87 | 0.89 / 0.94 |
| phrase | 300 | 1.00 / 0.99 | 0.99 / 0.99 | 0.99 / 0.98 |
| negated | 19 | 1.00 / 1.00 | 1.01 / 1.01 | 1.00 / 1.00 |
| must+should | 40 | 1.01 / 1.01 | 1.02 / 1.02 | 1.01 / 1.00 |
| single term | 1 | 0.99 | 1.06 | 1.02 |

The two ranked lanes carry the whole change. Phrases, negations, must+should, single terms and both
count lanes are at parity. Every number above comes from the one run on this head, so the table is
internally consistent rather than stitched from several.

All 962 queries return byte-identical results in both arms across all five commands, so rankings are
unchanged.

Per-change A/Bs at 5 M, 600 runs per arm, outputs byte-compared, each against the commit before it.
Conjunctions: driver screen `+the +book +of +life` TOP_10 0.43 and `+university +of +washington`
TOP_10 0.51; general-merge screen `+to +be +or +not +to +be` TOP_10 0.58; routing `+all +funds` 0.60,
`+american +funds` 0.68, `+new +york +population` 0.72. Unions: `new york population` TOP_10 0.69,
`tallest trees in the world` 0.72, `national book award` 0.74, `central community college` 0.77.
The sink opt-out: `remote +work` TOP_100 0.91 and TOP_1000 0.92, `+data privacy` TOP_100 0.91.

One experiment is included as a commit and its revert rather than dropped, because the measurement is
the useful part: stepping the two-term merge without an unpredictable branch wins only where the two
terms have comparable posting counts and loses on skewed pairs, where the branch predicts nearly
perfectly and the conditional moves serialise a loop the branchy form let the processor run ahead in.

## Tests

`superfile` FTS integration suite (136 tests, including the brute-force BM25 and fuzz oracles) and the
FTS unit tests (391) pass, as does `make check`. Adds an agreement test that runs the screened
membership walk against the unscreened flat merge over the same cursors, across a k ladder bracketing
a bar that bites from the first window through one that never fills, on a corpus that asserts it
routes to the screened kernel.

One note for review: the union completion now keeps its running score scalar rather than packed four
at a time, since a per-clause test needs a per-clause total. That reassociates the floating-point sum,
so an emitted score can move by an ulp. Exact ties are unaffected because those break on doc id, and
the brute-force oracles agree, but the ordering tables are the check to watch.


vinaykakade:
Confirmation run on the final head, all 962 queries, three engines in one run, two reps with the
engine order reversed and pooled. `infino-main` is `f2047564`, branch is `2720b011` (the head differs
only by a formatting commit). Same corpus and indexes for all three engines; index sizes 2.59 GiB
infino, 2.61 GiB Lucene.

Lane means against Lucene 10.5.0:

| lane | before | after |
|---|---:|---:|
| TOP_10 | 1.05 | **0.97** |
| TOP_100 | 1.02 | **0.98** |
| TOP_1000 | 0.96 | **0.95** |
| TOP_100_COUNT | 0.60 | **0.58** |
| COUNT | 0.90 | 0.90 |

Geometric mean of per-query ratios against Lucene over all five commands: 1.027 → **0.992**. Against
`main`: **0.966**.

The intersection tag, 300 queries, against Lucene:

| | before mean | after mean | before gmean | after gmean |
|---|---:|---:|---:|---:|
| TOP_10 | 1.60 | **1.02** | 1.21 | **0.86** |
| TOP_100 | 1.40 | **1.11** | 1.06 | **0.97** |
| TOP_1000 | 1.22 | **1.14** | 1.05 | **1.03** |

At TOP_10 by token count, against Lucene: two-token 1.44 → **0.97** (gmean 1.18 → **0.83**),
three-token 1.68 → **1.05** (gmean 1.26 → **0.93**), four-plus 1.81 → **1.10** (gmean 1.26 → **1.00**).

Against `main`, per tag, the intersection tag is 0.64 mean / 0.71 gmean at TOP_10, 0.78 / 0.89 at
TOP_100, 0.93 / 0.99 at TOP_1000. Every other tag is flat: unions 1.00–1.02, phrases 0.99, negations
0.99–1.00, single term 0.92–1.02, and both count lanes 0.97–1.00.

Mixed must+should is 1.02 at TOP_10 and 1.04 at TOP_100 and TOP_1000 against `main`, down from
1.04–1.08 before the last commit. On 40 queries that is at the edge of this harness's repeat-measure
noise; it is the one tag that has not fully returned to parity and it is called out rather than
rounded away.

Zero result mismatches across all 962 queries and all five commands in both arms, so rankings are
identical to `main` everywhere.

pratyushlokhande (approved):
lgtm.

PR #798, opened 2026-09-20
vector: gate the O(N) centroid-router fanout calibration (+ optimize phase timers)

muralikpbhat:
## What

Gates the O(N) centroid-router **fanout calibration** that runs inside `refresh_slow_vector_state`, and adds gated optimize phase timers.

Stacked on #793 (it uses `RecalibratePolicy`) — review/merge that first; base will retarget to `main` once it lands. Part of #783.

## Why

With probe-law recalibration already skippable (#793), an incremental-ingest `optimize()` was *still* super-linear per checkpoint. Phase timers traced it to a **second** full-corpus ground-truth scan, separate from `recalibrate_probe_laws`: `calibrate_centroid_router_fanout`, invoked from `refresh_slow_vector_state`. It exact-scores the query sample against every row (O(total rows)) to measure the router's `fanout_for_k`, and it ran **twice per optimize** — once at the drain tail, once in compaction — neither honoring the caller's recalibration policy. So a bulk ingest paid two whole-corpus scans per checkpoint even with recalibration skipped.

The drain-tail scan is also **redundant** within an optimize: compaction re-settles over the final membership and supersedes it.

## How

`refresh_slow_vector_state(inner, calibrate_fanout)`:
- `calibrate_fanout = false` → publish the membership blob (required for reopen) but **skip** the fanout GT scan, carrying the prior fanout / router-section ref forward. A later settle measures it.
- The **drain tail** always passes `false` (compaction recalibrates). **Compaction** passes it per `RecalibratePolicy` (`Skip` → false; `Force`/`Auto` → true). Checkpoint/pin stamps pass `false`.

So `Skip` now drops **both** O(N) calibrations (probe + fanout); `Force`/`Auto` recalibrates once, in compaction.

Also adds `diagnostics.optimize_phase_timers` (off by default, mirroring `drain_build_timers`): `[optphase]` drain / split / merge / recalibrate / settle / drain_settle / compact_total / router_cache, plus `[optmerge]` (merge splice-vs-rebuild cell counts + wall). The measuring stick that found this.

## Impact

On a 100k→1M local reproduction (single 16 vCPU / 64 GB box, 768-d) with `Skip`: the three O(N) scans (`drain_settle`, `settle`, `recalibrate`) drop to **0.0s at every checkpoint**, cutting total optimize ~**40–57%** per checkpoint — and the removed slice was the *fastest-growing*, so the saving widens with corpus size. Correctness floor unchanged: a `Force` (e.g. a final optimize before serving) recalibrates the fanout over the full corpus as before.

## Not in scope

The residual drain growth after this change is **per-cell overhead** — fine k-means + encode + spill + upload run once per touched cell, so drain scales with cell count `n_cent` (which split grows). `build_coarse_router` was ruled out by an A/B (fixed-grid drain is flat; growing-grid `build_coarse_router` stays ~0.03s even at n_cent=899). That per-cell overhead is a separate drain lever. Also separate: unifying the probe + fanout GT scans into one shared pass (a follow-up).


src/superfile/builder.rs
```diff
@@ -1521,6 +1532,17 @@ impl SuperfileBuilder {
                 }
                 all_stable_ids.extend_from_slice(&stable_ids);
                 packed_cells.push((cell_id, merged));
+                __rebuild_ns += __mt.elapsed().as_nanos();
+                __rebuild_cells += 1;
+            }
+            if __merge_timers {
+                eprintln!(
```
savannahar68, line 1539: lets move all the eprintln in the code to tracing library so that these logs show up in grafana.

src/supertable/compaction/mod.rs
```diff
@@ -301,14 +301,22 @@ impl Supertable {
...
+        let phase_timers = crate::config::global().diagnostics.optimize_phase_timers;
         Self::compact_one_table(self, cfg, recalibrate).await?;
         if matches!(
             self.inner().manifest.load().get_partition_strategy(),
             PartitionStrategy::VectorCell { .. }
         ) {
-            refresh_slow_vector_state(self.inner())
-                .await
-                .map_err(|error| CompactionError::Refresh(error.to_string()))?;
+            let __st = std::time::Instant::now();
```
savannahar68: we already import Instant at top. Lets use Instant::now()

src/superfile/builder.rs
```diff
@@ -1442,8 +1442,17 @@ impl SuperfileBuilder {
 
             let mut cell_ids: Vec<u32> = by_cell.keys().copied().collect();
             cell_ids.sort_unstable();
+            // [optmerge] per-phase split: cells taking the cheap byte-splice
+            // (I/O + memcpy) vs the full decode/re-cluster/re-encode rebuild, and
+            // the wall in each. Gated, off by default.
+            let __merge_timers = crate::config::global().diagnostics.optimize_phase_timers;
```
savannahar68, line 1448: lets fix the import and add it to the top

src/superfile/builder.rs
```diff
@@ -1442,8 +1442,17 @@ impl SuperfileBuilder {
...
+            let mut __splice_cells = 0usize;
+            let mut __rebuild_cells = 0usize;
+            let mut __splice_ns: u128 = 0;
+            let mut __rebuild_ns: u128 = 0;
             for cell_id in cell_ids {
                 let sources = by_cell.remove(&cell_id).expect("cell present");
+                let __mt = std::time::Instant::now();
```
savannahar68, line 1455: lets import this at top and use only Instant::now here

src/supertable/compaction/mod.rs
```diff
@@ -321,9 +329,16 @@ impl Supertable {
             // republish the entry blob and restamp. Hidden tables have no
             // manifest parts, so publication is required for reopen and a
             // failure must be visible to the caller.
-            refresh_slow_vector_state(hidden.inner())
-                .await
-                .map_err(|error| CompactionError::Refresh(error.to_string()))?;
+            let __st = std::time::Instant::now();
```
savannahar68: same

src/supertable/optimize/mod.rs
```diff
@@ -30,13 +30,34 @@ impl Supertable {
         )
     )]
     pub fn optimize(&self, opts: &OptimizeOptions) -> Result<(), OptimizeError> {
+        // Optimize phase timers ([optphase]); gated, off by default. A measuring
+        // stick for compaction scaling — see DiagnosticsSettings.
+        let phase_timers = crate::config::global().diagnostics.optimize_phase_timers;
+        let mut __t = std::time::Instant::now();
```
savannahar68: lets import at top and use Instant::now() everywhre

savannahar68 (approved):
lgtm
lets fix imports and eprintln everywhere

src/supertable/compaction/mod.rs
muralikpbhat: Done in 19fa31d5 — uses the top-level `Instant` import (already `use ...time::Instant`) with `Instant::now()`.

src/supertable/compaction/mod.rs
muralikpbhat: Done in 19fa31d5 — switched to `Instant::now()` here too.

src/supertable/optimize/mod.rs
muralikpbhat: Done in 19fa31d5 — added `use std::time::Instant;` at the top and switched all three call sites to `Instant::now()`; the [optphase] lines now log via `tracing::info!` instead of `eprintln!`.

src/superfile/builder.rs
muralikpbhat: Resolved by the rebase: this line was from the stacked view when #798 still carried #793's commits. #798 is now rebased onto main (which has #793), so its diff is just the fanout gate + phase timers and no longer touches builder.rs. No eprintln remain outside a pre-existing test bench.

src/superfile/builder.rs
muralikpbhat: Moot after the rebase — #798 no longer touches builder.rs (that was #793's code, now merged to main). The remaining phase-timer imports in optimize/compaction are hoisted in 19fa31d5.

src/superfile/builder.rs
muralikpbhat: Same — builder.rs is no longer in #798's diff post-rebase; the Instant imports that remain (optimize/compaction) are fixed in 19fa31d5.

PR #801, opened 2026-09-20
query: push a WHERE into bm25_search and hybrid_search, and fill k under it

### LINKED ISSUE #799: `bm25_search` / `hybrid_search` apply a `WHERE` after the global top-k; `vector_search` already does it right
## What happens

```sql
SELECT start_line
FROM hybrid_search('chunks', 'content', 'error failed', 'embedding', '<vec>', 6000)
WHERE path = 'one-file.log'
```

`k` is taken across the whole table first and the `WHERE` runs on the
survivors. Measured on an 11,279-row table of log chunks, asking for one
file's chunks:

| k | 83-line file (2 chunks) | 334-line file (9) | 5,553-line file (135) |
|---|---|---|---|
| 80 | 0 rows | 0 | 1 |
| 6,000 | 1 | 3 | 84 |
| 20,000 (> table) | 2 | 9 | 135 |

So "the top k of this file's chunks for this query" is only answerable by
setting `k` above the table's row count, and a `COUNT(*)` or `SUM(...)`
over such a relation silently undercounts. Any `WHERE` on a scalar column
does this - `path = …`, `path LIKE 'src/x/%'`, `lang IN (...)`.

## Why

`HybridSearchTable::scan` and `Bm25Table::scan` take `_filters` and drop
them, and neither implements `supports_filters_pushdown`, so DataFusion
keeps the predicate in a `FilterExec` above a global top-k scan.

`VectorSearchTable` is the counter-example in the same directory: it
reports every filter `Inexact`, lowers them to a `CandidatePlan`
(`candidate.rs`) and runs `vector_hits_filtered_by_plan`, so the kernel
ranks only rows that satisfy the bounded predicate and the `FilterExec`
re-verifies exactly. Its own doc comment names this exact fault as the
behaviour it replaced: "a post-filtered global top-k that underflows for
selective filters."

## What the semantics should be

A `WHERE` on a search relation means *the top k among rows satisfying the
predicate*, never *the predicate applied to the global top k*. That is what
`vector_search` does when the predicate is boundable, and what every
caller of the SQL surface assumes.

## Proposed implementation

1. **Same pushdown on all three TVFs.** `bm25_search` and `hybrid_search`
   report filters `Inexact` and lower them with the same `CandidatePlan`
   the vector exec builds (the lowering block in `vector_exec.rs::execute`
   belongs in `exec/common.rs`). The vector leg of `hybrid_search` reuses
   `vector_hits_filtered_by_plan`. The BM25 leg needs the matching
   kernel: score only doc ids in the candidate bitmap (postings ∩ bitmap,
   after `surviving_superfile_ids` prunes files), so k fills from allowed
   rows. RRF-fuse the two filtered lists. `FilterExec` stays above and
   re-verifies, so correctness never depends on the pushdown.

2. **Fill k when the plan is `Unbounded`.** A predicate on a non-FTS
   column, a `NOT`, a range: today the result silently underflows. The
   TVF can apply the predicate itself on the resolved rows and grow k
   until k rows survive or the table is exhausted (adaptive over-fetch),
   so the semantics hold everywhere at a cost proportional to selectivity.

3. **Make `path` boundable.** `CandidatePlan` bounds only FTS-indexed
   columns, and a chunks table typically indexes `content` alone, so
   `WHERE path = …` is `Unbounded` even on `vector_search` today. The
   cheap route is a loader-side one: index `path` as an FTS column
   (equality lowers to a token-AND superset; a `LIKE 'src/x/%'` prefix
   lowers under the #701 rules). The engine-side route is bounding
   equality and prefix on any dictionary-encoded string column from the
   Parquet statistics, with no FTS index.

## Tests

- A table of N rows over several `path` values; for each `path`,
  `hybrid_search(...) WHERE path = p` with `k` = the path's row count
  returns exactly those rows, ordered by score, for `k` far below N.
- The same for `bm25_search`.
- `COUNT(*)` over the filtered relation equals the path's row count.
- An `Unbounded` predicate still returns k rows when k exist.
- Parity: `vector_search` with the same predicate returns the same set as
  the vector leg of `hybrid_search`.

ekechinwokah:
A WHERE over a search table function means the top k among rows satisfying the predicate. bm25_search and hybrid_search ignored the filters DataFusion offered them, so the predicate ran above a global top-k and a selective one returned a fraction of k or nothing: measured on an 11,279-row table, asking one path's chunks for k = 80 returned 0 rows, and only a k above the table's row count returned them all. vector_search already lowered its filters to a candidate plan; the other two did not, and even that plan left a predicate the index cannot bound (a scalar column, a negation) to underflow.

Three changes, one mechanism each.

1. The admission gate the FTS kernels already consult for negated terms carries an allow-set: a superfile's candidate rows, resolved from the index. Every clause shape - the single-term walk, the union and intersection kernels, the phrase walk - admits a doc into its heap only if the set holds it, so the top-k fills from allowed rows; an unfiltered search never pays for the check. bm25_search_async gains a scoped form that narrows the fan-out to the scope's superfiles and hands each its rows; hybrid_search builds one scope and feeds both retrievers, so they rank within the same rows before fusion. Fixes #799

2. Whatever the index could not bound, the function makes up itself: the pushed predicate is compiled against the function's output, applied to its resolved rows, and when fewer than k survive the kernel is rerun for twice as many hits until k stand or the table is exhausted. All three functions and bm25_search_prefix take this path, so a scalar predicate fills k at a cost proportional to its selectivity instead of leaving the caller to guess an over-fetch. 

3. A scope's superfiles are the intersection of the plan's term-bloom survival with the manifest prune leaves the SQL scan itself lowers a WHERE to - scalar min/max, value sets, LIKE blooms and ranges, null counts - extracted from the provider into one function both callers use. A predicate on a column with no full-text index now opens only the superfiles whose statistics can hold a match. 

Filters are still reported Inexact and the FilterExec above re-checks every row, so correctness never depends on any of the three.


src/supertable/query/fts.rs
```diff
@@ -658,214 +701,227 @@ impl SupertableReader {
...
+                    let hits = match range {
+                        // Ranged units exist only for pure multi-should
+                        // queries (`fanout_for` never slices when a must
+                        // or negated clause exists).
+                        Some((start, end)) => {
+                            let cell = {
+                                let mut sets =
+                                    cursor_sets.lock().expect("cursor-set map lock poisoned");
+                                Arc::clone(sets.entry(suid).or_default())
+                            };
```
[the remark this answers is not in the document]
ekechinwokah: not taken. that line is main's, re-indented by the move into the match arm — the diff shows the same `expect` on both sides. a poisoned mutex here means a kernel panicked with the cursor-set map held, and the crate's answer to that everywhere (`SharedTopK` in this file, the scalar cache, the reader cache) is to stop, not to hand back a `QueryError` that reads like a query problem.

src/supertable/query/candidate.rs
```diff
@@ -77,11 +81,52 @@ use crate::{
...
+            allow: Some(HashMap::new()),
+        }
+    }
+
+    /// Whether `superfile` still has a candidate row under this scope.
+    pub(crate) fn admits_superfile(&self, superfile: &SuperfileEntry) -> bool {
+        self.allow
+            .as_ref()
+            .is_none_or(|allow| allow.contains_key(&superfile.uri))
+    }
```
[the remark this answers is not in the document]
ekechinwokah: taken, in another form. both callers iterate `scope.superfiles` and asked `admits_superfile` about each member, so nothing outside the scope was ever asked — but a method that only answers right for members is a footgun waiting for a third caller. it's gone: `CandidateScope::admitted_superfiles()` iterates the scope's own files and applies the row gate in one place, and both `bm25_search_scoped_async` and the hybrid vector leg use it. a unit test pins the two gates read together.

src/supertable/query/vector.rs
```diff
@@ -6489,6 +6497,44 @@ impl SupertableReader {
...
+            return Ok(CandidateScope::empty());
+        }
+        let allow = match plan {
+            CandidatePlan::Unbounded => None,
+            bounded => Some(
+                self.candidate_bitmaps_from_plan(&superfiles, bounded)
+                    .await?,
+            ),
+        };
+        Ok(CandidateScope { superfiles, allow })
```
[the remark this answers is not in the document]
ekechinwokah: not taken. `fanout_candidate_bitmaps` already drops a superfile whose candidate set came back empty, and every consumer now goes through `admitted_superfiles()`, which skips a file with no entry in `allow`. `superfiles` stays the statistics survivors on purpose: the two gates are readable apart (the fts_exec tests assert on the first alone), and the second is applied where the files are used.

ekechinwokah:
on the guide's two focus areas and its coverage gap:

prefix overfetch — not taken here. `bm25_search_prefix` has no scoped kernel yet, so under a `WHERE` it fills by over-fetching under the exact predicate, the way an unbounded scope does; `fill_top_k` doubles `k` a round, so the whole fill costs about twice its last round and a selective predicate pays log₂ rounds. correct today where it was wrong before (k applied table-wide before the WHERE); a scoped prefix kernel is its own change.

slow path trigger — taken. a bounded scope forced `FanOut::PerSuperfile` even when every kept superfile's candidate set was its whole row set. `CandidateScope::bounds_rows(&kept)` now says whether the scope narrows a row of the kept files at all, and the ranged term-union fast path stays open when it doesn't. tombstoned rows make a file read as bounded (the sets already have them subtracted), which is the conservative side. unit tests cover both.

coverage — the hybrid `WHERE` test now asserts the aggregate case too: `COUNT(*)` over the filtered `hybrid_search` counts the matching rows, not a fused top-k's survivors. `make check` clean and the query exec tests green on the head.

src/supertable/query/exec/common.rs
```diff
@@ -210,6 +223,179 @@ fn unknown_column_message(name: &str, output_schema: &SchemaRef) -> String {
...
+        // whole table: there is nothing more to fetch.
+        let exhausted = hits.len() < want || want >= total;
+        if kept.num_rows() >= k || exhausted {
+            let rows = kept.num_rows().min(k);
+            return kept
+                .slice(0, rows)
+                .project(&requested_positions)
+                .map_err(|e| DataFusionError::Execution(e.to_string()));
+        }
+        want = want.saturating_mul(OVER_FETCH_GROWTH).min(total);
```
vinaykakade: **high** The over-fetch loop has no ceiling: want doubles all the way to n_docs_total(), and every round calls resolve_hits with the full want hit list, materializing that many rows of the requested plus predicate columns into one in-memory RecordBatch.

Scenario: a 10M-row table, `SELECT * FROM bm25_search('body', 'the', 10) WHERE tenant_id = 'absent'`. tenant_id has no full-text index so the plan is Unbounded, and interleaved values mean the scalar min/max leaves prune no superfile. The loop then runs about 20 rounds and finishes by asking the kernel for a 10M-entry top-k heap and decoding every matching row's scalar columns into a single batch. Before this PR that query returned a short result cheaply; now it is an OOM and latency cliff. bm25_search_prefix is the worst case, since it gets no candidate_scope at all (fts_exec.rs:426) and leans entirely on this loop.

Suggested direction: cap want (a multiple of k, or a row/byte budget) and return short past the cap, rather than letting the table size be the only bound.

src/supertable/query/exec/common.rs
```diff
@@ -210,6 +223,179 @@ fn unknown_column_message(name: &str, output_schema: &SchemaRef) -> String {
...
+            .ok_or_else(|| {
+                DataFusionError::Internal(
+                    "a pushed-down search predicate did not evaluate to a boolean".into(),
+                )
+            })?;
+        let kept = filter_record_batch(&batch, mask)
+            .map_err(|e| DataFusionError::Execution(e.to_string()))?;
+        // The kernel returned fewer hits than asked, or was asked for the
+        // whole table: there is nothing more to fetch.
+        let exhausted = hits.len() < want || want >= total;
```
vinaykakade: **medium** `hits.len() < want` is read as "the kernel had nothing more to give", but tombstones are applied after the k-sized kernel heap, per unit, in dispatch::fanout_local_hits (dispatch.rs:512), and select_top_k_stable then truncates the merge to want. A round can therefore come back short purely because of deletes, not because matches ran out.

Scenario: one superfile, 1000 rows match 'rust', the 10 highest-scoring of them are tombstoned. `SELECT * FROM bm25_search('title', 'rust', 10) WHERE bucket = 1` asks for 10 in round one, the kernel finds exactly those 10, the tombstone filter drops all of them, so hits.len() == 0 < 10, exhausted is true and the function returns zero rows. That is the same underflow this PR exists to fix, just reached through the delete path.

Suggested direction: have the search closure report whether the kernel itself ran out of candidates (pre-tombstone heap not filled) rather than inferring it from the post-filter hit count.

src/supertable/query/exec/vector_exec.rs
```diff
@@ -207,11 +210,13 @@ impl TableProvider for VectorSearchTable {
...
     /// equality. For columns whose tokenization is 1:1 with the literal
     /// (keyword / categorical values) the pushdown is exact and the
-    /// `FilterExec` drops nothing. For free-text columns where the literal
-    /// is a sub-token of larger text, the `FilterExec` may trim below `k`
-    /// (mild underflow) — still far better than the pre-pushdown behavior,
-    /// which filtered the *global* top-k. The exact, no-`FilterExec` path is
-    /// the Rust `Supertable::vector_search_filtered` API.
+    /// `FilterExec` drops nothing. Where it is not — a literal that is a
+    /// sub-token of larger text, or a predicate no index bounds at all —
+    /// the exec applies the exact predicate to its own rows and widens the
```
vinaykakade, line 215: **medium** This doc comment claims the exec "widens the kNN until k survive ... so the function never comes out short of k while matching rows exist", but VectorSearchOptions::resolve (superfile/reader.rs:2087) returns a fixed nprobe that does not depend on k, and the Unbounded branch searches with the unfiltered default. The hit count is therefore bounded by the rows in the probed clusters.

Scenario: `vector_search('emb', q, 10) WHERE bucket = 1` on a large table where the bucket-1 rows sit outside the probed clusters. Once want exceeds the rows those clusters hold, `hits.len() < want` fires in fill_top_k, exhausted is true, and the function returns fewer than k rows even though matching rows exist in unprobed clusters.

Suggested direction: widen nprobe along with want on the fill path, or soften the claim here to say what the fill actually guarantees.

src/supertable/query/candidate.rs
```diff
@@ -77,11 +81,77 @@ use crate::{
...
+/// query runs (`hybrid_search` runs two): the superfiles that may hold a
+/// match, and — when the index can bound rows — the candidate rows in each.
+///
+/// Built by `SupertableReader::candidate_scope`. The superfile set is the
+/// intersection of the manifest's scalar / null / `IN` statistics pruning
+/// (the same leaves the SQL scan uses, so `path = 'x'` on a table whose
+/// superfiles are stats-disjoint on `path` opens only the file that can
+/// hold it) with the [`CandidatePlan`]'s term-bloom survival. The rows are
+/// the plan's per-superfile evaluation, tombstones already subtracted.
+#[derive(Debug, Clone, Default)]
```
vinaykakade: **low** The Default derive yields `superfiles: []` with `allow: None`, which reads as "no superfiles, but every row of every superfile is admitted": the opposite polarity from CandidateScope::empty(), which sets `allow: Some(HashMap::new())`. On a defaulted scope, admits_superfile returns true for every superfile.

Nothing constructs one today, so this is latent, but a future `..Default::default()` or `unwrap_or_default()` would silently produce an admit-everything scope and the row bound would vanish without any error.

Suggested direction: drop the Default derive, or hand-write it so it matches empty().

src/supertable/query/fts.rs
```diff
@@ -602,17 +635,29 @@ impl SupertableReader {
...
         let kept_refs: Vec<&Arc<SuperfileEntry>> = kept.iter().collect();
         // Phrase-bearing queries stay per-superfile: the ranged
-        // kernel is the pure term-union fast path.
-        let fanout = match has_phrases {
+        // kernel is the pure term-union fast path. So does a search
+        // with a per-row scope — the ranged union kernel carries no
+        // admission gate, only the clause kernels do. A scope whose
+        // candidate sets admit every row of every kept superfile gates
+        // nothing, so it is dropped here and the fast path stays open.
+        let per_row_scope: Option<Arc<HashMap<SuperfileUri, Arc<RoaringBitmap>>>> = scope
```
vinaykakade, line 643: **low** (cost) `per_row_scope.is_some()` forces FanOut::PerSuperfile, so every row-bounded bm25_search and hybrid_search loses sub-range slicing across the reader pool. On the compacted shape (one large superfile plus small remnants) that takes the FTS leg from pool_threads units down to one. Combined with the uncapped fill loop, a scoped query needing several rounds re-runs that single-threaded fan-out, plus its select_superfiles and global-idf wave, once per round.

Suggested direction: worth a bench number in the PR body against main for the scoped path, since the query path is covered by the no-regression bar. Longer term, teaching the ranged union kernel the admission gate would let scoped queries keep the slicing.

vinaykakade (commented):
Review of the WHERE pushdown. The core mechanisms hold up: every FTS kernel branch that pushes into the heap is guarded by filter.admits(doc), dispatch_or_algo declines the WAND path whenever a filter exists, build_work_units guarantees range: None under FanOut::PerSuperfile so the ranged kernel never sees an allow-set, and the manifest prune reuse is sound (collect_conjuncts drops OR, leaf_to_predicate keys on the bare column name so a table alias still prunes). Five comments below, on the fill loop and on two smaller things.

vinaykakade (approved):
LGTM after the review comments are addressed

src/supertable/query/exec/common.rs
ekechinwokah: taken. `want` doubled until it reached `n_docs_total()`, so the case you describe ended by asking the kernel for a table-sized heap and resolving every match into one batch — a query that was cheap before the pushdown existed.

`over_fetch_ceiling(k, total)` now bounds the growth: 64x `k`, floored at 10,000 so a small `k` still gets a real look before it gives up, capped at 100,000 so one round's resolved rows are bounded whatever `k` is, and never below the `k` actually asked for. past the ceiling the fill returns what it has.

returning short is the trade, and it's stated in the doc comment rather than left to be discovered. it is still strictly more than the caller used to get — before the pushdown `k` was applied table-wide and the `WHERE` ran after it. the ceiling is unit tested at each of its bounds.

prefix gets the same ceiling, which is where it was needed most: no `candidate_scope`, so the loop was the only thing standing between it and the whole table.

src/supertable/query/exec/common.rs
ekechinwokah: taken, in another form. the mechanism is exactly as you describe — `dispatch::fanout_local_hits` subtracts tombstones after each unit's k-sized heap, so a round comes back short from deletes alone, and reading that as exhaustion returned nothing.

i didn't plumb the pre-tombstone signal out of the kernel though: that changes the return type of `bm25_search_scoped_async` and the vector and hybrid equivalents — three call sites — to carry a bool whose only consumer is the loop. instead the loop stops inferring from a single round. asking for more can only return a superset, so exhaustion is a round that returns no more hits than the one before it. monotone, local to `fill_top_k`, no signature change, and it cannot end early for any reason, deletes included.

the cost is one extra round when the kernel genuinely does run dry, and that only happens on the unselective predicates the ceiling above already bounds. if you would rather have the exact signal, say so and i will plumb it.

your scenario is the regression test: ten highest-scoring rows deleted, twelve live below them, `k` of ten. reverting just the exhaustion line gives left: 0, right: 10.

src/supertable/query/exec/vector_exec.rs
ekechinwokah: taken. `VectorSearchOptions::resolve` takes no `k` at all — it returns `DEFAULT_NPROBE` or `FILTERED_DEFAULT_NPROBE` — so the fill widens `k` while the probe width stays put, and the candidates stay those of the probed cells. the claim was wrong as written; the doc now says what the fill actually guarantees, and that it is bounded.

not widening nprobe along with `want` here. that is a behaviour change to the vector path with its own recall and latency story, and it wants its own pr and its own bench rather than riding in on a doc fix.

src/supertable/query/candidate.rs
ekechinwokah: taken. derive dropped, with a comment on why so it does not come back: the derived one paired an empty `superfiles` with `allow: None`, which reads as 'no files, but every row of every file admitted' — the opposite polarity from `empty()`. nothing constructs one today, which is exactly when it is cheap to remove.

src/supertable/query/fts.rs
ekechinwokah: the a/b against main runs in ci on every push, so the scoped path's numbers come with this pr rather than being pasted into the body by hand.

on the mechanism — `bounds_rows(&kept)` already keeps the ranged fast path open when the scope admits every row of every kept superfile, so the fanout narrows only when the scope actually gates something. teaching the ranged union kernel the admission gate is the real fix, and it is not this pr.

the two interact the other way now as well: the ceiling bounds the number of rounds, so a scoped query can no longer re-run that single-threaded fan-out an unbounded number of times.

src/supertable/query/fts.rs
```diff
@@ -658,214 +703,227 @@ impl SupertableReader {
...
+                    let hits = match range {
+                        // Ranged units exist only for pure multi-should
+                        // queries (`fanout_for` never slices when a must
+                        // or negated clause exists).
+                        Some((start, end)) => {
+                            let cell = {
+                                let mut sets =
+                                    cursor_sets.lock().expect("cursor-set map lock poisoned");
+                                Arc::clone(sets.entry(suid).or_default())
+                            };
```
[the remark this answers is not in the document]
ekechinwokah: not taken, and it is the same line as #4057525155 above: it is main's, re-indented by the move into the match arm, and the diff shows the same `expect` on both sides. a poisoned mutex here means a kernel panicked holding the cursor-set map, and the crate's answer to that everywhere — `SharedTopK` in this file, the scalar cache, the reader cache — is to stop, not to hand back a `QueryError` that reads to the caller like a problem with their query.

src/supertable/query/exec/common.rs
```diff
@@ -210,6 +254,209 @@ fn unknown_column_message(name: &str, output_schema: &SchemaRef) -> String {
...
+        let exhausted = stalled || want >= ceiling;
+        if kept.num_rows() >= k || exhausted {
+            let rows = kept.num_rows().min(k);
+            return kept
+                .slice(0, rows)
+                .project(&requested_positions)
+                .map_err(|e| DataFusionError::Execution(e.to_string()));
+        }
+        previous_hits = Some(hits.len());
+        want = want.saturating_mul(OVER_FETCH_GROWTH).min(ceiling);
```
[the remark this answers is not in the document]
ekechinwokah: taken, and it is a real one. the plateau test was mine, from the commit before this, and it is wrong for exactly the reason given: a round is thinned after the kernel has run — tombstones are subtracted after each unit's k-sized heap, and fusion and stable-id de-duplication shrink a round too — so a flat count is not evidence the kernel is spent. twenty deleted rows above twelve live ones at k of ten empties two rounds in a row, and the fill returned nothing.

the ceiling is now the only stopping condition besides finding k, which is what the proposed code says. the cost is extra rounds on a query that was going to come up short anyway, and the ceiling bounds them. the signal that would end those rounds honestly is the kernel reporting its own saturation before tombstones and fusion touch the hits; that is not plumbed, and the doc comment says so rather than leaving it to be found again.

the regression test now deletes twenty rows instead of ten so two rounds come back empty. without the fix it is left: 0, right: 10.

src/supertable/query/exec/vector_exec.rs
```diff
@@ -395,50 +404,76 @@ impl ExecutionPlan for VectorSearchExec {
...
+                            reader
+                                .vector_hits_filtered_by_plan(column, query, want, options, bounded)
+                                .await?
+                        }
+                    };
+                    Ok(hits)
                 }
-            }
-            .map_err(search_query_df_error)?;
+            };
```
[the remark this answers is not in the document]
ekechinwokah: the mechanism named here is not the one that bites. widening `k` cannot reach rows in cells the query never probes, because `VectorSearchOptions::resolve` takes no `k` at all — it returns a fixed nprobe — so the fill widens the top-k while the probe width stays where it was. that is a real limit, it is in the thread above, and a scope would not move it.

what the comment does turn up is a real asymmetry: `fts_exec` and `hybrid_exec` both build a `candidate_scope` for any non-empty filter set, and this path builds only a `CandidatePlan`, so the vector tvf misses the manifest scalar-statistics superfile pruning the other two get. that is a cost win rather than a correctness fix and it lands on the path carrying the recall bar, so it is not riding into this pr on a review comment — it wants its own change with a bench behind it.

PR #802, opened 2026-09-20
fts: run a union window to its cap when one essential term has docs in it

vinaykakade:
A ranked union ends every window at the nearest block boundary among the terms that are still
essential, so the threshold refreshes once per block and the essential set can collapse early. When
a rare term is unioned with a very common one, that boundary is the common term's block, and most
windows hold no doc of the rare term at all: the window is one block of the common term, and the
per-window partition, bound and skip are paid to score nothing.

Three changes, all to the ranked-union kernel, none of which alters a result:

- **A window whose other essential terms have no doc before the window cap now runs to the cap.**
  The block boundary buys nothing there, because only one term contributes; its blocks are bounded
  by the non-essential terms alone.
- **Blocks under the bar are skipped by their metadata.** A run of blocks whose maxima cannot beat
  the threshold is walked in the skip table and left undecoded, instead of one decode and one skip
  per block. A block reaching past the window is scored up to it, so the next window's first doc
  still collects every contribution.
- **A window in which every term was accumulated admits straight off the presence bitmask.** Its
  scores are already final, so building the candidate list and filtering it is two passes where one
  compare per doc does.

Reader-only, no format change, no public API change.

## Benchmarks

All figures are this branch against its parent commit on a 5 M-document corpus, warmed, 600 runs
per arm, arms run base/new/new/base so drift cancels, outputs byte-compared on every row.

| query | ratio |
|---|---:|
| `the incredibles` TOP_1000 | **0.77** |
| `freedom tower` TOP_1000 | 0.97 |
| `american south` TOP_100 | 0.98 |
| `american south` TOP_10 | 0.98 |
| `the movement` TOP_1000 | 0.98 |
| `american south` TOP_1000 | 0.99 |
| `new york population` TOP_100 | 0.99 |
| `new york population` TOP_10 | 0.99 |
| `to be or not to be` TOP_100 | 0.99 |
| `the incredibles` TOP_100 | 1.01 |
| `+"the who" +uk` TOP_100 | 1.01 |
| `+to +be +or +not +to +be` TOP_1000 | 1.00 |
| `"the book of life"` TOP_100 | 1.00 |
| `+university +of +washington` TOP_10 | 1.00 |
| `new york population` COUNT | 1.00 |

Whole workload, 962 queries across five commands, two repetitions with the arm order reversed:

| measure | ratio |
|---|---:|
| geometric mean of per-query ratios, all commands | 0.999 |
| two-token unions, TOP_1000 | **0.96** |
| three-token unions, TOP_1000 | 0.99 |
| unions, TOP_100 | 0.99 |
| phrase queries, all ranked commands | 1.00 |
| conjunctions, all ranked commands | 1.00 |
| unranked counting | 1.00 |

The two-token unions at TOP_1000 account for 95 percent of the total improvement, which is where
the change was aimed. Nothing regresses. A control built twice from identical source puts this
harness's repeatability at 0.996 to 1.001, so the flat rows are flat.

## Tests

Three new paths, three regression tests, each written so that disabling the thing it covers makes
it fail. Treating every window as holding a single essential term, skipping one block further than
the bar allows, and dropping the guard that keeps a block reaching past the window from being
skipped all produce a wrong top-k under them. The last of those mattered: the guard is what keeps
the sole-essential bound, which counts no other essential term, from being applied to documents
past the window where those terms do have postings, and without the test it was invisible.

The FTS library suite and the `superfile` integration suite including the brute-force BM25 and
fuzz oracles pass; `make check` passes.


src/superfile/fts/reader/scorers.rs
```diff
@@ -2186,13 +2210,21 @@ impl FtsReader {
...
+                    // is scored up to it rather than skipped.
                     if prune && c.current_block != checked_block {
                         checked_block = c.current_block;
-                        let block_ub =
-                            c.current_block_max_bm25() + (total_term_ub - c.term_max_bm25);
-                        if block_ub <= threshold {
-                            let last = c.current_block_last_doc_id();
-                            c.skip_to(last.saturating_add(1));
+                        let others_ub = if sole {
+                            partial_max[f_essential]
```
pratyushlokhande, line 2244: nit: I shrank this bound by 10% and every test still passed — only a 50% cut gets caught, and by the older randomized test rather than the new one. A property test asserting a window bound never falls below a real score in that window would close it.

pratyushlokhande (approved):
Changes looks good. Ok to merge.
Added one nit comment for test.

vinaykakade:
You're right, and it was worse than the 10% you found. I swept twelve mutations over everything this
PR changes. Four survived:

| mutation | before |
|---|---|
| shrink the window bound 10% | survived |
| shrink the window bound 25% | survived |
| shrink only the sole-window bound 10% | survived |
| shrink the window bound 50% | caught, by the randomized test |
| land on the last skipped doc rather than past it | survived (not a bug: that doc is then scored normally, so it costs work and changes nothing) |
| use the sole-window bound in an ordinary window | caught |
| treat a doc in the last window slot as absent | caught |
| never track the second essential doc | caught |
| admit inline while non-essentials still owe a score | caught, as a hang rather than a failure |
| leave the presence word set after draining | caught |
| read the window score without clearing it | caught |
| admit negated docs in the inline path | caught |

The reason a shrink hides is that the ranking cannot see it. A bound that is slightly too small drops
only documents that were never going to place, so every returned top-k still matches the reference
walk; the cut has to get big enough to reach a document that would have placed before anything fails.
Asserting that a dropped document scored under the bar is no better, because a loose bound satisfies
that with room to spare. I tried that version first and the 10% cut passed it.

So the new test checks the bound on its own terms, as you suggested: every bound the walk computes is
recorded with the documents it claims to cover, and the test asserts it never falls below a real score
in that range, with the unpruned ranking as truth. Randomized over six corpora, and it fails if the
corpora ever stop exercising the bound, so it cannot quietly go vacuous.

Making it bite needed one more thing. A bound only witnesses a shrink where it is tight, so each block
now carries a planted document holding the common term at its highest frequency and the weak term, in
the shortest document in the corpus, so both reach their maximum weight on the same document. The
sole-window case needed its own fix too: that bound sums exactly the non-essential weights, and with
every term essential it was zero in every window, where a shrink of zero is still zero. A term in nine
documents out of ten gives it something to sum.

Where it lands: the global bound is now caught at a 1% shrink, and the sole-window bound at 10%. The
1% case on the sole-window bound still passes, so its slack is somewhere between the two.

src/superfile/fts/reader/scorers.rs
vinaykakade: Done (see comment below)

